PR #154, opened 2018-10-17
Feature/sl-72/implement mocking package/router

chris-miaskowski:
## Link to Jira task 

https://stoplightio.atlassian.net/browse/SL-86

## Change summary

It's an implementation of the router. The key file to look at here to understand how this works is the index.spec.ts file. 

The main assumption I made is that we will require baseUrl and path from the user (they must split the full url, not us).

Here is the code coverage ;)

selection_003


marbemac:
Loving this so far @chris-miaskowski! Looking forward to the finished candidate.

chris-miaskowski:
@marbemac 

Regarding 'relative' servers. In example: if someone spins up a mocker server on a `localhost:3000` but their OAS3 server configuration says `localhost:3001`.

Such user have three options:
1. Pass `localhost:3001` as a `baseUrl` to the router: `router.route(resources, { baseUrl: 'localhost:3001', path: aPathResolvedByServerRouting })`
2. Add the following server config to their OAS: ` { url: '' } ` | ` { url: '.' } ` | ` { url: '{any}' } ` and then invoke the router with basePath: '' | '.' | '*'

We could have made 'servers' optional and if `undefined` only match the path, but we would not be compliant with OAS (and probably with other specs).

Here is what OAS3 spec says about server's url:

> REQUIRED. A URL to the target host. This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the OpenAPI document is being served. Variable substitutions will be made when a variable is named in {brackets}.

Note is does not say: "url cannot be an empty string"

packages/http/src/router/index.ts
```diff
@@ -2,10 +2,74 @@ import { types } from '@stoplight/prism-core';
...
 import * as t from '../types';
+import { matchPath } from './matchPath';
+import { IMatch, MatchType } from './types';
+import { matchBaseUrl } from './matchBaseUrl';
 
 export const router: types.IRouter<IHttpOperation, t.IHttpRequest, t.IHttpConfig> = {
   route: async ({ resources, input, config }) => {
-    // given input and config, find and return the matching resource
-    throw new Error('Method not implemented.');
+    return Promise.resolve(route(resources, input));
```
StefanDywersant: When func is async you don't need to Promise.resolve

packages/http/src/router/index.ts
```diff
@@ -2,10 +2,74 @@ import { types } from '@stoplight/prism-core';
...
   },
 };
+
+function route(resources: IHttpOperation[], request: t.IHttpRequest) {
```
StefanDywersant: Is there a need to keep it as a separate func?
Could be moved to public route() method.

pytlesk4:
This looks really solid great work @chris-miaskowski, just needs a couple of changes.

- Throw an error when resource can't be instead of resolving/returning null. I updated the interface to support that.
- Support an option to disable matching by operation servers. When disabled requests method and path are used to match an HttpOperation.

chris-miaskowski:
Hi @pytlesk4, thank you for an insightful review. I will modify the code to throw an error as per request, but I disagree that we should _never_ return a null (w/o starting a discussion if returning null is ever a good idea).

Although I can appreciate the use case with errors I think throwing exceptions should be used sparsely and only in "exceptional" cases.

For instance, we should obviously throw an exception if we find that a user provided a malformatted data. However if the data is correct but an item just happens to not exist, we should return a valid response (null).

I have a couple of examples to support my argument:

- think about all the standard javascript functions like `find`, `match`, `indexOf`. They don't throw an exception if an element does not exist, they return normally
- and take RESTful APIs into account, if someone makes `GET /user/123` request and such user does not exist in the database, would you throw an error? I think people would normally expect to get an empty 200 response.

Anyway, having that said I will address the comments and we can change the decision later.

PR #156, opened 2018-10-18
Feature/sl-72/implement mocking package/mocker plus negotiator

chris-miaskowski:
## Link to a related Jira task

https://stoplightio.atlassian.net/browse/SL-72

## Change summary

This change introduces an implementation of Mocker package with default examples resolution behavior.


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -0,0 +1,262 @@
...
+        mediaType: responseWithSchema.mediaType,
+        schema: responseWithSchema.schema,
+      };
+    } else {
+      throw new Error('Data corrupted');
+    }
+  },
+};
+
+export default helpers;
```
chris-miaskowski: My :baby: :) 
I have "nothing to declare" ;)

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -0,0 +1,64 @@
+import { IMocker, IMockerOpts } from '@stoplight/prism-core/types';
+import { IHttpOperation } from '@stoplight/types';
+import { IHttpConfig, IHttpRequest, IHttpResponse } from '../types';
```
pytlesk4: Just a small stylistic thing, but lets import all npm modules first, and then a new line, and then all the file imports. I already made this change, and will be pushing it shortly. Just wanted to comment.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -0,0 +1,266 @@
...
```
pytlesk4: Can we move negotiator package outside of mocker again? It is a pretty cool package and we can use it to resolve other config options like if validation should be ran or not, or if the router should not match resources.servers hosts.

pytlesk4:
This looks really solid as well, good work guys! I made some changes already, mostly just style improvements/code clean up to get this into develop faster.

Only thing that I would like to see is the negotiator package moved outside of the mocker package. The negotiator package to me seems like it is perfect for negotiating all config options.

chris-miaskowski:
@pytlesk4 negotiator was outside initially but we decided to leave it in the mocker because it became pretty specific. 

I agree that at some point it may be nice to extract it but I like the YAGNI approach to things. I propose I merge the changes to `develop` and we'll extract mocker to a package as soon as we actually find a use case for it.

Let me know your thoughts I'm happy to open another pull request just to extract the negotiator.

PR #159, opened 2018-10-19
SL-71 HttpForwarder

StefanDywersant:
Jira task: https://stoplightio.atlassian.net/browse/SL-71

### DIFF
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -13,7 +13,8 @@
     "node": ">=8"
   },
   "dependencies": {
-    "json-schema-faker": "^0.5.0-rc16",
-    "@stoplight/prism-core": "0.0.0"
+    "@stoplight/prism-core": "0.0.0",
+    "axios": "^0.18.0",
+    "json-schema-faker": "^0.5.0-rc16"
   }
 }
--- a/packages/http/src/forwarder/HttpForwarder.ts
+++ b/packages/http/src/forwarder/HttpForwarder.ts
@@ -17,13 +17,15 @@ export class HttpForwarder
       throw new Error('Server list is missing in spec');
     }
 
+    const inputData = opts.input.data;
+
     const response = await axios({
-      method: opts.input.data.method,
-      url: this.resolveServerUrl(opts.resource.servers[0]) + opts.input.data.url.path,
-      params: opts.input.data.url.query,
+      method: inputData.method,
+      url: this.resolveServerUrl(opts.resource.servers[0]) + inputData.url.path,
+      params: inputData.url.query,
       responseType: 'text',
-      data: opts.input.data.body,
-      headers: opts.input.data.headers,
+      data: inputData.body,
+      headers: inputData.headers,
     });
 
     return {
--- a/packages/http/src/forwarder/HttpForwarder.ts
+++ b/packages/http/src/forwarder/HttpForwarder.ts
@@ -1,0 +1,50 @@
+import { IForwarder, IPrismInput } from '@stoplight/prism-core';
+import { IHttpOperation, IServer } from '@stoplight/types';
+import axios from 'axios';
+import { IHttpConfig, IHttpRequest, IHttpResponse } from '../types';
+
+export class HttpForwarder
+  implements IForwarder<IHttpOperation, IHttpRequest, IHttpConfig, IHttpResponse> {
+  public async forward(opts: {
+    resource?: IHttpOperation;
+    // todo: IHttpRequest should be enough
+    input: IPrismInput<IHttpRequest>;
+  }): Promise<IHttpResponse> {
+    if (!opts.resource) {
+      throw new Error('Missing spec');
+    }
+    if (!opts.resource.servers || opts.resource.servers.length === 0) {
+      throw new Error('Server list is missing in spec');
+    }
+
+    const response = await axios({
+      method: opts.input.data.method,
+      url: this.resolveServerUrl(opts.resource.servers[0]) + opts.input.data.url.path,
+      params: opts.input.data.url.query,
+      responseType: 'text',
+      data: opts.input.data.body,
+      headers: opts.input.data.headers,
+    });
+
+    return {
+      statusCode: response.status,
+      headers: response.headers,
+      body: response.data,
+    };
+  }
+
+  private resolveServerUrl(server: IServer) {
+    if (!server.variables) {
+      return server.url;
+    }
+
+    return server.url.replace(/{(.*?)}/g, (_match, variableName) => {
+      const variable = server.variables![variableName];
+      if (!variable) {
+        throw new Error(`Variable '${variableName}' is not defined, cannot parse input.`);
+      }
+
+      return variable.default || variable.enum![0];
+    });
+  }
+}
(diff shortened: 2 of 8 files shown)


packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -0,0 +1,50 @@
...
+  }): Promise<IHttpResponse> {
+    if (!opts.resource) {
+      throw new Error('Missing spec');
+    }
+    if (!opts.resource.servers || opts.resource.servers.length === 0) {
+      throw new Error('Server list is missing in spec');
+    }
+
+    const response = await axios({
+      method: opts.input.data.method,
```
chris-miaskowski: Minor stylistic comment: how about moving that `opts.input.data` to a constant? You'd avoid the repetition.

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -0,0 +1,50 @@
...
+      url: this.resolveServerUrl(opts.resource.servers[0]) + opts.input.data.url.path,
```
chris-miaskowski: Nothing wrong with `this.resolveServerUrl(opts.resource.servers[0])` but would be more generic if you passed `this.resolveServerUrl(opts.resource)` and let the function "decide" to choose first server.

Also `template strings` ;). Adding strings up is sooooo 2000 ;)

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    return {
+      statusCode: response.status,
+      headers: response.headers,
+      body: response.data,
+    };
+  }
+
+  private resolveServerUrl(server: IServer) {
```
chris-miaskowski: @StefanDywersant 
Curious: why not using the serverMatch function from the router module?

StefanDywersant: resolveServerUrl picks first server from list and eventually fills it with default values for variables. matchServer from router package finds server for given baseUrl. So The purpose of those methods is different -> I can't use matchServer.

packages/http/src/forwarder/HttpForwarder.ts
StefanDywersant: Template strings are nice feature, but which one is prettier in that case: `${a}${b}` or `a + b` ?
I vote on the latter ;-)

marbemac: I agree with @StefanDywersant :). Minor, but I generally use concatenation if its a simple combination of vars, and string interpolation otherwise.

marbemac:
### A couple of things:

- [ ] can we add a test for forwarding a file upload? (form-data request body with a file)
- [ ] add a test for a request that returns a non 2xx and you will find an issue :). axios throws errors for non 2xx status codes by default.

### Question:

We're not maintaining state in any of these classes (no `this.` usage) - do they need to be es6 classes? (HttpForwarder, HttpMocker, etc)

Example with plain object and pure functions: https://github.com/stoplightio/prism/blob/feature/sl-71/implement-forwarder/packages/http/src/router/index.ts

In any case, we should stay consistent on how we export and build up these structures. Example export inconsistency:

- exporting class instance: https://github.com/stoplightio/prism/blob/feature/sl-71/implement-forwarder/packages/http/src/forwarder/index.ts
- exporting class: https://github.com/stoplightio/prism/blob/feature/sl-71/implement-forwarder/packages/http/src/mocker/index.ts

I would suggest we export plain objects with pure functions where possible (like router). Thoughts?

StefanDywersant:
Mark,


It’s good to have that discussion now.
Mocker, Forwarder or Router are parts that I believe we can expose to the world as a separate modules. The „world” can wire them as they want to come out with some better ideas or implementation which better fits their needs. Having that said:

Classes are more easily extendable. Currently we just do new Mocker/Forwarder() and neither maintain the state nor have any constructor parameters. But are we sure that it will be like that forever? I can imagine an interface for plugins in Forwarder which are able to change slightly it’s behavior. Or caching layer? Or different adapter than axios? If such thing happens we will stand in front of such choice:
- `import { Forwarder } from ‚@stoplight/prism’; const f = new Forwarder(plugins?: IForwarderPlugin[]); f.forward(…)`
- `import { forwarder as f } from ‚@stoplight/prism’; f.registerPlugins(plugins); f.forward(…)`
- `import { forward as f } from ‚@stoplight/prism’; // not possible in ts / tricky :(`

As we are developing a library, I would vote on the classes solution. It leaves us with a possibility to retrofit forwarder implementation and be still backwards compatible.

Second solution assumes only one instance of forwarder, so we cannot have instances with some plugins while other instances have different set of plugins. Also it is harder to initialize it (the interface is not self-explanatory).

So I would stick to classes when the entity is doing some more complicated work. I would do functions or objects for utilities or algorithms.

—

Regarding entry point. You are completely right that we should choose one approach and stick to it. Here is my proposal: I would keep forwarder/HttpForwarder.ts and forwarder/index.ts separately:
- `forwarder/HttpForwarder.ts` gives us opportunity to perform dependency injection
- `forwarder/index.ts` allows us to quickly import preconfigured instance of HttpForwarder (factory)

StefanDywersant:
Mark,


It’s good to have that discussion now.
Mocker, Forwarder or Router are parts that I believe we can expose to the world as a separate modules. The „world” can wire them as they want to come out with some better ideas or implementation which better fits their needs. Having that said:

Classes are more easily extendable. Currently we just do new Mocker/Forwarder() and neither maintain the state nor have any constructor parameters. But are we sure that it will be like that forever? I can imagine an interface for plugins in Forwarder which are able to change slightly it’s behavior. Or caching layer? Or different adapter than axios? If such thing happens we will stand in front of such choice:
- `import { Forwarder } from ‚@stoplight/prism’; const f = new Forwarder(plugins?: IForwarderPlugin[]); f.forward(…)`
- `import { forwarder as f } from ‚@stoplight/prism’; f.registerPlugins(plugins); f.forward(…)`
- `import { forward as f } from ‚@stoplight/prism’; // not possible in ts / tricky :(`

As we are developing a library, I would vote on the classes solution. It leaves us with a possibility to retrofit forwarder implementation and be still backwards compatible.

Second solution assumes only one instance of forwarder, so we cannot have instances with some plugins while other instances have different set of plugins. Also it is harder to initialize it (the interface is not self-explanatory).

So I would stick to classes when the entity is doing some more complicated work. I would do functions or objects for utilities or algorithms.

—

Regarding entry point. You are completely right that we should choose one approach and stick to it. Here is my proposal: I would keep forwarder/HttpForwarder.ts and forwarder/index.ts separately:
- `forwarder/HttpForwarder.ts` gives us opportunity to perform dependency injection
- `forwarder/index.ts` allows us to quickly import preconfigured instance of HttpForwarder (factory)

packages/http/package.json
```diff
@@ -13,7 +13,8 @@
     "node": ">=8"
   },
   "dependencies": {
-    "json-schema-faker": "^0.5.0-rc16",
-    "@stoplight/prism-core": "0.0.0"
+    "@stoplight/prism-core": "0.0.0",
+    "axios": "^0.18.0",
```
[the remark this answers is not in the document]
pytlesk4: Other than that this pull request looks good, nice work!

marbemac:
@StefanDywersant sounds good! Let's consolidate to classes then for these components, and standardize the way we export them.

PR #161, opened 2018-10-22
SL-169: initial http server package implementation

marbemac:
Example end user usage in the readme. This is, for example, how the official CLI package would start up a server when needed.

- fastify as the underlying http server
- expose fastify options on creation
- expose fastify instance to end user, so they can add plugins or new routes if needed
- update the typings a bit so that instance user can specify loader input typings

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -1,28 +1,68 @@
-const express = require('express');
-import { createInstance } from '@stoplight/prism-http';
+import { createInstance, IHttpMethod, TPrismHttpInstance } from '@stoplight/prism-http';
+import * as fastify from 'fastify';
+import { IncomingMessage, Server, ServerResponse } from 'http';
+
 import { getHttpConfigFromRequest } from './getHttpConfigFromRequest';
+import { IPrismHttpServer, IPrismHttpServerOpts } from './types';
 
-const app = express();
-const port = 3000;
+export const createServer = (opts: IPrismHttpServerOpts = {}): IPrismHttpServer => {
+  const server = fastify<Server, IncomingMessage, ServerResponse>();
 
-// TODO: this is a trivial example, scratch code
-const prism = createInstance({
-  config: getHttpConfigFromRequest,
-})({
-  path: 'foo.json',
-});
+  const prism = createInstance({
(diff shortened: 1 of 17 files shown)


packages/http-server/src/server.ts
```diff
@@ -1,28 +1,73 @@
...
+
+    try {
+      const response = await prism.process({
+        method: (req.method || 'get') as IHttpMethod,
+        url: {
+          path: req.url || '/',
+          query: request.query,
+        },
+        headers: request.headers,
+        body: request.body,
```
StefanDywersant: Fastify parses body to object if it's application/json. It would most probably work, but I think we should ensure that we are sticking to raw string. The first example that comes to my head:
1. http-server receives request with content-length 100
2. http-server transforms body to object
3. forwarder calls axios with headers and body
4. axios transforms object to json leading to content-length mismatch

I believe there could be some more failures caused by parsing body.

packages/http-server/src/server.ts
```diff
@@ -1,28 +1,73 @@
...
+): fastify.RequestHandler<IncomingMessage, ServerResponse> => {
+  return async (request, reply) => {
+    const { req } = request;
+
+    try {
+      const response = await prism.process({
+        method: (req.method || 'get') as IHttpMethod,
+        url: {
+          path: req.url || '/',
+          query: request.query,
```
chris-miaskowski: One remark. `baseUrl` is optional here but if not provided `prism` will assume the path is relative (e.g. it will not take `servers` into account)

marbemac: Yeah, this is not really complete - see https://stoplightio.atlassian.net/browse/SL-216.

It is the reason that the test at https://github.com/stoplightio/prism/pull/161/files#diff-8dcd0f08ab8cb99f8132d693d8bf9682R72 is commented out.

Basically, pushing the URL parsing to the library user (http-server package in this case) is kind of annoying, and we might want to consider just accepting a simple URL string and handling the parsing internally in the http package. Going to close this one out for now, but feel free to respond in SL-216.

marbemac:
Welllll, was going to look at karols comment, but I suppose I can do that in a new branch - or even better you can add it to your list @pytlesk4  :).

PR #164, opened 2018-10-24
SL-84 basic CLI and build script rewrite

chris-miaskowski:
Link to Jira issue: https://stoplightio.atlassian.net/browse/SL-84

## Summary 

* added oclif as a dependency -> learnt that our build works fine only in compilation-time but fails in run-time
* redone the build process:
  * (kind of) replaced `paths` with `references` 
  * added dependencies to packages
  * de-composed packages and gave each its own tsconfig
* started working on the actual `run` command (WIP) 

The way it works is the following (I will use an example):
```
cd packages/cli
# compile current directory + automatically compile all dependencies
tsc -b
# that will take ./tsconfig.json and first compile all projects listed in the "references", once that's done it will compile the component itself
# each component produces an output to its local `lib` directory
# since each component is linked in node_modules with yarn's workspaces they will be available at compile-time (set proper "types" in package.json)
# they will also be available at run-time because each component's package.json now points to the correct "main" file
```


packages/http-server/src/server.ts
```diff
@@ -29,9 +28,7 @@ export const createServer = <LoaderInput = IFilesystemLoaderOpts>(
       return server;
     },
 
-    listen: async (...args) => {
-      return server.listen(...args);
-    },
+    listen: server.listen.bind(server),
```
marbemac: 👍 :)

packages/cli/package.json
```diff
@@ -1,19 +1,56 @@
...
   "version": "0.0.0",
-  "main": "lib/bundle.js",
-  "module": "lib/bundle.es.js",
   "author": "Stoplight <support@stoplight.io>",
+  "bin": {
+    "prism": "./bin/run"
+  },
+  "bugs": "https://github.com/chris-miaskowski/prism-cli/issues",
+  "dependencies": {
+    "@oclif/command": "^1",
```
marbemac: Can we peg deps with the `1.x.x` type syntax? And if not using that syntax elsewhere, applying the pattern there as well.

packages/core/src/utils/configMergerFactory.ts
```diff
@@ -1,25 +1,23 @@
...
-    return mergedConfigObject;
+    const resolvedConfigs = await Promise.all(
+      // remove any falsy configs
+      _compact(configs)
+        // resolve each config (resolveConfig is async)
+        .map((c: C) => resolveConfig<C, I>(input, c, defaultConfig))
+    );
+
+    // merge the configs over each other, in order
+    return _merge({}, ...resolvedConfigs);
```
chris-miaskowski: @marbemac I like the refactor, looks much nicer now. I see a problem with returning empty object, though.

First problem is: {} is not instance of C (interface says we will always return C, but we return {})
This may backfire at us here:
factory.ts#37
`const configObj: Config | undefined = await configMerger(input, defaultComponents.config);`
.
.
.
```
resource = await components.router.route(
              { resources, input, config: configObj },
              defaultComponents.router
            );
```
^ what if router expected to get undefined `config`? It will never be undefined because we return `{}`. It won't be instance of C too, so even though compilation passes we may have runtime errors. Imagine `type C = { x: number }`
x is required :(
and imagine we pass {} to that something that expects C.

marbemac: Hmm, good point - how about:

```ts
import { PrismConfig, PrismConfigFactory, resolveConfig } from '..';

const _compact = require('lodash/compact');
const _merge = require('lodash/merge');

/**
 * Merges all passed configs. Each next config wil override each previous config.
 */
export function configMergerFactory<C, I>(
  ...configs: Array<PrismConfig<C, I> | undefined>
): PrismConfigFactory<C, I> {
  return async (input: I, defaultConfig?: PrismConfig<C, I>): Promise<C> => {
    const resolvedConfigs =
      // remove any falsy resolved configs
      _compact(
        await Promise.all(
          // remove falsy config props
          _compact(configs)
            // resolve each config (resolveConfig is async)
            .map((c: C) => resolveConfig<C, I>(input, c, defaultConfig))
        )
      );

    if (!resolvedConfigs.length) {
      throw new Error('All configurations passed to the factory are undefined.');
    }

    // merge the configs over each other, in order
    return _merge({}, ...resolvedConfigs);
  };
}
```

packages/cli/package.json
chris-miaskowski: Sure, that was automatically generated by oclif cli ;)

chris-miaskowski:
@pytlesk4 @marbemac @StefanDywersant 

Hi guys, this PR is much bigger than the original story scope but it includes a already merged tsconfig rewrite.

As I was working on this story I noticed we will probably need to rewrite our `jest` config a little. 
Current configuration allows for only one tsconfig.json and assumes each submodule will have the exact same compilation process. This is not true for the CLI though. CLI compiles to a different target than other submodules and has a variant of a global tsconfig.

For that reason I was not able to write tests for the CLI (w/o getting into details CLI **must** compile to ES2017 which is not the case for other submodules. Current jest config does not allow that :( ).

I would like to merge that PR and take care of the jest config in SL-82.

Other than that `prism run` works like a charm :)

```
$ prism 
VERSION
  @stoplight/prism-cli/0.0.0 linux-x64 node-v8.9.0

USAGE
  $ prism [COMMAND]

COMMANDS
  help  display help for prism
  run   Start a server with the given spec file

$ prism run --help
Start a server with the given spec file

USAGE
  $ prism run

OPTIONS
  -m, --mock       Turn global mocking on or off
  -p, --port=port  [default: 4010] Port that Prism will run on.
  -s, --spec=spec  (required) File path or URL to the spec.

$ prism run --spec x
http://127.0.0.1:4010
```

marbemac:
Cool!

- `run` or `serve`? I typically think of `run` as a command that has a start and an end (a task), where-as `serve` is a long running process that you have to cancel (a server).
- Are we 100% happy with the term `spec`? Just for sake of discussion, what might the alternatives be?
- `File path or URL to the spec.` should probable be `Can be: path to the spec file, path to a folder full of spec files, or URL that points to a spec.`. Or something like that, suggestions?
- If `spec` is required, it should probably not be a flag but rather part of the command, ie `prism [command] [spec]`. But, should it be required? What if you just want to start up a server that logs requests somewhere?

pytlesk4:
> run or serve? I typically think of run as a command that has a start and an end (a task), where-as serve is a long running process that you have to cancel (a server).

- Agreed, think serve would be better.

> Are we 100% happy with the term spec? Just for sake of discussion, what might the alternatives be?

- If we can't come up with a better name to talk about something that can be mocked/validated against I am ok with spec. 

> If spec is required, it should probably not be a flag but rather part of the command, ie prism [command] [spec]. But, should it be required? What if you just want to start up a server that logs requests somewhere?

- I think we should just make it part of the command it self. And it can be a file, folder path, url ect... With that said, we should probably use a different term than spec, because a folder isn't a spec.

marbemac:
Suggestions?

pytlesk4:
Soon!

chris-miaskowski:
@marbemac @pytlesk4 

> > run or serve? I typically think of run as a command that has a start and an end (a task), where-as serve is a long running process that you have to cancel (a server).
> 
> * Agreed, think serve would be better.

No problemo, `serve` it be.

> > Are we 100% happy with the term spec? Just for sake of discussion, what might the alternatives be?
> 
> * If we can't come up with a better name to talk about something that can be mocked/validated against I am ok with spec.

I don't dislike the term `spec`. I could say: `schema` or `blueprint`? I think `spec` isn't bad, thought.

> > If spec is required, it should probably not be a flag but rather part of the command, ie prism [command] [spec]. But, should it be required? What if you just want to start up a server that logs requests somewhere?
> 
> * I think we should just make it part of the command it self. And it can be a file, folder path, url ect... With that said, we should probably use a different term than spec, because a folder isn't a spec.

After a second thought I agree with Marc that we probably don't need spec to be required. I would still be an option flag in such case, but I would remove the constraint.

> 
> The above is low hanging fruit, pull request looks great!

Thanks for your reviews guys!

package.json
```diff
@@ -22,58 +23,21 @@
...
     "@stoplight/types": "1.0.x",
     "@types/chance": "1.0.x",
-    "@types/express": "4.16.x",
     "@types/jest": "23.3.x",
-    "@types/node": "10.11.x",
+    "@types/node": "10.12.x",
     "ajv": "6.5.x",
     "chance": "1.0.x",
     "jest": "23.6.x",
+    "oclif": "^1.12.5",
```
StefanDywersant: Shouldn't it be runtime dependency of cli package?

chris-miaskowski: Oh, no, that's just a CLI for `building` the CLI :)

It's like the `@angular/cli`

StefanDywersant: :+1:

PR #166, opened 2018-10-25
SL-214 implement configMerger

chris-miaskowski:
merge user's config with default config with custom config

I'm asking for a PR to SL-84 as opposed to develop because these changes were derived from my CLI branch.

**Notes for reviewer**

To see examples take a look at :
- factory.spec.ts#"custom config function should merge over default config"
- configMergerFactory.spec.ts
- factory.ts#36

Long story short I am merging three configs: "process param config" over "custom component config" over "default component config".


packages/core/src/utils/configMergerFactory.ts
```diff
@@ -0,0 +1,25 @@
+import * as deepmerge from 'deepmerge';
```
marbemac: Is there a benefit of `deepmerge` versus `lodash.merge`? Asking simply because we generally stick to `lodash` ecosystem in the rest of the Stoplight ecosystem.

chris-miaskowski: @marbemac oh, not really, I didn't want to clog the dependencies with lodash (it's felt an overkill to include a whole library for 'just' deep merge.) I'm happy to swap to lodash if that's been the practice.

PR #168, opened 2018-10-31
SL-82: build and packaging

chris-miaskowski:
**Some remarks**

- `tslib` must be on the dependencies because we use `tsconfig/importHelpers:true`
- I had to break down `jest` configurations. This is the !key! of this pull request. I will explain why in a section below.
- I updated the code with some unrelated minor changes: specified node version in .nvmrc, updated .vscode with a typescript reference, updated readme with a `jest` hint

**Why did you split jest.config.js, Chris?**

In order to explain that I need to state some facts first:

- we use oclif to build CLI
- oclif **must** build to `es2017` - otherwise unit testing is impossible because `SomeCommand.run()` fails at compilation-time
- the `cli` package must define it's tsconfig.json file with target set to es2017
- we use `ts-jest` combined with `jest` to run tests
- jest.config.js has a special property called `globals.ts-jest.tsConfig` which tells jest+ts-jest which tsconfig.json to use (:!:)
- if `globals.ts-jest.tsConfig` is left empty -> jest will pick up the tsconfig.json file relative to the "currently processed package's directory"
- we want to be able to run all tests with one command, and generate all coverage with one command executed from the root directory
- because `globals.ts-jest.tsConfig` is a **global** configuration, ts-jest will leak incompatible tsconfig.json files between packages if we set this property explicitly...

So, in order to achieve the expected result I had to add `projects` property to the `jest.config.js` file, add an individual `jest.config.js` file to each package, remove the `globals.ts-jest` property, and leave the tsconfig.json files with their standard names.

There is another gotcha, though.

The way `jest` extends properties from one file to another is broken.

For example:
- if the base file has a `projects` field defined it will run all underlying jest tests and will carry **some** configurations. It will, for instance, carry the `moduleNameMapper` :| but I will not resolve `testMatch` property correctly.
- in order to fix that I had to create a separate file called `jest.base` which could be then extended.

So... `jest` is broken, but I have somehow manage to trick it and make the configuration work. Hopefully they will fix all those issues one day.


package.json
```diff
@@ -27,7 +28,7 @@
     "@types/node": "10.12.x",
     "ajv": "6.5.x",
     "chance": "1.0.x",
-    "jest": "23.6.x",
+    "jest": "^23.6.0",
```
StefanDywersant: Please change back to 23.6.x (convention)

packages/http/package.json
```diff
@@ -9,12 +9,16 @@
...
+  ],
   "engines": {
     "node": ">=8"
   },
   "dependencies": {
     "@stoplight/prism-core": "0.0.0",
     "axios": "^0.18.0",
-    "json-schema-faker": "^0.5.0-rc16"
+    "json-schema-faker": "^0.5.0-rc16",
+    "tslib": "^1.9.3"
```
StefanDywersant: 1.9.x

packages/http-server/package.json
```diff
@@ -12,9 +12,17 @@
...
   },
+  "files": [
+    "/lib"
+  ],
   "dependencies": {
     "@stoplight/prism-core": "0.0.0",
     "@stoplight/prism-http": "0.0.0",
-    "fastify": "1.12.x"
+    "fastify": "1.12.x",
+    "tslib": "^1.9.3"
```
StefanDywersant: 1.9.x

packages/http-server/package.json
```diff
@@ -12,9 +12,17 @@
...
+  },
+  "devDependencies": {
+    "jest": "^23.6.0",
```
StefanDywersant: 23.6.x

packages/core/package.json
```diff
@@ -9,11 +9,15 @@
...
+  "files": [
+    "/lib"
+  ],
   "engines": {
     "node": ">=8"
   },
   "dependencies": {
-    "lodash": "4.x.x"
+    "lodash": "4.x.x",
+    "tslib": "^1.9.3"
```
StefanDywersant: 1.9.x

packages/http-server/package.json
```diff
@@ -12,9 +12,17 @@
...
+    "ts-jest": "^23.10.4"
```
StefanDywersant: 23.10.x

packages/http/package.json
```diff
@@ -9,12 +9,16 @@
...
+    "/lib"
+  ],
   "engines": {
     "node": ">=8"
   },
   "dependencies": {
     "@stoplight/prism-core": "0.0.0",
     "axios": "^0.18.0",
-    "json-schema-faker": "^0.5.0-rc16"
+    "json-schema-faker": "^0.5.0-rc16",
```
StefanDywersant: No idea how to transform it to 0.x format ;)

package.json
chris-miaskowski: Not sure how I missed that, good spot!

chris-miaskowski:
@StefanDywersant changes addressed

chris-miaskowski:
Ignore the -0.6% coverage drop it's not related to my PR, codeclimate is being crazy

README.md
```diff
@@ -108,3 +108,12 @@ app.namespace('/v1').get('/', (req, res) => {
...
 });
 ```
+
+## Development
+
+### Common issues
+
+1. `jest --watch` throws ENOSPC error
+
+- [optional] Install `watchman` as per [documentation](https://facebook.github.io/watchman/docs/install.html#installing-from-source)
```
pytlesk4: @chris-miaskowski in order to run jest in watch mode I need to install watchman? If so, maybe we can add this to the yarn setup command? Our users will be happier.

chris-miaskowski: Nope, watchman won't help solving the issue it will only make the error read clearer. The actual error is related to linux OS strictly and the solution is to modify `fs.inotify.max_user_watches`.

Sometimes, e.g. when you have multiple VS Code instances open and those instances watch hundreds of files, jest --watch will throw an error because a maximum of watchers per OS is reached.

PR #170, opened 2018-11-01
General Clean up/Restructing of code

pytlesk4:
# Overview

 - Reorganized folder structure/files names. In general the file names were too verbose and it was hard navigating the code base.
	- General rule of thumbs:
		1. Files names should be camel case.
			- Examples
				JSONSchemaValidator.ts -> jsonSchemaValidator.ts
				ValidationRegistry.ts -> validationRegistry.ts

		2. use folder structure + file name to describe functionality. Class/function/interface follow a path-based-naming
			- Examples
				* folder/file: packages/http/src/validator/HttpValidator.ts -> (moved into) -> packages/http/src/validator/index.ts
					HttpValidator.ts is redundant because that file is nested under http and validator already.
				* index.ts exports class HttpValidator. The name comes from the folder structure because http and validator are in the path.
				  If HttpValidator is imported in another file, because of the way it is named it is easy to reason where it comes from.
		3. Changed helpers folder to utils. Convention at stoplight we use to name helper/utility functions.
		4. Moved utility function closer to actual usage. This is it keeps util folders from getting cluttered, can always be moved latter if other parts of the code need to use it.
			Example:
				validateAgainstSchema and convertAjvErrors are only being used by the validators package, so move into validators/utils.ts
		5. Updated folder names to reflect functionality of code underneath it.
			Example:
				- packages/http/src/validator/structure -> packages/http/src/validator/validators
				* After reviewing the code it looked like structure was just different implementations of validators like a request/response body, headers.

- Removed IHttpHeadersValidator, IHttpBodyValidator, and IHttpQueryValidator interfaces and replaced with IHttpValidator<Obj, Specs> interface

- Add types for http query objects, and headers
- reorganize helpers by what they are actually helping with
- move all interfaces into types file

Code explorer before restructuring:
image

Code explorer after restructuring:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "fmt": "prettier --write \"packages/*/src/**/*.{ts,js}\"",
     "lint": "tslint packages/*/src/**/*.{ts}",
     "build": "tsc --build tsconfig.all.json",
-    "build.clean": "rm -rf lib",
+    "build.clean": "find ./packages -name lib -type d -exec rm -rf {} \\;",
     "build.watch": "yarn build.clean && tsc --build --watch tsconfig.all.json",
     "test": "yarn lint && yarn test.fast",
     "pretest.fast": "yarn build",
--- a/yarn.lock
+++ b/yarn.lock
@@ -18,22 +18,22 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
-"@fimbul/bifrost@^0.11.0":
-  version "0.11.0"
-  resolved "https://registry.yarnpkg.com/@fimbul/bifrost/-/bifrost-0.11.0.tgz#83cacc21464198b12e3cc1c2204ae6c6d7afd158"
-  integrity sha512-GspMaQafpaUoXWWOUgNLQ4vsV52tIHUt0zpKPeJUYEyMvOSp7FIcZ1eQa7SK3GTusrEiksjMrDX/fwanigC3nQ==
+"@fimbul/bifrost@^0.15.0":
+  version "0.15.0"
+  resolved "https://registry.yarnpkg.com/@fimbul/bifrost/-/bifrost-0.15.0.tgz#f3a48dee3046681e926c1f970f0b1a67e29e088e"
+  integrity sha512-sHTwnwA9YhxcVEJkBlfKH1KLmGQGnNYPxk+09w5NnkXelYiiP8a5f351weYfxG0CUPLt1Fgkha20Y/9+jhjn/Q==
   dependencies:
-    "@fimbul/ymir" "^0.11.0"
-    get-caller-file "^1.0.2"
+    "@fimbul/ymir" "^0.15.0"
+    get-caller-file "^2.0.0"
     tslib "^1.8.1"
-    tsutils "^2.24.0"
+    tsutils "^3.1.0"
 
-"@fimbul/ymir@^0.11.0":
-  version "0.11.0"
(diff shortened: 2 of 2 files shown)


pytlesk4:
Last commit broke need this pull request from stoplight types: https://github.com/stoplightio/types

chris-miaskowski:
@pytlesk4 a general comment before I go any deeper. I generally think it's nice to have tests next to source files. I don't like storing tests under __tests__ or test/ directory. I know it was a convention for years, but I was easily converted when I first started putting tests next to source.

It's better for two reason imo: 
- because one can clearly see which files are not tested
- because we don't have to duplicate the folder structure

packages/http/src/types.ts
```diff
@@ -57,25 +57,29 @@ export interface IHttpConfig extends IPrismConfig {
...
 
+export interface IHttpNameValues {
+  [name: string]: string | string[];
+}
+
+export interface IHttpNameValue {
+  [name: string]: string;
+}
+
+export interface IHttpUrl {
```
chris-miaskowski: 👍

packages/http/src/validator/deserializers/__tests__/registry.spec.ts
```diff
@@ -1,11 +1,13 @@
-import { HttpParamDeserializerRegistry } from '..//HttpParamDeserializerRegistry';
-import { DeserializeHttpQuery, IHttpParamStyleDeserializer } from '../types';
+import { HttpParamStyles } from '@stoplight/types/http.d';
```
chris-miaskowski: @pytlesk4 why are you importing `http.d`? That won't work imo. Just remove `.d` :)

pytlesk4: That was a workaround to get it to work for me last night. I don't know why it did, but I could have been dreaming haha. Anyways, I have undone that. About to push my changes.

packages/http/src/validator/index.ts
```diff
@@ -1,15 +1,151 @@
...
+      );
+    }
+
+    if (config.headers) {
+      Array.prototype.push.apply(
+        results,
+        this.headersValidator.validate(
+          input.headers || {},
+          (resource.request && resource.request.headers) || [],
+          headerDeserializerRegistry,
```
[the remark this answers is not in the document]
pytlesk4: Passing in via constructors per your suggestion

pytlesk4:
> a general comment before I go any deeper. I generally think it's nice to have tests next to source files. I don't like storing tests under tests or test/ directory. I know it was a convention for years, but I was easily converted when I first started putting tests next to source.
>- because one can clearly see which files are not tested
  - Isn't that what code coverage is for?

>- because we don't have to duplicate the folder structure
  - Small cost for a cleaner director imo.

I have done both, where the tests live next to the source code, and in the tests folders. I like the later honestly, makes it easier to navigate files in a text editor imo. It is also how every project at stoplight is organized.

chris-miaskowski:
> > a general comment before I go any deeper. I generally think it's nice to have tests next to source files. I don't like storing tests under tests or test/ directory. I know it was a convention for years, but I was easily converted when I first started putting tests next to source.
> > 
> > * because one can clearly see which files are not tested
> 
> * Isn't that what code coverage is for?
> 
> > * because we don't have to duplicate the folder structure
> 
> * Small cost for a cleaner director imo.
> 
> I have done both, where the tests live next to the source code, and in the tests folders. I like the later honestly, makes it easier to navigate files in a text editor imo. It is also how every project at stoplight is organized.

:+1: no problemo, it's a matter of style so I won't delve into that :)

PR #176, opened 2018-11-07
SL-216 Stripping query part from url

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-216

### Notes to reviewer:
For fastify and it's request object we need to strip the query part from url. It looks better for express - there we have all necessary parameters prepared and in-place.

From the other side, we can't easily migrate to string-url solution (because of oas3 servers shape / variables / base path).

Maybe a good solution would be to support both. Url as a string and as an object. If user gets into trouble because of witchcraft he did in server list + base url + variables he can switch to `url: { path: string, basePath: string, query: {} }`. For most of the cases `url: string` will do the job.

What do you think?
@marbemac @pytlesk4 @chris-miaskowski

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -58,7 +58,7 @@ const replyHandler = <LoaderInput>(
       const response = await prism.process({
         method: (req.method || 'get') as IHttpMethod,
         url: {
-          path: req.url || '/',
+          path: (req.url || '/').split('?')[0],
           query: request.query,
         },
         headers: request.headers,
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -52,17 +52,22 @@ const replyHandler = <LoaderInput>(
     request: fastify.FastifyRequest<IncomingMessage>,
     reply: fastify.FastifyReply<ServerResponse>
   ) => {
-    const { req } = request;
-
     try {
+      const {
+        req: { method, url },
+        body,
+        headers,
+        query,
+      } = request;
+
       const response = await prism.process({
-        method: (req.method || 'get') as IHttpMethod,
+        method: (method ? method.toLowerCase() : 'get') as IHttpMethod,
         url: {
-          path: (req.url || '/').split('?')[0],
-          query: request.query,
+          path: (url || '/').split('?')[0],
+          query,
         },
-        headers: request.headers,
-        body: request.body,
+        headers,
+        body,
       });
 
       const { output } = response;
--- a/packages/http-server/src/__tests__/server.spec.ts
+++ b/packages/http-server/src/__tests__/server.spec.ts
@@ -81,9 +81,7 @@ describe('server', () => {
     expect(response.statusCode).toBe(201);
   });
 
-  // TODO: this test is failing because of our weird URL handling.
-  // should just be able to pass full url into prism, and prism should break out the parts
-  test.skip('should support choosing a response code', async () => {
+  test('should support choosing a response code', async () => {
     const response = await server.fastify.inject({
       method: 'POST',
       url: '/todos?__code=401',


packages/http-server/src/server.ts
```diff
@@ -58,7 +58,7 @@ const replyHandler = <LoaderInput>(
       const response = await prism.process({
         method: (req.method || 'get') as IHttpMethod,
         url: {
-          path: req.url || '/',
+          path: (req.url || '/').split('?')[0],
```
pytlesk4: Need to strip fragments, protocol, host, ect.. from the url so that is only the path.

pytlesk4:
> Maybe a good solution would be to support both. Url as a string and as an object. If user gets into trouble because of witchcraft he did in server list + base url + variables he can switch to url: { path: string, basePath: string, query: {} }. For most of the cases url: string will do the job.

I think `prism.process` should take a string or an object, so agree on that. But internally, if the url is a string, then we should parse it and convert it to a IHttpUrl.

packages/http-server/src/server.ts
marbemac: should add a test to verify what @pytlesk4 said.

also: 

- [ ] should not remove entire query string. we only need to remove our special query params (`__code`, `__dynamic`, `__contentType`, `__example`). Probably want to just create a const with those properties and use that wherever possible, so that things don't get missed if we add/change a new parameter in the future.
- [ ] add a test case to ensure those special query params are stripped out, and any others remain and are passed through to process

marbemac: actually @pytlesk4 that is part of the weirdness here. `url.path` is not the path, it needs to be the entire thing (proto and host included)

pytlesk4: @marbemac I think we should add properties on IHttpUrl for scheme, host and so on...

Also, surpriesd this isn't caught by typescript compiler, but the fastfitfy http query type is different than out query type: https://github.com/stoplightio/prism/blob/3ff4ff4e526ca20e287960b4b0a2235de9ff4f08/packages/http-server/src/server.ts#L62

marbemac:
In general, it seems like this edge case with the oas server variables + `basePath` is driving a less than ideal solution. What about just not supporting that edge case? It feels like bad practice in any case. Then process could just accept that simple url string. http-server would still need to strip out those reserved query parameters before passing string url to process (as described in inline commit), but it feels simpler to me?

If we want to continue with the object, then `url.path` should be renamed because it is misleading.

@chris-miaskowski @StefanDywersant @pytlesk4 what do you guys think?

packages/http-server/src/server.ts
marbemac: Is it? Looks like `type DefaultQuery = { [k: string]: any }` to me, which is compatible with our query typing.

pytlesk4: This is our query type: `[name: string]: string | string[]`

marbemac:
Another thought: if we just go with url string, we can always add a discrete `basePath` now or later that is separate from the url property. As it stands, a bunch of things feel weird about this api. The most recent example is I as a user ask myself "should path include basePath, or should basePath be pulled out of path and then prism will combine them?". 

```ts
export interface IHttpRequest {
  method: IHttpMethod;
  url: string;
  basePath?: string;
  headers?: {
    [name: string]: string;
  };
  body?: any;
}
```

pytlesk4:
I don't know how I feel about changing the IHttpRequest interface, now anytime I need query strings, or the protocol, ect... I have parse the url string.

marbemac:
By `I` do you mean you as a developer building the Prism internals? At the beginning of the `process` function we would just parse the incoming URL string into its parts and use that internally, which minimizes changes needed.

Perhaps two interfaces then, one for the options that `process` takes, and one for the parsed request object (similar to what we have now).

pytlesk4:
Correct I mean as me the developer making prism.

> Perhaps two interfaces then, one for the options that process takes, and one for the parsed request object (similar to what we have now).

Bingo!

packages/http-server/src/server.ts
chris-miaskowski: > Need to strip fragments, protocol, host, ect.. from the url so that is only the path.

@pytlesk4 @marbemac 
I don't think req.url will contain all that information. Fastify server doesn't know what `host` it is running on. E.g. when you run a web server on `localhost:3000` and do `curl localhost:3000 /somepath` fastify will not receive the `localhost:3000` part.

HTTP protocol provides **only** paths, the `host` header is for virtual servers. Therefore an HTTP server should **only** care about a path, not the whole URL.

Therefore the idea of passing whole URL to fastify or prism does not seem to fit into HTTP protocol.

chris-miaskowski:
> Need to strip fragments, protocol, host, ect.. from the url so that is only the path.

@pytlesk4 @marbemac 
I don't think req.url will contain all that information. Fastify server doesn't know what `host` it is running on. E.g. when you run a web server on `localhost:3000` and do `curl localhost:3000 /somepath` fastify will not receive the `localhost:3000` part.

HTTP protocol provides **only** paths, the `host` header is for virtual servers. Therefore an HTTP server should **only** care about a path, not the whole URL.

Therefore the idea of passing whole URL to fastify or prism does not seem to fit into HTTP protocol.

marbemac:
I think there is some miscommunication. Will try to clarify through example. 

Given this interface (which is what we have right now):

```ts
interface IHttpRequest {
  method: IHttpMethod;
  url: {
    baseUrl?: string;
    path: string;
    query?: {
      [name: string]: string;
    };
  };
  headers?: {
    [name: string]: string;
  };
  body?: any;
}
```

And the following request (`/v1` is the the basePath):

`GET https://api.example.com:3000/v1/users?skip=10&limit=5`

What should the `xxx` values be in the mapped requestObj below?

```ts
const requestObj: IHttpRequest = {
  method: 'get',
  url: {
    baseUrl: 'xxx',
    path: 'xxx',
    query: {
      skip: 10,
      limit: 5
    }
  }
}
```

---

Quick note: in general, I'm not sure how the consumer is supposed to specify that the `/v1` portion of the request path is actually the `basePath` - this edge case is freaking me out haha.

chris-miaskowski:
> I think there is some miscommunication. Will try to clarify through example.
> 
> Given this interface (which is what we have right now):
> 
> ```ts
> interface IHttpRequest {
>   method: IHttpMethod;
>   url: {
>     baseUrl?: string;
>     path: string;
>     query?: {
>       [name: string]: string;
>     };
>   };
>   headers?: {
>     [name: string]: string;
>   };
>   body?: any;
> }
> ```
> 
> And the following request (`/v1` is the the basePath):
> 
> `GET https://api.example.com:3000/v1/users?skip=10&limit=5`
> 
> What should the `xxx` values be in the mapped requestObj below?
> 
> ```ts
> const requestObj: IHttpRequest = {
>   method: 'get',
>   url: {
>     baseUrl: 'xxx',
>     path: 'xxx',
>     query: {
>       skip: 10,
>       limit: 5
>     }
>   }
> }
> ```
> 
> Quick note: in general, I'm not sure how the consumer is supposed to specify that the `/v1` portion of the request path is actually the `basePath` - this edge case is freaking me out haha.

It should be:
`baseUrl: xxx` = ""
`path: xxx` = "/v1/users"

`https://api.example.com:3000` is stripped by the HTTP server anyway.

If we wanted to **force** a different baseUrl (aka "server") we would need to do something like:

`GET https://api.example.com:3000/v1/users?skip=10&limit=5&__server=https://api.example.com:3000/v1`

and then
`baseUrl: xxx` = "https://api.example.com:3000/v1"
`path: xxx` = "/users"

marbemac:
Ah ok that makes sense - for some reason I was thinking we expected the entire URL in path (which would have ben weird, your description makes more sense).

In that case, looks good!

marbemac:
@pytlesk4 just waiting on you now.

PR #177, opened 2018-12-05
Update README.md

tbarn:
Made a few quick updates to the readme so it is less confusing and links to the new docs site


philsturgeon:
👍🏼👌🏼

PR #179, opened 2018-12-11
SL-732, SL-1251: integrate Prism with Graph

chris-miaskowski:
# Note to reviewers

Most of this code has been put together last time I was working on and I think was reviewed. This time I added a bunch of tests and fixed couple of bugs.


packages/core/src/utils/__tests__/graphFacade.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { GraphFacade } from '../../utils/graphFacade';
```
StefanDywersant: Shouldn't be graphFacade.**func**.ts?

packages/core/src/utils/graphFacade.ts
```diff
@@ -0,0 +1,66 @@
...
+      } else if (stat.isFile()) {
+        this.graphite.graph.addNode({
+          category: NodeCategory.Source,
+          type: FilesystemNodeType.File,
+          subtype,
+          path: fsPath,
+        });
+        this.fsBackend.readFile(fsPath);
+      }
+      return await this.graphite.scheduler.drain();
```
StefanDywersant: Don't need to await here.

packages/core/src/utils/graphFacade.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return null;
+  }
+
+  get httpOperations(): IHttpOperation[] {
+    const nodes = this.graphite.graph.virtualNodes.filter(node => node.type === 'http-operation');
+    return compact(nodes.map<IHttpOperation>(node => node.data as IHttpOperation));
```
StefanDywersant: What does it mean if node.data is empty? There was an error on graphite side? The graph hasn't been processed yet? Maybe we should throw here?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { IPrism } from '@stoplight/prism-core';
+import { IHttpOperation } from '@stoplight/types/http-spec';
+import { omit } from 'lodash';
+import { relative, resolve } from 'path';
+import { createInstance, IHttpConfig, IHttpRequest, IHttpResponse } from '../';
```
StefanDywersant: Completely not important, but most of the places import from `@stoplight/...`, not `'../'`

packages/core/src/utils/__tests__/graphFacade.spec.ts
chris-miaskowski: Yup, should be **int** :)u

packages/core/src/utils/graphFacade.ts
chris-miaskowski: Good question, not sure what to be honest. Looks like it's in the interface for no particular reason:

```
export interface IVirtualNode extends IBaseNode {
    category: NodeCategory.Virtual;
    data?: unknown;
}
```

In real-life this should not be undefined, though.

packages/http/src/__tests__/http-prism-instance.spec.ts
chris-miaskowski: Yes because it's actually a bit different ;)

packages/core/src/utils/graphFacade.ts
StefanDywersant: I'd throw error here since it shouldn't happen. If we keep silent here then we won't even notice that sth is going wrong or it will be very hard to find. If we do throw (keep in mind we are at early stage, things can happen) then we will be able to quickly identify the problem and add compact or resolve bug in graph and change the interface.

But it's just a non-blocking opinion ;)

chris-miaskowski: I will make a not to make that `data` field required instead of throwing error. There is no point in having virtual node w/o data imo.

StefanDywersant: It might have sense to data be empty from graph perspective (sth might not yet been processed). It does not from prism perspective (we did scheduler.drain() so we excpect everything to be resolved). That means that it is runtime exception.

marbemac: I think we can probably go with required data prop on virtual nodes - should be set to something when the node is created.

PR #187, opened 2019-02-11
SL-1619 - Prism forwarder can work without an API in place

XVincentX:
This PR will:

1. Change the Readme a little bit so that's more informative. There's still a *LOT* of work to do, but that's a start.
2. Transform `components.router.route` to a sync function, because there's nothing async in it (I hope it's the only occurrence but there might be other)
3. Remove the need of the `IHttpOperation` instance coming from graphite. In case it is absent, it'll read the data from the request object directly. Please note that, in my opinion, the IHttpOperation should be completely removed, because the only reason it's there it's because it needs to read the server/base Url — which can be done externally. If you agree with such change, I can make that happen.
4. Add/fix/rewrite tests
5. Fix TSLint that was not running properly on the packages

### DIFF
--- a/packages/http/src/forwarder/HttpForwarder.ts
+++ b/packages/http/src/forwarder/HttpForwarder.ts
@@ -7,21 +7,17 @@ export class HttpForwarder
   implements IForwarder<IHttpOperation, IHttpRequest, IHttpConfig, IHttpResponse> {
   public async forward(opts: {
     resource?: IHttpOperation;
-    // todo: IHttpRequest should be enough
     input: IPrismInput<IHttpRequest>;
   }): Promise<IHttpResponse> {
-    if (!opts.resource) {
-      throw new Error('Missing spec');
-    }
-    if (!opts.resource.servers || opts.resource.servers.length === 0) {
-      throw new Error('Server list is missing in spec');
-    }
-
     const inputData = opts.input.data;
 
     const response = await axios({
       method: inputData.method,
-      url: this.resolveServerUrl(opts.resource.servers[0]) + inputData.url.path,
+      baseURL:
+        opts.resource && opts.resource.servers && opts.resource.servers.length > 0
+          ? this.resolveServerUrl(opts.resource.servers[0])
+          : inputData.url.baseUrl,
+      url: inputData.url.path,
       params: inputData.url.query,
       responseType: 'text',
       data: inputData.body,
--- a/README.md
+++ b/README.md
@@ -4,12 +4,30 @@
 
 <a href="https://codeclimate.com/github/stoplightio/prism/test_coverage"><img src="https://api.codeclimate.com/v1/badges/f5e363a7eb5b8f4e570f/test_coverage" /></a>
 
-## Use cases
+## What's up?
+
+Prism is a set of packages that, given an API Description Document, can
+
+1. Spin up a mock server and respond according to the provided API Description Document
+2. Act as a proxy and forward your request to an upstream server
+3. Validate all the requests passing through against the provided API Description Document
+
+## How's that?
+
+Prims is a multi-package repository so divided:
+
+- `core:` basic interfaces and abstraction for API Description Documents
+- `http:` A Prism implementation to work with HTTP Requests
+- `http-server:` A _Fastify_ instance that uses Prism to validate/mock/respond and forward to http requests
+- `cli:` A CLI to spin up servers locally easily
+
+Look at the relative repositories' README for the specific documentation.
+
+## Examples
 
 ### Run a mock server locally to test requests
 
-```
-spec
+```yml
   - path: /
     server: http://x.com/v1
   - path: /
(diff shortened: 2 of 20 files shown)


XVincentX:
I couldn't hate Codeclimate more. It's making my tests fail because I'm not covering the code, but it won't say where the problem is.

image


I believe there's no such other trolling tool on the market.

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -7,21 +7,17 @@ export class HttpForwarder
...
     const inputData = opts.input.data;
 
     const response = await axios({
       method: inputData.method,
-      url: this.resolveServerUrl(opts.resource.servers[0]) + inputData.url.path,
+      baseURL:
+        opts.resource && opts.resource.servers && opts.resource.servers.length > 0
+          ? this.resolveServerUrl(opts.resource.servers[0])
+          : inputData.url.baseUrl,
+      url: inputData.url.path,
```
chris-miaskowski: This looks good but you'll need to add some integration tests to:

- packages/http/src/__tests__/http-prism-instance.spec.ts
- packages/http-server/src/__tests__/server.spec.ts

Verify that you can "have no resources, yet still be able to run prism and validation and routing doesn't fail". This is critical because `prism.process` will do routing and validation first (which rely on resources) and forwarding later.

chris-miaskowski:
> I couldn't hate Codeclimate more. It's making my tests fail because I'm not covering the code, but it won't say where the problem is.
> 
> image
> 
> I believe there's no such other trolling tool on the market.

Not sure what you are looking at but my codeclimate looks like this

selection_103

packages/http/src/forwarder/HttpForwarder.ts
XVincentX: That makes a lot of sense! I'll add the tests!

marbemac:
Heads up that once this is accepted and merged in, I will manually publish `@stoplight/prism-http` at 0.0.x for us to use asap.

packages/http/src/forwarder/HttpForwarder.ts
XVincentX: @chris-miaskowski The routing will fail since it seems like it requires resources to make it work properly, and I guess it's fine, isn't it?

XVincentX: @chris-miaskowski I'm also not so sure what I should be testing in the second file — since having a server without a document is not something we're aiming at, right?

chris-miaskowski: @XVincentX 
> The routing will fail since it seems like it requires resources to make it work properly.

That's what I was worried about. If `router` throws then `forwarder` won't even be executed.

But the question is **"what are you planning to use in you request maker eventually? `prism.process` or the `forwarded.forward`"**?

If the former then you must change your implementation (e.g. allow prism.process to take a special config parameter that allows no-resources), if the former then your tests are fine.

There is another problem I have realised, though.

What if someone uses `prism.process` but `router.router` returns empty array AND inputData.url.baseUrl OR inputData.url.path is undefined?

XVincentX: > That's what I was worried about. If `router` throws then `forwarder` won't even be executed.

That is correct; however the router can be skipped by providing an undefined instance when creating the Prism instance: https://github.com/stoplightio/prism/blob/feat/forwarder-fix/packages/core/src/factory.ts#L38

> If the former then you must change your implementation (e.g. allow prism.process to take a special config parameter that allows no-resources), if the former then your tests are fine.

I believe that right now both would work. I can either use the forwarder directly or call the process method: as long I do not provide the router component, I can still use that function call.

> What if someone uses `prism.process` but `router.router` returns empty array AND inputData.url.baseUrl OR inputData.url.path is undefined?

You get an exception, and you kind of deserved it I guess.

chris-miaskowski: > the router can be skipped by providing an undefined instance when creating the Prism instance

Yes but I wouldn't want Prism to rely on this sort of hacks and insider knowledge. Prism must prevent this situation from happening either by exposing the right interface or by throwing errors.

In other words forwarder should not be dependent on a combination of "you must provide router and resources, if you don't provide resources you should not provide router". This is just too complex.

> I believe that right now both would work. I can either use the forwarder directly or call the process method: as long I do not provide the router component, I can still use that function call.

If you want to use `forwarder` then go for it, but using using `prism.process` and relying on the complex condition "don't provide router if you don't provide resource" is a bad idea.

My suggestion is:
- either use forwarder
- or add explicit `config.allowNoResources` (or `config.forwardOnly`) prop and make sure router and validator aren't invoked.

> You get an exception, and you kind of deserved it I guess.

What exception? As far as I can see we don't throw any explicit exception.

XVincentX: > Yes but I wouldn't want Prism to rely on this sort of hacks and insider knowledge

> In other words forwarder should not be dependent on a combination of "you must provide router and resources, if you don't provide resources you should not provide router". This is just too complex.

If you think that is a hack, then there's a design problem. 

image

That insider knowledge is exposed as a top level property that can even be undefined so — it is a possibility. If you do not want that, then the library should not allow that behavior at all.

---

That said, I vote to use the `forwarder`, I'm ok with that.

XVincentX:
@marbemac Let's go!

PR #188, opened 2019-02-11
SL-396 npm release scripts

chris-miaskowski:
**Reviewer notes**

I have copied changes from https://github.com/stoplightio/prism/pull/173 that was never merged.
It was part of https://stoplightio.atlassian.net/browse/SL-396 which hasn't been finished (was closed as "won't fix") because (AFAIR) we weren't sure we should release prism to git.

Anyway, we don't have much choice but to run those release scripts and release prism packages.

Since this is a monorepo I expect (I wasn't able to verify it) that once we run `npm run release.wet` we there will be four packages released and tagged
- @stoplight/prism-http-v1.0.0
- @stoplight/prism-http-server-v1.0.0
- @stoplight/prism-cli-v1.0.0
- @stoplight/prism-core-v1.0.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,23 @@ jobs:
           name: Run build
           command: yarn build
 
+    release:
+      docker:
+        - image: circleci/node:8
+      steps:
+        - checkout
+        - run: yarn
+        - run: yarn build
+        - run: yarn release
+
 workflows:
   version: 2
   build_and_publish:
     jobs:
       - build
+      - release:
+          filters:
+            branches:
+              only: develop
+          requires:
+            - build
--- a/lerna.json
+++ b/lerna.json
@@ -1,0 +1,4 @@
+{
+  "packages": ["packages/*"],
+  "version": "0.0.0"
+}
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "branch": "chore/sl-83/prism-packaging-and-binary-deployment",
     "verifyConditions": [
       "@semantic-release/git",
-      "@semantic-release/github",
       "@semantic-release/npm"
     ],
     "prepare": [
@@ -79,14 +78,7 @@
       "@semantic-release/npm"
     ],
     "publish": [
-      "@semantic-release/github",
       "@semantic-release/npm"
-    ],
-    "success": [
-      "@semantic-release/github"
-    ],
-    "fail": [
-      "@semantic-release/github"
     ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -16,17 +16,28 @@
     "test.fast": "jest --env=node --maxWorkers=2",
     "test.coverage": "yarn test --coverage",
     "test.watch": "yarn test --watchAll",
-    "test.update": "yarn test.fast --updateSnapshot"
+    "test.update": "yarn test.fast --updateSnapshot",
+    "release.dry": "lerna exec -- semantic-release --dry-run -e semantic-release-monorepo --tag-format='${LERNA_PACKAGE_NAME}-v\\${version}'",
+    "release.wet": "lerna exec -- semantic-release -e semantic-release-monorepo --tag-format='${LERNA_PACKAGE_NAME}-v\\${version}'"
   },
   "devDependencies": {
+    "@commitlint/cli": "7.2.x",
+    "@commitlint/config-conventional": "7.1.x",
+    "@semantic-release/git": "7.0.x",
     "@stoplight/types": "3.x.x",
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -40,5 +51,42 @@
...
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+    }
+  },
+  "release": {
+    "branch": "chore/sl-83/prism-packaging-and-binary-deployment",
+    "verifyConditions": [
+      "@semantic-release/git",
+      "@semantic-release/github",
```
marbemac: Remove

package.json
```diff
@@ -40,5 +51,42 @@
...
+      "@commitlint/config-conventional"
+    ]
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+    }
+  },
+  "release": {
+    "branch": "chore/sl-83/prism-packaging-and-binary-deployment",
```
marbemac: `develop`?

package.json
```diff
@@ -40,5 +51,42 @@
...
+    "verifyConditions": [
+      "@semantic-release/git",
+      "@semantic-release/github",
+      "@semantic-release/npm"
+    ],
+    "prepare": [
+      "@semantic-release/git",
```
marbemac: Remove. in fact, not sure we need `prepare`, `publish`, `success`, or `fail`?

package.json
marbemac: Do we need `@semantic-release/commit-analyzer`?

package.json
chris-miaskowski: Oh yeah I have that commit locally but forgot to push :) one second!

package.json
chris-miaskowski: I use this for testing locally, will change to `develop`

package.json
chris-miaskowski: I reckon we do for stuff like "BREAKING CHANGE" etc.

marbemac: Looks like might be already included actually? https://github.com/Updater/semantic-release-monorepo#release-config - search that section for "set of plugins it wraps".

chris-miaskowski: Let me check, I know recall that the `monorepo-semantic-release` re-wrapped plugins and maybe that's why.

package.json
```diff
@@ -16,17 +16,28 @@
     "test.fast": "jest --env=node --maxWorkers=2",
     "test.coverage": "yarn test --coverage",
     "test.watch": "yarn test --watchAll",
-    "test.update": "yarn test.fast --updateSnapshot"
+    "test.update": "yarn test.fast --updateSnapshot",
+    "release.dry": "lerna exec -- semantic-release --dry-run -e semantic-release-monorepo --tag-format='${LERNA_PACKAGE_NAME}-v\\${version}'",
+    "release.wet": "lerna exec -- semantic-release -e semantic-release-monorepo --tag-format='${LERNA_PACKAGE_NAME}-v\\${version}'"
   },
   "devDependencies": {
+    "@commitlint/cli": "7.2.x",
```
XVincentX: I do not think we need this package

package.json
```diff
@@ -40,5 +51,34 @@
...
+    }
+  },
+  "release": {
+    "branch": "develop",
+    "verifyConditions": [
+      "@semantic-release/git",
+      "@semantic-release/npm"
+    ],
+    "prepare": [
+      "@semantic-release/git",
```
XVincentX: Would it be possible to move all this stuff in `.releaserc`? That'd help us to keep `package.json` clean

.circleci/config.yml
```diff
@@ -37,8 +37,23 @@ jobs:
           name: Run build
           command: yarn build
 
+    release:
+      docker:
+        - image: circleci/node:8
+      steps:
+        - checkout
+        - run: yarn
+        - run: yarn build
```
XVincentX: In theory `build` should be a `pre-release` task as well

lerna.json
```diff
@@ -0,0 +1,4 @@
+{
+  "packages": ["packages/*"],
+  "version": "0.0.0"
+}
```
XVincentX: Please add `"useWorkspaces": true` so that lerna will leverage yarn workspaces.

lerna.json
```diff
@@ -0,0 +1,4 @@
+{
+  "packages": ["packages/*"],
+  "version": "0.0.0"
```
XVincentX: Make sure to set yarn for npm client:

`"npmClient": "yarn",`

PR #190, opened 2019-03-08
fix: running `prism` cli threw exception [SL-2021]

chris-miaskowski:
- Class constructor Command cannot be invoked without 'new'

SL-2021

### DIFF
--- a/packages/cli/tsconfig.build.json
+++ b/packages/cli/tsconfig.build.json
@@ -2,7 +2,8 @@
   "extends": "../tsconfig.build.json",
   "compilerOptions": {
     "rootDir": "src",
-    "outDir": "lib"
+    "outDir": "lib",
+    "target": "es2017"
   },
   "include": ["src"],
   "references": [
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "module":"commonjs",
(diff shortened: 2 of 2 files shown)


XVincentX:
@chris-miaskowski 

Apparently it does not seem to be solving the problem. I've cloned this branch and ran `yarn build` and then the command.

 Also, given the bug I think it would be nice to at least run a test that the CLI runs — even simply printing the Prims version.

chris-miaskowski:
> @chris-miaskowski
> 
> Apparently it does not seem to be solving the problem. I've cloned this branch and ran `yarn build` and then the command.
> 
> Also, given the bug I think it would be nice to at least run a test that the CLI runs — even simply printing the Prims version.

Hmm interesting, are you sure you didn't have outdated `lib` folder in the CLI package?
I have tested this locally and it works...

chris-miaskowski:
> You're right! I cleaned the whole thing and it seems like it's working correctly now.
> 
> I'd still appreciate a minimal test on the CLI (simply printing the version or something) but it can be a next step.

I tried implementing a simple unit test once and it was a pain... Jest had some problems with closing the child process as far as I recall. But yeah we should have tests for this :(

XVincentX:
Alternatively it could be just a command on the CI ran after the tests. That'd at least check that the CLI loads correctly.

chris-miaskowski:
Oh, that's actually a good idea :) and we could check the shell's return code

XVincentX:
You got it!

PR #191, opened 2019-03-12
Chore/sl 2029/graphite upgrade

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2029

- upgraded graphite to 3.x
- enabled yaml support

### DIFF
--- a/packages/core/src/utils/graphFacade.ts
+++ b/packages/core/src/utils/graphFacade.ts
@@ -54,8 +54,7 @@ export class GraphFacade {
         });
         this.fsBackend.readFile(resourceFile);
       }
-      await this.graphite.scheduler.drain();
-      return;
+      return this.graphite.scheduler.drain();
     }
     return null;
   }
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/graphFacade.ts
```diff
@@ -42,25 +42,26 @@ export class GraphFacade {
...
           type: FilesystemNodeType.File,
           subtype,
-          path: fsPath,
+          path: resourceFile,
         });
-        this.fsBackend.readFile(fsPath);
+        this.fsBackend.readFile(resourceFile);
       }
-      return this.graphite.scheduler.drain();
+      await this.graphite.scheduler.drain();
```
philsturgeon: Should we add a void return type here to let folks know that nothing is being returned?

StefanDywersant: Good catch, thanks!

PR #196, opened 2019-03-14
[SL-2088] Let the body be empty if no schema exists to mock

philsturgeon:
https://stoplightio.atlassian.net/browse/SL-2088

## Change summary

When testing out basic prism functionality, it seems confusing that 
when I request a 404 it doesnt show the 404. Even if the 404 does not have
a schema defined, that should not matter, as not all errors need to have bodies.

If the code is requested, lets give it to the user. If a body is important, they
will add one, but returning a 200 instead is going to confuse a lot of folks.


packages/http-server/src/__tests__/server.spec.ts
```diff
@@ -26,6 +26,34 @@ describe('server', () => {
     expect(response.statusCode).toBe(200);
   });
 
+  test('will return requested error response even if no schema or examples defined', async () => {
```
philsturgeon: This test fails and it seems to be intentional from the code.

XVincentX:
@philsturgeon 

1. Missing snapshot update
2. Apparently the tests fail when run in bulk, but they pass when run singularly, like if there's some internal state that's preventing to serve successive requests
3. There's a react example that's bothering Jest — either we fix it or remove it (or at least remove It from the tests)

philsturgeon:
The problem exists when running this in the CLI myself, so its a case of the tests not asserting the bug well enough. :)

packages/http-server/src/__tests__/server.spec.ts
```diff
@@ -26,6 +26,34 @@ describe('server', () => {
...
+    const response = await server.fastify.inject({
+      method: 'GET',
+      url: '/pet/123?__code=404',
+    });
+
+    expect(response.statusCode).toBe(404);
+    expect(response.payload).toBe('');
```
XVincentX: Fixed.

packages/core/src/utils/resolveConfig.ts
```diff
@@ -1,13 +1,13 @@
 import { PrismConfig, PrismConfigFactory } from '../types';
 
-export async function resolveConfig<Config, Input>(
+export function resolveConfig<Config, Input>(
   input: Input,
   config: PrismConfig<Config, Input>,
   defaultConfig?: PrismConfig<Config, Input>
-): Promise<Config> {
+): Config {
```
chris-miaskowski: Why remove the Promise?
As far as I recall the config could be stored in a file. That's why the interface is async.

chris-miaskowski: E.g. some config may be coming from the code, some from request, some from file. They get merged in the end but can be async

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -40,7 +40,7 @@ export const getHttpConfigFromRequest: PrismConfigFactory<IHttpConfig, IHttpRequ
     if (typeof config.mock === 'boolean') {
       config.mock = httpOperationConfig;
     } else {
-      config.mock = Object.assign({}, config.mock, httpOperationConfig);
+      config.mock = Object.assign(config.mock, httpOperationConfig);
```
chris-miaskowski: afaik `Object.assign(config.mock, httpOperationConfig);` will mutate the object ;) If you don't care then that's OK.

packages/http-server/src/server.ts
```diff
@@ -78,12 +78,12 @@ const replyHandler = <LoaderInput>(
...
 
-        if (output.body) {
-          // body is already serialized
-          reply.serializer((payload: any) => payload);
-          reply.send(output.body);
-        }
+        reply.serializer((payload: unknown) => payload);
+        reply.send(output.body);
+      } else {
+        reply.code(500).send('Unable to find any decent response for the current request.');
```
chris-miaskowski: Note that in line 88 we have similar error handling. The idea was that Prism throws exceptions in such situations and Prism's clients handle those exceptions. How about throwing + handling exception instead? E.g. the error returned by Prism could have an error code we could check.

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -63,7 +63,7 @@ Here is the original validation result instead: ${JSON.stringify(input.validatio
     const example = negotiationResult.example as INodeExample;
     if (example && example.value !== undefined) {
       body = example.value;
-    } else {
+    } else if (negotiationResult.schema) {
```
chris-miaskowski: Would be nice to cover this with a unit test.

packages/http-server/src/server.ts
chris-miaskowski: Hmm actually looking at your implementation feel free to ignore the above comment.

XVincentX:
@chris-miaskowski This is not ready for review, please hold on.

chris-miaskowski:
> @chris-miaskowski This is not ready for review, please hold on.

@XVincentX 

Oh sorry my bad :( It wasn't marked as 'draft' nor WIP so thought it's ready.

XVincentX:
Yeah that was my fault; I misinterpreted what Phil was referring. This is going to require some other work though.

packages/core/src/utils/resolveConfig.ts
XVincentX: This is currently not happening and even if it would — this is not the right place to do that. This function would actually fetch and merge. We need to separate these two concerns: grab the sources from whenever you want, let the `resolve` function melt them together accordingly.

Moreover, the fact that this is async has been a very big source of pain while debugging the issue.

packages/http-server/src/getHttpConfigFromRequest.ts
XVincentX: Indeed I care a lot, that's why I changed.

The current line of code is mutating object anyway — so better make it explicit so I do not get a false sense of security.

packages/http/src/mocker/HttpMocker.ts
XVincentX: Good point; I'll check whether it makes sense and add it!

XVincentX:
I can't write in the PR title since I'm not the original author although I wrote the code — so putting the stuff here.

The following PR enables the required scenario by removing some limitations and let ultimately the body be empty in case there's no suitable example or schema to generate an example.

It also fixes a nasty bug where basically the `defaultConfig` — which was supposed to be fixed/stable over the time — was being modified keeping in memory specific calls parameters; making all the subsequent calls fail. Fixing this bug has required a lot of efforts because of the really complicated logic/chain of the `resolveConfig` function. I *HAD* to remove the async of that function (which was not being used anyway) to understand what was going on. It's going to be on my todo list to refactor that part as right now feels a little bit unmanageable.

I've also fixed a second bug where In case there's no body, the http server would *never* respond. I added at least a generic handler at the end of the chain that will at least return an error response and not get the server stuck

//cc @philsturgeon

philsturgeon:
+1 from me but GitHub doesnt care seeing as I opened this initially.

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -1,14 +1,18 @@
...
   req: IHttpRequest,
   defaultConfig?: PrismConfig<IHttpConfig, IHttpRequest>
 ) => {
   // For some reason this fixed the code coverage.
   let config: IHttpConfig = { mock: true };
+
   if (defaultConfig) {
-    config = await resolveConfig<IHttpConfig, IHttpRequest>(req, defaultConfig);
+    config = Object.assign(
+      { mock: true },
```
chris-miaskowski: Maybe reuse the config from line 9?

```suggestion
      config,
```

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -706,21 +706,6 @@ describe('NegotiatorHelpers', () => {
           schema: { type: 'string' },
         });
       });
-
-      it('and cannot find example and dynamic does not exist throw error', () => {
```
chris-miaskowski: I see you replaced the logic, with returning `{ code, mediaType }`.
Why remove this test VS update it to match what you return?

packages/http/src/mocker/generator/JSONSchemaExampleGenerator.ts
```diff
@@ -17,8 +17,8 @@ jsf.option({
 });
 
 export class JSONSchemaExampleGenerator implements IExampleGenerator<any> {
-  public async generate(schema: any, mediaType: string): Promise<string> {
-    const example = await jsf.resolve(JSON.parse(JSON.stringify(schema)));
+  public async generate(schema: unknown, mediaType: string): Promise<string> {
```
chris-miaskowski: :+1: for `unknown`

packages/core/src/utils/resolveConfig.ts
```diff
@@ -1,13 +1,17 @@
 import { PrismConfig, PrismConfigFactory } from '../types';
 
-export async function resolveConfig<Config, Input>(
+function isPrismConfigFactory<C, I>(val: PrismConfig<C, I>): val is PrismConfigFactory<C, I> {
```
chris-miaskowski: Nice :)

packages/http/src/mocker/HttpMocker.ts
chris-miaskowski: Did it make sense? :P

XVincentX:
> Can you point the line that did the job? 

@chris-miaskowski This is the commit fixing the issue:
https://github.com/stoplightio/prism/pull/196/commits/e481eeb419c3e4a3c54560301a616d0ce4ae5b7a

packages/http/src/mocker/HttpMocker.ts
XVincentX: Actually this is covered [here](https://github.com/stoplightio/prism/blob/develop/packages/http/src/mocker/__tests__/HttpMocker.spec.ts#L95-L139) and does not require a dedicated test. All 3 branches are tested and the test is exactly checking out that.

packages/http-server/src/getHttpConfigFromRequest.ts
XVincentX: Fair enough, although it's a little bit confusing. Done!

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
XVincentX: Restored.

XVincentX:
@chris-miaskowski Please have another round of review.

PR #200, opened 2019-03-16
Feature/sl 2035/cli url spec

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2035

Support passing a URL to the CLI specification


packages/cli/src/commands/serve.ts
```diff
@@ -28,25 +29,38 @@ export default class Serve extends Command {
...
+      return createServer({ path: spec }, { components });
+    }
+  }
+
+  private validateSpecPath(spec?: string) {
+    if (!spec || isHttp(spec)) {
       return;
     }
 
-    if (!existsSync(path)) {
```
philsturgeon: I wonder if the validateSpecPath is doing too much? It's being asked to know about HTTP, which means we are running the HTTP check twice, and we probably only need to run it once? 

Another _tiny_ thing but the existsSync and statSync are both a little bit race condition-prone anyway. Highly unlikely to be an issue, but a more robust approach is to attempt to read the file and handle the error if it does not work.

Maybe we can just do one of these and handle the rejections:

```
function readFileAsync(filename, encoding) {
    return new Promise((resolve, reject) => {
        fs.readFile(filename, encoding, (err, data) => {
            if (err) reject(err);
            else resolve(data);
        });
    });
}
```

This way we would not need a validateSpecPath method, but instead maybe a loadSpec method which returns the contents of the spec based on various inputs.

packages/core/src/loaders/filesystem/filesystemLoader.ts
```diff
@@ -12,7 +12,7 @@ export class FilesystemLoader {
    * There is no way to filter by a type in runtime :(
    */
   public async load(/*<Resource>*/ _opts?: IFilesystemLoaderOpts): Promise<IHttpOperation[]> {
-    const fsPath = _opts ? _opts.path : DEFAULT_PATH;
+    const fsPath = !_opts || !_opts.path ? DEFAULT_PATH : _opts.path;
```
philsturgeon: This line took me a minute to figure out. Maybe the inverse would be a little more clear:

```
const fsPath = _opts && _opts.path ? _opts.path : DEFAULT_PATH;
```

packages/core/src/loaders/http/httpLoader.ts
```diff
@@ -0,0 +1,26 @@
...
+import trimStart = require('lodash/trimStart');
+import { extname } from 'path';
+import { IHttpLoaderOpts } from '../../types';
+import { GraphFacade } from '../../utils/graphFacade';
+
+export class HttpLoader {
+  constructor(private graphFacade: GraphFacade) {}
+
+  public async load(opts?: IHttpLoaderOpts): Promise<IHttpOperation[]> {
+    if (!opts || !opts.url) return [];
```
philsturgeon: What is the benefit or loading this method with an empty array?

packages/core/src/loaders/filesystem/__tests__/filesystemLoader.unit.ts
```diff
@@ -11,12 +13,15 @@ describe('filesystemLoader', () => {
...
     createFileSystemNodeMock = graphFacadeMock.createFilesystemNode as jest.Mock;
     createFileSystemNodeMock.mockResolvedValue(true);
+
     Object.defineProperty(graphFacadeMock, 'httpOperations', {
       get: jest.fn().mockReturnValue(fakeHttpOperations),
     });
   });
 
   test('given no opts should delegate to graph facade for http operations', async () => {
+    jest.spyOn(fs, 'existsSync').mockReturnValueOnce(true);
```
philsturgeon: We removed these.

packages/core/src/loaders/http/httpLoader.ts
StefanDywersant: It's to be compliant with filesystemLoader. Filesystem loader returns empty array if spec is not set.

The other question is if it makes sense? For me not. But I don't think this PR is a good place for changing that. Changing that requires minor interface modifications.

packages/cli/src/commands/serve.ts
StefanDywersant: 1) Good spot on validateSpecPath. I refactored it to be more elegant.

2) Basically we cannot get rid of race condition here. Whatever we do here there will be still a read operation in graphite.

3) readFile (read-whole-file) is not a good approach for existence/dir checks in that particular situation. If we have big file / slow fs? Anyway it won't help use because of 2). It is also kind of temporary. I believe graphite should take care of loading http's.

PR #201, opened 2019-03-21
fix: SL-2191 mock switch fix

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2191

I made config.mock optional. When it was mandatory it was filled with default of false (e.g. while doing request without explicit __mock=?). That was always overriding what was passed to cli.

### DIFF
--- a/packages/http-server/src/__tests__/__snapshots__/getHttpConfigFromRequest.spec.ts.snap
+++ b/packages/http-server/src/__tests__/__snapshots__/getHttpConfigFromRequest.spec.ts.snap
@@ -34,17 +34,9 @@ Object {
 }
 `;
 
-exports[`getHttpConfigFromRequest() given no default config and no matching query should return my own default 1`] = `
-Object {
-  "mock": true,
-}
-`;
+exports[`getHttpConfigFromRequest() given no default config and no matching query should return my own default 1`] = `Object {}`;
 
-exports[`getHttpConfigFromRequest() given no default config and no query should return my own default 1`] = `
-Object {
-  "mock": true,
-}
-`;
+exports[`getHttpConfigFromRequest() given no default config and no query should return my own default 1`] = `Object {}`;
 
 exports[`getHttpConfigFromRequest() given no default config extracts code 1`] = `
 Object {
--- a/packages/http-server/src/__tests__/server.spec.ts
+++ b/packages/http-server/src/__tests__/server.spec.ts
@@ -6,7 +6,7 @@ describe('server', () => {
   let server: IPrismHttpServer<any>;
 
   beforeAll(async () => {
-    server = createServer({}, {});
+    server = createServer({}, { components: { config: { mock: true } } });
     await server.prism.load({
       path: relative(
         process.cwd(),
--- a/packages/http/src/__tests__/http-prism-instance.spec.ts
+++ b/packages/http/src/__tests__/http-prism-instance.spec.ts
@@ -9,7 +9,7 @@ describe('Http Prism Instance function tests', () => {
   let prism: IPrism<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig, { path: string }>;
 
   beforeAll(async () => {
-    prism = createInstance();
+    prism = createInstance({ config: { mock: true } });
     await prism.load({
       path: relative(
         process.cwd(),
--- a/packages/http/src/index.ts
+++ b/packages/http/src/index.ts
@@ -20,9 +20,7 @@ import { validator } from './validator';
 
 const createInstance = <LoaderInput>(overrides?: TPrismHttpComponents<LoaderInput>) => {
   return factory<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig, LoaderInput>({
-    config: {
-      mock: true,
-    },
+    config: {},
     loader: filesystemLoaderInstance,
     router,
     forwarder,
--- a/packages/http/src/types.ts
+++ b/packages/http/src/types.ts
@@ -32,7 +32,7 @@ export interface IHttpOperationConfig {
 }
 
 export interface IHttpConfig extends IPrismConfig {
-  mock: boolean | IHttpOperationConfig;
+  mock?: boolean | IHttpOperationConfig;
 
   security?: {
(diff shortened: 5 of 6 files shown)


chris-miaskowski:
@StefanDywersant 

I see how this can fix the issue however I have a couple of questions:
- what does `mock: undefined` means in terms of logic? Will it mock or not? How is that different than saying `mock: false`? 
- I'm not quite sure what the problem is. The config should be constructed in the following order: default config -> override config object -> dynamic config from a request. Isn't your code removing the first element in the chain?
- If default config is overriding the CLI config, wouldn't it make sense to actually fix it in the config merger?

packages/http/src/types.ts
```diff
@@ -32,7 +32,7 @@ export interface IHttpOperationConfig {
 }
 
 export interface IHttpConfig extends IPrismConfig {
-  mock: boolean | IHttpOperationConfig;
+  mock?: boolean | IHttpOperationConfig;
```
chris-miaskowski: Isn't `undefined` making the prism (the framework) unpredictable?

packages/http/src/index.ts
```diff
@@ -20,9 +20,7 @@ import { validator } from './validator';
 
 const createInstance = <LoaderInput>(overrides?: TPrismHttpComponents<LoaderInput>) => {
   return factory<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig, LoaderInput>({
-    config: {
-      mock: true,
-    },
+    config: {},
```
chris-miaskowski: I don't know why it's broken, but look at line 29 - `overrides` should override this configuration.
E.g. if someone passes their own `mock` in the CLI they should be able to change the settings **globally**.
I'm very surprised this doesn't work. Do we have unit/int tests confirming that?

StefanDywersant:
The actual behaviour was sth like that:
default = { mock: false }
cli = { mock: true }
req = not set in req so defaulted to: { mock: false }

The result after merging is: { mock: false } even though mocking was not explicitly set on req.

chris-miaskowski:
The idea was that:
1. `prism.createInstance` is created with a global `mock: false` by default
2. one can override the global setting with `prism.createInstance({mock: true})
3. one can temporarily override the settings in the request (__mock=true|false)

chris-miaskowski:
> The actual behaviour was sth like that:
> default = { mock: false }
> cli = { mock: true }
> req = not set in req so defaulted to: { mock: false }

We actually do not support setting __mock in the request.

> 
> The result after merging is: { mock: false } even though mocking was not explicitly set on req.

So it seems like a problem with merging. Did you try investigating that?

@StefanDywersant

chris-miaskowski:
Let's sum it up:

- be default we set the mock = true => all requests are mocked using the default `IHttpOperationConfig` setting
- on instantiation we should be able to override that default config to `false`, `true` or any other `IHttpOperationConfig` setting
- only if mock: false => all requests are forwarded
- on demand we should be able to enforce specific `mock` configuration: `code, mediaType, example, dynamic` (**note**: if we globally set the mock to `true`, **there is currently no way** to enforce `mock:false` on demand)
- if we instantiate prism with `mock: false` **there is currently no way** to set it to default `mock: true` configuration. However, it should be possible to enforce a specific mock configuration on demand (with __code etc)

packages/http-server/src/__tests__/server.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('server', () => {
   let server: IPrismHttpServer<any>;
 
   beforeAll(async () => {
-    server = createServer({}, {});
+    server = createServer({}, { components: { config: { mock: true } } });
```
chris-miaskowski: IHttpOperationConfig

packages/http/src/types.ts
philsturgeon: I agree, this changes it from a boolean/object to three-state-boolean/object and that's pretty complex.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -9,7 +9,7 @@ describe('Http Prism Instance function tests', () => {
   let prism: IPrism<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig, { path: string }>;
 
   beforeAll(async () => {
-    prism = createInstance();
+    prism = createInstance({ config: { mock: true } });
```
philsturgeon: What sort of instance do we have that is not a mock? The pass through proxy stuff that may or may not be implemented yet? 

Maybe we could have a createMockInstance() and createProxyInstance() to avoid some of this nested config object stuff?

packages/http-server/src/__tests__/__snapshots__/getHttpConfigFromRequest.spec.ts.snap
```diff
@@ -34,17 +34,9 @@ Object {
...
-}
-`;
+exports[`getHttpConfigFromRequest() given no default config and no matching query should return my own default 1`] = `Object {}`;
 
-exports[`getHttpConfigFromRequest() given no default config and no query should return my own default 1`] = `
-Object {
-  "mock": true,
-}
-`;
+exports[`getHttpConfigFromRequest() given no default config and no query should return my own default 1`] = `Object {}`;
```
XVincentX: Yeah I really think we should get rid of these snapshot here.

packages/http/src/types.ts
XVincentX: I'm also against this; anything is either something or is not. I struggle to imagine what `undefined` means in this context.

StefanDywersant: Actually undefined has a strong foundations.

The algorithm merges more or less those:
1. default config
2. components config
3. cli config
4. http request config

So if we did set mock: false in 3) and did set nothing in 4) then logically 4)'s config should have mock undefined (meaning we don't want to override state).

But I you were confused by that change that means that sth is wrong, I believe I should introduce Partial<Config> in some places to emphasize that these are partials.

StefanDywersant:
Duplicated by https://github.com/stoplightio/prism/pull/204 which has been merged already. Closing.

PR #204, opened 2019-03-28
Fix/sl 2191/separate config concept

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2191

Moved config out of components and made it partial.

The reasoning is that config logically should be completely optional. We cannot at every step (we have 4!) always decide what should be a value of parameters. However we need complete config at the end, when it is consumed by prism. This is now reflected in types. Most of configs are made partial and only the initial one (hardcoded inside prism) has all values required.

The original problem was caused by getHttpConfigFromRequest. When user did not forced __mock=true|false through query parameters the function was returning some default mock value anyway because it was enforced by types. No when user does not set __mock on request the config does not contain `mock` property -> so it is not overwriting values set in previous steps (e.g. via -m cli param).


philsturgeon:
My brain is stuck in Spectral today, I can't think. I have no immediate concerns with the code, but I defer to @XVincentX for this review.

packages/core/src/factory.ts
```diff
@@ -1,12 +1,15 @@
-import { configMergerFactory } from '.';
+import { configMergerFactory, PartialPrismConfig, PrismConfig } from '.';
 import { IPrism, IPrismComponents, IPrismConfig, IValidation } from './types';
 
 export function factory<Resource, Input, Output, Config, LoadOpts>(
+  defaultConfig: PrismConfig<Config, Input>,
```
chris-miaskowski: How about the opposite way? We could avoid passing `undefined` as first parameter.

```suggestion
  defaultComponents: Partial<IPrismComponents<Resource, Input, Output, Config, LoadOpts>>
  defaultConfig?: PrismConfig<Config, Input>,
```

packages/core/src/utils/resolveConfig.ts
```diff
@@ -1,17 +1,19 @@
...
 }
 
 export function resolveConfig<Config, Input>(
   input: Input,
-  config: PrismConfig<Config, Input>,
-  defaultConfig?: PrismConfig<Config, Input>
-): Config {
+  config: PartialPrismConfig<Config, Input>,
+  defaultConfig?: PartialPrismConfig<Config, Input>
+): Partial<Config> {
```
chris-miaskowski: Would 
```suggestion
): PartialPrismConfig<Config, Input> {
```
work?

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -41,7 +45,7 @@ export const getHttpConfigFromRequest: PrismConfigFactory<IHttpConfig, IHttpRequ
     if (typeof config.mock === 'boolean') {
       config.mock = httpOperationConfig;
     } else {
-      config.mock = Object.assign(config.mock, httpOperationConfig);
+      config.mock = Object.assign({}, config.mock, httpOperationConfig);
```
chris-miaskowski: :) @XVincentX removed this ~week ago

chris-miaskowski: I will leave it you to you guys ;)

StefanDywersant: Since `config.mock` may be undefined (which is basically concept of this PR) the above is needed

packages/core/src/utils/resolveConfig.ts
StefanDywersant: Nah, the purpose of resolveConfig is to return actual config. *PrismConfig is either actual config or factory.

packages/core/src/factory.ts
StefanDywersant: defaultConfig is mandatory and it is first arg. I don't see the point.

packages/core/src/__tests__/factory.spec.ts
```diff
@@ -8,7 +8,7 @@ describe('graph', () => {
     let defaultLoaderId = 0;
     let customLoaderId = 0;
 
-    const createInstance = factory<any, any, any, any, { id: number }>({
+    const createInstance = factory<any, any, any, any, { id: number }>(undefined, {
```
chris-miaskowski: Why do you have to pass `undefined` as first parameter?

StefanDywersant: Because of a hack someone did in order to simplify tests. See that bunch on `any` ?

PR #205, opened 2019-03-28
Fix/sl 2192/stringify examples

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2192

The cause of freeze was that fastify received object as a response body instead of string. This PR enforces response body to be stringified.


StefanDywersant:
> I do not think we need a new whole fixture; the functional snapshot changes look enough to me.

- it’s an fixture with static examples which caused actual problem
- I think it may be handy to have two different fixtures - pets and todos
- having that I could prepare a test which fails when bug exists and passes when it's fixed

StefanDywersant:
> This seems way more reasonable! Thanks!

You right :) Thanks!

PR #219, opened 2019-04-02
feat: add tsconfig paths to make the CLI work natively in TS SL-2369

XVincentX:
This PR hopefully completes the "make the CLI usable again" by using the tspath-register package that will correctly rewrite the paths so that ts-node (used by Oclif) will correctly require the internal workspace packages.

I'm in touch with the heroku Oclif team to make this happen automagically in the package.

This should help the developer testing the commands (and in general work with the codebase) without having to compile anything and stay in Typescript as much has possible.

---

Please do not forget the local development experience. It is important.


packages/cli/src/commands/serve.ts
```diff
@@ -13,6 +13,7 @@ export default class Serve extends Command {
     spec: oflags.string({
```
philsturgeon: Do we want to have required options? Is this one maybe something that should just be an argument? 

If you are trying to serve a think it seems like you are gonna definitely give it a spec file? Are there any other required options?

philsturgeon:
Nothing jumps out as a problem, but I don't know anything about tsconfig or any of that stuff. Maybe we need somebody else who can review it. Still learning over here.

packages/core/src/loaders/filesystem/__tests__/filesystemLoader.unit.ts
```diff
@@ -21,7 +21,7 @@ describe('filesystemLoader', () => {
 
   test('given no opts should delegate to graph facade for http operations', async () => {
     jest.spyOn(fs, 'existsSync').mockReturnValueOnce(true);
-    jest.spyOn(fs, 'statSync').mockReturnValueOnce({ isDirectory: () => false });
+    jest.spyOn(fs, 'statSync').mockReturnValueOnce({ isDirectory: () => false } as fs.Stats);
```
XVincentX: This needs to disappear but my mind is gone right now and I can't think of any possible solution. It's clear that we're mocking too much stuff and the API for this FileSystem Reader should probably be changed.

packages/core/package.json
```diff
@@ -16,12 +16,13 @@
...
-    "@stoplight/graphite": "4.6.x",
+    "@stoplight/graphite": "5.0.0",
     "axios": "0.x.x",
     "lodash": "4.x.x",
+    "mobx": "^5.9.4",
     "tslib": "1.9.x"
   },
   "devDependencies": {
-    "@types/lodash": "4.x.x"
+    "@types/lodash": "4.14.123"
```
philsturgeon: Do we want to be pugging to specific versions like this? I guess dependabot will shout at us anyway, but if we did `    "@types/lodash": "^4.14.123"` with the little ^ up there then it says anything above 4.14.123 and below 4.15. Is that better?

packages/core/src/loaders/http/httpLoader.ts
```diff
@@ -16,7 +16,7 @@ export class HttpLoader {
 
     await this.graphFacade.createRawNode(response.data, {
       type: FilesystemNodeType.File,
-      subtype: trimStart(extname(opts.url), '.'),
+      language: trimStart(extname(opts.url), '.'),
```
chris-miaskowski: In Graphite 5 language should be mapped to one of the `Language` enum values.

This is how it's done in Graphite:

```
const extensionToLanguage = {
  md: Languages.Markdown,
  yml: Languages.Yaml,
  js: Languages.JavaScript,
};
```

`extensionToLanguage[ext] || ext`

I suggest to keep a similar map here for the time being. We can centralize it in Graphite later.

packages/http/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@stoplight/prism-core": "0.0.0",
-    "ajv": "6.6.x",
+    "ajv": "6.10.0",
```
chris-miaskowski: why not `6.10.x`?

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
```diff
@@ -7,20 +7,21 @@ describe('HttpForwarder', () => {
   const forwarder = new HttpForwarder();
 
   beforeEach(() => {
-    jest.spyOn(axios, 'default').mockImplementation(jest.fn);
+    jest.spyOn(axios, 'default');
```
chris-miaskowski: The intention or the `mockImplementation` was to actually stub that function not to spy and pass through it. Why have you decided to change this?

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
```diff
@@ -7,20 +7,21 @@ describe('HttpForwarder', () => {
...
   });
 
   describe('forward()', () => {
     describe("parameters haven' been provided", () => {
       it('proxies request correctly', async () => {
-        jest.spyOn(axios, 'default').mockImplementation(() => ({
+        jest.spyOn(axios, 'default').mockResolvedValue({
```
chris-miaskowski: Is that the test that should return promise but returned actual value that you mentioned in Slack?
Looking at the "expect" I'm not surprised it passed... It's not really resting the result, just the fact that it calls `default`

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -11,9 +11,9 @@ describe('HttpMocker', () => {
 
   beforeEach(() => {
     httpMocker = new HttpMocker(mockExampleGenerator);
-    jest.spyOn(helpers, 'negotiateOptionsForValidRequest').mockImplementation(jest.fn);
-    jest.spyOn(helpers, 'negotiateOptionsForInvalidRequest').mockImplementation(jest.fn);
-    jest.spyOn(mockExampleGenerator, 'generate').mockImplementation(jest.fn);
+    jest.spyOn(helpers, 'negotiateOptionsForValidRequest');
```
chris-miaskowski: Similar issue: AFAIK spyOn will not mock the implementation but use the actual implementation. What was the reason you changed this?

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -217,20 +205,15 @@ describe('NegotiatorHelpers', () => {
   });
 
   describe('negotiateOptionsBySpecificCode()', () => {
-    let negotiateOptionsBySpecificResponseMock: jest.Mock;
-    let negotiateOptionsForDefaultCodeMock: jest.Mock;
+    let negotiateOptionsBySpecificResponseMock: jest.SpyInstance;
```
chris-miaskowski: :+1:

packages/core/package.json
chris-miaskowski: @philsturgeon the versioning "policy" is that we should use the `1.2.x` pattern (but avoid the `^`). Not sure why @XVincentX didn't use `x`. Probably an oversight.

packages/core/src/loaders/filesystem/__tests__/filesystemLoader.unit.ts
chris-miaskowski: @XVincentX it's a unit test and it should mock all dependencies, why do you think it's mocking too much?

packages/core/package.json
XVincentX: I just let `yarn` set the thing in the way it wants, but if I recall correctly yesterday @P0lip was doing a similar change to have a smaller `node_modules` directory. I can change it if required, I do not mind.

packages/core/src/loaders/filesystem/__tests__/filesystemLoader.unit.ts
XVincentX: You're using a single property from an entire gigantic object; this flag could be passed somehow as a parameter so we can mock that thing in particular and not the whole `statSync` returned object.

packages/http/package.json
XVincentX: `yarn` sets this automatically, but I'm ok with changing it
```suggestion
    "ajv": "6.10.x",
```

packages/core/package.json
P0lip: Dev dependecies are totally irrelevant since they are trimmed out during the build process, so it can stay as is.

And yeah, I wanted to make the versioning stricter.
We usually use either `^` or `1.x.x` for regular dependencies, but in my case I wanted to make it less relaxed.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
P0lip: Yeah, it won't mock the method by default, unless you instrument it do so by calling methods such as `mockImplementation` or `mockReturnValue` etc. etc.

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
XVincentX: 1. If your intention is only to check if a function was called, you need a `spy` and not a `mock`. I understand that touching the network is a PITA; the function gets mocked anyway below, so there's no reason to mock it in a before each (which can actually be completely removed)

2. This code was wrong anyway — the correct call should have been
```diff
- jest.spyOn(axios, 'default').mockImplementation(jest.fn);
+ jest.spyOn(axios, 'default').mockImplementation(jest.fn());
```

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
XVincentX: It's still incorrect though, right?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: 1. As mentioned above, the code is incorrect
2. Why mock something that just works?

packages/core/package.json
XVincentX: Done in 32fb515

packages/core/src/loaders/http/httpLoader.ts
philsturgeon: Dont forget to support yml and yaml :)

packages/core/src/loaders/filesystem/__tests__/filesystemLoader.unit.ts
chris-miaskowski: :+1: makes sense

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
chris-miaskowski: > 1. If your intention is only to check if a function was called, you need a `spy` and not a `mock`. I understand that touching the network is a PITA; the function gets mocked anyway below, so there's no reason to mock it in a before each (which can actually be completely removed)

The reason it was in the before each is to avoid problems with new tests (e.g. someone adds a test and don't realize they use an actual network). Better be safe than sorry ;)

> 2. This code was wrong anyway — the correct call should have been
> 
> ```diff
> - jest.spyOn(axios, 'default').mockImplementation(jest.fn);
> + jest.spyOn(axios, 'default').mockImplementation(jest.fn());
> ```

:+1:

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
chris-miaskowski: @XVincentX because it's a unit test. If there is one thing that I will strongly hold to is that unit test should **always mock/stub all** dependencies.

If `negotiateOptionsForValidRequest` stops working I don't want to see 100 tests files failing, but only the `helpers.spec.ts`. So unless it's an integration test (it's not in this case ;)) I don't see a reason to not mock that function.

chris-miaskowski:
@XVincentX note that I'm working on a change in Graphite 5.1.0 that will help you with the `Language` thingy.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: > If negotiateOptionsForValidRequest stops working I don't want to see 100 tests files failing, but only the helpers.spec.ts

I disagree but I have no forces and time right now to fight, so I'll make the change. To keep it short though tests should be designed to fail, not to pass. If a test passes every time and never fail, it tells me _nothing_. It should fail when *both* the dependencies *OR* the behaviour changes *OR* the dependency protocol changes.

XVincentX: Restored in f8d018a

packages/core/src/loaders/http/httpLoader.ts
XVincentX: I'll address this once graphite 5.1 is out

chris-miaskowski:
@XVincentX I have just found that `Language` is not properly exported in Graphite (is `const enum` should be `enum`) and I release a fix soon. Shouldn't block you.

chris-miaskowski:
Good to go @XVincentX! One day we'll have that tests discussion, one day! ;)

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
philsturgeon: I am currently very blocked on a hotfix because of the incredibly heavy amount of mocking involved in these very files. Unless the dependency is hitting the wire, the database or the filesystem (or something else similarly outside the control of the test suite), there is no reason to mock it. If you do not own it, definitely do not mock it. Mocking dependencies generally leads to code that does not work thinking it works really well. 

Please do not continue with the "mock all the things" approach or we will have a very rough time. If a test code is hitting so many dependencies that it is slow, it probably means our code is too tightly coupled, and shout be refactored.

philsturgeon:
Are there any JS tools which can sniff for network requests and error? Ruby uses something called VCR for recording HTTP snapshots (a lot like jest snapshots but for HTTP only) and that has the benefit of blocking all HTTP traffic outbound unless specifically whitelisted. Is this a thing? 

--
Phil Sturgeon
@philsturgeon

PR #226, opened 2019-04-02
Invalid Request - Missing Params

philsturgeon:
Added a bunch more integration tests, and used a failing sample specification as a fixture for a few more. 

This PR also moves a lot of snapshots to inline, and tries to remove them altogether in other instances. All done in an attempt to improve readability, and effectiveness of our test suite.

Hidden in here is also a change that will return a JSON 500 error if the incoming request is missing a required header, query param, or body param. 

_We are going to have a biiiig old chat about if we should be doing "500 Prism is panicking because we dunno what response to give" or "Hey we saw that you like 400s so we put a 400 in your response" or something else. Will update this PR with the outcome of that chat._

### DIFF
--- a/packages/http/src/__tests__/__snapshots__/http-prism-instance.spec.ts.snap
+++ b/packages/http/src/__tests__/__snapshots__/http-prism-instance.spec.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Http Prism Instance function tests returns correct response for an existing route 1`] = `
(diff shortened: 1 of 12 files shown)


packages/http/src/__tests__/__snapshots__/http-prism-instance.spec.ts.snap
```diff
@@ -26,39 +26,3 @@ Object {
   },
 }
 `;
-
```
philsturgeon: Removing reliance on snapshots for some of the most important tests because it is hard to read them.

XVincentX:
@philsturgeon I think I'm going to close this and take over.

PR #236, opened 2019-04-08
Pre-release automation

XVincentX:


philsturgeon:
So this will look at develop and do... what? How does it know if it should release a beta or a final.

XVincentX:
@philsturgeon Right now this is a WIP — the idea is to release every PR merged in develop in a NPM package on the `beta` tag so that basically nobody will be able to install it (yet) — but it's going to be important because of the integrations we want to do.

>  How does it know if it should release a beta or a final.

It does not. Right now it will always release on "beta" when something goes to develop. Once we're ready to ship this for real, I'll remove this limitation.

P.S: semantic-release 16 will effectively support multi branch releases.

philsturgeon:
Awesome! Totally understand the why, was just unsure on the how, and yep this won't last forever I'm with you. Trying to learn more about semantic release bot in general.

PR #237, opened 2019-04-08
Deployment script primer SS-31

XVincentX:
1. Removed examples
2. Set up Lerna publish with no git tags
3. Added licenses in all the projects
4. Regenerated the CLI Readme

Requires #239 merged


XVincentX:
P.S: Merging this will trigger the release process. Good luck.

philsturgeon:
This is going to conflict with #243, which is about to add some command tests. We should get that one into develop first, I think.

package.json
```diff
@@ -1,14 +1,12 @@
 {
   "private": true,
   "name": "prism",
-  "license": "UNLICENSED",
-  "version": "0.0.0",
+  "license": "Apache-2.0",
```
chris-miaskowski: Why Apache-2.0? Have that been confirmed with @marbemac / @brianmrock ?
I know little about licenses, so maybe that's good but I remember Marc or Brian said that we'll need to be careful with this.

packages/cli/src/commands/serve.ts
```diff
@@ -20,6 +20,7 @@ export default class Serve extends Command {
       char: 'm',
       description: 'Turn global mocking on or off',
       default: true,
+      required: true,
```
chris-miaskowski: I don't like mixing stuff in a PR very much. I would limit this PR to packaging only - this looks like a feature.

packages/http/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@stoplight/prism-http",
-  "version": "0.0.0",
+  "version": "3.0.0-pre",
```
chris-miaskowski: Hmm, do we need to start from v3?
I mean `@stoplight/prism-http` has never been release so it should start from 1 right?
IIRC lerna with release ``prism-http-1.0.0` anyway so should be no collision with existing tags (correct?)

This obviously applies to all packages.

packages/http/package.json
```diff
@@ -16,7 +16,7 @@
     "node": ">=8"
   },
   "dependencies": {
-    "@stoplight/prism-core": "0.0.0",
+    "@stoplight/prism-core": "3.0.0-pre",
```
chris-miaskowski: Hmm, curious how this will work.
Let's say you publish this, am I right that:

`lerna` will look at the remote tags, will find that latest tag is `2.x.y` and will bump to `2.x+1.y`?

OR

will `lerna` look at the `package.json#version` and bump to `3.0.0-pre.1` (or similar) then tag and release `3.0.0-pre.1`? But at the same time `@stoplight/prism-core: 3.0.0-pre` will still remain unpublished thus this code will not build?

package.json
XVincentX: I've asked @philsturgeon — and he mandated to use the same in Spectral, which is Apache 2.

BTW in case you're curious about the differences: https://tldrlegal.com

This was super beneficial to me

packages/cli/src/commands/serve.ts
XVincentX: Yeah it is, it was to make sure the argument was correctly displayed in the CLI; in any case this is ditched in #243 so when rebasing this will fade away :)

packages/http/package.json
XVincentX: Ah, this is a good point. Technically we have never published this on npm so it's effectively a 1.0 — but on the other hand it's branded as "Prism 3.0"; 

@philsturgeon how do we want to go with the versioning? 1.x or 3.x?

I *think* we should go with 3.0; we can theoretically use the 2.x tags on npm to download the old prims binary so that we offer a "streamlined" experience, but I'm open to alternatives.

package.json
chris-miaskowski: @XVincentX good link definitely clears up a lot, thanks! Still I don't have opinion, I will let someone else approve that part.

packages/cli/src/commands/serve.ts
chris-miaskowski: :+1:

packages/http/package.json
chris-miaskowski: Ok, a crazy idea, maybe 3.0 for prism-cli and 1.0 for the framework stuff? 

Meh :p 

Or just go with 3 :)

XVincentX: In theory it's possible to have different versions but that might complicate the things a little bit; unless it's an hard requirement, I'd stick with the current 3.x stack (or start again with 1.x)

chris-miaskowski: Yup stick to 3.x, let's not be purists.

packages/http/package.json
XVincentX: Good question. Let's break down the current publish command

`yarn lerna publish prerelease --yes --dist-tag beta --conventional-commits`

* `prerelease`: add the `-alpha` to the final version
* `--yes`: do not ask for any confirmation
* `--dist-tag beta`: when publish on NPM, put it on the beta tag (I should change it to alpha).
* `--conventional-commits`: use angular style commits to determine the next version (the same we're using in semantic-release)


Let's make an example of what's going to happen right now:

lerna will parse the current version in package.json (3.0.0-pre -> 3.0.0), realise that it's a prerelease and so add the `-alpha` tag and then make a minor bump. Final result: `3.1.0-alpha`

Since the `dist-tag` is `beta`, the package will be installable only when specifying it explicitly

`npm i @stoplight/prism-cli`  -> No result
`npm i @stoplight/prism-cli@beta` -> OK

This will make sure that nobody installs and uses the software until we feel comfortable.

Does that clarify?

chris-miaskowski: That helps a lot, still have some questions :)

> lerna will parse the current version in package.json (3.0.0-pre -> 3.0.0), realise that it's a prerelease and so add the -alpha tag and then make a minor bump. Final result: 3.1.0-alpha

- Will lerna update package.json#version? I think workspaces rely on that value for local development, right?
- Why even `3.1.0`? Shouldn't we at least release `3.0.0` first? Isn't lerna simply adding `-alpha` w/o bumping minor if it notices it's a prerelease?

> Since the dist-tag is beta, the package will be installable only when specifying it explicitly

And when would a `stable` release be triggered? Do you plan to add it too?

> npm i @stoplight/prism-cli@beta -> OK

Still, even though `@stoplight/prism-core` is now released as `3.1.0-alpha`, @stoplight/prism-http` points to `@stoplight/prism-core@3.0.0-pre` right? If I `yarn add @stoplight/prism-http` it won't be able to find `@stoplight/prism-core@3.0.0-pre` and will break.

XVincentX: > Will lerna update package.json#version?

Yes it does for the package itself and for all the dependants packages it's indeed required and we can't escape that.

> Isn't lerna simply adding -alpha w/o bumping minor if it notices it's a prerelease?

You're actually right, I wrote the version incorrectly in the comment

image


> And when would a `stable` release be triggered? Do you plan to add it too?

The default tag is `latest`, and that's the one we're going to release when the stuff gets in `master`, but this is definitely something that'll be coming next.

> Still, even though @stoplight/prism-core is now released as 3.1.0-alpha, @stoplight/prism-http points to @stoplight/prism-core@3.0.0-pre right? If I yarn add @stoplight/prism-http@beta it won't be able to find @stoplight/prism-core@3.0.0-pre and will break.

I'm not too sure I'm following here; all the versions are updated in the relative package.json files before they are published, no mismatch should be happening.

Moreover, you can test the changes locally on your computer before publishing by doing 

`yarn lerna publish prerelease --dist-tag beta --conventional-commits`

notice the missing yes parameter:

```diff
- yarn lerna publish prerelease --yes --dist-tag beta --conventional-commits
+ yarn lerna publish prerelease --dist-tag beta --conventional-commits
```

and then answering no — you should be able to have a picture of what's going on.

chris-miaskowski: Ok, all sounds good!

.circleci/config.yml
```diff
@@ -36,9 +36,30 @@ jobs:
...
+      - checkout
+      - run:
+          name: Install Dependencies
+          command: yarn
+      - run:
+          name: Build all code to JavaScript
+          command: yarn build
+      - run:
+          name: Publish
+          command: yarn lerna publish prerelease --yes --dist-tag beta --conventional-commits
```
chris-miaskowski: you wanted to change `beta` to `alpha` iirc

XVincentX: Done in bf05503

chris-miaskowski:
@XVincentX :crossed_fingers: 

Run the license stuff through someone pls when you get a chance.

PR #240, opened 2019-04-09
fix: SO-80 default to empty body, match even if no servers

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SO-80

Changes:
- match paths even if no servers provided (making servers truly optional)
- return empty text/plain response if no example/schema found (making examples in response truly optional)


packages/http/src/router/index.ts
```diff
@@ -23,26 +22,25 @@ export const router: IRouter<IHttpOperation, IHttpRequest, IHttpConfig> = {
...
-      const { servers = [] } = resource;
-
-      if (servers.length) noServerProvided = false;
       if (pathMatch !== MatchType.NOMATCH) nonePathMatched = false;
 
+      const { servers = [] } = resource;
       let serverMatch: MatchType | null = null;
 
-      if (!ignoreServers) {
+      if (!ignoreServers && servers.length > 0) {
```
chris-miaskowski: What if we **don't** want to ignore servers AND no servers is found?

You are missing a logical branch here.

```
if(!ignoreServers) {
  if (servers.length > 0) {
  } else {
    // this is missing
  }
}

Comment (chris-miaskowski) [reply packages/http/src/router/index.ts]: In case described below we should still throw the error you have removed

File: packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts [hunk 2]
```diff
@@ -536,11 +536,14 @@ describe('NegotiatorHelpers', () => {
         headers: [],
       };
 
-      expect(() => {
-        helpers.negotiateDefaultMediaType(partialOptions, response);
-      }).toThrow(
-        'Could not generate response for provided content type or no content type provided. Tried to fallback to application/json, but no definition found.'
-      );
+      expect(helpers.negotiateDefaultMediaType(partialOptions, response)).toEqual({
```
chris-miaskowski: Please read my other comment first about this change and if we decide to push this forward please add an integration test.

For all future commits I'm thinking about adding this as a mandatory step. WDYT @StefanDywersant @philsturgeon @XVincentX ? We have missed too many bugs because we didn't have them.

packages/http/src/router/__tests__/index.spec.ts
```diff
@@ -42,31 +41,14 @@ describe('http router', () => {
     });
 
     describe('given a resource', () => {
-      test('should not match if no server defined', () => {
+      test('should match even if no server defined', () => {
```
chris-miaskowski: Please add an integration test.

packages/http/src/router/index.ts
StefanDywersant: All branches are covered here. The code you wrote is not equivalent of the above.

`if (A && B) 1; else 2;`  is a shorthand of  `if A { if B 1; else 2; } else { if B 2; else 2; }`

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -125,11 +125,17 @@ const helpers = {
...
+    } else {
+      // no httpContent found, returning empty body
+      return {
+        code,
+        mediaType: 'text/plain',
+        example: {
+          externalValue: '',
+          key: 'default',
+        },
+      };
```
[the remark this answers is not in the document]
StefanDywersant: First of all, the key 'default' is ignored in further processing (it's just forced by type).

Questions:
1) there is a content but it's not found by negotiation and not `application/json`
a) return any content (pick first)
b) return default (text/plain '')
c) throw error

2) there is no content at all
a) return default (text/plain '')
b) throw error

For me, for both it feels more natural to just throw errors (the way it was implemented).

However from user perspective it might be more handy to return empty text/plain. If so, I'd return it for both cases to keep consistent (like it's implemented now). User can start the server faster with less effort and problems.

@philsturgeon - curious your opinion

packages/http/src/router/index.ts
chris-miaskowski: Ok, I see but I'm saying we need the following logic (I overdid it on purpose):

```
if(!ignoreServers) {
  if (servers.length > 0) {
    // ...
    noServerProvided = false;
    noneServerMatched = false;
  } else {
    // Next line is important. We have explicitly asked to match servers but no server was provided. This code removes this functionality.
    noServerProvided = true;
    noneServerMatched = true;
  }
} else {
   noServerProvided = false;
   noneServerMatched = false;
}
```

How do you want to achieve the same logic with two branches?

```
if(!ignoreServers && servers.length > 0) {
   // ...
} else {
   noneServerMatched = false;
}
```

StefanDywersant: This is the point of this change. If none server provided it should still work (see the jira issue). That is why I removed noServerProvided errors. That's not relevant anymore.

Btw. see that line: https://github.com/stoplightio/prism/pull/240/files#diff-f9a10b37616fb5669ecd5218fc8535c9R27

chris-miaskowski: > This is the point of this change. If none server provided it should still work (see the jira issue). That is why I removed noServerProvided errors. That's not relevant anymore.
> 
> Btw. see that line: https://github.com/stoplightio/prism/pull/240/files#diff-f9a10b37616fb5669ecd5218fc8535c9R27

If I understand it correctly the purpose of this task is to provide response **if there is no servers (servers = []) AND if we did not ask to validate servers (ignoreServers=true)**.

This is different from the case I'm making. I'm telling **I explicitly ask you to validate server (ignoreServers=false) AND there is no servers provided (servers = [])**.

The two are different.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
philsturgeon: I am not sure what the intention of the example, or externalValue are doing here. 

When there is literally no response defined in OpenAPI, we still need to respond:

```
openapi: 3.0.1
info:
  title: API
  version: 1.0.0
paths:
  /:
    get:
      responses:
        '200':
          description: OK
```

I was not really trying to suggest we change how defaults work, as I am not sure how the specific code should be written. The feature request was just to make sure this example responds without failing.

Our mime type logic does seem a bit funny. We should be looking for one that matches the `Accept` header, if that is not there we should be looking for `default`, if that is not there we should _technically_ error. 

However that is probably not very useful for most of our users, who may well expect it to use JSON by default (as so many people use JSON without ever even checking Accept). So, having application/json as another check cant hurt. If you put Accept: application/xml it should certainly return you XML.

But when you define nothing in responses, you should get nothing. That's what I'd expect. "Nothing" is defined as text/plan with an empty string.

packages/http/src/router/index.ts
philsturgeon: Definitely very confusing. 

```
 const ignoreServers: boolean = requestBaseUrl === undefined;
```

I thought we completely ignored base paths, and I'm not sure why we are turning ignoreServers on or off based on wether or not a base path exists? Base path is not even a concept in OpenAPI v3.

Do we have server validation? Like, if i pass in ?__server=foo.com it should error if foo.com does not exist in the spec, right? 

- if (user specified a server - server validation is on)
  - ❌ no servers exist, so fail (cant validate against nothing)
  - ❌ servers exist and this is not one of them
  - ✅ servers exist and this is one of them
- else
  - there is no else nobody cares about servers yaaaaay

Am I missing something?

philsturgeon: Also `noneServerMatched` is a really confusing name for a variable. negative variable names for booleans lead to double negatives that are hard to understand. 

can we have positive variable names? 

- `serverValidationEnabled`
- `serverFound`

that sort of thing.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
pytlesk4: > I am not sure what the intention of the example, or externalValue are doing here.
> 
> When there is literally no response defined in OpenAPI, we still need to respond:
> 
> ```
> openapi: 3.0.1
> info:
>   title: API
>   version: 1.0.0
> paths:
>   /:
>     get:
>       responses:
>         '200':
>           description: OK
> ```
> 
> I was not really trying to suggest we change how defaults work, as I am not sure how the specific code should be written. The feature request was just to make sure this example responds without failing.
> 
> Our mime type logic does seem a bit funny. We should be looking for one that matches the `Accept` header, if that is not there we should be looking for `default`, if that is not there we should _technically_ error.
> 
> However that is probably not very useful for most of our users, who may well expect it to use JSON by default (as so many people use JSON without ever even checking Accept). So, having application/json as another check cant hurt. If you put Accept: application/xml it should certainly return you XML.
> 
> But when you define nothing in responses, you should get nothing. That's what I'd expect. "Nothing" is defined as text/plan with an empty string.

💯, content negotiation for the win!

packages/http/src/router/index.ts
chris-miaskowski: > Definitely very confusing.
> 
> ```
>  const ignoreServers: boolean = requestBaseUrl === undefined;
> ```
> I thought we completely ignored base paths, and I'm not sure why we are turning ignoreServers on or off based on wether or not a base path exists? Base path is not even a concept in OpenAPI v3.

OpenAPI doesn't have base path _per se_ but it has `servers`. Servers in OAS3 are pretty flexible and can be templated btw. `requestBaseUrl` is basically a flag to turn on/off server/basePath validation.


> 
> Do we have server validation? Like, if i pass in ?__server=foo.com it should error if foo.com does not exist in the spec, right?

AFAIK we do or we have partially implemented it (at the framework level). Anyway we **should** have it. 

> 
> * if (user specified a server - server validation is on)
>   
>   * x no servers exist, so fail (cant validate against nothing)
>   * x servers exist and this is not one of them
>   * white_check_mark servers exist and this is one of them
> * else
>   
>   * there is no else nobody cares about servers yaaaaay
> 
> Am I missing something?

Sounds about right!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
chris-miaskowski: Wrapping it all up:
- try to match any existing content type before we we fallback to `text/plan`
- support `Accept` header
- if no mediaType is matched generate `text/plain` ""
- make `example` optional. Returning "default" example looks like a hack and is relying on internal knowledge of the implementation detail. Tight coupling.
- if neither `example` nor `schema` is defined generate empty message.

This looks like a candidate for a separate PR and separate story (IIRC SO-80 isn't about media types anyway but thought that is a small change so didn't mention it before)

@philsturgeon @StefanDywersant sounds good?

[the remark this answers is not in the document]
StefanDywersant: > make example optional. Returning "default" example looks like a hack and is relying on internal knowledge of the implementation detail. Tight coupling.
The "default" is not an internal knowledge. It's just a label because all examples need to have label. It's not being used anywhere. I can change it to "Krzysztof" and it will work ;)

> if neither example nor schema is defined generate empty message.
Btw. that's the reason I'm using example. text/plain has no schema.

@chris-miaskowski  If I understand correctly you suggesting to do all those in separate story?

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
StefanDywersant: I'd be more cautious with integration tests. They are slow so we should do integration test for every tiny feature. Or we should establish a framework which lets use decide when to run which tests (units on every occasion, integration not).

We don't want to end with a tests running 10mins on local machine (@chris-miaskowski remember callinize server?)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
chris-miaskowski: yes

packages/http/src/router/index.ts
chris-miaskowski: @StefanDywersant this is still relevant btw ^ My first comment is still valid.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
chris-miaskowski: @StefanDywersant we can cherry pick integration tests. Right now we can't really complain on abundance of integration tests so I'd them to be added.

StefanDywersant: So, from what you are saying, now we need to add integration _and_ unit tests _always_, for every change ?

StefanDywersant: And btw. I can't add full integration test here. The same reason as with SO-82. Can't prepare jest environment to run a webserver, do a request and exit cleanly.

packages/http/src/router/index.ts
philsturgeon: Check out the README, I put a FAQ section in there: 

> **The API description defines a base path of `/api` (using OpenAPI v2 `basePath` keyword, or in
> OpenAPI v3 using a path in `servers.url`), but requests seem to fail when using it?**
> 
> Base paths are completely ignored by the Prism HTTP server, so they can be removed from the request.
> If you have a base path of `/api` and your path is defined as `hello`, then a request to
`http://localhost:4010/hello` would work, but `http://localhost:4010/api/hello` will fail.

Let me know if that is not true. 

If we do not yet have any server validation then this code is even more confusing :D

philsturgeon: > `requestBaseUrl` is basically a flag to turn on/off server/basePath validation.

Maybe that variabke could be `serverValidationEnabled` or something that starts with `is, so its a bit more clear :) I dont want to bikeshed over variable naming but this whole thing is confusing to the larger intent, and without docs we have to read the code, and I cant read this code! 😓

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
philsturgeon: awesome summary on all points @chris-miaskowski, different JIRA issues for each of them, and can we shove in some tests to make that clear its whats happening. :)

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
philsturgeon: Right, instead of approaching the test from:

1. CLI starts a web server
2. web server figures out which HTTP response to return
3. we test the web server has the right response

We skip step 1 and test that the http library is doing the right thing. This is actually when you _do_ want mocks. The tests on the CLI only need to confirm that when we put in --foo=abc that it passes on the correct things to the http library, or errors as we expect to, etc. Trying to do all the things via the CLI tests wil become a mess.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -527,7 +527,7 @@ describe('NegotiatorHelpers', () => {
       expect(actualOperationConfig).toBe(fakeOperationConfig);
     });
 
-    it('given no default contents should throw', () => {
+    it('given no default contents should return default response', () => {
```
philsturgeon: Sorry to quibble on the English here, but we are using the word "default" twice to mean two different things. 

> when no default response return text/plan with empty body

packages/http/src/router/index.ts
```diff
@@ -17,32 +17,36 @@ export const router: IRouter<IHttpOperation, IHttpRequest, IHttpConfig> = {
   route: ({ resources, input }) => {
     const matches = [];
     const { path: requestPath, baseUrl: requestBaseUrl } = input.url;
-    const ignoreServers: boolean = requestBaseUrl === undefined;
+    const serverValidationEnabled = !!requestBaseUrl;
```
philsturgeon: Excellent!

packages/http/src/router/index.ts
StefanDywersant: Summary of changes made:
- introduced `any[Method,Path,Server][Matched,Provided]` for clarity what those vars do
- fixed logic, handling `servers exist and this is not one of them` case
- updated tests

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
chris-miaskowski: > And btw. I can't add full integration test here. The same reason as with SO-82. Can't prepare jest environment to run a webserver, do a request and exit cleanly.

We already have integration tests, we call them "functional" tests. It's find if you just add them on the `http-server` layer.

> So, from what you are saying, now we need to add integration and unit tests always, for every change ?

I think it would make sense until we have a solid base of integration tests.

StefanDywersant: What does it mean "solid base of integration tests" ?

I have a feeling that if we add integration tests to every small thing we do we end with multiplication of every unit test case for every component used there. And we invest lots of time. Both together is bad ;)

Also remember that we have full coverage of those cases via unit tests. Full!

I'm opposing because I think we should have reasonable amount of tests for each kind (~100% units, <100% func, <<100% integration). In this particular case I don't see any added value of duplicating the unit test with func one. But your single next comment is enough for me to feel overvoted and just do it ;)

philsturgeon: This conversation is getting a bit off the rails. 

@chris-miaskowski Yep, we have tests called functional, e.g: packages/http/src/mocker/__tests__/functional.spec.ts

Good. These tests seem to be unit tests to me, we put a variety of things in and see what comes out. I think your definition of integration test is just something which doesnt have any mocks or stubs, so by that definition yes these are integration tests.

@StefanDywersant we definitely dont want to be repeating unit tests and integration tests for everything, correct. I think you are using a similar definition of integration tests to me, which is something which hits multiple layers in a more realistic way.

Ignore the term "integration test" for a minute, Chris is correct that we need tests to confirm the new functionality, but this should be done at the appropriate boundary. Ignoring the CLI and the HTTP server, we want to make sure the HTTP mocker is doing the right thing. 

Karol has said we cannot (for technical reasons) have tests which start at the CLI, launch a server, then see what the server is responding with under specific situations, which is fine, because that's mostly a job for unit tests. 

We do however need tests which show that things are being passed from boundary to boundary correctly. 

- Unit tests for CLI which show it is talking to the HTTP server correctly under various circumstances
- Unit tests which show HTTP server is talking to HTTP correctly under various circumstances

Then a handful of integration tests which show the CLI can spin up a working instance, quickly poke a few things over HTTP, and close it down again. I would call these integration tests, and we only need a small number of them to confirm the CLI is doing its job, because the unit tests are confirming everything else. 

Sound good?

chris-miaskowski: The problem is that we have 100% unit test coverage and Phil keeps finding bugs. There is even a story @philsturgeon created specifically to create integration tests.

We now have so little integration tests that before we reach 'solid base' there will be plenty of time to move away from this idea.

philsturgeon: Ahhh, we are agreeing with different words.

Yes our unit tests are mocking far too much which is why we keep getting errors. 

You are asking for more integration tests, I am asking for unit tests that mock way less, these two things are the same but we have slightly different definitions of the term integration test.

philsturgeon: Also 100% to more integration tests for things that are not technically awful. The "CLI spawns web server" is hard, but "less mocky unit tests or integration tests whatever you call them" for everything else is definitely something we need.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
StefanDywersant: Done

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
StefanDywersant: I totally agree with adding more functionals and more integration tests.
But I also want them to be good quality. Not just adding "on every occasion". This time the test it a total duplicate. All integration-related-stuff is tested in many ways and we are just changing the spec (which is exactly covered by unit test).

But to keep the train running, I just added the "test of contention" ;)

packages/http-server/src/__tests__/server.spec.ts
```diff
@@ -98,4 +98,17 @@ describe('server', () => {
 
     expect(response.statusCode).toBe(200);
   });
+
+  test('should return response even if there is no content defined in spec', async () => {
+    server = createServer({}, { components: {}, config: { mock: true } });
+    await server.prism.load({
+      path: relative(process.cwd(), resolve(__dirname, 'fixtures', 'no-responses.oas2.yaml')),
```
philsturgeon: btw I keep seeing this pop up and it can be simplified massively:

```
path: resolve(__dirname, 'fixtures', 'no-responses.oas2.yaml'),
```

This test looks great! I do not consider this an integration test, so maybe i dunno what words mean, but I like this test. A++

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
chris-miaskowski: @StefanDywersant can you pls create a story and move those changes to a separate branch/pr pls?
and add a reference to that new story somewhere in this PR, cool?

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
chris-miaskowski: @philsturgeon yes, integration tests to me (and I think by the definition) are tests that do not mock but are still testing code. They may seem similar to unit tests (which are 100% mocked) but they aren't.

What you are talking about is an e2e test, also known as functional tests: where you spin up the whole product and test similarly to how a user would test.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
StefanDywersant: Without above code this story is not done. It will throw error when you try a request with almost empty spec.

However I will create a new story for those things we've been talking about - Accept, content negotiation etc.

Are you ok with that?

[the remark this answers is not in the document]
chris-miaskowski: Hmm, I thought this story was about `servers` not about undefined responses? Where am I wrong? Are the two things related somehow?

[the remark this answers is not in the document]
StefanDywersant: Yes they are. There were two barricades to cross. First when there was no servers defined, second when there was no proper response (which was actual jira issue).

Like I said, I'd leave it in this PR and create another one which extends that behaviour with actually looking into Accept: and available responses. Extend, not change. Empty response will stay for one of the cases.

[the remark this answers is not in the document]
chris-miaskowski: All right! Sounds good!

chris-miaskowski:
Approved to unblock ya. Fix the test and feel free to merge.

XVincentX:
Hold on on merging, I swear I am on this literally right now.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -125,11 +125,17 @@ const helpers = {
         },
         httpContent
       );
+    } else {
+      // no httpContent found, returning empty body
+      return {
+        code,
+        mediaType: 'text/plain',
+        example: {
+          externalValue: '',
```
XVincentX: I think this should not be an `INodeExternalExample`, but a `INodeExample`, since this is not coming from an external source but rather generated here internally.

Moreover, I think it'd be more correct if the value would be `undefined`

```suggestion
          value: undefined,
```

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
XVincentX: I can cope with this — although there's another step we should remember — OAS2 defines a global [produces](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) key where there are defaults; these should be respected in case nothing is defined.

This change, though, should be done in graphite and somehow it should arrive here.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -125,11 +125,17 @@ const helpers = {
...
+          key: 'default',
+        },
+      };
     }
-    // a httpContent for default mediaType does not exist
-    throw new Error(
```
XVincentX: I'm wondering whether this should not be controllable through flags `--strict` or something.

XVincentX:
@StefanDywersant In case you want to release this on nom as well, just do

`yarn release` from the main package and then push it up!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
chris-miaskowski: @XVincentX Well yeah that's actually a very good point!

Now thinking about this a bit more I'm leaning towards a bit different solution.

This whole thing should IMO be **validation**. E.g. if I request "text/plain", but such media type cannot be produced we should return 4xx.

@philsturgeon WDYT?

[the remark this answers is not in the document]
XVincentX: The problem here is that this global property does not exist in OAS3; in that case, we're back to the original issue. 

Therefore I think what we can do is to use that property in OAS2 documents (and so Graphite shall somehow bring it here), otherwise, we'll have to agree on something (and I'm ok with `text/plain`)

packages/http/src/router/__tests__/index.spec.ts
```diff
@@ -42,31 +41,14 @@ describe('http router', () => {
...
-      });
-
-      test('should not match if no servers arrays provided', () => {
-        const method = pickOneHttpMethod();
-        const path = randomPath();
-
-        return expect(() =>
-          router.route({
-            resources: [createResource(method, path, [])],
+            resources: [expectedResource],
```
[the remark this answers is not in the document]
StefanDywersant: This test verifies whether router responds with a resource or not in certain conditions. This is not about selecting particular resource out of many.

But you are completely right - it's misleading. I changed the test so it only checks whether there was no error thrown without caring what resource has been returned.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
StefanDywersant: Good spot, INodeExample was sth I've been looking for but for unknown reason did not found it. Also returning undefined.

packages/http/src/router/__tests__/index.spec.ts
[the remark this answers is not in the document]
StefanDywersant: > As I mentioned multiple times, tests should fail by default, not pass, otherwise they do not tell us anything.

I'm bit unsure if I understand this correctly. How this particular test should look like?

packages/http-server/src/__tests__/server.spec.ts
StefanDywersant: Done

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
StefanDywersant: Hm, I'd extend a bit that idea of returning text/plain.

If we are at the point where we need to return default response, then let's try to match the response with Accept header. We could prepare several types of empty responses:
- text/plain: ''
- application/json: null
- application/xml: ''

How about that?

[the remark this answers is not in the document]
StefanDywersant: I created task for Accept header and refining the flow of picking example https://stoplightio.atlassian.net/browse/SO-98

packages/http/src/router/__tests__/index.spec.ts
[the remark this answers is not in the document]
XVincentX: Ok, the new version clarifies what's going on. Resolved!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: SO-98 created, so this is resolved.

PR #244, opened 2019-04-10
https://github.com/stoplightio/prism/pull/243 (continued)

StefanDywersant:


README.md
```diff
@@ -59,9 +59,10 @@ content-length: 98
 content-type: application/json
 
 {
-    "id":-93918115,
-    "name":"magna",
-    "tag":"pariatur"
+    "id": 13137069,
+    "name": "doggie",
+    "photoUrls": [ "ad fugiat sunt", "ea" ],
+    ...
```
StefanDywersant: @XVincentX how about that?

README.md
```diff
@@ -72,10 +73,29 @@ Responses will be mocked using realistic data that conforms to the type in the d
...
+content-length: 101
+content-type: application/json; charset=utf-8
+
+{
+    "error": "Internal Server Error",
+    "message": "Route not resolved, none path matched.",
+    "statusCode": 500
+}
+```
+
```
StefanDywersant: @chris-miaskowski are you ok with such example?

chris-miaskowski: Yup, good enough!

README.md
```diff
@@ -72,10 +73,29 @@ Responses will be mocked using realistic data that conforms to the type in the d
 This will run a proxy server with validation enabled according to given spec.
 
 ```bash
-$ prism server examples/petstore.json
+$ prism server examples/petstore.oas2.json
 > http://127.0.0.1:4010
 ```
 
+Now let's call our server with a non-existing url to see validation in action:
```
philsturgeon: Is this validation? If so, what type of validation, path validation? 

I think we might be overusing the term validation, maybe we could say "to see what happens when you specify an invalid route"

PR #249, opened 2019-04-11
fix: SL-2377 host/forwarded headers support

StefanDywersant:
Issue: https://stoplightio.atlassian.net/browse/SL-2377

The issue was caused by invalid host header. The host header stated that it is 127.0.0.1 while the target server responded with a certificate for another domain (like example.com in this case).

Solution:
- if there is host header set, override it with hostname[:port] of server which we are actually doing request
- move original Host: header to Forwarded: host=<value> (according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)


packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -32,6 +38,22 @@ export class HttpForwarder
...
+  private updateHostHeaders(baseUrl: string, headers?: IHttpNameValue) {
+    // no headers? do nothing
+    if (!headers) return headers;
+
+    // host header provided? override with actual hostname
+    if (headers.hasOwnProperty('host')) {
+      return {
+        ...headers,
+        host: urijs(baseUrl).host(),
+        forwarded: `host=${headers.host}`,
```
XVincentX: Shouldn't this be capitalised?
```suggestion
        Forwarded: `host=${headers.host}`,
```

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -32,6 +38,22 @@ export class HttpForwarder
     };
   }
 
+  private updateHostHeaders(baseUrl: string, headers?: IHttpNameValue) {
+    // no headers? do nothing
+    if (!headers) return headers;
```
XVincentX: Shouldn't we fill these headers even if there are no headers? My gut says you're doing this mostly because of Typescript, isn't it?

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -32,6 +38,22 @@ export class HttpForwarder
...
+
+    // host header provided? override with actual hostname
+    if (headers.hasOwnProperty('host')) {
+      return {
+        ...headers,
+        host: urijs(baseUrl).host(),
```
XVincentX: You can avoid installing a new library and just use builtin `URI`
```suggestion
        host: (new URL(baseUrl)).host(),
```

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
```diff
@@ -1,26 +1,25 @@
 import * as axios from 'axios';
 
-import { httpOperations, httpRequests } from '../../__tests__/fixtures';
+import { httpInputs, httpOperations, httpRequests } from '../../__tests__/fixtures';
 import { HttpForwarder } from '../HttpForwarder';
 
+jest.mock('axios', () => ({
+  default: jest.fn(async () => ({ status: 200 })),
```
XVincentX: You can use `mockResolvedValue` which is a little bit cleaner

```suggestion
  default: jest.fn().mockResolvedValue({status:200})
```

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -32,6 +38,22 @@ export class HttpForwarder
     };
   }
 
+  private updateHostHeaders(baseUrl: string, headers?: IHttpNameValue) {
+    // no headers? do nothing
+    if (!headers) return headers;
+
+    // host header provided? override with actual hostname
+    if (headers.hasOwnProperty('host')) {
```
XVincentX: I _think_ that `host` is a mandatory http header or something — basically the client can't route the request at all without it. If this code is executed, the `host` header is definitely there.

Maybe this check can be removed (but maybe Typescript will start to complain?)

packages/http/src/forwarder/HttpForwarder.ts
StefanDywersant: Nope, `Host: ` header is not necessary for webserver to process the request. I'd assume that client knows better whether Host: should be there or not. And prism should follow that decision.

packages/http/src/forwarder/HttpForwarder.ts
StefanDywersant: No necessarily - http headers are case insensitive. I made it lowercase because it looks better when constructing JS object - and just because of that.

packages/http/src/forwarder/__tests__/HttpForwarder.spec.ts
StefanDywersant: 👍

packages/http/src/forwarder/HttpForwarder.ts
StefanDywersant: That's what I was looking for, happy to remove urijs dependency :)

packages/http/src/forwarder/HttpForwarder.ts
XVincentX: I think that without the host header the client won't be able to route the request at all as it does not know the destination. It seems indeed to be mandatory for http 1.1

Quoting from [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)

> A Host header field must be sent in all HTTP/1.1 request messages. A 400 (Bad Request) status code will be sent to any HTTP/1.1 request message that lacks a Host header field or contains more than one.

I'm fine with leaving the code as it is right now, but I think in the future we can modify the `headers` parameters so that it's not optional and has at least the `HOST` header.

packages/http/src/forwarder/HttpForwarder.ts
StefanDywersant: Just checked (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host).

Host: is required for all HTTP/1.1 reqs. That means that there _are_ some requests in the world that do not set Host header ;)

Additionally I bet all webservers will respond properly when Host: is not set (e.g. apache and nginx will respond with default virtual server).

XVincentX: Right, that is a good point. `io-ts` would be really what we need here :trollface:

PR #250, opened 2019-04-12
refactor: update Prims to use official stoplight types SO-101

XVincentX:
This PR removes the types that were homemade back in the time to use the official one we ship as a package


packages/core/src/types.ts
```diff
@@ -1,26 +1,7 @@
...
- * The below few interfaces (until end comment) should be defined in the @stoplight/types repo
- */
-
-export interface IDisposable {
-  dispose: () => void;
-}
-
+import { IDiagnostic, Omit } from '@stoplight/types';
+export type INoRangeDiagnostic = Omit<IDiagnostic, 'range'>;
 export type Path = Array<string | number>;
```
P0lip: `JsonPath` is exported by `@stoplight/types`

XVincentX: This type was not even used. 😱

PR #252, opened 2019-04-12
test: add some oas2 integration tests SO-102

XVincentX:
- Consolidates and adds some tests for OAS2 (actually what's in the file currently seems to be good enough)
- Adds a skipped test that we need to figure out
- Polishes the Ajv code so that we get typings back
- Fixes the jest configuration so tests do not get ran twice
- Moves the tests on Node 10


packages/http-server/src/__tests__/server.oas2.spec.ts
```diff
@@ -54,6 +73,24 @@ describe('server', () => {
...
+  test('should not mock a request that is missing the required query parameters with no default', async () => {
+    const response = await server.fastify.inject({
+      method: 'GET',
+      url: '/pet/findByTags',
+    });
+
+    expect(response.statusCode).toBe(400);
+  });
+
+  test.skip('should automagically provide the parameters when not provided in the query string and a default is defined', async () => {
```
XVincentX: @philsturgeon @chris-miaskowski 

Right now Prism does not use the `default` parameter value in case it's defined in the OAS document, so this test will fail. We need to decide whether this is something we want to support or not.

Moreover, it does not check for the `collectionFormat` property in the parameter, so that's another validation that's currently not happening.

philsturgeon: Yes when a specific code is requested and that code is not defined we can look for a body in `default`. @chris-miaskowski and I both got to agreeing on that.

philsturgeon: Can you make another JIRA issue for collectionFormat?

XVincentX: SO-105 created.

PR #253, opened 2019-04-15
OAS3 integration tests and fixes SO-103

XVincentX:
- Adds some integration tests on OAS3 documents
- Changes the response negotiation to consider the presence of a default response (which gets overridden with the requested status code)
- Changes the media type negotiation to not throw in case there's no content for that (return empty instead)
- Uses the `noEmit` flag for typechecks in CI so checks time halves.


examples/petstore.oas3.json
```diff
@@ -176,12 +240,121 @@
...
+        "summary": "List all pets",
+        "operationId": "listPets",
+        "tags": [
+          "pets"
+        ],
+        "parameters": [
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "How many items to return at one time (max 100)",
```
philsturgeon: Can we trim these fixtures down by removing anything we dont need? It'll take a bit of time but makes reading much easier.

XVincentX: This has been done in 8b93461 — I can give it another pass though.

XVincentX:
@philsturgeon I've gone ahead, unified the two documents (oas2 and 3) and used the same test assertions; then I wrote the required specific assertions for oas3.

Moreover, while writing these tests I literally lost 2-3 hours in understanding a subtle bug I was not able to track — ultimately it turn out that since the graphite instances created by the loaders for Prism are singleton, when loading a different file _even though creating a new Prims instance_ the new resources get stacked up on top of the last loaded ones — basically creating a whole `oas2+oas3` resources array, confusing the whole Prism server. See a33b860 for more details. //cc @chris-miaskowski this made me lose my mind for a *while*

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -2,25 +2,22 @@ import { relative, resolve } from 'path';
 import { createServer } from '../';
 import { IPrismHttpServer } from '../types';
 
-describe('server', () => {
+describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', file => {
```
philsturgeon: Awesome!

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -135,16 +123,58 @@ describe('server', () => {
...
+        shipDate: '12-01-2018',
+        status: 'placed',
+        complete: true,
+      },
     });
+    expect(response.statusCode).toBe(400);
+  });
 
-    const response = await server.fastify.inject({ method: 'GET', url: '/' });
+  if (file === 'petstore.oas2.json') {
```
philsturgeon: I cant figure out why we are splitting this with an if? What is v2 or v3 specific about thesew tests?

XVincentX: This test in particular is testing something against an `oas2` file, and so it's oas2 specific.
The oas3 one is testing the `default` responses that apparently are not a thing in OAS2.

Does that clarify?

packages/cli/src/util/__tests__/createServer.unit.ts
```diff
@@ -26,7 +26,7 @@ describe('server command', () => {
 
     expect(createPrismServer).toHaveBeenLastCalledWith(
       { url: 'http://path.to/spec.oas2.yaml' },
-      { components: { loader: httpLoaderInstance }, config: { mock: false } }
+      { components: { loader: expect.any(HttpLoader) }, config: { mock: false } }
```
chris-miaskowski: :+1:

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -64,19 +61,10 @@ describe('server', () => {
     expect(payload).toHaveProperty('name');
   });
 
-  test('will return 500 with error when an undefined code is requested', async () => {
```
chris-miaskowski: Would be nice to have a _positive_ test in place for this one. E.g. 
`should return requested status code with an error message even if error code not defined`

packages/http-server/src/__tests__/server.oas.spec.ts
chris-miaskowski: `if`s in tests are evil :P 
why not simply move it outside the `each`?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -165,7 +177,10 @@ const helpers = {
           httpContent
         );
       } else {
-        throw new Error('Requested content type is not defined in the schema');
+        return {
+          code,
+          mediaType: 'text/plain',
```
chris-miaskowski: If user requests a specific `mediaType` but we couldn't find it don't we want to return a problem+json response?

packages/http/src/mocker/__tests__/functional.spec.ts
```diff
@@ -180,16 +177,18 @@ describe('http mocker', () => {
...
+            mocker.mock({
+              resource: httpOperations[0],
+              input: Object.assign({}, httpRequests[0], {
+                data: Object.assign({}, httpRequests[0].data, {
+                  headers: { 'Content-type': 'application/yaml' },
+                }),
               }),
-            }),
+            })
+          ).resolves.toMatchObject({
```
chris-miaskowski: I'd lie to see we match against the body too. It's equally important i think

packages/http/src/mocker/__tests__/functional.spec.ts
```diff
@@ -25,20 +24,18 @@ describe('http mocker', () => {
...
+              },
             },
-          },
-        });
-
-        return expect(rejection).rejects.toEqual(
-          new Error('Requested content type is not defined in the schema')
-        );
+          })
+        ).resolves.toMatchObject({ headers: { 'Content-type': 'text/plain' } });
```
chris-miaskowski: Let's match the response body too. I'd like us to be explicit about it.

packages/http-server/src/__tests__/server.oas.spec.ts
chris-miaskowski: Oh ok, I see you've done it several lines below

packages/http/src/index.ts
```diff
@@ -25,7 +25,7 @@ const createInstance = <LoaderInput>(
   return factory<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig, LoaderInput>(
     { mock: true },
     {
-      loader: filesystemLoaderInstance,
+      loader: new FilesystemLoader(),
```
chris-miaskowski: Does that fix this 'singleton' bug? Would be nice to write a unit test to cover this fix (test fails w/o this fix and passes with the fix - standard TDD).

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: Yeah that's a good point, I'll move these outside. The reason I kept these inside is because I didn't want to reinit Prism again, but it's worth the duplication. Will fix!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Yes, but this is another story I think @StefanDywersant will tackle, right?

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: Done in 5e5d65f

packages/http/src/mocker/__tests__/functional.spec.ts
XVincentX: Done in d6dc96f

packages/http/src/mocker/__tests__/functional.spec.ts
XVincentX: Done in d6dc96f

packages/http/src/index.ts
XVincentX: Added in 6fd666b

packages/http-server/src/__tests__/server.oas.spec.ts
philsturgeon: Good thing you moved them, but I still dont understand what is version specific about them. When I asked what is version specific sabout them you answered saying that they were version specific but... why. 🤣

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
philsturgeon: If we just dont mention a mediaType will it default to text plan or will it try and JSON? I know fastify has some opinions.

XVincentX: I can't do that, the `mediaType` is mandatory for the `IHttpNegotiationResult` (which is the returned object)

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: Ah sorry — it turns out default responses are a thing in OAS2 as well. Fixed in 27b2dc8

PR #260, opened 2019-04-17
feat: init prism from an existing graphite instance

XVincentX:
# WIP


XVincentX:
This is not happening anymore, we'll take care of it from Studio directly.

PR #265, opened 2019-05-06
405 —  Method not allowed — SO-184

XVincentX:
This PR will make sure to throw a special exception in case the current path has been matched, but it wasn't possible to find an appropriate method.

We'll catch these exceptions and return the appropriate status codes in a follow up PR.

Highlights:

1. Dry some functions that were unexpectedly complicated and/or accumulating state over time (`matchServer`, `randomPath`)
2. Do not creare `Error` objects in place where they're not thrown, otherwise the stack trace is wrong and you have no idea where the exception happened. (This made me lose some time)
3. Refactored the `route` method to not accumulate the state over time with global `booleans`, but rather create a `matches` object and find the correct thing to do using a series of filters
4. Added tests for the new scenario.

SO-184

### DIFF
--- a/packages/http/src/router/__tests__/index.spec.ts
+++ b/packages/http/src/router/__tests__/index.spec.ts
@@ -1,11 +1,10 @@
 import { IHttpOperation, IServer } from '@stoplight/types';
 import { Chance } from 'chance';
 import {
+  NO_PATH_MATCHED_ERROR,
   NO_RESOURCE_PROVIDED_ERROR,
   NO_SERVER_CONFIGURATION_PROVIDED_ERROR,
-  NONE_METHOD_MATCHED_ERROR,
-  NONE_PATH_MATCHED_ERROR,
-  NONE_SERVER_MATCHED_ERROR,
+  NO_SERVER_MATCHED_ERROR,
 } from '../errors';
 import { router } from '../index';
 import { pickOneHttpMethod, pickSetOfHttpMethods, randomPath } from './utils';
@@ -99,7 +98,7 @@ describe('http router', () => {
               },
             },
           })
-        ).toThrow(NONE_METHOD_MATCHED_ERROR);
+        ).toThrow(NO_PATH_MATCHED_ERROR);
       });
 
       describe('given matched methods', () => {
@@ -126,7 +125,7 @@ describe('http router', () => {
                 },
               },
             })
-          ).toThrow(NONE_PATH_MATCHED_ERROR);
+          ).toThrow(NO_PATH_MATCHED_ERROR);
         });
 
         test('given a concrete matching server and matched concrete path should match', async () => {
@@ -252,7 +251,7 @@ describe('http router', () => {
                 },
               },
             })
-          ).toThrow(NONE_PATH_MATCHED_ERROR);
+          ).toThrow(NO_PATH_MATCHED_ERROR);
         });
 
         test('given a concrete servers and mixed paths should match concrete path', async () => {
@@ -378,7 +377,7 @@ describe('http router', () => {
                 },
               },
             })
-          ).toThrowError(NONE_SERVER_MATCHED_ERROR);
+          ).toThrowError(NO_SERVER_MATCHED_ERROR);
         });
 
         test('given empty baseUrl and empty server url it should match', async () => {
--- a/packages/http/src/router/__tests__/index.spec.ts
+++ b/packages/http/src/router/__tests__/index.spec.ts
@@ -1,6 +1,8 @@
+import { IHttpMethod } from '@stoplight/prism-http';
 import { IHttpOperation, IServer } from '@stoplight/types';
 import { Chance } from 'chance';
 import {
+  NO_METHOD_MATCHED_ERROR,
   NO_PATH_MATCHED_ERROR,
   NO_RESOURCE_PROVIDED_ERROR,
   NO_SERVER_CONFIGURATION_PROVIDED_ERROR,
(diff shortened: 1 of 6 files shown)


packages/http/src/router/__tests__/index.spec.ts
```diff
@@ -99,7 +100,7 @@ describe('http router', () => {
               },
             },
           })
-        ).toThrow(NONE_METHOD_MATCHED_ERROR);
+        ).toThrow(NO_PATH_MATCHED_ERROR);
```
chris-miaskowski: NO_PATH_MATCHED_ERROR?
You meant NO_METHOD_MATCHED_ERROR?

chris-miaskowski: Ignore, I see what you did

packages/http/src/router/__tests__/utils.ts
```diff
@@ -30,19 +40,15 @@ interface IRandomPathOptions {
 }
 
 export function randomPath(opts: IRandomPathOptions = defaultRandomPathOptions): string {
-  opts = Object.assign({}, defaultRandomPathOptions, opts);
+  defaults(opts, defaultRandomPathOptions);
```
chris-miaskowski: Nice, I didn't know `defaults`. I'm not a huge fan of mutating arguments, though. Is this not a problem?

XVincentX: Me neither, but you were mutating them anyway in the previous statement.

`opts` is passed externally as an argument — and then they get overridden with `opts = Object.assign({…})`

philsturgeon:
It's not related to this change, but something to keep in mind, is that these tests are complicated to read due to not really following describe-it syntax. They confiuse context and assertion in the title.

For example this:

```
 test('given a concrete matching server and unmatched templated path should not match', () => {
```

Should be:

```
describe('when server is valid', () => {
    describe('and templated path is nonsense', () => {
        it('throws a NO_METHOD_ERROR, () => {
```

Stuff like that. Then you can group constants for shared setup, and have more assertions in more clearly labeled its. 

Learned all this in rspec land, I love the website http://betterspecs.org/ but not 100% of it is relevant to Jest ofc.

packages/http/src/router/__tests__/utils.ts
```diff
@@ -30,19 +40,15 @@ interface IRandomPathOptions {
...
-  let leadingSlash = chance.pickone(['/', '']);
-  let trailingSlash = chance.pickone(['/', '']);
-  if (opts.leadingSlash !== undefined) {
-    leadingSlash = opts.leadingSlash ? '/' : '';
-  }
-  if (opts.trailingSlash !== undefined) {
-    trailingSlash = opts.trailingSlash ? '/' : '';
-  }
+
+  const leadingSlash = opts.leadingSlash ? '/' : '';
```
chris-miaskowski: As long as all tests pass this is good! Probably was overly complex.

packages/http/src/router/__tests__/utils.ts
XVincentX: f671266

packages/http/src/router/index.ts
```diff
@@ -3,89 +3,94 @@ import { IHttpOperation, IServer } from '@stoplight/types';
...
-    if (!anyPathMatched) {
-      throw NONE_PATH_MATCHED_ERROR;
+      if (matches.every(match => !!match.serverMatch && match.serverMatch === MatchType.NOMATCH)) {
+        throw new Error(NO_SERVER_MATCHED_ERROR);
+      }
     }
 
-    if (!anyServerProvided) {
-      throw NO_SERVER_CONFIGURATION_PROVIDED_ERROR;
+    if (!matches.some(match => match.pathMatch !== MatchType.NOMATCH)) {
```
chris-miaskowski: I like your refactoring in general but whether it's more readable or not it's rather debatable ;)

Maybe you could increase the readability by extracting those one liners to small functions. For instance:

```suggestion
    if (noPathMatched(matches)) {
```

philsturgeon:
Interestingly I don't see `405`in this PR, is there something else to be done? Can we test the status code? Or is there a nickname for the status code in Node?

packages/http/src/router/index.ts
XVincentX: Yeah, I think is a good idea! Will do!

XVincentX: 3b4b9cf

XVincentX:
@philsturgeon You're correct, the PR title is a little bit misleading. This PR throws an exception when the condition to return a 405 happen. The real status code is on a followup PR; the reason for this is that this already includes some refactors that are beneficial for the codebase; the whole thing together would just be too long to review.

XVincentX:
> Looks like we already had that error being returned.

No, this is incorrect. If you check the `errors.ts` constants you can see a new value has been added and the condition detected by reword the conditions order.

chris-miaskowski:
> Interestingly I don't see `405`in this PR, is there something else to be done? Can we test the status code? Or is there a nickname for the status code in Node?

That's what I said in the #prism channel ;) This PR doesn't return 405 it cleans some old code up which already threw an error in case method was not matched. If I understood correctly @XVincentX will return the 405 status code as part of resolving SO-141. Correct @XVincentX? 
The thing I was confused about is why SO-185 was marked as won't fix and linked to SO-141 (which I think are two different things). As long as the end result is correct I don't care about ticket numbers though ;)

XVincentX:
Yeah I take a lot of blame on this, I made some confusion with the Jira stuff.

PR #270, opened 2019-05-07
Problem+Json for error messages

XVincentX:


package.json
```diff
@@ -22,15 +22,15 @@
...
-    "@types/node": "12.0.0",
+    "@types/jest": "24.x.x",
+    "@types/node": "12.x.x",
     "chance": "1.0.x",
-    "jest": "24.8.0",
+    "jest": "24.x.x",
     "lerna": "3.13.4",
     "ts-jest": "24.x.x",
     "tslint": "5.16.0",
-    "tslint-config-stoplight": "1.2.x",
```
philsturgeon: I notice that everyone has their own personal preference and they change them as they go. It concerns me that we change these so much, and that we do not seem to notice the technical differences behind them.

Pegging to specifics = not good, we miss security updates.
Using x = we are opening up the possibility of a dependency rolling back a dependency if it has strict requirements and we would only notice if we keep a close eye on the lock file

I generally prefer ^24.0.12 because that says "I'll take anything in the 24.0.x range so long as its over 12, because 11 had a bug that was messing us up and there is a reason I said ^24.0.12"

@chris-miaskowski you are one of the folks I see changing these things a lot too, would you agree with the above? :)

XVincentX: My preference is: _whatever yarn puts_, I am just sticking to the rules people are giving me

chris-miaskowski: @philsturgeon I can go with either as long as we are consistent. `.x` was the way to go so far, I'm happy to move away from it.

packages/http/src/mocker/errors.ts
```diff
@@ -0,0 +1,8 @@
+import { Omit } from '@stoplight/types';
+import { ProblemJson } from '../types';
+
+export const NO_INVALID_RESPONSE_TEMPLATE: Omit<ProblemJson, 'detail'> = {
+  name: 'NO_INVALID_RESPONSE_TEMPLATE',
+  title: 'Missing invalid response or response example for the current invalid request',
```
philsturgeon: This is a bit of a mouthful! What does "Missing invalid response" mean and what does the other half mean? 

Maybe we can just call this INVALID_REQUEST we do not need to talk about the response template being missing, thats an implementation choice we currently make which might change. The error is just that thr request sucks, and we can talk about response definitions in the `details` perhaps.

package.json
philsturgeon: Sweeet, let's try and stick to ^ as it says "or newer", and if we do not care at all which minor version or patch is being used then do ^ and 0 instead of an x so its clear.

packages/http/src/mocker/errors.ts
XVincentX: This error means:

1. The user sends a request to Prism
2. Prism's been able to find a matching path and method
3. Prism has validated the provided payload against the JSON Schema defined in the document
4. The validation has *failed*
5. Prism looks for a 4xx response to return to the user as an error
6. No such error is found and Prism complains with this error

Does this clarify what's going on? I'm open to better title and messages, I admin they're not very helpful.

philsturgeon: Sure I understand the flow that lead to the error, but I think we are being to specific about why we decided to throw that. 

The error is really "validation failed" and the information should be about that. `details` can mention _why_ they are seeing this error and not their own validation messages, and that will also be in the logs, but I think we are being too specific.

Imagine if we decide to change this, so prism does not look for a user defined 400 by default, or is moved to a config switch. It would mean our overly specific error messages are now inaccurate.

packages/http/src/router/index.ts
```diff
@@ -63,24 +65,32 @@ export const router: IRouter<IHttpOperation, IHttpRequest, IHttpConfig> = {
...
+          NO_SERVER_CONFIGURATION_PROVIDED_ERROR,
+          `No server configuration has been provided, although ${requestBaseUrl} as base url`
+        );
       }
 
       if (matches.every(match => !!match.serverMatch && match.serverMatch === MatchType.NOMATCH)) {
-        throw new Error(NO_SERVER_MATCHED_ERROR);
+        throw ProblemJsonError.fromTemplate(
+          NO_SERVER_MATCHED_ERROR,
+          `The base url ${requestBaseUrl} hasn't been matched with any of the provided servers`
```
philsturgeon: Is it a base URL?

packages/http/src/mocker/errors.ts
XVincentX: f36ccb2

packages/http/src/router/index.ts
XVincentX: Hmm, hard to say. Technically yes it's a base url, such as

`test.com/stuff`, but since we ignore the `basePath` or whatever it's called in OAS2, it's rather an `origin`

packages/http/src/mocker/errors.ts
```diff
@@ -0,0 +1,8 @@
+import { Omit } from '@stoplight/types';
+import { ProblemJson } from '../types';
+
+export const INVALID_REQUEST_PAYLOAD: Omit<ProblemJson, 'detail'> = {
+  name: 'INVALID_REQUEST_PAYLOAD',
+  title: 'Invalid request body payload',
+  status: 412,
```
philsturgeon: Is 412 the right code here? We are talking about validation right? 

422 I would understand, that is common for validation or 400 as we've done before, but 412 is a whole other thing.

XVincentX: I'm sorry that was the indeed the intention

package.json
XVincentX: Ok so shall I modify all the dependencies here now?

philsturgeon: Sorry, please. Then let’s merge this beast.

XVincentX:
@philsturgeon Correct, there are not effectively that many changes. I'll wait @chris-miaskowski to have a look into before merging.

packages/http/src/types.ts
```diff
@@ -83,3 +75,21 @@ export interface IHttpResponse {
...
+};
+
+export class ProblemJsonError extends Error {
+  public static fromTemplate(template: Omit<ProblemJson, 'detail'>, detail?: string): ProblemJsonError {
+    return new ProblemJsonError(template.name, template.title, template.status, detail || '');
+  }
+
+  constructor(readonly name: string, readonly message: string, readonly status: number, readonly detail: string) {
+    super(message);
+    Error.captureStackTrace(this, ProblemJsonError);
```
chris-miaskowski: Nice, didn't know about this :)

packages/http/src/mocker/errors.ts
```diff
@@ -0,0 +1,8 @@
+import { Omit } from '@stoplight/types';
+import { ProblemJson } from '../types';
+
+export const INVALID_REQUEST_PAYLOAD: Omit<ProblemJson, 'detail'> = {
+  name: 'INVALID_REQUEST_PAYLOAD',
+  title: 'Invalid request body payload',
+  status: 422,
```
chris-miaskowski: Looks like you've swapped `400` to `422`. You think it matches better than `400`? If so maybe we could name the errors in a way that would reflect the status code?

How about `UNPROCESSABLE_ENTITY`?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -38,16 +38,10 @@ export class HttpMocker
...
-          statusCode: 400,
-          headers: {
-            'Content-type': 'text/plain',
-          },
-          body: `ERROR: Your request is not valid.
-We cannot generate a sensible response because your '400'
-response has neither example nor schema or is not defined.
-Here is the original validation result instead: ${JSON.stringify(input.validations.input)}`,
-        };
+        throw ProblemJsonError.fromTemplate(
```
chris-miaskowski: I'm trying to think if we could be more specific here? E.g. we don't check what the `error` is.

For instance, the original error message says:

> We cannot generate a sensible response because your '400'	
> response has neither example nor schema or is not defined.	
> Here is the original validation result instead: ${JSON.stringify(input.validations.input)}`,

Which leads me to think that there is more than one failure scenario. Maybe the original message didn't make sense but it looks like it's two different things: invalid request body VS undefined example/schema. Have you been able to verify which one it actually is?

packages/http/src/types.ts
```diff
@@ -83,3 +75,21 @@ export interface IHttpResponse {
...
 }
+
+export type ProblemJson = {
+  name: string;
+  title: string;
+  status: number;
+  detail: string;
+};
+
+export class ProblemJsonError extends Error {
```
chris-miaskowski: Can you add a simple unit test showing what we expect ProblemJsonError to look like? E.g. in other tests you do `expectToThrow(ProblemJsonError.fromTemplate(etc))`, but would be nice to see what `ProblemJsonError.fromTemplate(etc)` is meant to be.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -64,13 +59,14 @@ describe('Http Prism Instance function tests', () => {
...
-    });
-    expect(response).toMatchSnapshot();
+    expect(
+      prism.process({
+        method: 'get',
+        url: {
+          path: '/pet/findByStatus',
+        },
+      })
+    ).rejects.toThrowError(INVALID_REQUEST_PAYLOAD.title);
```
chris-miaskowski:
```suggestion
    ).rejects.toThrowError(ProblemJsonError.fromTemplate(INVALID_REQUEST_PAYLOAD.title));
```

packages/http-server/src/server.ts
```diff
@@ -75,14 +91,22 @@ const replyHandler = <LoaderInput>(
...
+      const status = 'status' in e ? e.status : 500;
+      reply
+        .type('application/problem+json')
+        .serializer(problemJsonSerializer)
+        .code(status)
+        .send({
+          type: e.name || 'https://stoplight.io/prism/errors#UNKNOWN',
+          title: e.message,
+          status,
+          detail: e.detail || '',
```
chris-miaskowski: Isn't `detail` set to empty by default in the JsonProblemError constructor?

packages/http-server/src/server.ts
```diff
@@ -1,11 +1,30 @@
 import { configMergerFactory } from '@stoplight/prism-core';
 import { createInstance, IHttpMethod, TPrismHttpInstance } from '@stoplight/prism-http';
+import * as buildSerializer from 'fast-json-stringify';
 import * as fastify from 'fastify';
 import { IncomingMessage, Server, ServerResponse } from 'http';
-
 import { getHttpConfigFromRequest } from './getHttpConfigFromRequest';
 import { IPrismHttpServer, IPrismHttpServerOpts } from './types';
 
+const problemJsonSerializer = buildSerializer({
```
chris-miaskowski: Maybe add a comment that this should be in sync with the JsonProblemError type?

packages/http-server/src/server.ts
```diff
@@ -75,14 +91,22 @@ const replyHandler = <LoaderInput>(
...
       }
     } catch (e) {
-      reply.code(500).send(e);
+      const status = 'status' in e ? e.status : 500;
+      reply
+        .type('application/problem+json')
+        .serializer(problemJsonSerializer)
+        .code(status)
+        .send({
+          type: e.name || 'https://stoplight.io/prism/errors#UNKNOWN',
```
chris-miaskowski: I like this :) can we make all error names follow this syntax?

packages/http/src/mocker/errors.ts
XVincentX: Makes sense. b5eb0ea

packages/http-server/src/server.ts
XVincentX: Yes! The idea would also be to create a real webpage with bookmark so that we can also provide suggestions on how to fix it.

XVincentX: Fixed in 4a481c8

XVincentX:
> Remember to implement the 405

This is implemented in this PR, see the `NO_METHOD_MATCHED_ERROR` error

packages/http-server/src/server.ts
XVincentX: I was considering the idea of ditching this completely. This is only required to speedup the error responses but I think the effort of keeping this in sync and the code required to set this up is not worth the effort. What do you think?

XVincentX: Yeah, let's get rid of that 531ff60

package.json
XVincentX: Careting done

XVincentX: 8091c09

packages/http-server/src/server.ts
XVincentX: Good catch! cb18613]

packages/http/src/types.ts
XVincentX: Isn't this assured by the types?

```ts
public static fromTemplate(template: Omit<ProblemJson, 'detail'>, detail?: string): ProblemJsonError
```

`ProblemJsonError` has `name, message,status,detail`, so there's no way it has *not* such fields. Does that clarify?

packages/http/src/mocker/HttpMocker.ts
XVincentX: The original error was more specific, @philsturgeon asked to be generic because we might want to change the way Prism responds.

Ideally, instead of saying "Hey I can't find an error response for this invalid request", we could in the future generate such error response automatically.

PR #278, opened 2019-05-09
fix: look for 422 for invalid requests

XVincentX:
This PR will make sure Prims will look for `422` responses in the document for invalid requests first; if not defined, Prism will generate one and not use the `default` response, if defined.

This replace the following behaviour: look for `400` response and if not defined, use the `default` response and, if not defined, then throw a `422`

### DIFF
--- a/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
+++ b/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
@@ -32,11 +32,14 @@ function findResponseByStatusCode(
     return candidate;
   }
 
+  return createResponseFromDefault(statusCode, responses);
+}
+
+function createResponseFromDefault(statusCode: string, responses: IHttpOperationResponse[]) {
   const defaultResponse = responses.find(response => response.code === 'default');
   if (defaultResponse) {
     return Object.assign({}, defaultResponse, { code: statusCode });
   }
-
   return undefined;
 }
 
--- a/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
+++ b/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
@@ -32,8 +32,7 @@ function findResponseByStatusCode(
   if (candidate) {
     return candidate;
   }
-  if (useDefault)
-    return createResponseFromDefault(statusCode, responses);
+  if (useDefault) return createResponseFromDefault(statusCode, responses);
   return undefined;
 }
 
--- a/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
+++ b/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
@@ -15,7 +15,7 @@ function findHttpContentByMediaType(response: IHttpOperationResponse, mediaType:
 }
 
 function findLowest2xx(httpResponses: IHttpOperationResponse[]): IHttpOperationResponse | undefined {
-  const generic2xxResponse = findResponseByStatusCode(httpResponses, '2XX');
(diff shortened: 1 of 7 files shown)


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -245,12 +249,12 @@ const helpers = {
...
 
   negotiateOptionsForInvalidRequest(httpResponses: IHttpOperationResponse[]): IHttpNegotiationResult {
-    // currently only try to find a 400 response, but we may want to support other cases in the future
-    const code = '400';
-    const response = findResponseByStatusCode(httpResponses, code);
-    // TODO: what if no 400 response is defined?
+    // currently only try to find a 422 response, but we may want to support other cases in the future
+    const code = '422';
+    const response = findResponseByStatusCode(httpResponses, code, false);
+    // TODO: what if no 422 response is defined?
```
philsturgeon: Ok, so I know I said:

> the order is always: "we want to throw an X, if they have an X use it, if they have a default use it, they have neither so here is our version of an X"

but I think we should double this up for the validation use case because so many people use 400.... if validation fails can we look for X or Y?

> we want to throw an 422, if they have a 422 use it, if they have 400 that'll do, if they have a default use it, they have none of that so bugger here is our version of an 422

Is that bonkers?

XVincentX: This makes sense, and it's happening in eb4cc18. Tests included.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -245,12 +246,12 @@ const helpers = {
...
 
   negotiateOptionsForInvalidRequest(httpResponses: IHttpOperationResponse[]): IHttpNegotiationResult {
-    // currently only try to find a 400 response, but we may want to support other cases in the future
-    const code = '400';
-    const response = findResponseByStatusCode(httpResponses, code);
-    // TODO: what if no 400 response is defined?
+    const response =
+      findResponseByStatusCode(httpResponses, '422') ||
+      findResponseByStatusCode(httpResponses, '400') ||
+      createResponseFromDefault(httpResponses, '422');
```
XVincentX: @philsturgeon Here are the relevant bits. Invalid request:

Find 422 -> Find 400 -> Find a default -> Generate a 422 automagically.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -143,16 +143,64 @@ describe('NegotiatorHelpers', () => {
...
+              ],
+            },
+          ])
+          .instance();
+
+        const actualConfig = helpers.negotiateOptionsForInvalidRequest(httpOperation.responses);
+        expect(actualConfig).toHaveProperty('code', '400');
+      });
+
+      test('but a default response exists', async () => {
```
philsturgeon: The “but a foo exists” should be a describe. Describe is to setup the context then test is for the assertion to explain what you are actually asserting.

XVincentX: I agree, but I kept the style that was there. I'm all for the change and refactor opportunistically when the occasion comes.

philsturgeon: My bad I thought you added this, merge away!

PR #299, opened 2019-05-14
Prism should read yml files too SO-200

XVincentX:
1. Correctly reveals yml files as yaml for graphite
2. Handles startup exceptions and prints them out
3. Updates the README

### DIFF
--- a/packages/core/src/utils/graphFacade.ts
+++ b/packages/core/src/utils/graphFacade.ts
@@ -4,6 +4,7 @@ import {
   FileSystemBackend,
   FilesystemNodeType,
 } from '@stoplight/graphite/backends/filesystem';
+import { filenameToLanguage } from '@stoplight/graphite/backends/filesystem/utils';
 import { NodeCategory } from '@stoplight/graphite/graph/nodes';
 import { createGraphite } from '@stoplight/graphite/graphite';
 import { createOas2HttpPlugin } from '@stoplight/graphite/plugins/http/oas2';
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/graphFacade.ts
```diff
@@ -64,3 +66,8 @@ export class GraphFacade {
     return compact(nodes.map<IHttpOperation>(node => node.data as IHttpOperation));
   }
 }
+
+function transformLanguage(lang: string) {
```
marbemac: Suggest a simple map over function:

```ts
import { Languages } from '@stoplight/graphite/graph';

const langMap = {
  json: Languages.Json,
  yaml: Languages.Yaml,
  yml: Languages.Yaml,
}
```

This way can also check that `language` exists in `langMap` and report error if it does not.

```ts
language = langMap[language];
if (!language) {
  // report error and stop or whatever
}
```

XVincentX: Ah interesting idea, I didn't think of reusing that part of graphite!

XVincentX: @marbemac how about importing the function that does that in graphite and use it?

`import { filenameToLanguage } from '@stoplight/graphite/backends/filesystem/utils'`

…
`language: filenameToLanguage(path)`

marbemac: Even better! Still suggest reporting an error if language is undefined.

XVincentX: The language will never be undefined; the function is using a proxied map returning the string you set in case it's not in the map 

https://github.com/stoplightio/graphite/blob/master/src/backends/filesystem/utils.ts#L15-L26

marbemac: Ah, got it - ok well then 🤷‍♂, hopefully folks don't feed prism files that are not json or yaml.

XVincentX:
@philsturgeon How about you follow the guide and try to release Prism?

PR #301, opened 2019-05-14
--dynamic flag for CLI SO-217

XVincentX:
1. Adds the `--dynamic` CLI flag
2. Prints something on the CLI if the dynamic is true
3. Changes the configuration object to be either "false" or a full configuration object (no more hidden defaults)
4. Adds tests

image

### DIFF
--- a/packages/cli/src/commands/mock.ts
+++ b/packages/cli/src/commands/mock.ts
@@ -5,7 +5,7 @@ import { createServer } from '../util/createServer';
 
 export default class Server extends Command {
   public static description = 'Start a mock server with the given spec file';
-  public static flags = { port: FLAGS.port };
+  public static flags = { port: FLAGS.port, dynamic: FLAGS.dynamic };
   public static args = [ARGS.spec];
 
   public async run() {
@@ -14,11 +14,11 @@ export default class Server extends Command {
     signaleInteractiveInstance.await('Starting Prism…');
 
     const {
-      flags: { port },
+      flags: { port, dynamic },
       args: { spec },
     } = this.parse(Server);
 
-    const server = createServer(spec, { mock: true });
+    const server = createServer(spec, { mock: { dynamic } });
     try {
       const address = await server.listen(port);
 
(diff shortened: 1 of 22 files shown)


packages/cli/src/const/options.ts
```diff
@@ -15,4 +15,11 @@ export const FLAGS = {
     default: 4010,
     required: true,
   }),
+
+  dynamic: oflags.boolean({
+    char: 'd',
+    description: 'Dynamically generate examples.',
+    default: true,
```
philsturgeon:
```suggestion
    default: false,
```

packages/cli/src/commands/mock.ts
```diff
@@ -14,11 +14,15 @@ export default class Server extends Command {
     signaleInteractiveInstance.await('Starting Prism…');
 
     const {
-      flags: { port },
+      flags: { port, dynamic },
       args: { spec },
     } = this.parse(Server);
 
-    const server = createServer(spec, { mock: true });
+    if (dynamic) {
```
philsturgeon:
```suggestion
    if (true || dynamic) {
```

XVincentX:
I think @chris-miaskowski should have a look here as well before merging this in.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -191,5 +173,48 @@ describe('HttpMocker', () => {
...
+
+          expect(response.body).toBeDefined();
+
+          const allExamples = flatMap(mockResource.responses, res =>
+            flatMap(res.contents, content => content.examples),
+          ).map(x => {
+            if ('value' in x) return x.value;
+          });
+
+          allExamples.forEach(example => expect(response.body).not.toEqual(example));
```
chris-miaskowski: This test checks that the response is defined and that it's not one of the defined one, but it doesn't tell us anything about the nature of the response. 

Have you thought about treating this like a white box and maybe stubbing the response? It looks like this file is mostly about unit tests, whereas this test is more of an integration test.

XVincentX: Done in 007e6a6

PR #303, opened 2019-05-15
refactor: get rid of node-tmp

XVincentX:
Gets rid of `node-tmp` entirely and uses Graphite to init the stuff from memory

### DIFF
--- a/packages/core/src/loaders/http/httpLoader.ts
+++ b/packages/core/src/loaders/http/httpLoader.ts
@@ -1,8 +1,6 @@
 import { FilesystemNodeType } from '@stoplight/graphite/backends/filesystem';
 import { IHttpOperation } from '@stoplight/types';
 import axios from 'axios';
-import { trimStart } from 'lodash';
-import { extname } from 'path';
 import { IHttpLoaderOpts } from '../../types';
 import { GraphFacade } from '../../utils/graphFacade';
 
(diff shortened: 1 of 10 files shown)


packages/core/src/utils/graphFacade.ts
```diff
@@ -60,6 +60,18 @@ export class GraphFacade {
     await this.graphite.scheduler.drain();
   }
 
+  public async createRawNode(raw: string, { type, path }: Pick<ISourceNode, 'type' | 'path'>) {
+    this.graphite.graph.addNode({
+      category: NodeCategory.Source,
+      type,
+      language: filenameToLanguage(path),
+      path: '/' + path,
```
XVincentX: Note: I hadd to add the `/` otherwise Graphite complains because the source node is not an absolute path. However, we need the path (in this case an URL) there.

packages/core/src/loaders/http/httpLoader.ts
```diff
@@ -1,24 +1,23 @@
...
 
-tmp.setGracefulCleanup();
-
 export class HttpLoader {
   constructor(private graphFacade: GraphFacade = new GraphFacade()) {}
 
   public async load(opts?: IHttpLoaderOpts): Promise<IHttpOperation[]> {
-    if (!opts || !opts.url) return [];
+    if (!opts) return [];
+    if (!opts.url) return [];
```
marbemac:
```ts
if (!opts || !opts.url) return [];
```

PR #307, opened 2019-05-17
[SO-145] Prism v3 Release Validation - Add test harness

lag-of-death:
## Checklist

- ~[ ] Tests have been added (for bug fixes / features)~

Feature.

No test harness in `master`.

A base for `e2e` tests (with a couple of example ones) is introduced.

No breaking changes.


.circleci/config.yml
```diff
@@ -21,14 +21,29 @@ jobs:
       - run:
           name: Install Dependencies
           command: yarn
+      - run:
+          name: Build all code to JavaScript
+          command: yarn build
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
```
XVincentX: This line is only required when publishing on `npm`, which we're not doing here.

.circleci/config.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
+      - run:
+          name: Create CLI binaries
+          command: npx pkg --out-path ./cli-binaries ./packages/cli/
```
XVincentX: I am assuming you want to test the binaries with the binaries.

I think we can just test the stuff directly from the source code, however if you want to go with the binaries that'll do as well; however we should:

1. Cache the binaries we download from `pgk`, otherwise we're going to spend 2 more minutes every time we run the CI
2. Depending how long these tests take to run, maybe the harness should only run when we merge in `master`
3. Probably the harness should be a separate step.

package.json
```diff
@@ -16,7 +16,10 @@
     "test.ci": "jest --maxWorkers=2",
     "test.coverage": "yarn test --coverage",
     "test.watch": "yarn test --watchAll",
+    "run.binary": "cd cli-binaries && ./$BINARY mock ../examples/$SPEC",
```
XVincentX: Prefer longer paths over cd in directories:
```suggestion
    "run.binary": "./cli-binaries/$BINARY mock ../examples/$SPEC",
```

package.json
```diff
@@ -16,7 +16,10 @@
...
     "test.update": "yarn test.fast --updateSnapshot",
+    "build.binaries": "yarn && yarn build && npx pkg --out-path ./cli-binaries ./packages/cli/",
```
XVincentX: Use pre scripts instead of multiple commands. Also, I think you can assume the basic yarn command has been ran already.

.circleci/config.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
       - save_cache:
           name: Save Yarn Package Cache
           key: yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
           paths:
             - node_modules/
+      - run:
+          command: SPEC=petstore.oas2.json BINARY=prism-cli-linux yarn run.binary
```
XVincentX: To make the harness cross platform, you can avoid to use the binary and use the source code directly:

```suggestion
          command: yarn workspace @stoplight/prism-cli cli mock ./examples/petstore.oas2.json
```

.circleci/config.yml
philsturgeon: Yep the goal here is definitely to test the actual binaries. We want to make sure the built product is working as expected, as there have been a few issues with the build process in the past.

philsturgeon: As for branches / master, we can worry about that later. This PR will be putting them everywhere and when it annoys us we can change it.

.circleci/config.yml
philsturgeon: Doubling down on binaries, lets pass the entire path in instead of a partial, so the script is more flexible.

```suggestion
          command: SPEC=petstore.oas2.json BINARY=./cli-binaries/prism-cli-linux yarn run.binary
```

.circleci/config.yml
```diff
@@ -21,14 +21,29 @@ jobs:
...
+          background: true
+      - run:
+          shell: /bin/sh
+          command: |
+            wget --retry-connrefused --waitretry=1 --read-timeout=20 --timeout=15 -t 10 http://localhost:4010
+            :
       - run:
           name: Run tests
-          command: yarn test.coverage --verbose
```
philsturgeon: We can do both :D

philsturgeon: Sorry what I meant was we should have normal tests and binary tests as two different steps because they likely won’t be used together forever.

test-harness/README.md
```diff
@@ -24,6 +24,8 @@ Gold master files contain data about both request and response.
 `yarn test.binary`
 or
 `SPEC=./examples/petstore.oas2.json BINARY=./cli-binaries/prism-cli-linux yarn test.binary`
+or
```
philsturgeon: What happens when multiple specs are provided?

lag-of-death: https://github.com/stoplightio/prism/pull/307/files#diff-389656a19bda4636b1c0c52a8281822aR28

.circleci/config.yml
lag-of-death: does `likely` satisfy investing time now to separate these tests?

philsturgeon: It’s more that we should not be confusing these two very different test suites. We will likely separate them in Circle very soon but we will definitely be running them separately locally. 

yarn test is for source code which devs will run all the time 
yarn test.binary is for the built product which developers will run rarely (before submitting PR or a release maybe, or maybe not) 

It’s better to keep separate test suites seperate, so we can reference them however we want on circle.

test-harness/README.md
```diff
@@ -0,0 +1,36 @@
+## Generating binaries
+
+For anything to work, the binaries have to be generated first with: `yarn build.binaries`. This will generate three binaries.
+
+### Alternatively, run any of the following:
+
+1. `yarn build.binary` will assume the host machine
+2. `HOST=host yarn build.binary` will generate a binary for the current host machine
+3. `HOST=node10-linux-x64 yarn build.binary` will generate a binary for node10-linux-x64
```
XVincentX: We should probably tell the user what are the possible choices or at least put a link to where the choices are (`pkg` docs should be good).

test-harness/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+*Before running any of the above, make sure that `yarn` was run beforehand*. If `yarn` was not run, you might experience `Error: command mock not found` issue.
```
XVincentX: So what we _really_ need is to make sure the dependencies are installed, and that can be done by either using `yarn` or `npm` — for this reason I'd rather say "make sure to have all the dependencies installed" rather than pointing to a specific command. Today yarn is a thing, maybe tomorrow it will not.

test-harness/README.md
```diff
@@ -0,0 +1,36 @@
...
+##  To run tests, run any of the following:
+
+*There is no need to manually start prism binary to run tests.*
+
+1. `yarn test.binary`
+2. `SPEC=./examples/petstore.oas2.json BINARY=./cli-binaries/prism-cli-linux yarn test.binary`
+3. `SPEC=./examples/petstore.oas2.json,./examples/petstore.oas3.json BINARY=./cli-binaries/prism-cli-linux yarn test.coverage --verbose` (`SPEC` can take comma delimited paths to specs)
+4. `RUN_V2_TESTS=1 SPEC=./examples/petstore.oas2.json,./examples/petstore.oas3.json yarn test.binary`
+
+Adding `RUN_V2_TESTS=1` will additionally run the tests against Prism v2 binary. Please run `yarn download.prism.v2` first.
```
XVincentX: I'd rather say "make sure to have Prism installed". Most of the folks in Stoplight have it installed anyway because of the local development stuff.

test-harness/createMasterFiles.js
```diff
@@ -0,0 +1,30 @@
...
+
+  try {
+    fs.writeFileSync(
+      `${__dirname}/gold-master-files/${constructMasterFileName({
+        path,
+        method,
+        headers,
+        body,
+      })}.json`,
+      JSON.stringify(reqRes, null, 2)
```
XVincentX: Can you make sure to put a new line after the stringified content, so GitHub stops to complain?

test-harness/createMasterFiles.js
```diff
@@ -0,0 +1,30 @@
...
+  } catch (err) {
+    console.error(err);
+  }
+}
+
+(async function() {
+  for (const request of requests) {
+    const { dynamic, ...req} = request;
+
+    await recordMasterFile(req);
```
XVincentX: I think all these request can be run using `Promise.all`, decreasing the run time kind of significantly

```ts
await Promise.all(requests.map(recordMasterFile))
```

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
+const fetch = require('node-fetch');
+const requests = require('./requests');
+
+const { exec } = require('child_process');
+const { makeRequest, constructMasterFileName, readFile } = require('./helpers');
+
+async function waitForPrism(done) {
```
XVincentX: For this kind of operations is usually better to attach yourself to the `stdout` stream you can get from the `exec` function and check when the output is coming.

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
...
+
+  const reqRes = await makeRequest(request);
+
+  return {
+    reqRes,
+    masterFile,
+  };
+}
+
+function killPrism(done = () => null) {
```
XVincentX: Hmm, I'm generally not a great fan of spawning processes from an application; I can live with it since it's testing code, but if there's another way, it'd be way better.

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
+const fetch = require('node-fetch');
+const requests = require('./requests');
+
+const { exec } = require('child_process');
+const { makeRequest, constructMasterFileName, readFile } = require('./helpers');
+
+async function waitForPrism(done) {
+  try {
+    await fetch('http://localhost:4010');
```
XVincentX: If you're listening on `4010`, send this parameter explicitly when spawning Prism so we do not get surprises in case the port gets changed for any reason.

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
...
+
+  return {
+    reqRes,
+    masterFile,
+  };
+}
+
+function killPrism(done = () => null) {
+  exec('fuser -k 4010/tcp', () => {
+    done();
```
XVincentX: In any case — you should use the `subprocess.kill()` function that the `exec` function is returning when spawning prism

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
...
+  });
+}
+
+const spec = `${process.env.SPEC || `${__dirname}/../examples/petstore.oas2.json`}`;
```
XVincentX: Use `path.join` so you can avoid the string interpolation.
```suggestion
const spec = process.env.SPEC || path.join(__dirname, '/../examples/petstore.oas2.json');
```

packages/core/src/utils/__tests__/__snapshots__/graphFacade.int.ts.snap
```diff
@@ -2,6 +2,1418 @@
...
+              ],
+              "type": "array",
+            },
+          },
+        ],
+        "description": "",
+        "headers": Array [],
+      },
+      Object {
+        "code": "401",
```
XVincentX: Note to myself: I want to get rid of this snapshot, it's simply unmanageable.

test-harness/index.js
```diff
@@ -0,0 +1,270 @@
...
+      });
+
+      xdescribe('When a form data is sent with application/x-www-form-urlencoded to create a single resource', () => {
+        test(
+          [
+            'Should answer with JSON representing the resource',
+            'prism v3 does not respect x-ww-form-urlencoded',
+          ].join(),
+          async () => {
+            const { reqRes, masterFile } = await runTest(requests[5]);
```
XVincentX: Maybe use [each](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout) instead of looping through all the tests manually?

pytlesk4:
>
>     2. Food for thoughts: maybe a lot of the hassle of writing the file on the disk and styff could be avoided by just using Jest Snapshots — I think this is really a legit use case and I'd try to see if it's possible to pursue such opportunity. For example. https://github.com/jest-community/snapshot-diff could be used to write the snapshot of the output differences between Prism V2 and V3
>

@XVincentX snapshot diff looks cool.

test-harness/index.js
lag-of-death: @XVincentX, I don't see any merit it using `path.join` instead of string interpolation. How exactly is `path.join` better here?

test-harness/index.js
lag-of-death: Yeah, did that. Didn't work.

test-harness/index.js
lag-of-death: Cool, how do I tell prism to start on port X?

examples/petstore.oas2.json
```diff
@@ -42,6 +42,289 @@
     "http"
   ],
   "paths": {
+    "/no_auth/pets/findByStatus": {
```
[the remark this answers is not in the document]
lag-of-death: Coming from? They were added for the tests.

examples/petstore.oas3.json
```diff
@@ -36,6 +36,299 @@
     }
   ],
   "paths": {
+    "/no_auth/pets": {
```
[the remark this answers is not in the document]
lag-of-death: Please see the comment above :)

package.json
```diff
@@ -16,7 +16,15 @@
     "test.ci": "jest --maxWorkers=2",
     "test.coverage": "yarn test --coverage",
     "test.watch": "yarn test --watchAll",
+    "download.prism.v2": "curl -L https://github.com/stoplightio/prism/releases/download/v2.0.17/prism_linux_amd64 -o prism && chmod +x ./prism",
```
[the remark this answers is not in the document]
lag-of-death: Okay, but why I shouldn't be able to run this locally? Why take this away from a dev? And even if you have prism locally, it doesn't matter, since here we are downloading to prism project root folder.

lag-of-death:
@XVincentX, please see my answers before I do any more coding on this. And gold master files are recorded with `./test-harness/createMasterFiles.js` as described in [`./test-harness/README.md`](https://github.com/stoplightio/prism/blob/7869d8819af49a9a4d96c6bc64a69f7219cc719c/test-harness/README.md). There is no hassle with it. So what's wrong with that?

test-harness/index.js
lag-of-death: In cases when we do not assert `expect(reqRes).toStrictEqual(masterFile);`, where do I put assertions then? Since I cannot just `expect(reqRes).toStrictEqual(masterFile);` for a dynamic response, how do I now change this? Do you think all of the assertions should be moved to [`requests.js`](`https://github.com/stoplightio/prism/blob/7dab5507392b1108b7e4438177e5c9b6ab5403af/test-harness/requests.js`)? And also, where do I move `describe`s and `test`s content (descriptive strings)?

test-harness/README.md
philsturgeon: Oh this is so cool! Sounds good.

examples/petstore.oas2.json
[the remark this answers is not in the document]
XVincentX: Yeah I was wondering whether you found these new resources;
1. somewhere in the internet
2. Prism v2 repository
3. authored them personally

test-harness/index.js
XVincentX: It is documented in the CLI page. The main Readme file has a link. https://github.com/stoplightio/prism/blob/master/packages/cli/README.md

I have no problems to add it to the main file as well though!

test-harness/index.js
XVincentX: Ok that's weird, I'll probably look into and see what's wrong.

examples/petstore.oas2.json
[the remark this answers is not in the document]
lag-of-death: So `/no_auth` paths are taken from the already existing ones but without security (so, in each of the `/no_auth` paths there is `security: []`). For example, `/no_auth/pets/findByStatus` is just `/pets/findByStatus` without any security. Hope it clears this up ;)

test-harness/README.md
lag-of-death: So I did some changes to this now, you can check it out :)

test-harness/createMasterFiles.js
lag-of-death: thanks, great tip :+1:

test-harness/index.js
lag-of-death: @XVincentX, I think this is good for now, we can always change it when it turns out that the current implementation is not good enough

test-harness/index.js
lag-of-death: Hey @XVincentX, so I was also trying to spawn `prism` with `circleci` step, but was having: https://github.com/circleci/circleci-docs/issues/1037
for some, it was "resolved" with adding a delay of 10 seconds or something, but I didn't find this to always be so

.circleci/config.yml
```diff
@@ -21,11 +21,21 @@ jobs:
       - run:
           name: Install Dependencies
           command: yarn
+      - run:
+          name: Create Linux CLI binary
+          command: HOST=host yarn build.binary
```
philsturgeon: what does this host bit mean?

examples/petstore.oas2.json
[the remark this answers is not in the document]
philsturgeon: Ok this is fine.

test-harness/index.js
philsturgeon: There are supposedly some windows support benefits or something. This is the way it's generally done everywhere else so we might as well for consistency even if nothing else.

.circleci/config.yml
lag-of-death: just updated the readme with:
```
`HOST` variable is used to specify what binary we want to build.
`HOST` might be set to anything that is considered a target in `pkg` [supports](https://github.com/zeit/pkg#targets).
```

https://github.com/stoplightio/prism/blob/SO-145_add_test_harness/test-harness/README.md#alternatively-run-any-of-the-following

LMK if still not clear :)

lag-of-death:
@philsturgeon, what suggested change do you have in mind? There's a couple of them. @XVincentX, please take a look if you find the time, thaaanks! :)

XVincentX:
Let me give another round of checks, I'll make it today.

test-harness/index.js
XVincentX: Yeah exactly, in general `path.join` is going to be always better than any implementation we can come up with. It's clearly not blocking, just saying.

package.json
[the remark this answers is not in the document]
XVincentX: I think that the scripts sections should really be stuff related to the real package development. 

For me (but that's a personal opinion) this is just distracting noise. If you really want these kind of scripts, I would create a `scripts.sh` file and put that stuff here. Not in `package.json` for sure. That's for Node, not for other things.

We can keep it here if you want, it's not a problem — but expect me to transfer these in the `.circleci/config` file once we testing harness goes in "maintenance" mode.

test-harness/index.js
philsturgeon: @lag-of-death this was the suggestion i was suggesting you implemented.

XVincentX:
> lease see my answers before I do any more coding on this. And gold master files are recorded with ./test-harness/createMasterFiles.js as described in ./test-harness/README.md. There is no hassle with it. So what's wrong with that?

In theory, there's nothing wrong. My point is that `jest` is providing us a great tool (snapshot) with great support for diffs, battle tested, great developer experience, pointing out the diffs and stuff, which is exactly what we need to do.

Personally for me I'd have leveraged them with no brainer instead of writing them manually.

In general, I always think (with rare occasions) that other people write better software than me, so I prefer to use other solutions.

If @philsturgeon is fine and you feel comfortable with the solution you have —  I guess it's ok anyway.

package.json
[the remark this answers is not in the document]
philsturgeon: I dont imagine that a developer will be doing this on a regular basis, or even ever. 

Asking why we shouldn't add code is backwards, we should ask why we should add code. Otherwise we get a kitchen sink. :)

XVincentX:
I'm checking out your branch to find out what's wrong with the `subprocess.kill`. Hold on!

lag-of-death:
@XVincentX, are we OK here? Or do you want to test that process killing?

XVincentX:
Let me have one last view and I'll merge this in an hour or so.

test-harness/index.js
```diff
@@ -0,0 +1,265 @@
...
+  };
+}
+
+const spec = process.env.SPEC || join(__dirname, '/../examples/petstore.oas2.json');
+const specs = spec.split(',');
+
+const createSpec = (specPath, prismCmd) => {
+  return () => {
+    describe(specPath, () => {
+      let prismProcessHandle;
```
lag-of-death: @XVincentX, the previous version didn't need any `let`. I don't see how is this and later mutation with `prismProcessHandle = exec(prismCmd);` better? I'd rather go back 2 commits.

XVincentX: Let me finish first :)

lag-of-death:
@XVincentX, I've just check it on my local linux machine, and now `jest` is not exiting.

XVincentX:
@lag-of-death Try that again, I moved the (unfortunate) required flag on the command level.

lag-of-death:
Still not OK, try to reproduce:
1. be sure `prism` is not running, ie: `fuser -k 4010/tcp`
2. `yarn test.binary`
3. `SPEC=./examples/petstore.oas2.json BINARY=./cli-binaries/prism-cli-linux yarn run.binary`

3rd step is now failing because `prism` is still running.

lag-of-death:
thanks @XVincentX, I'm merging this, let's address Windows at a later time

pytlesk4:
Too late to the party, but I have some questions.

- To add a new test case, do I need to update test-harness/index.ts? Pretty sure the answer is yes, so then my next question is why wouldn't this be table driven?

- Is there any way to run a one-off test?
  - This would be useful when the test harness fails in ci, and I want to focus on fixing one test at a time locally on my machine.

- Is it possible to specify the version of Prism? Would be a nice to see what was broken per-version.

- Why aren't we using typescript?

- It looks like it isn't possible to run these tests in parallel, is that correct?

lag-of-death:
Hi @pytlesk4,

1. Yes. I'm not sure what you mean by `table`, is it DB one? Here I'll assume that you mean `js array`. So yeah, it could be done like so. We could move all `describe`s and `test`s content (the descriptions) to the array. And we could also move the assertions there (this one is a bit tricky because of static/dynamic responses from `prism`). But yeah, doable, and please, feel free to create a card for it under https://stoplightio.atlassian.net/browse/SO-223

2. You can easily just do `test.only` in `test-harness/index`.

3. So for now `prism` v3 is always run. You can optionally run `prism` v2 when you set [`RUN_V2_TESTS=1`](https://github.com/stoplightio/prism/blob/master/test-harness/README.md#to-run-tests-run-any-of-the-following)

4. It was intended to write this in TS, but eventually for SO-145 it was agreed that JS is OK for now. Feel free to create a card for it under https://stoplightio.atlassian.net/browse/SO-223

5. These tests are single file now, and from what I understand, should run sequentially. Also, when we run both `prism`s (v2 and v3) we first start up v3 binary and then v2 binary, so this also happens sequentially. The same happens when we have 2 specs for v3 (oas2 and oas3), we first start up `prism` v3 with one spec, kill it, and then start it with the other spec. We could possibly start all `prism`s on different ports and have it paralleled. So could be done in the future.

PR #312, opened 2019-05-21
chore: move dev deps to root

marbemac:
Open for discussion, but IMHO it makes sense to move the devDependencies to the root workspace rather than duplicating things like `typescript` and `@types/lodash` across packages (and having to keep their versions in sync, etc).


XVincentX:
I am generally pro for this, but not for *all* the dependencies.

For example, while I agree that having a centralised Jest and Typescript and TSLint configurations definitely makes sense (and we should do that, 👍) I really do not think we should put `oclif` (which is a specific dev-dependency of the CLI) as a shared dependency.

My counter points:

1. Really makes understanding what packages need way more difficult. Imagine one day we want to separate these packages — that will be a problem. Reverse engineer the specific dev dependency won't be fun.
2. There's no real disk space benefit, since Yarn workspaces use a single `node_modules` directory anyway
3. I haven't verified, but I _think_ TypeScript could start screaming about this organisation.

---

@marbemac Is this PR attempting to solve a specific issue?

marbemac:
> @marbemac Is this PR attempting to solve a specific issue?

Yeah, in a separate branch I noticed the Typescript version was behind, and thus updated it. However, I did not realize that some of the other packages also declare typescript as a dev dependency, so I started getting some weird errors in vscode. When I noticed that we're declaring dev deps in multiple packages, I figured I'd open a PR that consolidates them down to root.

> My counter points:

1. 🤷‍♂ do you reeeeally think it's a possibility that we break this repo apart in the next year? I understand this hypothetical, but it doesn't seem like a realistic one imho.
2. Correct, was never about disk space.
3. Works fine for me.

We can indeed go with "sometimes put devDependencies in root workspace, sometimes declare them in their specific package (unless they're used by more than one package?)" message, but to me that is more complicated to explain than "devDependencies go in the root package.json file".

All that said, I swear I'm not passionate about this one way or the other :). I just feel like a "dev deps go in root" is a simpler message for internal and external contributors. Will leave to you and @philsturgeon to decide what to do!

XVincentX:
Ok, I think we can move with this then; can we at least to add to our tslint configuration file [this rule](https://palantir.github.io/tslint/rules/no-implicit-dependencies/) so that we do not require stuff that's not declared?

PR #314, opened 2019-05-21
feat: Implement header mocking functionality SO-227

### LINKED ISSUE #311: Not mocking response headers
Hi - 

I'm updating to openAPI 3.0, and prism doesn't seem to be sending response header examples.  My JSON looks like this:

```"responses": {
          "200": {
            "headers": {
              "X-Entitlements": {
                "schema": {
                  "type": "string",
                  "example": "lesterlesterlester"
                }
              }
            },
            "content": {
              "application/json": {
                "description": "A successful request",
                "schema": {
                  "type": "object",
                  "properties": {
```

I get the response, but the headers are mising. Am I doing something wrong or is this not supported yet?

```HTTP/1.1 200 OK
content-type: application/json
content-length: 1227
Date: Mon, 20 May 2019 18:15:27 GMT
Connection: keep-alive

{"userSession":{"user":```

### PR DESCRIPTION
**Pull request by XVincentX**
Closes #311 
Requires https://github.com/stoplightio/graphite/pull/179

### CODE REVIEW

File: packages/http-server/src/__tests__/server.oas.spec.ts [hunk 1]
```diff
@@ -206,6 +206,30 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
...
     checkErrorPayloadShape(response.payload);
   });
+
+  test('should mock the response headers', async () => {
+    const response = await server.fastify.inject({
+      method: 'GET',
+      url: '/user/login?username=foo&password=foo',
+    });
+
+    // OAS2 does not support examples for Headers, to they MUST be always generated automagically,
```
philsturgeon: Does graphite, and therefore prism, support `x-example` for OAS2? If so it would be ok to have default behavior tested like this and then a targeted unit test showing the support.

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -206,6 +206,30 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
...
+      'x-rate-limit': file === 'petstore.oas3.json' ? '1000' : expect.any(String),
+      'x-stats': file === 'petstore.oas3.json' ? '1500' : expect.any(String),
+      'x-expires-after': expect.any(String),
+    };
+
+    for (const headerName of Object.keys(expectedValues)) {
+      expect(response.headers).toHaveProperty(headerName, expectedValues[headerName]);
+    }
+
+    // Make sure that the generated examples foolow the 'format' keyword in any case.
```
philsturgeon:
```suggestion
    // Make sure that the generated examples follow the 'format' keyword in any case.
```

packages/http-server/src/__tests__/server.oas.spec.ts
philsturgeon: That can be a follow up feature btw no need to scope creep this one.

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -49,22 +51,52 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
...
 }
+
+function isINodeExample(nodeExample: INodeExample | INodeExternalExample | undefined): nodeExample is INodeExample {
+  return !!nodeExample && 'value' in nodeExample;
+}
+
+function computeMockedHeaders(headers: IHttpHeaderParam[], ex: IExampleGenerator): Promise<Dictionary<string>> {
+  const headerWithPromiseValues = mapValues(keyBy(headers, h => h.name), async header => {
+    if (header.content && header.content.schema) {
+      return ex.generate(header.content.schema, 'application/json');
```
philsturgeon: Everything makes sense to me apart from this bit, why are we saying that headers are JSON? Does this do anything or is the value essentially ignored because they can only really have strings in headers anyway?

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: At the moment no, x-example is not supported in graphite.

packages/http/src/mocker/HttpMocker.ts
XVincentX: Good question. That parameter is indeed useless and it's basically cheating the example generator, which is requiring this parameter and it MUST be application/json otherwise it throws an exception.

https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/generator/JSONSchemaExampleGenerator.ts#L27

I think I'll just destroy it as part of the work to use OpenAPI Sampler.

philsturgeon:
Perfect. Thank you. 

--
Phil Sturgeon
@philsturgeon

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -219,7 +232,7 @@ const helpers = {
       findResponseByStatusCode(httpResponses, '400') ||
       createResponseFromDefault(httpResponses, '422');
     if (!response) {
-      throw new Error('No 422 response defined');
+      throw new Error('Nor 422, 400 or default response defined');
```
philsturgeon: "Neither Frank, nor Bob, knew how OpenAPI examples work."
```suggestion
      throw new Error('No 422, 400, or default responses defined');
```

XVincentX:
@philsturgeon @chris-miaskowski Ready for review!

packages/http/src/mocker/HttpMocker.ts
XVincentX: @philsturgeon I've dropped the bomb: https://github.com/stoplightio/prism/compare/feat/simple-generator?expand=1

PR #315, opened 2019-05-23
[SO-194] Prism (v3) : Integration tests for ?__example=examplename

lag-of-death:
## Checklist

- ~[ ] Tests have been added (for bug fixes / features)~
- ~[ ] Docs have been added / updated (for bug fixes / features)~

Adding tests for an already existing functionality.

No change in the functionality itself.


philsturgeon:
Agree with @XVincentX we want to show example selection is doing its thing, we should also have a third test which shows what happens when an invalid example is requested.

Does it error or just ignore it and pretend you didn’t even request an example?

XVincentX:
You should receive an error message if the example you're selecting is not existing.

lag-of-death:
@phil, @XVincentX, I think that the concerns are addressed now. Can one of you approve if OK?

philsturgeon:
Nicely done, merge away!

PR #322, opened 2019-05-25
refactor: Payload Generator improvements SO-232

XVincentX:
This PR revisits a little bit the way we generate the examples from JSON Schemas.

1. Remove the interface+class boilerplate (for a single method) and export a simple function signature somebody can implement
2. Remove the serialisation responsibility from the schema generator — this is something the framework should do somewhere else
3. Refactor the tests to use less snapshots and more assertions

Note: While refactoring the code I also spotted a test passing incorrectly because of a spy that wasn't reset. https://github.com/stoplightio/prism/pull/322/files#diff-e53e98c10a813d2d271cc0fa5c55ddcaR201


packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { generate } from '../JSONSchema';
+
+describe('JSONSchemaExampleGenerator', () => {
```
philsturgeon: This name is wrong now?

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -217,10 +217,10 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
     // accorging to the schema
 
     const expectedValues = {
-      'x-rate-limit': file === 'petstore.oas3.json' ? 1000 : expect.any(String),
-      'x-stats': file === 'petstore.oas3.json' ? 1500 : expect.any(String),
+      'x-rate-limit': file === 'petstore.oas3.json' ? 1000 : expect.any(Number),
+      'x-stats': file === 'petstore.oas3.json' ? 1500 : expect.any(Number),
       'x-expires-after': expect.any(String),
-      'x-strange-header': file === 'petstore.oas3.json' ? 'string' : '{}',
+      'x-strange-header': file === 'petstore.oas3.json' ? 'string' : {},
```
philsturgeon: This but confuses me. Why was it ‘{}’ before and why is it {} now? They both seem like funny things to happen, don’t we just want an empty string for a header?

packages/http-server/src/server.ts
```diff
@@ -71,9 +71,9 @@ const replyHandler = <LoaderInput>(
           reply.headers(output.headers);
         }
 
-        reply.serializer((payload: unknown) => payload).send(output.body);
```
philsturgeon: This workaround was a fix for double encoded JSON, as fasting seems to assume it should JSON encode whatever you pass it. Is this no longer happening?

XVincentX: Not exactly.

The reason this workaround was required was because the example generator was also taking care of the serialisation — so we had to bypass the one that the framework (Fastify) was offering.

Now that I separated the concepts (the serialisation is something the example generator should not care about) — we can remove the "do not serialise" part and let the framework do its job.

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: Good question, I actually wanted to brief/ask you about this anyway.

In OpenAPI 3 documents the schema has a separate property in the Header object (`example` or `examples`) — in OAS2 the schema is "melted" with the header definition itself.

For this reason, JSON Schema Faker is kind of confused and generates an empty object for a schema that's not really complete/defined.

I can add a small workaround to standardise the behaviour. Up to you!

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -217,10 +217,10 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
     // accorging to the schema
 
     const expectedValues = {
-      'x-rate-limit': file === 'petstore.oas3.json' ? 1000 : expect.any(String),
-      'x-stats': file === 'petstore.oas3.json' ? 1500 : expect.any(String),
+      'x-rate-limit': file === 'petstore.oas3.json' ? 1000 : expect.any(Number),
+      'x-stats': file === 'petstore.oas3.json' ? 1500 : expect.any(Number),
       'x-expires-after': expect.any(String),
-      'x-strange-header': file === 'petstore.oas3.json' ? 'string' : '{}',
+      'x-strange-header': 'string',
```
philsturgeon: When we spoke on slack I meant this should be an empty string, because an empty string is more logical than an empty object.

This should follow the same rules as the response bodies, which are so far not quite behaving as wed expect. 

static: should try the various sources of example, then fall back to type based defaults if required, or null if not required.
dynamic: lorum ipsum or random numbers etc

When we implement static as the default, and have dynamic on the switch, I would expect these headers to follow those same rules. Does that make sense?

XVincentX: Sure! 9a80dcb

PR #323, opened 2019-05-27
Logging Prism's decisions.

XVincentX:
All right, be prepared.

#### TL; DR

1. Prism can now log stuff from the negotiation. 
2. The negotiation and logging process can't make Prism crash anymore for any reason.

#### Long version

The following PR, in theory, implements logging for the negotiator. Practically, though, this PR is laying down the foundation for the logging for the future hosted version as well and hopefully start moving Prism's code in a different direction.

There were some main principles I've been keeping in mind while writing the code for this stuff:

1. Prism should not process/format/react to the logs. Somebody else should do it; doing stuff on logs is usually slow and Prism should instead answer all the requests as fast as possible.
2. The negotiation process in Prism is currently composed by numerous and nested function calls and last thing I wanted to do is to carry over this `logger` parameter in all the function and having to deal with it everywhere
3. The logging cannot be a singleton defined somewhere — because the logging instance will be provided externally (Prism Http Server, in this case)
4. The logging process and the negotiation process should never make Prism crash. In any circumstance. I repeat, in *ANY* circumstances.

Let's now see how I've tamed them down.

1. Use [Pino](https://getpino.io). If you go on their website the write that "it's the fastest logger on the world" but you can ignore that, that was not the reason why I went with it. The main points were:
    - It's included with [fastify](https://fastify.io) — and so we're not introducing a new dependency. We've been shipping it since forever.
    - Pino does not include any way to process/alert/react to logs; it simply outputs the log statements as JSON on the stdout (or any stream you provide to it). It means that all the processing must be off process and I think this is great. It means that, by design, there's no way somebody can make messes with logs and do any action that's logging dependant. You can't break what's not even included.

So where's the log processing happening in our case? In the CLI. You can see from the code that, whenever the production environment is detected (or the `-m` flag is passed) the [CLI will effectively use the cluster module to fork the process](https://github.com/stoplightio/prism/pull/323/files#diff-354729d6ad7943965bb0088ba7d8f7d8R36), run Prism Server and funnel its stdout logs back to the CLI, where Signale will pretty print them. When Prism will be hosted somewhere, we'll likely do the same.

In case you're running the things locally to write code and test the stuff, you'll see that the fork is avoided and all is done in process. This is done fundamentally because it's going to be way easier to debug stuff in a single process. This is exactly the same that Jest does with the `--runInBand` flag.

image

You can see from the image what when the `-m` flag is specified, you'll find an additional process with node running — this is exactly the forking happening.

Note: if `NODE_ENV===production` the forking is enabled by default, so that if you use the CLI in a Docker Environment you get what the production behaviour.

2. Inject the logger through currying. Thanks to the partial application I can basically write a function that takes the _regular_ arguments which returns a function needing an additional dependency to run.
`const negotiate = (arg1, arg2, arg3) => logger => { // real body }`. For this, I've employed `fp-ts`'s reader implementation, that allows me to do exactly that: write code that has regular arguments and returns something that will need a logger to be executed. By leveraging `chain` and `map` method, intermediate functions can easily introspect and modify the result in the middle without having to worry about the logging presence. It'll be put as last step. Moreover, if one day we will want to make the Router start outputting logs as well — it's really easy. All we need to do is make the router spit up a Reader and chain it with the following one in the mocker. All good, profit.

3. See point 2. With Currying I can pretend the inner functions will eventually have a logger, it's resolution is happening in a _completely different npm package_; this is important to understand why I could not simply have a shared instance to require from somewhere.

4. This happened as an effect of the reader. Since the logger in injected at the end, the traditional try catch in the middle of the code to decorate the errors does not work anymore. I could have workaround this but I decided instead to change the approach and make sure the negotiation and the logging process never throws exceptions. Therefore the code has been modified to use the `Either` implementation that's included in `fp-ts`, which is not that hard to understand. If you check out the `NegotiatorHelper.ts` you can see there are not that many changes and more importantly nested exceptions are now been flattened — which is good. Clearly there are 1-2 things that aren't *that* cool but I'm on all of them. In particular:
    * Jest should support wrapped helpers so I can remove my home made helpers https://github.com/facebook/jest/issues/8540
    * A built-in `ReaderEither` could let me remove some [nested maps](https://github.com/stoplightio/prism/pull/323/files#diff-e18554f8d62727c21aaa7a2ea6de4a72R65). I'm quite sure @gcanti will make it happen soon :trollface: 

---

### Outstanding points
- [x] Refactor `mock.ts` to look a little bit more human
- [x] Resolve the null-ness of `logger` component
- [x] Understand whether we're logging enough or too much
- [x] CLI Readme Update
- [x] Deduplicate the testing helpers
- [x] General review and cleanup

SO-231


packages/http/src/mocker/HttpMocker.ts
```diff
@@ -32,48 +40,68 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
       throw new Error('Http request is not defined');
     }
 
-    // setting default values
-    const acceptMediaType = input.data.headers && caseless(input.data.headers).get('accept');
-    config = config || { mock: false };
-    const mockConfig: IHttpOperationConfig =
-      config.mock === false ? { dynamic: false } : Object.assign({}, config.mock);
+    return new Reader<Logger, IHttpOperationConfig>(logger => {
```
philsturgeon: Does this feel a bit invasive to you? I am not sure about it. Do we need to embed our code in a function just to get logger doing its job?

packages/http/src/types.ts
```diff
@@ -15,7 +15,7 @@ export type TPrismHttpComponents<LoaderInput> = Partial<
 >;
 
 // TODO: should be complete | and in the @stoplight/types repo
-export type IHttpMethod = 'get' | 'put' | 'post' | 'delete' | 'options' | 'head' | 'patch' | 'trace'; // ... etc
+export type IHttpMethod = HttpMethod | 'trace';
```
philsturgeon: Wait why is trace not in there?

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,97 @@
...
+    logInstance.fatal(e.message);
+    cluster.worker.kill();
+  }
+}
+
+function pipeOutputToSignale(stream: Readable) {
+  stream.pipe(split(JSON.parse)).on('data', (logLine: LogDescriptor) => {
+    const logLevelType = logLevels.labels[logLine.level];
+
+    let prefix = LOG_COLOR_MAP[logLine.name].black(`[${logLine.name}]`);
```
lag-of-death: no need for `let`, let's have `const prefix = constructPrefix(...)` so that:
1. there is no mutation
2. `pipeOutputToSignale` is shorter

packages/core/src/factory.ts
```diff
@@ -81,14 +80,22 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         let output: Output | undefined;
         if (resource && components.mocker && (configObj as IPrismConfig).mock) {
           // generate the response
```
lag-of-death: `// generate the response` -> `generateTheResponse`
and later (if it must be mutated): `output = generateTheResponse`
I don't believe adding in comments justify bigger functions' bodies

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -37,46 +37,71 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
...
       }
-    } else {
-      negotiationResult = helpers.negotiateOptionsForValidRequest(resource, mockConfig);
-    }
 
-    const mockedBody = computeBody(negotiationResult, payloadGenerator);
-    const mockedHeaders = computeMockedHeaders(negotiationResult.headers || [], payloadGenerator);
+      return mockConfig;
+    })
+      .chain(mockConfig => {
```
lag-of-death: can the function passed to `.chain` be moved out?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -37,46 +37,71 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
...
+                ),
+              ),
+          );
+        } else {
+          return withLogger(logger =>
+            logger.success('The request passed the validation rules. Looking for the best response'),
+          ).chain(() => helpers.negotiateOptionsForValidRequest(resource, mockConfig));
+        }
+      })
+      .chain(result => {
```
lag-of-death: can the function passed to `.chain` be moved out?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -37,46 +37,71 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
...
-    const acceptMediaType = input.data.headers && caseless(input.data.headers).get('accept');
-    config = config || { mock: false };
-    const mockConfig: IHttpOperationConfig =
-      config.mock === false ? { dynamic: false } : Object.assign({}, config.mock);
+    let payloadGenerator: PayloadGenerator = generateStatic;
 
-    if (!mockConfig.mediaTypes && acceptMediaType) {
-      mockConfig.mediaTypes = acceptMediaType.split(',');
+    if (config && typeof config.mock !== 'boolean' && config.mock.dynamic) {
+      payloadGenerator = generate;
```
lag-of-death: can't we just do:
```
const payloadGenerator = (config && typeof config.mock !== 'boolean' && config.mock.dynamic) && generate;
```
?

lag-of-death:
any chance on dividing this into smaller PRs? Ie there is something only VS-code related and I think it could be easily move out of this PR, no?

packages/http/src/mocker/HttpMocker.ts
XVincentX: Not exactly, but I've been able to refactor it in a way very similar. ffffb9d

packages/cli/src/commands/mock.ts
XVincentX: Good point! Done in 8d9f6d2

packages/http/src/mocker/HttpMocker.ts
XVincentX: Yes! I've created an `assembleResponse` function with the code. b95eef9

packages/http/src/mocker/HttpMocker.ts
XVincentX: Yes! Done in 34638c0 with the new `negotiateResponse` function!

lag-of-death: there even is `if/else`! I love it :+1:

packages/http/src/mocker/HttpMocker.ts
XVincentX: I'm resolving this because the code has been refactored.

packages/http/src/types.ts
XVincentX: It's in the `HttpExtendedMethod` type but it has some other verbs we do not care about.

packages/core/src/factory.ts
XVincentX: I fail to understand what you're suggesting. Can you elaborate more?

XVincentX:
>  there is something only VS-code related

That is correct, but that's kind of required. The change in particular is to make sure VSCode is able to attach the debugger to the child processes when running Prism CLI in production mode

https://github.com/stoplightio/prism/pull/323/files#diff-3090e25fed17a6bd57433c6bbadc451aR27

packages/core/src/factory.ts
lag-of-death: Sure. So the code that is now marked with the comment could just go to an external function. And then we could populate `output` with the result of running this function. Maybe we could even make this `output` a `const`?

XVincentX: Aaah ok I got it, let me see what we can do here!

lag-of-death: cool :+1:

lag-of-death:
@XVincentX, I'm a bit confued here. What about other IDEs then? Is it that `prism` is only to be developed with `VScode`? Maybe there is some policy on IDEs and `prism`?

XVincentX:
@lag-of-death Oh no no — we can also remove the whole .vscode directory from the git repository and keep it locally if you prefer, I do not mind; I found it there, versioned and I modified to make sure it works property with the multi process use case.

lag-of-death:
@XVincentX, people use different `IDE`s and for those, that don't use `VScode`, the mentioned changes are just unnecessary. Since this is OSS, and there are a couple of IDEs out there, I don't see a need to keep `.vscode` in the repo. The benefit? Less code changes in PRs. Unless of course there is already people depending on the config

XVincentX:
@lag-of-death yeah, I'm fine with it; I modified it for the people, I do not use VSCode, I use Sublime.

@philsturgeon thoughts?

philsturgeon:
@lag-of-death yep @XVincentX said "people use other IDEs than VS Code" to me today too. We're all telling each other even though we all already know. 

You've both made good suggestions, let's remove it.

packages/core/src/factory.ts
lag-of-death: @XVincentX, any ETA on this?

XVincentX: I'm on it now, I'm trying to see if there's a way to make it clear :)

XVincentX: @lag-of-death I've tried but I haven't been able to find anything that would really satisfy me. The best I got was having two functions taking 3000 parameters anyway and not making the code any clear.

I think this needs a more in depth refactor which maybe it's not worth doing now.

If you have a specific suggestion/idea in mind I'm absolutely open; I haven't personally found anything that appealing. Let me know your ideas!

lag-of-death: @XVincentX, I didn't have anything specific in mind. Thanks for looking into it. Let's then have it as it is, but only for now. After what I saw, I think that refactoring `factory`/`process` deserves a separate card.

XVincentX:
@chris-miaskowski Totally, I was going to wait for you in any case

packages/core/src/factory.ts
```diff
@@ -81,14 +80,22 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
+                config: configObj,
+              },
+              defaultComponents.mocker,
+            )
+            .run(components.logger.child({ name: 'MOCKER', input }))
+            .fold(
+              e => {
+                throw e;
+              },
+              r => r,
```
chris-miaskowski: `r`?  how about `x`? :trollface: 
Does it stand for `result`?

packages/core/src/logger.ts
```diff
@@ -0,0 +1,33 @@
+import * as pino from 'pino';
```
chris-miaskowski: General inquiry: does `pino` work in browser context? Remember that one of the Prism's requirements is to be able to work both in nodejs and browser (e.g. we want to use it in request maker).

It caught my attention because I wasn't sure if `core` is actually a good place to store this stuff in. It looks that `pino` is mostly CLI specific (?)

packages/http/package.json
```diff
@@ -23,6 +23,8 @@
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
+    "fp-ts": "^1.18.2",
```
chris-miaskowski: I hate to say this but `fp-ts` seem to require some work on the documentation end. Cool to use new tools but just FYI it may be a steep learning curve for contributors.

packages/core/src/logger.ts
XVincentX: 1. Yes! http://getpino.io/#/docs/browser

2. Not entirely correct: the CLI is the place where the logs get captured from the stream and processed/visualised, but its usage spread in the whole application. Moreover, Pino implements the log4j interface so it should be, if needed, easily replaceable with anything we like in the future (Winston, Buyan). Does that clarify?

packages/core/src/factory.ts
XVincentX: Yes, it's a shortcut for result, but I can change it, I do not mind

lag-of-death: This clearly is `identity`. Can use `_.identity`

packages/http/package.json
lag-of-death: Personally, when I saw it was a big positive! I didn't use `fp-ts`, but I'd like to see fp used more in our codebase (even if it was only limiting side effects). I'm all for keepking `fp-ts`. This is clearly a good sign.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
+      if (!result) {
+        logger.info(`Unable to find a ${code} response definition`);
+        return createResponseFromDefault(httpOperation.responses, code);
       }
-    }
-    // if no response found under a status code throw an error
-    throw new Error('Requested status code is not defined in the schema.');
+
+      return result;
+    }).chain(responseByForcedStatusCode => {
```
chris-miaskowski: What's the benefit of using chain vs if?

packages/core/src/factory.ts
XVincentX: @lag-of-death I can make the change, but is it going to improve the readability? At first glance it makes me scratch a little bit more my head instead of clarifying. What do you think?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
+      if (responseByForcedStatusCode) {
+        return helpers
+          .negotiateOptionsBySpecificResponse(httpOperation, desiredOptions, responseByForcedStatusCode)
+          .chain(either =>
+            either.fold(
```
chris-miaskowski: One of the goals was to improve readability but I can't tell this did the trick.
We're a increasing the cognitive complexity. Contributing to Prism is already hard (TypeScript, monorepo, workspaces/build and now we're adding functional programming on top of it - I have the feeling we're narrowing potential contributors to 0.00001% of developers ;) )

chris-miaskowski: There is nothing to do with this comment other than maybe not using `chains`?

packages/http/package.json
XVincentX: > some work on the documentation end

This is true in general, the documentation is lacking. Speaking with the authors they're working on it as part of the next release but, in general, it's not the library documentation you need, but a mini training on the concepts. I'll have a presentation about this in [August in Columbus](https://www.javascriptandfriends.com/speakers), you might want to watch the recording :trollface:

> but I'd like to see fp used more in our codebase (even if it was only limiting side effects)

It's great to have you into this stuff! My ultimate goal would be to make Prism a functional mock server if possible. For now I started with logging which is a small project and the risk is contained. As you probably noticed, at a certain point we need to go out from the functional world and throw an exception, but as this "virus" expands, little by little the `throw` statements will hopefully start to disappear!

packages/core/src/factory.ts
lag-of-death: @XVincentX, honestly I don't care, `x => x`, `r => r`, `_.identity` are all the same to me. The fact is though, that `x => x` and `r => r` might be more understandable for those, that don't know about `identity`. Up to you, if it was up to me, I'd most probably left it as is ;)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
+              () =>
+                helpers
+                  .negotiateOptionsForDefaultCode(httpOperation, desiredOptions)
+                  .map(innerEither =>
+                    innerEither.mapLeft(error => new Error(`${error}. We tried default response, but we got ${error}`)),
```
chris-miaskowski: Ok, will skip that part for now. You lost me on the `innerEither` ;)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
+                  ),
+              result => reader.of(right(result)),
```
chris-miaskowski: No idea what `reader.of` does. Wanted to read the documentation, here is what it says... https://gcanti.github.io/fp-ts/modules/Reader.ts.html#of-constant

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,101 @@
 import { Command } from '@oclif/command';
+import { createLogger, logLevels } from '@stoplight/prism-core';
+import chalk from 'chalk';
```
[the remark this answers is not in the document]
XVincentX: Good point, this could be a good candidate to extraction when we will have new commands. I can do it now. I'd defer this when we'll have the opportunity though.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
-      };
-    } else {
-      throw new Error(
-        `Request invalid but mock data corrupted. Neither schema nor example defined for ${response.code} response.`,
-      );
-    }
+      logger.success(`Found response ${result.code}. I'll try with.`);
+      return result;
+    }).chain(response => {
+      return withLogger(logger => {
```
chris-miaskowski: Note to self: second time I see `withLogger` in this function. I'd just remove the `chain` honestly. I think it just adds more nesting then we used to have.

packages/cli/README.md
```diff
@@ -70,6 +73,12 @@ OPTIONS
 _See code: [src/commands/mock.ts](https://github.com/stoplightio/prism/blob/v3.0.0-alpha.16/src/commands/mock.ts)_
 <!-- commandsstop -->
 
+## Running in production
+
+When running in development mode (which happens when the `NODE_ENV` environment variable is not set to `production`) or the `-m` flag is set to false, both the HTTP Server and the CLI (which is responsible of parsing and showing the received logs on the screen) will run within the same process.
+
+Processing logs slows down the process significantly. If you're planning to use the CLI in production (for example in a Docker Container) we recommend to run the CLI with the `-m` flag or set the `NODE_ENV` variable to `production`. In this way, the CLI and the HTTP server will run on two different processes, so that logs processing, parsing and printing does not slow down the http requests processing.
```
chris-miaskowski: Awesome idea, great that you thought about it!

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,101 @@
...
+          prefix: chalk.bgWhiteBright.black('[CLI]'),
+          message: `Dynamic example generation ${chalk.green('enabled')}.`,
+        });
+      }
 
-      if (server.prism.resources.length === 0) {
-        signaleInteractiveInstance.fatal('No operations found in the current file.');
-        this.exit(1);
+      if (multiprocess) {
+        signale.star({
```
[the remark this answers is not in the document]
XVincentX: Good point, effectively the whole prefix can be handled automatically. 4b486e3

packages/cli/src/commands/mock.ts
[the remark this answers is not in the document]
philsturgeon: Sounds good to me.

packages/http/package.json
lag-of-death: @XVincentX, so very cool. From `Professor Frisby's Mostly Adequate Guide To Functional Programming`:
```
It may come as a shock, but throw/catch is not very pure. When an error is thrown, instead of returning an output value, 
we sound the alarms! The function attacks, spewing thousands of 0's and l's like shields & spears in an electric battle 
against our intruding input. With our new friend Either , we can do better than to declare war on input, we can respond with 
a polite message.
```
And I do know there is `Either` somewhere in this PR! ;)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
+      return helpers
+        .negotiateByPartialOptionsAndHttpContent(
+          {
+            code,
+            dynamic,
+            exampleKey,
+          },
+          httpContent,
+        )
+        .map(contentNegotiationResult => ({
```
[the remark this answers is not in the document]
lag-of-death: hm.. why would mapping on `Either` be confusing?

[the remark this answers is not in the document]
XVincentX: I understand @chris-miaskowski's confusion. I had the same when I started to look into this in November.

Thing is the `map `we use on daily basis on an arrays and objects is a specialisation of a more general map that is basically the introspection in an "effect".

In case of an `Array`, the effect is the "repetition" of values
In case of an `Object`, the effect is the presence of multiple properties
On an `Either` though, the effect is that "something went correctly". 

Array and Either are effectively wrapping/representing an effect.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -53,236 +60,283 @@ function findResponseByStatusCode(
...
-        return {
-          headers: response.headers,
-          ...contentNegotiationResult,
-        };
-      } else {
-        throw ProblemJsonError.fromTemplate(
-          NOT_ACCEPTABLE,
-          `Could not find any content that satisfies ${mediaTypes.join(',')}`,
-        );
+            mediaType: 'text/plain',
```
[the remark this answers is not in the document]
XVincentX: Ah now I realise the reason of that PR — I can merge that branch and see how it goes when we settle down on the code here.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: `chain` is a fundamental piece — I'll try to explain it in another of the comment you did.

You have a point though, this is indeed a case-limit that will be addressed soon by the library itself by providing a more specific interface to handle this use case. So you're not wrong here!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Same as above, this will be beautifully solved when fp-ts will have `ReaderEither` — which is soon!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Good point, I'm opening a PR right now to make it more clear.

The definition is really simple though

`of: a => new Reader(() => a)`

Is fundamentally a shortcut to create a `Reader` returning a value directly without having to manually write the function.

Think of it as

`jest.fn().mockImplementation(()=> t)` vs `jest.fn().mockReturnValue(t)`

The reason this is required is because composition happens when types align — and so we need to make sure we always return a `Reader` to make the `chain` mechanism work.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: I'm resolving this in favour of another comment about `chain` that should clarify what's going on.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: `if` and `chain` do two different things, they cannot be interchanged.

To keep it short, `chain` is just like a `flatMap`

`[1,2,3,4].map(t=>t*2) = [2,4,6,8]`
`[1,2,3].flatMap(t=>[t, t * 2, t * 3]) = [1, 2, 3, 2, 4, 6, 3, 6, 9]`

---

In this context in particular, `chain` unwraps the effect for you so if you have an Either — you can concatenate it with another Either

```ts

const operationThatMightFail = getDataFromInternet();

operationThatMightFail
  .map(operation => operation.result)
  .chain(result => postInternetDataSomewhere(result))
```

In this case you can see:

1. You receive an `Either` from a function that gets data from the internet or fails.
2. In case it passes, it will call the `map` function that will effectively transform the result
3. Pass the data to chain; a chain function can return another `Either` that will be "executed"/"unwrapped" for you automatically, flattening the final Either you have to deal with.

Does that help?

CHANGELOG.md
```diff
@@ -18,7 +18,6 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ### Features
 
 - Prism now has got a static example fallback in case the `dynamic` flag is not enabled #347
-- Prism can now validate servers #351
```
philsturgeon: Oops!

README.md
```diff
@@ -173,7 +173,8 @@ Will give you the following error:
 ## What's next for Prism?
 
 - [x] Server Validation
-- [x] Content Negotiation
+- [x] Accept header validation
+- [x] Content header validation
```
philsturgeon: Oops!

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,97 @@
...
+import * as cluster from 'cluster';
+import { LogDescriptor, Logger } from 'pino';
 import * as signale from 'signale';
-import { ARGS, FLAGS } from '../const/options';
+import * as split from 'split2';
+import { PassThrough, Readable } from 'stream';
+import { ARGS, FLAGS, LOG_COLOR_MAP } from '../const/options';
 import { createServer } from '../util/createServer';
 
+function logEnabledFeature(message: string) {
```
philsturgeon: What does log enabled feature mean?

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,97 @@
...
+      flags: { port, dynamic, host, multiprocess },
       args: { spec },
     } = this.parse(Server);
 
-    signaleInteractiveInstance.await('Starting Prism…');
-
-    if (dynamic) {
-      signale.star('Dynamic example generation enabled.');
-    }
+    if (cluster.isMaster) {
```
philsturgeon: This change has so many new terms in it. cluster, stream, multipurpose, pino, master, signale, there is so much not mocking command leaking into the mocking command. I fear this will confuse the heck out of anyone who works on it.

Could you maybe make a stiry with sone notes how to refactor this? It's not just about reuse for a 2nd command, its a problem for understanding this one command.

README.md
XVincentX: Aren't we doing these?

CHANGELOG.md
XVincentX: This is correct, the server validation is not in alpha 15, but in alpha 16

packages/cli/src/commands/mock.ts
XVincentX: `logCLIMessage should be better 5734296

README.md
philsturgeon: I replaced Accept / Content with "Content Negotiation" in #357 because the two things are both part of content negotiation. If you did this intentionally can you revert it, and if you didnt mean to do this... could you look out for other accidental reverts?

packages/cli/src/commands/mock.ts
XVincentX: Yeah ok, this is definitely confusing; let me try to come up with a better way to reorganise the thing.

README.md
XVincentX: You were right. Fixed.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
lag-of-death: I never liked `chain` / `bind` names for it. `flatMap` is so obvious a name! :)

packages/cli/src/commands/mock.ts
```diff
@@ -1,43 +1,97 @@
...
+    });
+    logInstance.start(`Prism is listening on ${address}`);
+  } catch (e) {
+    logInstance.fatal(e.message);
+    cluster.worker.kill();
+  }
+}
+
+function pipeOutputToSignale(stream: Readable) {
+  function constructPrefix(logLine: LogDescriptor): string {
```
lag-of-death: hm.. if I'm not mistaken, we can just move out `constructPrefix` from `pipeOutputToSignale` (?)

XVincentX: Yeah, I'll probably refactor this command and offload most of the shit in the `createServer` file and go from there, hopefully that's going to help a lot in this command specifically. Stay tuned for updates.

packages/core/src/logger.ts
```diff
@@ -0,0 +1,33 @@
...
+      note: 10,
+      success: 11,
+      start: 12,
+    },
+    level: 'note',
+    base: {},
+    timestamp: false,
+  };
+
+  if (destination) return pino(options, destination);
```
lag-of-death: sorry, I had to... maybe a ternary ;) ?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -1,82 +1,106 @@
...
-    let payloadGenerator: PayloadGenerator = generateStatic;
+  }: IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>): Reader<Logger, Either<Error, IHttpResponse>> {
+    const payloadGenerator: PayloadGenerator =
+      config && typeof config.mock !== 'boolean' && config.mock.dynamic ? generate : generateStatic;
 
-    if (config && typeof config.mock !== 'boolean' && config.mock.dynamic) {
-      payloadGenerator = generate;
-    }
+    return withLogger(logger => {
+      // setting default values
```
lag-of-death: would it be much work to make this a separate function? Just like `negotiateResponse` ?

packages/cli/src/commands/mock.ts
lag-of-death: :+1:

packages/cli/src/commands/mock.ts
XVincentX: @philsturgeon I've teared out stuff as much as possible here. 72336ba

XVincentX: It's still a monster, but it should be very well enclosed in two nice functions.

packages/cli/src/commands/mock.ts
lag-of-death: @XVincentX, I've seen that `pipeOutputToSignale` is now in another file. But can we also `constructPrefix` move out of `pipeOutputToSignale`? I don't see a need for `constructPrefix` to be defined inside `pipeOutputToSignale` (but maybe there is (?)). And also, how about `pipeOutputToSignale` -> `pipeOutputToSignal` ?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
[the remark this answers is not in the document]
chris-miaskowski: @XVincentX do you still have that on your radar?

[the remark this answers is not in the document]
XVincentX: I have, I'm a little bit overwhelmed but I'll get it done.

Issue #324, opened 2019-05-27
Error 422, Missing Authorization header param


GuilhermeRizzottoLis:
When my requisition is with a header, the response is a 422, saying Missing Authorization Header param, but i put in the requisition -H "Authorization: bearer blabla". Is that an issue? When i use the Authorization as query, it works

lag-of-death:
hi @GuilhermeRizzottoLis, so basically, security validation is not yet supported and it should not matter if you send the token as a header or as a part of the query. Please see https://github.com/stoplightio/prism/issues/267#issue-440752688 for more on this.

I'm guessing that you have a specification that has something like:
```json
"parameters": [
  {
    "name": "Authorization Header",
    "in": "query",
    "required": true,
    "type": "string"
  }
]
```
(or `schema` instead of `type` if it is OAS v3)

If the above is true, then everything is OK (unless my assumption is wrong of course). It is OK because you're sending the required parameter. In the case, when you do not send it, you get `422` and it is, in fact, the desired behavior.

Does this clarify the things? If not, please attach the spec and the requests (the one with the header and the one when you have `Authorization Header` in the query?

GuilhermeRizzottoLis:
`"parameters": [
  {
    "name": "Authorization",
    "in": "header",
    "required": true,
    "type": "string"
  }
]`

in my case is in: header, when i put in: query it works

GuilhermeRizzottoLis:
the problem is not about the security, it could be another parameter passed by header, and it doesent works

GuilhermeRizzottoLis:
`curl -X GET -s -D "/dev/stderr" -H "Authorization = dasds" http://127.0.0.1:4010/templates | json_pp
`

lag-of-death:
Hi @GuilhermeRizzottoLis, thanks for the feedback. I've just checked it and indeed when we provide a required header param in a request, we end up with `Missing <<HEADER_NAME>> header param` in the response. Thanks for reporting this!
@philsturgeon, https://github.com/stoplightio/prism/pull/325 is not related to the bug reported by @GuilhermeRizzottoLis.

GuilhermeRizzottoLis:
Dou you guys pretend to fix quickly? Thanks for the support.

XVincentX:
Hey @GuilhermeRizzottoLis ,

thanks a lot for reporting this!

I've tried to reproduce the issue with a very minimal document; according to the conversation I inferred you're using an OpenAPI 2.0 document — so I did the same:

```json
{
[...]

PR #325, opened 2019-05-27
[SO-135] Fix falsely required param bug

lag-of-death:
[SO-135](https://stoplightio.atlassian.net/browse/SO-135)

## Checklist

- ~[ ] Docs have been added / updated (for bug fixes / features)~

Bug fix.

When all of the required query parameters are provided (for a request), we get a false bug:
`{"path":["query"],"code":"type","message":"should be string","severity":0}`

No.


philsturgeon:
Is this related to #324 ?

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -18,6 +22,12 @@ describe('HttpValidator', () => {
       it('returns validation errors for whole request structure', async () => {
         expect(await validator.validateInput({ resource: httpOperations[2], input: BAD_INPUT })).toMatchSnapshot();
       });
+
+      describe('when all required params are provided', () => {
+        it('returns no validation errors', async () => {
+          expect(await validator.validateInput({ resource: httpOperations[0], input: GOOD_INPUT })).toMatchSnapshot();
```
StefanDywersant: Please use `.toEqual([])` instead of `.toMatchSnapshot()` for simple expectations. E.g. this time the expectation is an empty array, so it would be way more readable when the expected result was in test code directly.

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -41,13 +48,22 @@ describe('HttpValidator', () => {
...
+
+      describe('when all required params are provided', () => {
+        it('returns no validation errors', async () => {
+          expect(
+            await validator.validateInput({
+              resource: httpOperations[0],
+              input: GOOD_INPUT,
+              config: { mock: false, validate: { request: { headers: false, query: true, body: false } } },
+            }),
+          ).toMatchSnapshot();
```
StefanDywersant: Please use `.toEqual([])` instead of `.toMatchSnapshot()` for simple expectations. E.g. this time the expectation is an empty array, so it would be way more readable when the expected result was in test code directly.

lag-of-death:
@philsturgeon, no, not related to https://github.com/stoplightio/prism/issues/324

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -84,12 +94,7 @@ describe('HttpValidator', () => {
   describe('validateOutput()', () => {
     describe('all validations are turned on', () => {
       it('returns validation errors for whole request structure', async () => {
-        expect(
-          await validator.validateOutput({
-            resource: httpOperations[1],
-            output: BAD_OUTPUT,
-          }),
-        ).toMatchSnapshot();
+        expect(await validator.validateOutput({ resource: httpOperations[1], output: BAD_OUTPUT })).toMatchSnapshot();
```
XVincentX:
```suggestion
        expect(validator.validateOutput({ resource: httpOperations[1], output: BAD_OUTPUT })).resolves.toMatchSnapshot();
```

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -18,6 +22,12 @@ describe('HttpValidator', () => {
       it('returns validation errors for whole request structure', async () => {
         expect(await validator.validateInput({ resource: httpOperations[2], input: BAD_INPUT })).toMatchSnapshot();
       });
+
+      describe('when all required params are provided', () => {
+        it('returns no validation errors', async () => {
+          expect(await validator.validateInput({ resource: httpOperations[0], input: GOOD_INPUT })).toEqual([]);
```
XVincentX:
```suggestion
          expect(validator.validateInput({ resource: httpOperations[0], input: GOOD_INPUT })).resolves.toEqual([]);
```

lag-of-death: @XVincentX, If only `husky` allowed me to do this ;>

XVincentX: I'm not following, what has husky to do with this?

lag-of-death:
```
ERROR: /home/lag-of-death/RubymineProjects/prism/packages/http/src/validator/__tests__/functional.spec.ts:28:11 - Promises must be handled appropriately
ERROR: /home/lag-of-death/RubymineProjects/prism/packages/http/src/validator/__tests__/functional.spec.ts:58:11 - Promises must be handled appropriately
husky > pre-commit hook failed (add --no-verify to bypass)
```

XVincentX: It's not husky, that's ESLint and it's actually right. I've modified the suggestion.

StefanDywersant: + you can remove `async` since there is no `await` inside

lag-of-death: @XVincentX, but now I have to know to return that `promise`. With `async/await` it is much more clear.

lag-of-death: I'm gonna go with `async/await` unless you think it's crucial to change it. And btw, all of the other tests are `async/await`. So let's stay consistent here as the alternatives are not in any way better.

XVincentX: It's not crucial, it can stay as it is!

philsturgeon:
You don’t need to return a promise from a test, you _can_ return a promise, but it’s not required. 

The rule should be: only add async when the function has an await”, not “always have async for consistency.” 

--
Phil Sturgeon
@philsturgeon

XVincentX:
@philsturgeon

> you _can_ return a promise

Not exactly. This still works because Jest has an internal queue (when using `expect(promise).resolves`) that will be used to track the status of the promise internally a wait before moving to the next test. This, in mocha, would be a dangling handle.

 The truth, in general, for other test frameworks and for the software in general — you should always return or store the promise.

lag-of-death:
@philsturgeon, I don't quite understand. I used `await` because `validator.validateOutput` returns a promise. And since I used `await` then I decorated the function with `await`. So, yeah, what are you trying to say?

philsturgeon:
The threading was lost because I replied via email. Don’t worry, it’s already merged. 👍🏼

PR #326, opened 2019-05-27
Add issue templates

philsturgeon:


marbemac:
1) I think we need to ask for the relevant OpenAPI file? It's very difficult to diagnose a bug without this context.
2) Do we want feature requests to be here or in the forum? If here, when are feature requests "closed"? If the answer is "when they are completed", or "when we are 100% sure we will never implement the feature", then won't this lead to an open issue count that balloons over time?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: 'bug: '
+labels: bug
+assignees: ''
```
XVincentX: Put my name! I want to die because of the issues :trollface:

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. macOS 10.14.4]
+ - Prism Version [e.g. v3.0.0]
+ - Install method [e.g. binary, npm, yarn, docker]
+
+**Additional context**
```
XVincentX: I'd really remove this — if people need to put additional details, they'll put them anyway, regardless of whether we'll write this or not.

tbarn:
@marbemac I think it will be hard in an open source project context to fully move over feature requests to the forum. It would be nice if they were there, but I think it would be more work to push people to post them there. 

Rather, we can focus on doing a good job of setting expectations for feature requests in issues and also make it easier for people outside of Stoplight to pick up requests if they have an interest in seeing something implemented.

.github/ISSUE_TEMPLATE/bug_report.md
tbarn: Agreed.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+**Note: For support questions, please use the [Stoplight Community forum](https://community.stoplight.io)**. This repository's issues are reserved for feature requests and bug reports. If you are unsure if you are experiencing a bug, the [Community forum](https://community.stoplight.io) is a great place to start.
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
+1. When I run this command: 'prism mock http://gist.github.com/an-esxample-spec-file.yaml --foo --bar`
```
philsturgeon: @marbemac I was implying it here but should I call it out way more explicitly? Suggestion welcome. I can also fix the typo.

.github/ISSUE_TEMPLATE/bug_report.md
philsturgeon: On your head be it!

.github/ISSUE_TEMPLATE/bug_report.md
philsturgeon: Ok I just made it more better.

philsturgeon:
@marbemac Just pushed some changes making it more clear we need an example. 

As for 2. I think asking developers to sign up for an account just to recommend an improvement will probably mean we just don't get that feedback at all. Frankly I would not if I was in their shoes. 

Some users will report feature requests and bugs on the forums and we should send them over here. 

Some users will ask for support over here and we should do our best to send them to the forums.

Some users will report bugs, which turn out to not be bugs, so we should help them here and close things. 

Some users will report feature requests that we do not want to implement for various reasons, and that is usually going to be because there is a different way to do the thing. We will just close those, that's fine, we don't have to do everything a users asks for, that would be unrealistic. We should leave feature requests that we _do_ want to implement open. @XVincentX will make a Jira ticket for anything we think we should do, and then it goes in the backlog and we prioritize accordingly. 

A lot of this is just up to @XVincentX's discretion. It's just humans having conversations after all. :)

PR #327, opened 2019-05-28
fix: make sure the proposed schema is valid before calling it

### LINKED ISSUE #324: Error 422, Missing Authorization header param

XVincentX:
This PR will make sure that the provided JSON Schema is effectively valid before trying to use it to validate the payload.

Closes #324 
SO-234

### DIFF
--- a/packages/http/src/validator/__tests__/index.spec.ts
+++ b/packages/http/src/validator/__tests__/index.spec.ts
@@ -188,8 +188,8 @@ describe('HttpValidator', () => {
     });
 
     describe('invalid schema provided', () => {
-      it('should not call the validate method', async () => {
-        const validateSpy = jest.spyOn(validatorUtils, 'validateAgainstSchema');
+      it('should not call the validateAgainstSchema method', async () => {
+        const validateAgainstSchemaSpy = jest.spyOn(validatorUtils, 'validateAgainstSchema');
 
         await httpValidator.validateInput({
           resource: {
@@ -223,7 +223,7 @@ describe('HttpValidator', () => {
         });
 
         expect(httpHeadersValidator.validate).toHaveBeenCalled();
-        expect(validateSpy).not.toHaveBeenCalled();
+        expect(validateAgainstSchemaSpy).not.toHaveBeenCalled();
       });
     });
   });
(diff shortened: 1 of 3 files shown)


packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -178,6 +186,46 @@ describe('HttpValidator', () => {
         it("validates query assuming it's empty", test(undefined, { url: { path: '/' } }));
       });
     });
+
+    describe('invalid schema provided', () => {
+      it('should not call the validate method', async () => {
```
lag-of-death: `should not call validateAgainstSchema method` ?
because we expect `.validate` to be called

packages/http/src/validator/validators/params.ts
```diff
@@ -29,7 +29,7 @@ export class HttpParamsValidator<Target, Spec extends IHttpParam> implements IHt
       }
 
       const resolvedStyle = spec.style || style;
-      if (spec.content && spec.content.schema && target[spec.name]) {
+      if (spec.content && spec.content.schema && validateJSONSchema(spec.content.schema) && target[spec.name]) {
```
lag-of-death: why not just: `ajv.validateSchema(spec.content.schema)`
introducing `validateJSONSchema` looks a bit too much for this simple case

packages/http/src/validator/__tests__/index.spec.ts
XVincentX: That is correct, I didn't pick a great selection of names.

Refactored in 653221e

packages/http/src/validator/validators/params.ts
XVincentX: The problem is that the `ajv` instance in that function is shared and configured ad hoc. For example, it includes the draft-4 meta schema which is required for multiple parts of OAS2 validations. We need to use exactly that instance.

Does that answer your question?

XVincentX:
We're going to make this happen in Graphite.

packages/http/src/validator/validators/params.ts
lag-of-death: Yes, thank you ! :)

PR #329, opened 2019-05-29
fix: lower case parameters for presence checks

XVincentX:
Lowercase all the parameters for comparison, because Fastify lowers all the headers by default.

This code is probably going to be completely removed when we will have Graphite emit JSON Schema even for parameters and stuff around, however this will fix the things in meantime.

SO-233

### DIFF
--- a/packages/http/src/validator/validators/__tests__/headers.spec.ts
+++ b/packages/http/src/validator/validators/__tests__/headers.spec.ts
@@ -66,6 +66,19 @@ describe('HttpHeadersValidator', () => {
 
                 expect(validateAgainstSchemaModule.validateAgainstSchema).toHaveBeenCalled();
               });
+
+              it('ignores the header casing', () => {
+                expect(
+                  httpHeadersValidator.validate({ 'X-Test-Header': 'abc' }, [
+                    {
+                      required: true,
+                      name: 'x-test-header',
+                      style: HttpParamStyles.Simple,
+                      content: { schema: { type: 'string' }, examples: [], encodings: [] },
+                    },
+                  ]),
+                ).toEqual([]);
+              });
             });
           });
         });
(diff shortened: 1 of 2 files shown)


packages/http/src/validator/validators/params.ts
```diff
@@ -1,22 +1,25 @@
...
     private _style: HttpParamStyles,
   ) {}
 
   public validate(target: Target, specs: Spec[]): IPrismDiagnostic[] {
+    const loweredCaseTarget = mapKeys(target, (_value, key) => key.toLowerCase());
+
     const { _registry: registry, _prefix: prefix, _style: style } = this;
     return specs.reduce<IPrismDiagnostic[]>((results, spec) => {
-      if (!(spec.name in target) && spec.required === true) {
+      if (!(spec.name in loweredCaseTarget) && spec.required === true) {
```
lag-of-death: I think we need to do this for https://github.com/stoplightio/prism/blob/451a5a43db85cfcf62a49fabd151fdf5385143bc/packages/http/src/validator/validators/params.ts#L35 too

packages/http/src/validator/validators/__tests__/headers.spec.ts
```diff
@@ -66,6 +66,19 @@ describe('HttpHeadersValidator', () => {
 
                 expect(validateAgainstSchemaModule.validateAgainstSchema).toHaveBeenCalled();
               });
+
+              it('ignores the header casing', () => {
```
lag-of-death: we should probably add a similar test that has `required: false`

XVincentX: If required is false the check is not happening at all. What would you like me to test in particular?

lag-of-death: Hm.. ok, so maybe I don't understand. There now is:
`if (!(spec.name in loweredCaseTarget) && spec.required === true)`. And later there is `if (spec.content && spec.content.schema && target[spec.name])`. Don't we need a test for `if (spec.content && spec.content.schema && target[spec.name])`?

XVincentX:
Yeah I think we'll hold on with this for now. We decided to take a different direction that will probably invalidate all this stuff — we'll probably remove all these custom validators and use a regular JSON Schema.

I'm sorry for the annoyances  @lag-of-death — but It was not really easy to come up with a solution for this.

PR #333, opened 2019-06-03
Respect Accept header when processing requests

### LINKED ISSUE #288: Accept Header
As an Prism user I want it to respect Accept header provided in the request to Prism mocker, which will let me select the content type I would like in the response. 

If the user picks a valid path item and a valid method, but picks an invalid content type in the Accept header (the content type they define is not defined in the response, and the response does not have a */* defined) then respond with a 406.

When seeking for a content type, make sure to be RFC compliant and support `application/whatever+json` as valid JSON.

- Accept: application/json and there is application/json: use application/json
- Accept: application/vnd.api+json and there is application/json: 406 error sorry bud
- Accept: application/json and response has application/vnd.api+json: use application/vnd.api+json
- Accept: application/vnd.api+json and response has **both**: use the more specific match.

[Content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) is wild. 

If we can find an existing node module tool to help, we should use that.

*Acceptance criteria*

- Responses for generic JSON, XML and CSV are defined and user can pick whichever
- Responses can define various flavors of JSON and have that flavour specifically picked
- Vague acceptance matching means if a specific flavour is the only JSON defined and they want JSON, give them that flavour of JSON
- If they request a XML and response doesnt have it, 406 error
- If they request a specific flavour of JSON and response doesnt have it, even if it has generic JSON, 406 error

### LINKED ISSUE #332: Prism is not respecting the Accept header
**Describe the bug**
I'm getting empty responses from Prism mock server.

**To Reproduce**

Using this example specification:
https://gist.github.com/guilleiguaran/3cb198f59f804acae754124c57304ef6

1. When I run this command: 
`prism mock openapi.yml`

2. Then make the following HTTP request: 
`curl -i -H "Accept: application/vnd.api+json" "http://127.0.0.1:4010/languages"`

3. Then I got this response:

```
HTTP/1.1 200 OK
content-type: text/plain
content-length: 0
Date: Fri, 31 May 2019 22:57:57 GMT
Connection: keep-alive
```

**Expected behavior**
I expect one response like the one displayed by redoc-cli for same endpoint:
```
{
  "data": [
    {
      "type": "language",
      "id": "1",
      "attributes": {
        "name": "english"
      }
    }
  ]
}
```

**Desktop (please complete the following information):**

- OS: macOS 10.14.4
- Prism Version: 3.0.0-alpha.11.
- Install method: yarn

XVincentX:
* Run test harness *AFTER* the normal tests.
* Modify the OAS example document to put JSON first, otherwise it will try to produce xml when requesting `*/*` (and this is something we do not to support yet)
* Use `fastify-accept-serialiser` to use the normal `stringify` not only on `application/json` but to anything that can be serialised as JSON anyway
* Correctly end the request in case there's an error while serialising the payload
* Correctly use the accept header when selecting the content

Closes #332
Closes #288


packages/http/src/mocker/HttpMocker.ts
```diff
@@ -33,7 +33,7 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
     }
 
     // setting default values
-    const inputMediaType = input.data.headers && caseless(input.data.headers).get('content-type');
```
philsturgeon: The variable name was correct before, it was just being used poorly. The content-type is the inputMediaType. 

Maybe we could rename the variable to acceptHeader so there is no confusion. It might not be a single media type string there could be multiple values, so better to be clear about it being an Accept header.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -10,11 +11,18 @@ function findExampleByKey(httpContent: IHttpContent, exampleKey: string) {
...
   response: IHttpOperationResponse,
   mediaType: string,
 ): IMediaTypeContent | undefined {
-  return response.contents.find(content => content.mediaType === mediaType);
+  return (
+    response.contents.find(content => content.mediaType === mediaType) ||
+    response.contents.filter(content => content.mediaType !== '*').find(content => {
+      const parsedRequestedMediaType = mediaTypeParser.fromString(mediaType);
+      const parsedCandidateMediaType = mediaTypeParser.fromString(content.mediaType);
+      return parsedRequestedMediaType.suffix === parsedCandidateMediaType.subtype;
```
philsturgeon: This looks annoying. 

When give `text/plain` the property `media.subtype` is `'plain'`, but when given `'application/vnd.company.app.entity-v2+xml` the `media.subtype` is `vnd.company.app.entity-v2` and xml is on `media.suffix`.

This means your code only match exact matches `content => content.mediaType === mediaType` or mismatched specifics when the client requests `application/foo+json` but only `application/json` is defined in the spec. 

If the user requested application/json and the spec had `application/json; charset=utf8` this would also fail. 

I think you want to find a content negotiation tool which will handle the comparisons for you, because this is a can of worms.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -102,8 +110,9 @@ const helpers = {
     response: IHttpOperationResponse,
   ): IHttpNegotiationResult {
     const { code, dynamic, exampleKey } = partialOptions;
-    const mediaType = 'application/json';
-    const httpContent = findHttpContentByMediaType(response, mediaType);
+    const httpContent =
+      findBestHttpContentByMediaType(response, '*/*') || findBestHttpContentByMediaType(response, 'application/json');
```
philsturgeon: Maybe the fallback header string could be an array of strings, like `['*/*', 'application/json']` because calling this function twice will execute the lookup code twice. :)

packages/http/src/mocker/HttpMocker.ts
XVincentX: Done!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: I wish there would be a package handling all this stuff seriously — unfortunately I haven't been able to find a single one doing this. I'll give it another round…but no luck so far.

XVincentX:
@philsturgeon 

Given the requirement changes, if we can figure out WHEN exactly send a 406, this could close #288 as well

packages/http-server/src/server.ts
```diff
@@ -1,15 +1,26 @@
...
 export const createServer = <LoaderInput>(
   loaderInput: LoaderInput,
   opts: IPrismHttpServerOpts<LoaderInput>,
 ): IPrismHttpServer<LoaderInput> => {
-  const server = fastify<Server, IncomingMessage, ServerResponse>();
+  const server = fastify().register(fastifyAcceptsSerializer, {
+    serializers: [
+      {
+        regex: /json$/,
+        serializer: JSON.stringify,
```
pytlesk4: Should this use `safeStringify`, from `@stoplight/json`?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -1,5 +1,6 @@
 import { IHttpContent, IHttpOperation, IHttpOperationResponse, IMediaTypeContent, Omit } from '@stoplight/types';
-
+// @ts-ignore
+import * as accepts from 'accepts';
```
pytlesk4: `yarn add --dev @types/accepts`

philsturgeon:
So the "returning an empty body" thing is when we have confirmed that a code should be sent back and we've just failed to find an example. 

We want to add a layer in, before the example/empty stuff even comes into play, which is "are they accepting for a valid content type. If they ask for a content type the spec doesnt support according to accepts package, errors. 

If they ask for a content type that is supported but doesnt have any example or schema, then empty body.

Does that make sense?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: I did it! The problem is that, although the `accept` functions wants just an object with a `headers` property, the types request the whole `req` object that I clearly do not have in Prism's internals. Either I ignore it here, or I will have to cast the stuff to a request.

I've also tried to fix the types but this package is so dependant that I'd have to fix half of the ecosystem. I'm fine with both solutions, I went with this one for no particular reason.

Does that clarify?

packages/http-server/src/server.ts
XVincentX: According to the documentation, the only thing `safeStringify` does more than the regular function is to handle circular references — which is a thing you can't express in an OAS document (as in a text file in general), so I skipped that. Does that make sense?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
pytlesk4: Yup, totally fine with the current way then!

packages/http-server/src/server.ts
pytlesk4: Yup, all good then!

XVincentX:
@pytlesk4 Good point on the multiple http headers — I'll write a test right now and fix it if necessary.
@philsturgeon Yes, that explains. On it!

XVincentX:
@pytlesk4 3090fb3 and 2969c4d are checking out multiple accept headers.

XVincentX:
@philsturgeon d120bad implements the requested behaviour. I'd suggest you to pull out the branch and check it out really quickly.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Done!

philsturgeon:
Really close! 
```
$ http GET  http://127.0.0.1:4010/no_auth/pets\?name\=234 'Accept: image/jpeg;q=0.8, application/xml;q=0.3, text/plain;q=0.2'
HTTP/1.1 406 Not Acceptable
Connection: keep-alive
Date: Mon, 03 Jun 2019 19:05:00 GMT
content-length: 89

{
    "error": "Not Acceptable",
    "message": "Allowed: /json$/,application/json",
    "statusCode": 406
}
```

It should be returning a problems JSON and that message is looking a bit funky.

XVincentX:
@philsturgeon Try it again, it should be ok now.

packages/http/src/mocker/errors.ts
```diff
@@ -6,3 +6,9 @@ export const UNPROCESSABLE_ENTITY: Omit<ProblemJson, 'detail'> = {
   title: 'Invalid request body payload',
   status: 422,
 };
+
+export const NOT_ACCEPTABLE: Omit<ProblemJson, 'detail'> = {
+  type: 'NOT_ACCEPTABLE',
+  title: 'The server cannot produce a response matching your content negotiation header',
```
philsturgeon:
```suggestion
  title: 'The server cannot produce a representation for your accept header',
```

XVincentX:
@philsturgeon Harness for 406 response added.

test-harness/gold-master-files/__path____user_username___method___GET__.json
```diff
@@ -10,7 +10,7 @@
     "statusText": "OK",
     "headers": {
       "content-type": [
-        "application/json; charset=utf-8"
```
philsturgeon: This reminds me, do we have any tests where the spec has `application/json; charset=utf-8` but they request has `accept: application/json`? I'd like to be very sure that works.

XVincentX: Added in 2ebcaef

XVincentX: 5bc26a9

philsturgeon:
amazing!

PR #338, opened 2019-06-04
fix: correctly derive a valid JSON Schema from an OAS parameter

### LINKED ISSUE #324: Error 422, Missing Authorization header param

XVincentX:
This PR updates Graphite in Prism to use the latest version which includes the JSON Schema normalisation features.

1. Update packages (Typescript, Types, Graphite)
2. Cleaned up the examples from optional properties. In the new graphite a lot of fields are not mandatory and this requires an additional round of checks in Prism. An example is [here](https://github.com/stoplightio/prism/pull/338/files#diff-52fafd78c752aca126e4fd8c4ce567a6R79)
3. The new graphite pointed out some inconsistencies in the previous handmade validator — which are now fixed since it has been offloaded to Ajv
4. Added a specific test to make sure the validator is not case sensitive (headers) — comically speaking, this worm hole started from *this* specific issue

P.S: There are some other weird things happening in the deserialisation that I want to lookup and fix, but not here, not now.

Closes #324


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -13,20 +13,31 @@ function findExampleByKey(httpContent: IHttpContent, exampleKey: string) {
...
-  return response.contents.find(content =>
-    accepts({
-      headers: {
-        accept: mediaType.join(','),
-      },
-    }).type(content.mediaType),
-  );
+  return response.contents
+    .map(t => {
+      if (t.mediaType === '*') t.mediaType = '*/*';
```
XVincentX: @chris-miaskowski Do you know/recall why the default media type has been marked with `*` instead of `*/*`? If there's no particular reason I could probably fix this in graphite and avoid this remap.

chris-miaskowski: Hey @XVincentX, no reason - was picked arbitrarily.

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -46,7 +46,7 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
     let negotiationResult: IHttpNegotiationResult;
     if (input.validations.input.length > 0) {
       try {
-        negotiationResult = helpers.negotiateOptionsForInvalidRequest(resource.responses);
+        negotiationResult = helpers.negotiateOptionsForInvalidRequest(resource.responses!);
```
chris-miaskowski: `resource.responses!`
Is this an oversight?

packages/http/src/__tests__/fixtures/index.ts
```diff
@@ -246,9 +222,6 @@ export const httpOperations: IHttpOperation[] = [
     id: 'todos',
     method: 'post',
     path: '/todos',
-    servers: [],
```
chris-miaskowski: Does this mean that servers can now be `undefined`?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -79,15 +79,14 @@ function isINodeExample(nodeExample: INodeExample | INodeExternalExample | undef
 
 function computeMockedHeaders(headers: IHttpHeaderParam[], ex: PayloadGenerator): Promise<Dictionary<string>> {
   const headerWithPromiseValues = mapValues(keyBy(headers, h => h.name), async header => {
-    if (header.content) {
```
chris-miaskowski: Out of interest, why did we flattened it?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -202,7 +196,7 @@ describe('HttpMocker', () => {
             expect(response.body).toBeDefined();
 
             const allExamples = flatMap(mockResource.responses, res =>
-              flatMap(res.contents, content => content.examples),
+              flatMap(res.contents, content => content.examples || []),
```
chris-miaskowski: I'm looking at the `mockResource.responses` and I can't see a content with an undefined `examples` array. Does this `|| []` stem from a type change? Can `examples` be undefined?

If it can I'm curious why. My personal opinion about types is that if something is an array we should not let it be undefined. If there is no value it should be an empty value in my opinion. In the end what what we want to say is: there is no examples. Allowing it to be `undefined` is like saying: there can be no examples OR there can be no such thing as examples (which is confusing most of the time and forces stuff like `|| []`)

packages/http/src/mocker/negotiator/types.ts
```diff
@@ -1,12 +1,13 @@
...
+import { IHttpHeaderParam, INodeExample, INodeExternalExample } from '@stoplight/types';
+import { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
 
 export interface IHttpNegotiationResult {
   code: string;
   mediaType: string;
   bodyExample?: INodeExample | INodeExternalExample;
-  schema?: ISchema;
-  headers: IHttpHeaderParam[];
+  schema?: JSONSchema4 | JSONSchema6 | JSONSchema7;
```
chris-miaskowski: Have you thought about creating a `type JSONSchema = JSONSchema4 | JSONSchema6 | JSONSchema7` and exporting it in `@stoplight/types`?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Ah ok nice — then maybe we can make it happen with https://github.com/stoplightio/graphite/pull/191 and remove this?

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -40,6 +41,45 @@ describe('HttpValidator', () => {
...
+                ],
+              },
+            },
+            input: {
+              method: 'get',
+              url: {
+                path: '/hey',
+              },
+              headers: {
+                api_key: 'ha',
```
chris-miaskowski: If I'm not mistaken this test will only verify that the request headers are case insensitive, but not that the `input.headers` are. 

E.g. this does not confirm that `input.headers.Api_key` works fine.

In theory the actual code could looks something like this:
```ts
if (request.headers.name.toLowerCase() === Object.keys(input.headers)[0])
```

packages/http/src/validator/validators/__tests__/body.spec.ts
```diff
@@ -64,11 +65,11 @@ describe('HttpBodyValidator', () => {
 
       describe('validator for given media type exists', () => {
         it('return validation errors', () => {
-          const mockSchema = { type: 'string' };
+          const mockSchema: JSONSchema4 = { type: 'string' };
 
           jest.spyOn(validatorRegistry, 'get').mockImplementation(mediaType => {
             expect(mediaType).toBe('application/json');
-            return (content: any, schema: ISchema) => {
+            return (content: any, schema: JSONSchema4 | JSONSchema6 | JSONSchema7) => {
```
chris-miaskowski: Yup, we defo need a union :)

packages/http/src/mocker/HttpMocker.ts
XVincentX: Yes! 88bb828

packages/http/src/__tests__/fixtures/index.ts
XVincentX: That is correct

packages/http/src/mocker/HttpMocker.ts
XVincentX: There has been kind of a long discussion with @marbemac, who was kind of concerned of the size of the operation created by graphite.

Basically his rationale was — instead of having a lot of empty things that effectively take space, let's set them as possibly undefined.

packages/http/src/mocker/negotiator/types.ts
XVincentX: Yes, 2-3 times, but at the end I didn't do it because hopefully we'll come up with the right schema draft we want to support and we'll probably remove all the other.

I understand it's kind of a big overload though, so I can do the change. In another PR :trollface:

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Do not count this specific example, it's a leftover I have to fix — since there's a way to tell TypeScript "hey this specific thing is *NOT* undefined.

As your personal opinion, I am inclined to be on your side — however everybody here has one — and mine is even different. null/undefined should never appear in your code **no matter the cost** 

No:
```ts
const str: string | undefined = getTheStringThatCouldBeUndefined();
```

Yes:
```ts
const str: Option<string> = getTheStringThatCouldBeUndefined();
```

packages/http/src/validator/validators/params.ts
```diff
@@ -15,45 +16,42 @@ export class HttpParamsValidator<Target, Spec extends IHttpParam> implements IHt
...
+    );
+
+    return validateAgainstSchema(parameterValues, schema, prefix).concat(deprecatedWarnings);
   }
 }
+
+function createJsonSchemaFromParams(params: IHttpParam[]): JSONSchema4 | JSONSchema6 | JSONSchema7 {
+  const schema: JSONSchema4 | JSONSchema6 | JSONSchema7 = {
+    type: 'object',
+    properties: pickBy(mapValues(keyBy(params, p => p.name.toLowerCase()), 'schema')) as JSONSchema4,
```
chris-miaskowski: The more I look at the `x.toLowerCase` the more I come to a conclusion that this should be done at a class level.
At some point someone will forget to lower case a value and we'll get an error. If, instead, we used a class that would override setters (and/or getters) and lowercase everything it would be much simpler to write this code w/o risking hard to find errors. 
I don't think it's worth refactoring now, but WDYT as a general rule?

packages/http/src/validator/__tests__/functional.spec.ts
XVincentX: This is a good point — the reason I didn't handle this case is that the nodejs HTTP server will *always* put the headers in lowercase, so this is a scenario that should never happen.

That said, given we will probably use Prism directly from the client, it might indeed make sense to make this normalisation in Prism itself. Let me check it out and I'll probably put another commit to fix this.

Good catch! 👍

packages/http/src/validator/validators/params.ts
XVincentX: I like the idea of performing this automagically — but I do not really follow the example practically (through a class) — can you clarify/make a minimal example?

I was thinking of using [caseless](https://www.npmjs.com/package/caseless) again, which we're already employing in the codebase.

chris-miaskowski: Do you need to use `pickBy` here? What's the benefit? IIRC it would work the same way w/o it.

XVincentX: `pickBy` will automatically remove any property that's undefined

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
chris-miaskowski: Gotcha, I get the argument about taking up space in DB / memory though. I can live with it ;p

I'm with you on that topic btw ;)

packages/http/src/mocker/negotiator/types.ts
chris-miaskowski: Cool!

packages/http/src/validator/__tests__/functional.spec.ts
chris-miaskowski: > the nodejs HTTP server will always put the headers in lowercase

Every server or just the current implementation? (rhetorical question I guess, since it must be implementation specific).

Yeah like you said I think it still makes sense to patch/test it here.

packages/http/src/validator/validators/params.ts
chris-miaskowski: So we currently mostly use interfaces only (which is awesome!) and we don't use classes.
Imagine that you pass a param `params: IHttpParam[]` but the instance of it is in fact `params: CaselessHttpParam[]` (or whatever the name, could be simply `HttpParam`

Then the implementation would simply be:
```ts
class HttpParam implements IHttpParam {
   // I made that decorator up but it would effectively implement a setter and a getter and lowercase stuff.
   @caseless
   public name: String;
}
```

or if possible

```ts
@caseless
class HttpParam implements IHttpParam {}
```

chris-miaskowski: Then in code you'd simply do 
`keyBy(params, p => p.name)`
In fact, you wouldn't even have to because all would be lowercase in the first place and there would be no dups.

chris-miaskowski: Oh cool stuff, good to know!

packages/http/src/validator/__tests__/functional.spec.ts
XVincentX: Every server done in NodeJs will have lowercased headers.

https://nodejs.org/api/http.html#http_message_headers

> Header names are lower-cased.

packages/http/src/validator/validators/params.ts
XVincentX: Ok clear, let me see if I can come up with something!

XVincentX: I'll have to pass on on this one unfortunately — it's really too much long to do know, I'll come up with something better when we will revisit the deserialisers/validators part.

XVincentX:
@philsturgeon @chris-miaskowski 

Please have another look.

packages/http/src/validator/deserializers/style/form.ts
```diff
@@ -9,14 +10,21 @@ export class FormStyleDeserializer implements IHttpQueryParamStyleDeserializer {
     return style === HttpParamStyles.Form;
   }
 
-  public deserialize(name: string, parameters: IHttpNameValues, schema: ISchema, explode: boolean = true) {
-    const { type } = schema;
+  public deserialize(
+    name: string,
+    parameters: IHttpNameValues,
+    schema?: JSONSchema4 | JSONSchema6 | JSONSchema7,
```
philsturgeon: We seem to creating this list in a lot of places. Can we make an alias that we control, some sort of list of "supported versions" or something that we can pass around without needing to hint this piped list?

XVincentX: Ok done

XVincentX:
> My last point is that I couldn't find a test case for the originally reported issue (the header param).
Maybe we could take the exact spec provided in the original issue report and use it to feed the test?

Can you elaborate more? What is the test that's missing? The original issue was on the camel case of the headers, which now have a test.

philsturgeon:
@XVincentX what I meant about the tests, is that we should probably add some cases to the functional tests for changes as we go. packages/http/src/mocker/__tests__/functional.spec.ts didnt seem to get any love?

XVincentX:
@philsturgeon That test file is testing the mocker, which is *after* the validation step. I understand what you mean though, I'll find the relevant file and add the stuff you asked for.

PR #319, opened 2019-05-24
WIP: OpenAPI Sampler for "Static Examples"

philsturgeon:
## Checklist

- [ ] Tests have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## What kind of change does this PR introduce

Bug fix, feature, or docs update

## What is the current behavior

You can also link to an open issue here.

## If this is a feature change, what is the new behavior

It can now make a nice cup of tea.

## Does this PR introduce a breaking change

What changes might users need to make in their application due to this PR?

### Other information

(e.g. detailed explanation, related issues, links for us to have context, eg. Stack Overflow, issues outside of the repo, forum, etc.)

PR #339, opened 2019-06-05
Make dynamic examples disabled by default, to be enabled via flag `-d`

### LINKED ISSUE #263: Static examples by default, with dynamic CLI/JS/Query option
As a Prism user with experience using v2, I expect similar behavior from v3, and as such examples should be static by default, with an option to enable dynamic responses if I want to.

**Acceptance Criteria**

- Make the default responses be consistently static and update integration tests

- Add --dynamic CLI option to the mock command so all responses served are dynamic (same as current behavior)

- Add  ?__dynamic=true/false query string param which will make that one request dynamic or not, and override the CLI option

Order should be: 

- if ?__dynamic=true then dynamic response

- else if --dynamic in CLI then dynamic response

- else look for examples, use the first examples it finds that matches the mime type or star/star

- If no examples then go through the schema, and for every property and use 1.) example 2) default, 3) if nullable use null, 4) if not nullable then datatype relevant static values: "string", 0 etc.

beldar:
Fixes #263

PR Title

## Checklist

Bug fix, feature, or docs update

You can also link to an open issue here.

It can now make a nice cup of tea.

What changes might users need to make in their application due to this PR?

### Other information

(e.g. detailed explanation, related issues, links for us to have context, eg. Stack Overflow, issues outside of the repo, forum, etc.)

### DIFF
--- a/packages/cli/src/commands/mock.ts
+++ b/packages/cli/src/commands/mock.ts
@@ -22,7 +22,7 @@ export default class Server extends Command {
       signale.star('Dynamic example generation enabled.');
     }
 
-    const server = createServer(spec, { mock: { dynamic: true || dynamic } });
+    const server = createServer(spec, { mock: { dynamic: false || dynamic } });
     try {
       const address = await server.listen(port);
 


XVincentX:
@beldar 

Thanks for the Pull Request!

This would indeed fix the issue you mentioned; thing is we're forcing (temporarily) the dynamic examples right now for a reason: we still have to come out with a good static example generation strategy. Once this will be documented and implemented, we'll switch it back to the regular flag.

Is this preventing you from using the tool at all?

philsturgeon:
Related #319

beldar:
> @beldar
> 
> Thanks for the Pull Request!
> 
> This would indeed fix the issue you mentioned; thing is we're forcing (temporarily) the dynamic examples right now for a reason: we still have to come out with a good static example generation strategy. Once this will be documented and implemented, we'll switch it back to the regular flag.
> 
> Is this preventing you from using the tool at all?

Hi there, 

Not sure what you mean with static example generation, my assumption was that the **spec** passed to prism would contain the static examples and that's what prism would return?

XVincentX:
@beldar

Yes! That is indeed the intent, but we want to be smart in case some models do not contain *any* example and you have the dynamic generation disabled. Throwing an exception it's probably fine, 

We want to generate an example somehow but still have it predictable. For this to happen we're in the phase of evaluating other generators libraries. Once we'll make the choice, we will fix the bug and restore the original flag evaluation.

---

P.S: I visited your offices last April and I'm often in Barcelona, probably at the end of June. We should connect! Hablamos?

philsturgeon:
@beldar check out #319 it might give you some context. Right now we do not have a static example mode, we are working on adding it.

beldar:
OK, I understand. We were planning to use this tool for contract testing, but if it cannot serve the examples defined on the contract it's not gonna work for us.

Thanks for the explanation though!

philsturgeon:
@beldar it absolutely uses the examples defined, there are a lot of different types of example. If you have defined response body examples then those will get used. You can even pick specific examples via the query string. 

The static vs dynamic is something different. When there is no full response example we have to construct the mock response from the schema, and there are two ways we can do that. 

1. Look at the property type, min, max, length, etc and faker things (dynamic)
2. Look at the property default, example, take the first enum, etc (static) 

We have 1, we are adding 2 now. Again, all that should be in #319.

Maybe subscribe to that PR so you know when it's done.

beldar:
Ah ok, I misunderstood then. That makes sense. I'll follow that PR then.

javigomez:
By the way @XVincentX 

> P.S: I visited your offices last April and I'm often in Barcelona, probably at the end of June. We should connect! Hablamos?

Happy to have you in our office. Definitely ping us!!! ^_^

XVincentX:
@javigomez @beldar 

Me dejáis un contacto? Estaré en Barcelona este fin de Junio!

javigomez:
Justo te he enviado un email a tu cuenta de @icloud.com

PR #340, opened 2019-06-05
Custom host cli option

adriengibrat:
Add optional custom host cli option

## Checklist

- [x] Tests have been added (features)

Add optional custom host cli option to allow to listen to another host than 127.0.0.1

Currently [fastify](https://github.com/fastify/fastify#fastify-v1x)
> .listen binds to the local host, localhost, interface by default (127.0.0.1 or ::1) [...]
when the host is not specified

This is an issue to easily deploy prism or run it inside docker.

This PR introduce an optional `--host` / `-h` prism cli option that will be passed to '@stoplight/prism-http-server' & then fastify `.listen`, thus allowing to listen to specific host or '0.0.0.0'

No.

### Other information

It could be interesting to reproduce the fastify note https://github.com/fastify/fastify#note in the readme.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -59,6 +59,7 @@ ARGUMENTS
 
 OPTIONS
   -d, --dynamic    Dynamically generate examples.
+  -h, --host=host  [default: 127.0.0.1] Host that Prism will listen to.
   -p, --port=port  (required) [default: 4010] Port that Prism will run on.
 ```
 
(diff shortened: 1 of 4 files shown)


philsturgeon:
@adriengibrat thank you! Can you add that note too? Save me doing up a whole PR

philsturgeon:
I'll add the note I was being lazy 😴 

Thank you for this PR!

PR #344, opened 2019-06-06
fix: add generic */*+json parser

### LINKED ISSUE #341: Prism failing with custom Content-Type header
**Describe the bug**
I'm getting an `Unsupported Media Type: application/vnd.api+json` error from Prism mock server.

**To Reproduce**

Using this example specification:
https://gist.github.com/guilleiguaran/3cb198f59f804acae754124c57304ef6

1. When I run this command: 
`prism mock openapi.yml`

2. Then make the following HTTP request: 
```
curl -X POST -H 'Content-Type: application/vnd.api+json' -H 'Accept: application/vnd.api+json' -i http://localhost:4010/users --data '{
  "data": {
    "type": "user",
    "attributes": {
      "email": "phil@fakemail.com",
      "password": "mypassword1$",
      "first_name": "Phil",
      "last_name": "Sturgeon",
      "terms_accepted": true
    }
  }
}'
```

3. Then I got this response:

```
{
  "statusCode": 415,
  "code": "FST_ERR_CTP_INVALID_MEDIA_TYPE",
  "error": "Unsupported Media Type",
  "message": "FST_ERR_CTP_INVALID_MEDIA_TYPE: Unsupported Media Type: application/vnd.api+json"
}
```

**Expected behavior**
I expect one response like the one displayed by redoc-cli for same endpoint:
```
{
  "data": {
    "type": "user",
    "attributes": {
      "email": "angelica.acosta@fakemail.com",
      "password": "nulla eu est",
      "first_name": "Angelica",
      "last_name": "Acosta",
      "registration_state": "onboarding_completed"
    },
    "id": "1"
  }
}
```

**Desktop (please complete the following information):**

- OS: macOS 10.14.4
- Prism Version: 3.0.0-alpha.12.
- Install method: yarn

XVincentX:
1. Restores default Fastify serialisation for regular JSON
2. Uses `type-is.is` to detect a flavour of JSON and parse/serialise it
3. Removes an old negotiation test that does not really make that much sense and it's already tested inside there

Closes #341


packages/http-server/src/server.ts
```diff
@@ -14,13 +15,24 @@ export const createServer = <LoaderInput>(
...
-        regex: /json$/,
+        regex: /application\/\S*json$/,
         serializer: JSON.stringify,
       },
     ],
     default: 'application/json',
   });
 
+  server.addContentTypeParser('*', { parseAs: 'string' }, (req, body, done) => {
+    if (typeIs(req, ['application/*+json'])) {
```
chris-miaskowski: I'm looking at the typeIs docs page and it seems that doing 
```suggestion
    if (typeIs(req, ['json'])) {
```
Should do the trick. `+json` is perhaps to strict. I think this code will not match `application/json` for instance, will it?

XVincentX: Unfortunately simply putting `['json']` won't work.

I'm adding some tests to make sure it works in all the known conditions.

packages/http-server/src/server.ts
```diff
@@ -14,13 +15,24 @@ export const createServer = <LoaderInput>(
   const server = fastify().register(fastifyAcceptsSerializer, {
     serializers: [
       {
-        regex: /json$/,
+        regex: /application\/\S*json$/,
```
chris-miaskowski:
```
crapplication/json
application/mjson
application/vnd.xoxo.json
application/vnd.api+xml;q=json
```

Heads up that it will also match the above. Maybe a good candidate for the unit tests.

XVincentX: I'm trying to see if I can replace this part and use is-type as well, the regex can fail and I do not feel exactly that comfortable.

packages/http-server/src/server.ts
```diff
@@ -14,11 +15,29 @@ export const createServer = <LoaderInput>(
...
     serializers: [
       {
-        regex: /json$/,
+        /*
+          In theory this should effectively be a regexp, but since it's JavaScript I thought it would
+          be a good idea to fake it and use typeIs.is function since it's way more reliable than any
+          regexp I can think of.
+        */
+        regex: {
+          test: (value: string) => !!typeIs.is(value, ['application/*+json']),
```
chris-miaskowski: Huehue I like the hack ;)

test-harness/gold-master-files/__path____pets_123___method___POST___headers____content-type___application_vnd.api+json____body____\_hello\__\_world\____.json
```diff
@@ -0,0 +1,32 @@
...
+    "method": "POST",
+    "headers": {
+      "content-type": "application/vnd.api+json"
+    },
+    "path": "/pets/123",
+    "host": "http://localhost:4010"
+  },
+  "response": {
+    "status": 500,
+    "statusText": "Internal Server Error",
```
chris-miaskowski: Hmm wait, I'm not sure I follow. Is that the expected response? You sure we want 5xx?
Not sure if this help is useful. Maybe it should at least reply with a correct 200?

XVincentX: yeah that is confusing, I'm open to better ideas.

Basically the issue this PR is fixing was at Fastify level, where the framework was refusing the request, completely. The fact that you got a message error from Prism means that the request went through correctly. Does that clarify?

XVincentX: Actually I have a better idea, hold on

test-harness/gold-master-files/__path____no_auth_pets_10___method___GET___headers____accept___application_idonotexist___.json
```diff
@@ -12,20 +12,19 @@
...
       ]
     },
     "body": {
-      "type": "https://stoplight.io/prism/errors#NOT_ACCEPTABLE",
-      "title": "The server cannot produce a response matching your content negotiation header",
-      "status": 406,
-      "detail": "Could not find any content that satisfies application/idonotexist"
+      "statusCode": 406,
+      "error": "Not Acceptable",
+      "message": "Allowed: [object Object],application/json"
```
philsturgeon: What is happening here, is this meant to be object?

test-harness/gold-master-files/__path____no_auth_pets_123___method___POST___headers____Content-Type___application_x-www-form-urlencoded____body___name=just_a_name__.json
```diff
@@ -9,24 +9,24 @@
     "host": "http://localhost:4010"
   },
   "response": {
-    "status": 415,
-    "statusText": "Unsupported Media Type",
+    "status": 500,
```
philsturgeon: Is this meant to be 500 or 415?

XVincentX: Fixed in 399d276bcca3a5fcdd0c7bcfaff47d339d96f72c. This is yet another problem of the test harness, it does not have any automatic invalidation system and so they can become stale and we do not notice it. We really need to find another way here.

test-harness/gold-master-files/__path____no_auth_pets_10___method___GET___headers____accept___application_idonotexist___.json
XVincentX: Very nice catch! The fake regex was missing a `toString()` method, added in 9177b9b9c8f1388babdf6db65557e8fe78d961e2

test-harness/gold-master-files/__path____no_auth_pets_123___method___POST___headers____Content-Type___application_x-www-form-urlencoded____body___name=just_a_name__.json
philsturgeon: I gotcha. Lets pair up sometime on Monday to chat about what we think we'd like to do to improve test harness.

packages/http-server/src/server.ts
```diff
@@ -14,11 +15,33 @@ export const createServer = <LoaderInput>(
   const server = fastify().register(fastifyAcceptsSerializer, {
     serializers: [
       {
-        regex: /json$/,
+        /*
+          In theory this should effectively be a regexp, but since it's JavaScript I thought it would
```
philsturgeon:
```suggestion
          This is a workaround, to make Fastify less strict in its json detection. It expects a regexp, but instead we are using typeIs.
```

test-harness/index.js
```diff
@@ -253,6 +253,25 @@ const createSpec = (specPath, prismCmd) => {
             expect(reqRes.response.status).toBe(422);
           });
         });
+
+        describe('content type parser', () => {
+          test.each(
+            [['application/json', false], ['application/vnd.api+json', false], ['application/xml', true],
```
philsturgeon: Maybe moving this to a const on the line above would be nice for readability.

PR #347, opened 2019-06-12
Static Examples generation

XVincentX:
This PR implements the strategy we agreed for static examples generation.

1. Removes the forced `true` flag
2. Uses OpenAPI Sampler in case the static example generation is enabled
3. I _had_ to move the `generate` argument for `HttpMocker` because of the weird configuration merger thing which I do not want to touch right now
4. Adds a bunch of tests

---

As you can see from the code, I had to add an additional function that takes a regular JSON Schema and transforms it to something that OpenAPI Sampler can effectively use to generate static examples.

Maybe we could do it the other way around, but this is a followup though.

SO-225

---

If this gets merged:
https://github.com/Redocly/openapi-sampler/pull/10

I can remove the horrible function I had to wrote myself to walk through the schema and put the first `examples` in the `example`

### DIFF
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -23,6 +23,7 @@
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
     "json-schema-faker": "^0.5.0-rc17",
+    "openapi-sampler": "^1.0.0-beta.14",
     "tslib": "^1.9.0"
   },
   "publishConfig": {
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -22,7 +22,7 @@
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
-    "json-schema-faker": "^0.5.0-rc17",
+    "@stoplight/json-schema-faker": "",
     "openapi-sampler": "^1.0.0-beta.14",
     "tslib": "^1.9.0"
   },
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -16,13 +16,13 @@
     "node": ">=8"
   },
   "dependencies": {
+    "@stoplight/json-schema-faker": "^0.5.0-rc18",
     "@stoplight/prism-core": "^3.0.0-alpha.14",
     "accepts": "^1.3.7",
     "ajv": "^6.0.0",
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
-    "@stoplight/json-schema-faker": "",
     "openapi-sampler": "^1.0.0-beta.14",
     "tslib": "^1.9.0"
   },
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -16,13 +16,13 @@
     "node": ">=8"
   },
   "dependencies": {
-    "@stoplight/json-schema-faker": "^0.5.0-rc18",
     "@stoplight/prism-core": "^3.0.0-alpha.14",
     "accepts": "^1.3.7",
     "ajv": "^6.0.0",
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
+    "json-schema-faker": "^0.5.0-rc17",
     "tslib": "^1.9.0"
   },
   "publishConfig": {
--- a/packages/http/src/mocker/HttpMocker.ts
+++ b/packages/http/src/mocker/HttpMocker.ts
@@ -77,7 +77,10 @@ function isINodeExample(nodeExample: INodeExample | INodeExternalExample | undef
   return !!nodeExample && 'value' in nodeExample;
 }
 
-function computeMockedHeaders(headers: IHttpHeaderParam[], ex: PayloadGenerator): Promise<Dictionary<string>> {
+function computeMockedHeaders(
+  headers: IHttpHeaderParam[],
+  payloadGenerator: PayloadGenerator,
+): Promise<Dictionary<string>> {
   const headerWithPromiseValues = mapValues(keyBy(headers, h => h.name), async header => {
     if (header.schema) {
       if (header.examples && header.examples.length > 0) {
@@ -86,7 +89,7 @@ function computeMockedHeaders(headers: IHttpHeaderParam[], ex: PayloadGenerator)
           return example.value;
         }
       } else {
-        const example = await ex(header.schema);
+        const example = await payloadGenerator(header.schema);
         if (!(isObject(example) && isEmpty(example))) return example;
       }
     }
@@ -98,12 +101,12 @@ function computeMockedHeaders(headers: IHttpHeaderParam[], ex: PayloadGenerator)
 
 async function computeBody(
   negotiationResult: Pick<IHttpNegotiationResult, 'schema' | 'mediaType' | 'bodyExample'>,
-  ex: PayloadGenerator,
+  payloadGenerator: PayloadGenerator,
 ) {
   if (isINodeExample(negotiationResult.bodyExample) && negotiationResult.bodyExample.value !== undefined) {
     return negotiationResult.bodyExample.value;
   } else if (negotiationResult.schema) {
-    return ex(negotiationResult.schema);
+    return payloadGenerator(negotiationResult.schema);
   }
   return undefined;
 }
--- a/packages/http/src/mocker/HttpMocker.ts
+++ b/packages/http/src/mocker/HttpMocker.ts
@@ -2,7 +2,7 @@ import { IMocker, IMockerOpts } from '@stoplight/prism-core';
 import { Dictionary, IHttpHeaderParam, IHttpOperation, INodeExample, INodeExternalExample } from '@stoplight/types';
 
 import * as caseless from 'caseless';
-import { fromPairs, isEmpty, isObject, keyBy, mapValues, toPairs } from 'lodash';
+import { isEmpty, isObject, keyBy, mapValues } from 'lodash';
 import {
   IHttpConfig,
   IHttpOperationConfig,
@@ -17,11 +17,11 @@ import helpers from './negotiator/NegotiatorHelpers';
 import { IHttpNegotiationResult } from './negotiator/types';
 
 export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpConfig, IHttpResponse> {
-  public async mock({
+  public mock({
     resource,
     input,
     config,
-  }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): Promise<IHttpResponse> {
+  }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): IHttpResponse {
     let payloadGenerator: PayloadGenerator = generateStatic;
 
     if (config) {
@@ -66,10 +66,8 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
       negotiationResult = helpers.negotiateOptionsForValidRequest(resource, mockConfig);
     }
 
-    const [body, mockedHeaders] = await Promise.all([
-      computeBody(negotiationResult, payloadGenerator),
-      computeMockedHeaders(negotiationResult.headers || [], payloadGenerator),
-    ]);
+    const mockedBody = computeBody(negotiationResult, payloadGenerator);
+    const mockedHeaders = computeMockedHeaders(negotiationResult.headers || [], payloadGenerator);
 
     return {
       statusCode: parseInt(negotiationResult.code),
@@ -77,7 +75,7 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
         ...mockedHeaders,
         'Content-type': negotiationResult.mediaType,
       },
-      body,
+      body: mockedBody,
     };
   }
 }
@@ -86,11 +84,8 @@ function isINodeExample(nodeExample: INodeExample | INodeExternalExample | undef
   return !!nodeExample && 'value' in nodeExample;
 }
 
-function computeMockedHeaders(
-  headers: IHttpHeaderParam[],
-  payloadGenerator: PayloadGenerator,
-): Promise<Dictionary<string>> {
-  const headerWithPromiseValues = mapValues(keyBy(headers, h => h.name), async header => {
+function computeMockedHeaders(headers: IHttpHeaderParam[], payloadGenerator: PayloadGenerator): Dictionary<string> {
+  const headerWithPromiseValues = mapValues(keyBy(headers, h => h.name), header => {
     if (header.schema) {
       if (header.examples && header.examples.length > 0) {
         const example = header.examples[0];
@@ -98,17 +93,17 @@ function computeMockedHeaders(
           return example.value;
         }
       } else {
-        const example = await payloadGenerator(header.schema);
+        const example = payloadGenerator(header.schema);
         if (!(isObject(example) && isEmpty(example))) return example;
       }
     }
     return '';
   });
 
-  return resolvePromiseInProps(headerWithPromiseValues);
+  return headerWithPromiseValues;
 }
 
-async function computeBody(
+function computeBody(
   negotiationResult: Pick<IHttpNegotiationResult, 'schema' | 'mediaType' | 'bodyExample'>,
   payloadGenerator: PayloadGenerator,
 ) {
@@ -118,9 +113,4 @@ async function computeBody(
     return payloadGenerator(negotiationResult.schema);
   }
   return undefined;
-}
-
-async function resolvePromiseInProps(val: Dictionary<Promise<string>>): Promise<Dictionary<string>> {
-  const promisePair = await Promise.all(toPairs(val).map(v => Promise.all(v)));
-  return fromPairs(promisePair);
 }
--- a/packages/http/src/mocker/HttpMocker.ts
+++ b/packages/http/src/mocker/HttpMocker.ts
@@ -12,17 +12,26 @@ import {
   ProblemJsonError,
 } from '../types';
 import { UNPROCESSABLE_ENTITY } from './errors';
+import { generate, generateStatic } from './generator/JSONSchema';
 import helpers from './negotiator/NegotiatorHelpers';
 import { IHttpNegotiationResult } from './negotiator/types';
 
 export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpConfig, IHttpResponse> {
-  constructor(private _exampleGenerator: PayloadGenerator) {}
-
   public async mock({
     resource,
     input,
     config,
   }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): Promise<IHttpResponse> {
+    let payloadGenerator: PayloadGenerator = generateStatic;
+
+    if (config) {
+      if (typeof config.mock !== 'boolean') {
+        if (config.mock.dynamic) {
+          payloadGenerator = generate;
+        }
+      }
+    }
+
     // pre-requirements check
     if (!resource) {
       throw new Error('Resource is not defined');
@@ -58,8 +67,8 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
     }
 
     const [body, mockedHeaders] = await Promise.all([
-      computeBody(negotiationResult, this._exampleGenerator),
-      computeMockedHeaders(negotiationResult.headers || [], this._exampleGenerator),
+      computeBody(negotiationResult, payloadGenerator),
+      computeMockedHeaders(negotiationResult.headers || [], payloadGenerator),
     ]);
 
     return {
--- a/packages/http/src/mocker/HttpMocker.ts
+++ b/packages/http/src/mocker/HttpMocker.ts
@@ -24,12 +24,8 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
   }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): IHttpResponse {
     let payloadGenerator: PayloadGenerator = generateStatic;
 
-    if (config) {
-      if (typeof config.mock !== 'boolean') {
-        if (config.mock.dynamic) {
-          payloadGenerator = generate;
-        }
-      }
+    if (config && typeof config.mock !== 'boolean' && config.mock.dynamic) {
+      payloadGenerator = generate;
     }
 
     // pre-requirements check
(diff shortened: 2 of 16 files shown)


packages/http/src/mocker/HttpMocker.ts
```diff
@@ -12,17 +12,26 @@ import {
...
+  public mock({
     resource,
     input,
     config,
-  }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): Promise<IHttpResponse> {
+  }: Partial<IMockerOpts<IHttpOperation, IHttpRequest, IHttpConfig>>): IHttpResponse {
+    let payloadGenerator: PayloadGenerator = generateStatic;
+
+    if (config) {
+      if (typeof config.mock !== 'boolean') {
```
philsturgeon: maybe you could use some && instead of 3 levels of nesting?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
             expect(response.body).toEqual((selectedExample as INodeExample).value);
           });
         });
+
+        describe('and the resource has an example, but the request does not specify any', () => {
+          it('should return the first viable example', () => {
```
philsturgeon: could we find a way to replace viable with some words which explain what that means? What constitutes viable?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
...
+            const selectedExample = flatMap(mockResource.responses, res =>
+              flatMap(res.contents, content => content.examples || []),
+            )[0];
+
+            expect(selectedExample).toBeDefined();
+            expect(response.body).toEqual((selectedExample as INodeExample).value);
+          });
+        });
+
+        describe('and the resource has not an example', () => {
```
philsturgeon:
```suggestion
        describe('and the resource has no response examples', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
             expect(response.body).toEqual((selectedExample as INodeExample).value);
           });
         });
+
+        describe('and the resource has an example, but the request does not specify any', () => {
```
philsturgeon:
```suggestion
        describe('no response example is requested specifically, and a response example is defined', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
...
+
+          const response = httpMocker.mock({
+            input: mockInput,
+            resource: resourceOperation,
+            config: { mock: { dynamic: false } },
+          });
+
+          describe('the property has an example key', () => {
+            it('should return the example key', () => expect(response.body).toHaveProperty('name', 'Clark'));
+            describe('the property has an example and a default key', () => {
```
philsturgeon: no need to repeat `the property has an example` because it is nested inside a describe that says that.
```suggestion
            describe('and a default key', () => {
```

packages/http/src/mocker/HttpMocker.ts
XVincentX: 5aed8b8

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
...
+              it('should still prefer the example', () => expect(response.body).toHaveProperty('middlename', 'J'));
```
philsturgeon: specs are about assertions so be assertive! :D 

```suggestion
              it('prefers the example', () => expect(response.body).toHaveProperty('middlename', 'J'));
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
...
+            });
+
+            describe('with a objects', () => {
```
philsturgeon: not sure what this means.

XVincentX: It basically checks whether it works also with nested objects. https://github.com/stoplightio/prism/pull/347/files#diff-e53e98c10a813d2d271cc0fa5c55ddcaR273

Maybe `with a nested object`?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Actually it's just the first example.

XVincentX:
```suggestion
          it('should return the first example', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
philsturgeon:
```suggestion
            describe('and the property containing the example is deeply nested', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,100 @@ describe('HttpMocker', () => {
...
+              expect(response.body).toHaveProperty('surname', 'Kent');
+            });
+          });
+
+          describe('the property has nor default, nor example', () => {
+            describe('is nullable', () => {
+              it('should be set to null', () => expect(response.body).toHaveProperty('age', null));
+            });
+
+            describe('and is not nullable', () => {
```
philsturgeon: What about if it is a non null enum string. 

What about if the property is oneOf/allOf/anyOf what does it show.

philsturgeon: Does it support $ref

XVincentX: > What about if it is a non null enum string.

Returns the first enum value

> oneOf

Returns the first value

> allOf

Merges and returns (as long they're compatible)

> anyOf

Returns the first value

> $ref

Apparently not supported.

I'm writing the tests for this as we speak.

XVincentX: @philsturgeon You have them now.

https://github.com/stoplightio/prism/pull/347/commits/9872b0a09cc50ec912cb7c07a1eba55d89a117bc

packages/http/package.json
```diff
@@ -16,13 +16,14 @@
...
   },
   "dependencies": {
-    "@stoplight/json-schema-faker": "^0.5.0-rc18",
     "@stoplight/prism-core": "^3.0.0-alpha.14",
     "accepts": "^1.3.7",
     "ajv": "^6.0.0",
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
+    "@stoplight/json-schema-faker": "",
```
philsturgeon: We had to bring this chap back?

XVincentX: Unfortunately yes; it's still because of the damn JSONpath library that it's used internally. I've tried to get rid of it but I decided to give up…for now. The feature is more important right now.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
             expect(response.body).toEqual((selectedExample as INodeExample).value);
           });
         });
+
+        describe('no response example is requested specifically, and a response example is defined', () => {
```
philsturgeon:
```suggestion
        describe('a response example is defined, but no response example is requested', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+          it('should return the first example', () => {
```
philsturgeon:
```suggestion
          it('returns the first example', () => {
```

Trying to be more assertive with our assertions. Clean up any old should's you find laying around 🗡

XVincentX: Roger!

chris-miaskowski:
Before I even start, can you add in the jira link (anywhere really)?

philsturgeon:
@chris-miaskowski no we dont do that on OSS remember? :D

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+          it('returns the first example', () => {
+            const response = httpMocker.mock({
+              input: mockInput,
+              resource: mockResource,
+              config: { mock: { dynamic: false } },
+            });
+
+            expect(response.body).toBeDefined();
+
+            const selectedExample = flatMap(mockResource.responses, res =>
```
chris-miaskowski: Mind simplifying it to

```suggestion
            const selectedExample = mockResource.responses[0].contents[0].examples[0]
```

?

Using flat map is cool but not correct if we had other responses defined in that object (400, 500).

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+              flatMap(res.contents, content => content.examples || []),
+            )[0];
+
+            expect(selectedExample).toBeDefined();
+            expect(response.body).toEqual((selectedExample as INodeExample).value);
+          });
+        });
+
+        describe('and the resource has no response examples', () => {
+          const resourceOperation: IHttpOperation = {
```
chris-miaskowski: Ok I see what you did here, but that object seems to be too big.
Wouldn't it make better less coupled tests if you split it into smaller, more targeted resource operations?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+                      required: ['name', 'surname', 'age', 'email'],
+                    },
+                    encodings: [],
+                  },
+                ],
+              },
+            ],
+          };
+
+          const response = httpMocker.mock({
```
chris-miaskowski: Before each? Maybe you're not mutating the response object in your tests but how do I know that? :)
Better be safe than sorry!

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+            input: mockInput,
+            resource: resourceOperation,
+            config: { mock: { dynamic: false } },
+          });
+
+          describe('the property has an example key', () => {
```
chris-miaskowski: By the time of reading this line scrolling up and down to see what the given object was is becoming painful. I can feel my mouse scroll temperature increased!

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+              {
+                code: '200',
+                headers: [],
+                contents: [
+                  {
+                    mediaType: 'application/json',
+                    schema: {
+                      type: 'object',
+                      properties: {
+                        name: { type: 'string', examples: ['Clark'] },
```
chris-miaskowski: I think you're missing examples: [] and examples: ['a', 'b']

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+                expect(response.body).toHaveProperty('paymentStatus', 'completed'));
+              it('should return the first anyOf value', () =>
+                expect(response.body).toHaveProperty('creditScore', 1958));
+              it('should return the first oneOf value', () =>
+                expect(response.body).toHaveProperty('paymentScore', 'string'));
+              it('should return the first allOf value', () =>
+                expect(response.body).toHaveProperty('walletScore', 'hello'));
+            });
+          });
+        });
```
chris-miaskowski: Note to self: do we actually need to test any of this? If I'm not mistaken we are using the library to generate this content which means we're effectively testing a library here... If so we should probably just mock the library and see it got called (?)

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,118 @@ describe('HttpMocker', () => {
...
+            });
+          });
+
+          describe('the property has nor default, nor example', () => {
+            describe('is nullable', () => {
+              it('should be set to null', () => expect(response.body).toHaveProperty('age', null));
+            });
+
+            describe('and is not nullable', () => {
+              it('should return the default string', () => expect(response.body).toHaveProperty('email', 'string'));
```
chris-miaskowski: Hmmm wait, I think that's incorrect.
The assumption was "dynamic = false"
`email: { type: 'string' }` does not have a static example...
I don't think returning `{ email: 'string' }` makes sense.
Email should not be included in the body at all.

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,12 +14,41 @@ jsf.option({
...
-  useExamplesValue: true,
-  useDefaultValue: true,
   maxItems: 20,
   maxLength: 100,
 });
 
-export async function generate(source: unknown): Promise<unknown> {
-  return jsf.resolve(cloneDeep(source));
+export function generate(source: JSONSchema): unknown {
+  return jsf.generate(cloneDeep(source));
```
chris-miaskowski: What's the difference here? `generate` vs `resolve`?

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,12 +14,41 @@ jsf.option({
...
+export function generateStatic(source: JSONSchema): unknown {
+  const transformedSchema = toOpenAPIJSONSchemaesque(source);
+  return sampler.sample(transformedSchema);
+}
+
+function hasExamples(source: JSONSchema): source is JSONSchema6 | JSONSchema7 {
+  return 'examples' in source;
+}
+
+function toOpenAPIJSONSchemaesque(schema: JSONSchema): any {
```
chris-miaskowski:
```suggestion
function toOpenApiJsonSchemaesque(schema: JSONSchema): any {
```

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,12 +14,41 @@ jsf.option({
...
+
+      if (hasExamples(innerProp) && innerProp.examples) {
+        Object.assign(innerProp, { example: innerProp.examples[0] });
+      }
+
+      return toOpenAPIJSONSchemaesque(innerProp);
+    });
+  });
+
+  return returnedSchema;
```
chris-miaskowski: I'm surprised to see you haven't use reduce ;P

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,12 +14,41 @@ jsf.option({
...
+
+function hasExamples(source: JSONSchema): source is JSONSchema6 | JSONSchema7 {
+  return 'examples' in source;
+}
+
+function toOpenAPIJSONSchemaesque(schema: JSONSchema): any {
+  const returnedSchema = cloneDeep(schema);
+
+  ['properties', 'anyOf', 'allOf', 'oneOf'].forEach(property => {
+    if (!schema[property]) return;
```
chris-miaskowski: Maybe for consistency
```suggestion
    if (!returnedSchema[property]) return;
```

packages/http/src/mocker/generator/JSONSchema.ts
chris-miaskowski: Ok, I read half of the function and it would really use some unit tests to show what it's actually doing.

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,12 +14,41 @@ jsf.option({
...
+
+    const mapFn = Array.isArray(returnedSchema[property]) ? map : (mapValues as (obj: unknown) => unknown[] | unknown);
+
+    returnedSchema[property] = mapFn(schema[property], innerProp => {
+      if (typeof innerProp === 'boolean') return innerProp;
+
+      if (hasExamples(innerProp) && innerProp.examples) {
```
chris-miaskowski: This is also true for `innerProp.examples = true`

chris-miaskowski:
> @chris-miaskowski no we dont do that on OSS remember? :D

Aaaargh! Please at least in the branch name! Or comment, anywhere :P

packages/http/src/mocker/generator/JSONSchema.ts
XVincentX: Nice catch! The second condition is meant to be `Array.isArray(examples)`. Fixed in 9f96a15

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Absolutely, I think this is a result of a copy paste that I adapted.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Good question. The thing is that the horrible function you just saw will be removed when openapi-sampler's PR will be merged and released (hopefully very soon).

This test will effectively make sure that the examples generated by the updated library still match with what we want. Does it clarify?

packages/http/src/mocker/generator/JSONSchema.ts
XVincentX: `generate` is the sync-version, with partial dereferencing through given refs — the idea is that Graphite will resolve all the references for us so we do not need to go through them again.

packages/http/src/mocker/generator/JSONSchema.ts
XVincentX: it's `JSONSchema`, right?

image

XVincentX: I can, but this is going to be removed soon (literally not later than Monday) — this function was just to "close" the card and make the tests pass while waiting for openapi sampler to be updated.

packages/http/src/mocker/generator/JSONSchema.ts
XVincentX: See https://github.com/stoplightio/prism/pull/347/files#r293477569

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: It really depends what we want to do with non required properties — this is a call for @philsturgeon

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Can you elaborate? I fail to understand

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: I need you to elaborate more, I fail to understand what you mean.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Done!

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: As in you had to go up and forth too many times to check out the property you were testing?

XVincentX: Makes sense, I can probably refactor it

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: Ah ok maybe you mean to create smaller schemas with the single properties and test them accordingly, right?

packages/http/src/mocker/generator/JSONSchema.ts
XVincentX: https://github.com/Redocly/openapi-sampler/pull/10

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: 7123262ed82371f969c4af2a8d98e4ccfa3af588

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: 7123262ed82371f969c4af2a8d98e4ccfa3af588

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: c7896e8

XVincentX:
@philsturgeon @chris-miaskowski Please give it another round.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -229,6 +228,188 @@ describe('HttpMocker', () => {
...
+                properties: {
+                  middlename: { type: 'string', examples: ['WW', 'JJ'] },
+                },
+              });
+
+              it('prefers the first example', () =>
+                expect(responseWithMultipleExamples.body).toHaveProperty('middlename', 'WW'));
+            });
+
+            describe('with no multiple example values in the array', () => {
```
philsturgeon:
```suggestion
            describe('with an empty `examples` array', () => {
```

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
chris-miaskowski: I can live with it. For now! :P

XVincentX:
@philsturgeon Do you want to give it another quick look before I merge this?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -210,43 +210,44 @@ describe('HttpMocker', () => {
...
-            expect(selectedExample).toBeDefined();
-            expect(response.body).toEqual((selectedExample as INodeExample).value);
+              expect(selectedExample).toBeDefined();
+              expect(response.body).toEqual((selectedExample as INodeExample).value);
+            });
           });
         });
 
-        describe('and the resource has no response examples', () => {
+        describe('and the response has not an examples', () => {
```
philsturgeon: This wording was better before I think. We’re trying to clarify what type of examples it doesn’t have. response examples was meant to mean content examples or something, to avoid confusing it with property examples or something else

PR #351, opened 2019-06-12
feature: add server validation the Http Client

chris-miaskowski:


chris-miaskowski:
@XVincentX @philsturgeon @lag-of-death 
Feel free to start reviewing folks. I have to add a bunch of e2e tests and update the readme but other than that it's done.

Pay special attention to tests so see whether we're all in alignment.

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -228,6 +253,67 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
...
+          expect(response.payload).toEqual(
+            '{"type":"https://stoplight.io/prism/errors#NO_SERVER_MATCHED_ERROR","title":"Route not resolved, no server matched.","status":404,"detail":"The base url https://petstore.swagger.io/v2 hasn\'t been matched with any of the provided servers"}',
+          );
+        });
+
+        test(`when the server is invalid return error`, async () => {
+          const response = await server.fastify.inject({
+            method: 'GET',
+            url: '/store/inventory?__server=https://notvalid.com',
+          });
```
philsturgeon: This all looks great! An idea for an extra rest: can we define a server with variables then see if a URL with a valid and invalid server according to those validations performs as expected?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -1,114 +1,268 @@
...
 import { IHttpOperation } from '@stoplight/types/http-spec';
+import * as nock from 'nock';
 import { relative, resolve } from 'path';
 import { createInstance, IHttpConfig, IHttpRequest, IHttpResponse, ProblemJsonError } from '../';
 import { forwarder } from '../forwarder';
 import { UNPROCESSABLE_ENTITY } from '../mocker/errors';
-import { NO_PATH_MATCHED_ERROR } from '../router/errors';
+import { NO_BASE_URL_ERROR, NO_PATH_MATCHED_ERROR, NO_SERVER_MATCHED_ERROR } from '../router/errors';
+
+const fixturePath = (filename: string) => relative(process.cwd(), resolve(__dirname, 'fixtures', filename));
```
philsturgeon: I think the relative part is superfluous, the __dirname is an absolute path so you can just resolve(__dirname, ...

XVincentX:
@chris-miaskowski I fixed the build for you.

packages/core/src/factory.ts
```diff
@@ -37,15 +39,31 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         // build the config for this request
         const configMerger = configMergerFactory(defaultConfig, customConfig, c);
         const configObj: Config | undefined = configMerger(input);
+        let inputValidations: IPrismDiagnostic[] = [];
 
         // find the correct resource
         let resource: Resource | undefined;
         if (components.router) {
-          resource = components.router.route({ resources, input, config: configObj }, defaultComponents.router);
+          try {
```
lag-of-death: `process` function starts to look very scary. It's very lengthy and full of mutations. If we keep adding to it, it's gonna be unmanageable (or it is already). @chris-miaskowski, can you do a bit of scouting and refactor this before it explodes?

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -11,14 +11,39 @@ function checkErrorPayloadShape(payload: string) {
...
+async function instantiatePrism(specPath: string) {
+  const server = createServer({}, { components: {}, config: { mock: { dynamic: false } } });
+  await server.prism.load({
+    path: relative(process.cwd(), specPath),
+  });
+  return server;
+}
+
+describe('given getOperationWithBody.oas2.json', () => {
+  test('when GET /pet with invalid body returns correct error message', async () => {
```
lag-of-death: so `when` should go to `describe` as it describes the conditions, but maybe there is some agreement on this I'm not aware of?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -1,114 +1,268 @@
...
+      describe('given invalid path', () => {
+        const request: IHttpRequest = {
+          method: 'get',
+          url: {
+            path: '/x-bet',
+            baseUrl,
+          },
+        };
+
+        test('returns input warning', async () => {
```
lag-of-death: so this `test` doesn't have `when` or `given`

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -18,7 +19,7 @@ export class HttpForwarder implements IForwarder<IHttpOperation, IHttpRequest, I
         : inputData.url.baseUrl;
 
     if (!baseUrl) {
-      throw new Error('Either one server in spec or baseUrl in request must be defined');
+      throw ProblemJsonError.fromTemplate(NO_BASE_URL_ERROR);
```
lag-of-death: Just FYI, I always have a problem with this. It is so very surprising to see a `return` at the end of the function and then, later, learn that we might never get to this `return`. I'd personally change this `if` to `if/else`. In `if` we throw, in `else`, we `return`. But I'm aware that some use this `early return` practice. I don't agree with it, it's so confusing.

packages/http/src/router/index.ts
```diff
@@ -63,36 +63,42 @@ export const router: IRouter<IHttpOperation, IHttpRequest, IHttpConfig> = {
       };
     });
 
+    matches = matches.filter(match => match.pathMatch !== MatchType.NOMATCH);
```
lag-of-death: this `route` function is so long and full of `if`s, mutations and throwing errors. The cognitive load one has to invest to understand it is a bit too much. Can this be refactored before it's too late?

XVincentX: I want to ditch it completely and use a routing library, such as https://github.com/delvedor/find-my-way

Happy to see this refactored though.

XVincentX:
P.S: I'm jumping on this in the afternoon.

packages/http-server/src/__tests__/server.oas.spec.ts
chris-miaskowski: Sure! Good idea, will add it in a moment.
It's actually interesting to know that we have test cases covering this :) 
FYI each package has its own set of unit + integration tests. E.g. testing that a server matches is currently tested with: a unit test testing the actual matcher function implementation, a unit test testing some higher level component using that function, an integration test testing the Http Client package and now an integration test in the Http Server package :D 

I have to add it to the test-harness (e2e) too :)

chris-miaskowski: Btw I've been thinking about this for a while now and a lot of the integration test "kind of" repeat. 
For instance I can have 10 integration tests checking server validation at a Http Client level (`http-prism-instance.spec.ts`).
I could have 10 exact same integrate tests at a Http Serevr level (`server.oas.spec.ts`).
Finally one could argue that we should have the same 10 tests a the test harness level...

But, even though, they sometimes spot the issue I think it's an overkill. Any thoughts on how to stay sane?

packages/core/src/factory.ts
chris-miaskowski: This is a piece of code that we don't want to touch right now. You are right that it requires refactoring  but it's a rabbit hole - trust my word on it, I've been here were this function was created.

The best I could think of right now is wrapping some of the internals in helper functions but not sure is that any good. It's kind of like touching linux core ;)

Let me know if you have any specific refactoring suggestions, I eager to hear your thoughts on that.

packages/http-server/src/__tests__/server.oas.spec.ts
chris-miaskowski: `when` doesn't describe the _condition_ but the _action_.
`given` describes the condition
So for instance
`given` condition
`when` action
`then` result

packages/http/src/__tests__/http-prism-instance.spec.ts
chris-miaskowski: I considered 'when request processed then returns input warning' to be too verbose.
Your thoughts?

packages/http/src/forwarder/HttpForwarder.ts
chris-miaskowski: Oh yes it's been debated many times and I alone have participated in such debates many times :) 
To else or not to else that is the question!

Now after all the years writing JS code I think both styles are correct and I personally let people choose their liking.
I think in general these days styling issues should be picked up by linter to be quite honest.
I remember times when 90% of code reviews would revolve around code style and in the end no one could really win the argument leaving less time to focus on functional or design flaws.

I personally like to reduce the nesting, therefore I don't add the 'else' statement if it's not logically required.

packages/http/src/router/index.ts
chris-miaskowski: @XVincentX to your point on the library I'm happy to replace the code if you find a library that does what we need. We have a very special use case here - we not only route but check the server validity.

@lag-of-death I'm happy to see what I can do if you can suggest a refactoring.

packages/http-server/src/__tests__/server.oas.spec.ts
lag-of-death: yeah, sure it would make sense if we were using cucumber. So I understand that's agreed to put `when` inside `test`/`it`?

packages/http/src/__tests__/http-prism-instance.spec.ts
lag-of-death: I was only about inconsistency, the other `tests` have `when`.

packages/http/src/forwarder/HttpForwarder.ts
lag-of-death: But it's not about styling. Let's see:
`if x then y else z`
```
if x then y
z
```
The first example clearly shows logical flow (how it is branching out). In the second one, there is nothing that shows this connection. As I said, I prefer the first one and deem it clearer, but won't argue here since this is so old and unsettled issue.

packages/core/src/factory.ts
lag-of-death: Yeah, I totally understand. But even the comments inside, they clearly show that these should be separate functions. Why even have these comments? Distinctive function names would make it clear enough what's being done. So, at least, we could change all the `let`s to `const`s and have these consts populated in this way `const x = getX(...)`. So, for example, instead of:
```typescript
// build output
        let output: Output | undefined;
        if (resource && components.mocker && (configObj as IPrismConfig).mock) {
```
let's have:
`const output = buildOutput(...)`

packages/http/src/router/index.ts
lag-of-death: at least we could move all these `if`s that throw errors and make an array of "tuples" like so:
`[{predicate: () => false, action: () => throw 'A'}, {predicate: () => true, action: () => throw 'B'}]`
and move this array out of `route`

packages/http-server/src/__tests__/server.oas.spec.ts
philsturgeon: It’s not about wether we are using cucumber or not, describe it syntax works the same in any test suite. I’ve used describe-it syntax with rspec, goconvey, kahlan, mocha, jasmine, etc. 

Chris your comment looks good but what you’ve put into the test suite looks incorrect according to the guide you just set out. 

Let’s take cucumber as a guide https://cucumber.io/docs/guides/overview/

But you don’t need to write “test(then” because you can use “it (does a thing” 

That’s what the it function is there for :)

philsturgeon: You definitely don’t want test(when

The it’s are assertions, setup goes elsewhere.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -1,114 +1,268 @@
...
+          {
+            method: 'get',
+            url: {
+              path: '/pet',
+            },
+          },
+          config,
+        );
+
+        expect(result.output!.statusCode).toEqual(200);
```
XVincentX: ditto, make sure the output is defined

README.md
```diff
@@ -119,6 +119,79 @@ Connection: keep-alive
 
 This error shows the request is missing a required property `name` from the HTTP request body.
 
+#### Server Validation
+
+You can simulate a request to a real server and verify whether it is valid according to the spec
```
philsturgeon:
```suggestion
OpenAPI lets API spec authors make only certain servers available, and they also allow certain operations to be restricted to certain servers. Make sure the server URL you plan to use is a valid server this the particular operation you are attempting. 
```

README.md
```diff
@@ -119,6 +119,79 @@ Connection: keep-alive
...
+by providing a `__server` query param.
+
+Take this minimalistic spec (oas3) example:
+
+```json
+{
+  "openapi": "3.0.0",
```
philsturgeon: It's a guess here but I think people would rather see YAML examples. Less squiggles more point.

chris-miaskowski:
Thanks for the review everyone! I hear you all!
As much as I'd love to clean up the code I'm not sure if I will have the time to boyscout it now.
Will see what I can do but the priority for me now is to get the feature in.

packages/http/src/forwarder/HttpForwarder.ts
XVincentX: Ha! I always love these discussions where everybody is right and nobody is wrong.

I won't refrain from adding my (useless) two cent:

Historically the early return was used because some compilers could effectively generate branchless code for it:

```ts
if (a) 
  return 0;
return 1;
```

Some compilers would effectively rewrite the output to be more like this:

```ts
const results = [0, 1];
return results[a];
```

Which effectively is a branchless code.

Today this is probably not true anymore — we're not writing code that's compiled, we're writing suggestions to the interpreter. Therefore clarity *can* be preferred.

README.md
```diff
@@ -119,6 +119,79 @@ Connection: keep-alive
...
+    }
+  ]
+}
+```
+
+You can make a request enforcing server validation this way
+
+```bash
+curl -X GET localhost:4010/pet?__server=http://stoplight.io/api
+# will give you a lorem ipsum response, like so:
```
philsturgeon: These comments do not make it clear what I will be seeing if this runs successfully. Why are we talking about lorum ipsum? Maybe we could just say "It will give you a mocked response as usual" without getting into dynamic or static or any other sidepoints?

packages/http-server/src/__tests__/server.oas.spec.ts
philsturgeon: Yep there is crossover of some things, but it usually goes a little like this.

image

You will not want to test every single possible thing in every single possible way in the test harness, but we will be testing some of the same things sure. 

And the HTTP Server and HTTP Client both integrate with the mocker logic. We need to make sure that integration works. Yes.

packages/core/src/factory.ts
philsturgeon: Yeah, I think once we document the HTTP Client API (this sprint?) we are going to start wanting to make the API nicer too, and refactoring the guts can be part of that general cleanup.

examples/petstore.oas3.json
```diff
@@ -1377,11 +1285,7 @@
           "status": {
             "type": "string",
             "description": "pet status in the store",
-            "enum": [
-              "available",
-              "pending",
-              "sold"
-            ]
+            "enum": ["available", "pending", "sold"]
```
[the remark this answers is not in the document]
chris-miaskowski:
```
{
  "editor.tabSize": 2,
  "editor.renderWhitespace": "boundary",
  "editor.cursorStyle": "line",
  "editor.rulers": [
    100
  ],
  "editor.formatOnSave": true,
  "editor.formatOnSaveTimeout": 3000,
  "npm.packageManager": "yarn",
  "tslint.packageManager": "yarn",
  "tslint.jsEnable": true,
  "prettier.tslintIntegration": true,
  "eslint.enable": false,
  "files.trimTrailingWhitespace": true,
  "files.insertFinalNewline": true,
  "files.watcherExclude": {
    "**/.git/objects/**": true,
    "**/node_modules/**": true,
    "**/dist/**": true,
    "**/.cache-loader/**": true
  },
  "files.exclude": {
    "**/.git": true,
    "**/.DS_Store": true,
    "**/lib": true,
  },
  "search.exclude": {
    "**/.git/objects": true,
    "**/node_modules": true,
    "**/yarn.lock": true,
    "**/dist": true,
    "**/.cache-loader/**": true
  },
  "typescript.tsdk": "./node_modules/typescript/lib"
}
```

CONTRIBUTING.md
```diff
@@ -95,6 +95,10 @@ rm -rf packages/**/.tsbuildinfo
 
 …then try again. If it does not work, you can blame Vincenzo and write him.
 
+3. VS Code complains that it cannot find `@stoplight/prism-*` package
```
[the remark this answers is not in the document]
XVincentX: I confirm and in any case if that happens it's something I really want to fix instead of writing we can't make it work.

packages/http/src/__tests__/http-prism-instance.spec.ts
chris-miaskowski: Yup, spot on @lag-of-death! Commit is on its way.

CONTRIBUTING.md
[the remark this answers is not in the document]
chris-miaskowski: You cool leaving it in before we fix it though?

README.md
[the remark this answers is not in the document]
chris-miaskowski: Commited @philsturgeon 's suggestion ;)

CONTRIBUTING.md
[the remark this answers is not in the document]
XVincentX: I meant I confirm it's not an issue — are you still receiving this problem?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -1,114 +1,268 @@
...
+              severity: DiagnosticSeverity.Warning,
+            },
+          ]);
+        });
+
+        test('make a http request anyway', async () => {
+          // note that we are 'nocking' the request in beforeEach
+          const result = await prism.process(request, config);
+
+          expect(result.output!.statusCode).toEqual(200);
```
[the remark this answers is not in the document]
chris-miaskowski: Meh :P I don't see that much of a value. Cool do add the change but I wouldn't mind leaving it the way it is.

CONTRIBUTING.md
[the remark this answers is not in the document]
chris-miaskowski: ah, not anymore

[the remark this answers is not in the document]
chris-miaskowski: will fix

packages/core/src/factory.ts
```diff
@@ -37,15 +39,31 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
+            resource = components.router.route({ resources, input, config: configObj }, defaultComponents.router);
+          } catch (error) {
+            const { message, name, status } = error as ProblemJsonError;
+            // rethrow error we if we're attempting to mock
+            if ((configObj as IPrismConfig).mock) {
```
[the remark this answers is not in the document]
chris-miaskowski: Hmm, I don't get it.

> Shouldn't we check that it's a server mismatch error?

If **mock is true** we should re-throw **all** errors (server mismatch, path mismatch, method mismatch whatever).

If **mock is false** (aka. forwarding is turned on) don't throw errors but put the on the input validation errors list.

> Otherwise this won't throw the "route not matched" in case mock is enabled

Wait, isn't my code doing the exact opposite? If mock is enabled it **will** throw "route not matched"

packages/http/package.json
```diff
@@ -27,5 +27,9 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
chris-miaskowski: Damn! Now you know I didn't read the contributing docs :trollface:

packages/http/src/router/index.ts
chris-miaskowski: Hey I'd like to leave that for some other time. I'm usually all for boy scouting but lots of things will change in the code base in next weeks/months so that may soon be irrelevant.

README.md
```diff
@@ -119,9 +119,69 @@ Connection: keep-alive
 
 This error shows the request is missing a required property `name` from the HTTP request body.
 
+#### Server Validation
+
+OpenAPI lets API spec authors make only certain servers available, and they also allow certain operations to be restricted to certain servers. Make sure the server URL you plan to use is a valid server this the particular operation you are attempting.
+by providing a `__server` query param.
```
XVincentX: I guess this should be on the same line

packages/core/src/factory.ts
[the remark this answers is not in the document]
XVincentX: Yeah, I just didn't read it correctly

packages/core/src/factory.ts
```diff
@@ -37,23 +39,41 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         // build the config for this request
         const configMerger = configMergerFactory(defaultConfig, customConfig, c);
         const configObj: Config | undefined = configMerger(input);
```
XVincentX: This is probably an old leftover.

```suggestion
        const configObj: Config = configMerger(input);
```

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -11,19 +11,86 @@ function checkErrorPayloadShape(payload: string) {
...
+    it('returns 404 and problem json payload', () => {
+      return expect(requestPetGivenServer(serverUrl)).resolves.toMatchObject({
+        statusCode: 404,
+        payload: expectedPayload(serverUrl),
+      });
+    });
+  });
+
+  const expectedPayload = (serverUrl: string) =>
+    `{"type":"https://stoplight.io/prism/errors#NO_SERVER_MATCHED_ERROR","title":"Route not resolved, no server matched.","status":404,"detail":"The base url ${serverUrl} hasn\'t been matched with any of the provided servers"}`;
```
XVincentX: I'm wondering if there's a way to check out this stuff without using this templated string.

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -11,19 +11,86 @@ function checkErrorPayloadShape(payload: string) {
...
+      method: 'GET',
+      url: '/pet',
+      payload: {
+        id: 'strings are not valid!',
+      },
     });
+
+    expect(response.statusCode).toBe(422);
+    expect(response.payload).toEqual(
+      '{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid: [{\\"path\\":[\\"body\\"],\\"code\\":\\"type\\",\\"message\\":\\"should be object\\",\\"severity\\":0}]"}',
```
XVincentX: Yeah, it might be worth parse this content

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -11,19 +11,86 @@ function checkErrorPayloadShape(payload: string) {
...
+    );
+    await server.fastify.close();
```
XVincentX: This should be in a `after` statement

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -228,6 +295,67 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
     }
   });
 
+  describe('server validation: given __server query param', () => {
```
XVincentX: According to Phil (or at least, if I understood correctly), these should be two nested describe blocks

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -228,6 +295,67 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
...
+      const response = await server.fastify.inject({
+        method: 'GET',
+        url: '/pets/10?__server=https://petstore.swagger.io/v2',
+      });
+
+      expect(response.statusCode).toBe(200);
+    });
+
+    // oas2 does not support overriding servers
+    if (file === 'petstore.oas3.json') {
```
XVincentX: This is ok for now, but maybe we could have oas3 specific tests instead of branching

README.md
```diff
@@ -119,9 +119,69 @@ Connection: keep-alive
...
+
+```bash
+curl -X GET localhost:4010/pet?__server=http://stoplight.io/api
+# This will return a 200 response
+```
+
+On the other hand, putting a server which is not defined in the specification, for example:
+
+```bash
+curl -X GET localhost:4010/pet?__server=ftp://acme.com/api
```
philsturgeon: Oh this is weird that we introduced FTP here just without any explanation, just snuck it in there. I honestly do not know if FTP is even allowed in OpenAPI, it doesn't mentioned either way but... it has never come up. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#serverObject

PR #355, opened 2019-06-14
Docs/http api

mallachari:
## Checklist

Http api docs update

Currently  http module is not documented


README.md
```diff
@@ -119,6 +119,49 @@ Connection: keep-alive
 
 This error shows the request is missing a required property `name` from the HTTP request body.
 
+### HTTP API
```
philsturgeon: Do we want this in the main README? It’s getting rather full.

Maybe we could have one example in the main readme then link to the package README for the rest of it? Would that be a decent compromise?

README.md
```diff
@@ -119,6 +119,49 @@ Connection: keep-alive
...
+
+Prism provides [HTTP API][http-docs] as separate module for forwarding and mocking requests.
+
+#### Forwarding example
```
philsturgeon: We’re trying to hide forwarding until we can figure out how we want to do it properly, not advertise it. Could we remove this one?

packages/http/README.md
```diff
@@ -1,3 +1,214 @@
 # Prism HTTP
 
-TBD
+This package provides HTTP APIs for forwarding and mocking requests.
```
philsturgeon: The HTTP client package is meant to be a generic HTTP client which has the ability to also mock responses. We should be advertising it like that, and we want to avoid talking about forwarding because that’s an internal implementation detail of a proxy we no longer use.

README.md
chris-miaskowski: That's a good plan @philsturgeon I will pick it up from @mallachari and make the adjustements

README.md
chris-miaskowski: I will look into it and reword it.

README.md
```diff
@@ -119,6 +119,49 @@ Connection: keep-alive
...
+
+```ts
+const data = await forwarder.forward({
```
chris-miaskowski: FYI @mallachari we should not be using this module as a standalone tool anyway. I will describe how to use it in a bit (not that if used this way you won't get stuff like input/output validation)

packages/http/README.md
```diff
@@ -1,152 +1,148 @@
...
+            enum:
+              - eat
+              - drink
+      responses:
+        200:
+          description: Get Todo Items
+```
 
-### Forwarder props
+## I want to use Prism to mock responses
```
philsturgeon: Could we add a quick sentance on why? 

Talk about how it can let you mock without needing to hit an actual server instance, which is great for browser-based mocking, etc.

packages/http/README.md
```diff
@@ -1,152 +1,148 @@
...
-  validations: {
-    input: IPrismDiagnostic[];
+```bash
+{ statusCode: 200,
+  headers: { 'Content-type': 'text/plain' },
+  body: undefined }
 ```
 
-Validations are not supported at the moment for Http Forwarder.
+## I want Prism to make a http request a server
```
philsturgeon: This might not be clear enough for people without context. We want to explain that Prism HTTP Client is a generic HTTP client first, and it also happens to have mock mode enabled. 

Imagine we are releasing a axios competitor, this README https://www.npmjs.com/package/axios should be an inspiration. 

One of the features of our HTTP client is that it can shortcut and return mocks instead. 

Instead of talking to localhost:4010 which might confuse people and make them think they are talking to Prism HTTP Server, why don't we make calls to https://pokeapi.co/ or api.example.org or literally anything other than localhost.

chris-miaskowski:
@XVincentX @philsturgeon I'm done, please review. Feel free to point out all silly grammar and all that non-native mistakes I made.

@mallachari I ended up having to change the readme quite significantly but I hope you don't mind. I have focused on the higher level function (process) vs the mock and forward. Although we will document the latter at some point as well, right now we need to focus on simpler cases.

README.md
```diff
@@ -216,3 +224,8 @@ If you are interested in contributing to Prism itself, check out our [contributi
 [cli]: https://www.npmjs.com/package/@stoplight/prism-cli
 [2.x]: https://github.com/stoplightio/prism/tree/2.x
 [cli-docs]: packages/cli/README.md
+[http-docs]: packages/http/README.md
+
+```
```
tbarn: Any particular reason this is here?

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
-# Prism HTTP
+# Http Client
```
tbarn: "HTTP" over "Http"

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
 
-TBD
+This package provides a http client featuring the ability to:
+
+- mock responses based of OAS spec
```
tbarn: "OpenAPI" over "OAS"

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+              - eat
+              - drink
+      responses:
+        200:
+          description: Get Todo Items
+```
+
+## I want to mock responses of operations defined in an OAS file
+
+[Try it out!](https://repl.it/@ChrisMiaskowski/prism-http-client-basic-mocking)
```
tbarn: This is cool!

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+
+- mock responses based of OAS spec
+- make requests to a http server
+- validate the request and/or the response according to the provided OAS
+
+The goal of this document is to provide you with some basic code examples to get you started and to cover some of the advanced scenarios.
+
+# Installation
+
+`yarn add @stoplight/prism-http`
```
philsturgeon:
```suggestion
```
yarn add @stoplight/prism-http
```
```

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+- turn mocking on and off
+- influence Prism's mocked response generation strategy
+
+The actual interface looks like this (but rather than explain each property I will give you some examples)
+
+```ts
+export interface IHttpConfig extends IPrismConfig {
+  mock: false | IHttpOperationConfig;
+
+  validate?: {
```
XVincentX: This is stuff that non TypeScript people might struggle to understand.

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+
+### Config Property
+
+Prism's config object (`IHttpConfig`) allows you to manipulate Prism's behaviour in many ways. For instance:
+
+- turn validation on and off
+- turn mocking on and off
+- influence Prism's mocked response generation strategy
+
+The actual interface looks like this (but rather than explain each property I will give you some examples)
```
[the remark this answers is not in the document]
chris-miaskowski: I did that intentionally. I felt a reference would be less useful. I think most developers would rather see a bunch of examples rather than description of types. Often those descriptions are not telling much.

Plus I think if we were to simply document types... We can automatically generate such documentation with jsdoc or whatever.

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+
+This will enforce a 403 response (given that such response is defined in your OAS file).
+
+## Loading specs
+
+When you load a spec you "teach" prism how your API looks like.
+You can run prism without loading a spec, but features such as validation will be disabled.
+
+```javascript
+const loaderConfig = { path: '/absolute/file/path.oas3.yaml' };
```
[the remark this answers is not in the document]
chris-miaskowski: Sure, lots of this stuff will eventually change :) Let's update once it does

packages/http/README.md
chris-miaskowski: Any suggestions? The reason I copied this he is I didn't want to introduce yet another format for documenting our interfaces.

chris-miaskowski:
Thanks you for your time everyone. Great comments, really appreciate it! Let me know once you're donedone and I will start addressing.

philsturgeon:
Done, I click submit review when done so they all come at once. 

--
Phil Sturgeon
@philsturgeon

XVincentX:
I'm done too, I use the review feature!

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+## I want to mock responses of operations defined in an OAS file
+
+[Try it out!](https://repl.it/@ChrisMiaskowski/prism-http-client-basic-mocking)
+
+```javascript
+const Prism = require('@stoplight/prism-http');
+const path = require('path');
+
+// Create Prism instance and configure it as a mocker generating static examples
+const config = { mock: { dynamic: false } };
```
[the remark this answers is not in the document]
XVincentX: It does not.

Mock is either `disabled` or configured accordingly.

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+    console.log(prismResponse.output);
+  });
+```
+
+Output
+
+```bash
+{ statusCode: 200,
+  headers: { 'Content-type': 'text/plain' },
+  body: undefined }
```
XVincentX:
```suggestion
  body: undefined 
}
```

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+```
+
+## I want to make a http request a server
```
XVincentX:
```suggestion
## Proxying requests to an upstream server
```

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+
+In this use case we assume we have a server running at `http://localhost:4010`
```
XVincentX: This is a little bit confusing; this is usually the address Prism listens to. Maybe let's assume the server is running on 3030 or 8080?

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+    },
+  })
+  .then(prismResponse => {
+    console.log(prismResponse.output);
+  });
+```
+
+In response you'll get whatever the server responds with.
+
+## I want all my requests to talk to a server unless I override it per a reqeust
```
XVincentX:
```suggestion
## Overriding forwarding to mock for a single request
```

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+```
+
+There are two (both optional) arguments you can supply `createInstance` with:
+
+- config (of `IHttpConfig` type)
+- components (of `TPrismHttpComponents` type)
+
+We will cover the `config` argument in next section and we'll leave `components` for some other time (overriding default `components` is the _ultimate advanced stuff_).
+
+### Config Property
```
XVincentX:
```suggestion
### Config Object
```

packages/http/README.md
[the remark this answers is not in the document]
XVincentX: Automatically generated docs: nobody will never look at it. Do you remember our fantastic TSDocs for Graphite and other projects? Did somebody ever look at them?

We eliminated them at the end.

I'm ok with the format; we're going to need a complete reference probably at some point though.

packages/http/README.md
XVincentX: Eh, not right now unfortunately. Maybe revisiting the API could help us to make it better, so I guess we'll keep it as it is right now.

packages/http/README.md
```diff
@@ -1,3 +1,376 @@
...
+
+```javascript
+const config = {
+  mock: {
+    example: 'key',
+  },
+};
+```
+
+In OAS3 one can construct such response that will have multiple examples mapped by keys.
```
XVincentX:
```suggestion
In OpenAPI Specification 3 one can construct such response that will have multiple examples mapped by keys.
```

packages/http/README.md
```diff
@@ -1,152 +1,148 @@
...
 
-```ts
-interface IPrismInput<IHttpRequest> {
-  data: IHttpRequest;
-  validations: {
-    input: IPrismDiagnostic[];
+```bash
+{ statusCode: 200,
+  headers: { 'Content-type': 'text/plain' },
+  body: undefined }
```
philsturgeon: Do we really want an undefined body as the primary example for a response mocking tool? If feels like we are missing an opportunity to actually show off what prism does.

packages/http/README.md
philsturgeon: @XVincentX i was suggesting a slightly different direction, to make it even clearer, we could use not localhost.

> Instead of talking to localhost:4010 which might confuse people and make them think they are talking to Prism HTTP Server, why don't we make calls to https://pokeapi.co/ or api.example.org or literally anything other than localhost.

packages/http/README.md
chris-miaskowski: > Instead of talking to localhost:4010 which might confuse people and make them think they are talking to Prism HTTP Server, why don't we make calls to https://pokeapi.co/ or api.example.org or literally anything other than localhost.

This is exactly what I did :) I tested it using the prism CLI

We could use api.example.org or pokeapi.co buuuuut api.example.org does not return stuff... pokeapi.co doesn't do anything relevant either.

Instead, if someone runs the Prism CLI with the spec provided at the top of the documentation they will be able to actually test this stuff.

philsturgeon: Ok, I like the idea that this should definitely actually run. Pokeapi is an API with JSON data so if you take a second look you'll see you can construcvt URLs.

https://pokeapi.co/api/v2/pokemon/ditto/

JSON! 

I respect that using the CLI is a good way to have a server running locally that we know they have, but can you see how it is confusing for the user? Using prism to talk to prism is meta and conflates two different components and their use.

packages/http/README.md
chris-miaskowski: Hey, I have built similar under personal account for Spectral. let's get a SL one and I will replace these at that point.

packages/http/README.md
chris-miaskowski: Yeah good point :P

packages/http/README.md
chris-miaskowski: Gotcha! Let's GO with Pokemons! ;)

chris-miaskowski: Actually, I picked cats :P

chris-miaskowski:
@tbarn @philsturgeon @XVincentX all done. mind re-reviewing? hope to get this merged today.

philsturgeon:
The intention was more guide/example than API Reference you are absolutely right. If we want ref docs for the JS API we can use JS API reference doc generates, and link to it from below our quick examples here.

packages/http/README.md
```diff
@@ -1,3 +1,405 @@
...
+  });
+```
+
+Output
+
+```bash
+{
+  statusCode: 200,
+  headers: { 'Content-type': 'text/plain' },
+  body: 'Write lit OpenAPI specs!'
```
philsturgeon:
```suggestion
  body: 'Write literally any other word than lit OpenAPI specs!'
```

packages/http/README.md
```diff
@@ -1,3 +1,405 @@
...
+### Server Validation
+
+Consider a request
+
+```javascript
+prism.process({
+  method: 'get',
+  url: {
+    path: '/path', // must be prefixed with slash
+    baseUrl: 'http://localhost:4010',
```
philsturgeon: Are we keeping localhost:4010 here for a reason? We changed the other instances to a real server for clarity, is this different ?

chris-miaskowski: we are not, I'm dumb :P

packages/http/README.md
```diff
@@ -1,3 +1,405 @@
...
+          schema:
+            enum:
+              - eat
+              - drink
+      responses:
+        200:
+          description: Get Todo Items
+          content:
+            'text/plain':
+              example: Write lit OpenAPI specs!
```
philsturgeon:
```suggestion
              example: Write dope OpenAPI specs!
```

@chris-miaskowski you missed one

PR #356, opened 2019-06-15
Make the build process decent

XVincentX:
I was fundamentally tired of the TypeScript server complaining multiple times per day because it wasn't able to find something or stuff around, so I decided that destroying the current build process and redoing it from scratch was a good idea.

1. Output directories for all the projects is now `dist` instead of `lib` — this should align with what other projects do, `lib` often refers to source code.
2. Revisited jest configuration — some options were not taking in account how Jest really works in the resolutions.
3. Redone all the ts configuration files — they're minimal now and every project is getting only the references it really needs. This made me discover a leak reference (basically the `core` package was requesting something from the upper level `http` package). We were **very** lucky because the required thing was just a type definition, so it got stripped by TypeScript but otherwise the published package would have been broken. e2ccfd6

Please do not blindly accept this — pull down the branch, build it, use it a bit and we'll go from there.


philsturgeon:
@XVincentX looks great! When you say pull it down and play around, what am I meant to be doing to check this works?

XVincentX:
I guess the best thing to do would be to use the branch while developing some features and see if the TypeScript Server in VSCode goes nuts or not.

Apparently all works fine on Sublime and when compiling the software.

XVincentX:
Update: the TS Server in VSCode *still* goes nuts sometimes, and apparently there's nothing we can do about it.

XVincentX:
image

PR #359, opened 2019-06-17
Devops/docker

MikeLebeau:
## Checklist

Add Dockerfile in the repository to make docker use obvious/easier for users.

The official docker image is not up-to-date and does not point to any repository.

Adding the dockerfile to this repository may be the easy way to keep docker image in sync.

Nope :D


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:18.10
```
XVincentX: We do not need to use such big default image, we can probably go with a minimal based on node.

```suggestion
FROM node:12-alpine
```

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+RUN apt-get -qq update \
```
XVincentX: We're not going to need these updates

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+&& apt-get -qq -y install curl \
+&& curl -L https://raw.githack.com/stoplightio/prism/master/install | sh
```
XVincentX: This needs a discussion. What should our Docker image be based on? The binaries we provide or the npm package?

If it's the former, we're going to need node as base image, otherwise we can probably just go with the scratch one.

philsturgeon:
All great points @XVincentX.

Just one thing:

> This is missing the CI automation to publish the new image when we release Prism.

I think that would be a job for us to handle separately. Making a docker image means people can clone this down and run it with docker, that is a good step on its own. Then we can push those images out somewhere else, maybe closer to v3.0 final.

XVincentX:
All right so maybe what we can do here is to create and merge a `Dockerfile` that builds the image starting from the `source code` in the repository directly of trying to create a production image from the artefacts, whose process to get accepted is going to be a little bit harder.

@MikeLebeau What do you think? If you're up for the task, I can give you some other suggestions and make it happen!

MikeLebeau:
Yo !
Yeah! I am OK to take this point 😄 
Other suggestions ? I'm listening 😄 

OK, I have seen the git dependency on `jsonpath`.
What do you recommend :
    - Build prism before build docker ? Or build prism inside docker ?
    - Once I have build prism (with `yarn build`), how can I embed the libs within docker and make it work ?

philsturgeon:
jsonpath is a special case, because we use a fork pegged to a specific commit. We are erradicating this from our ecosystem one project at a time but we gotta do it for prism too. https://github.com/stoplightio/spectral/pull/257

Fancy a challenge? 😅

XVincentX:
Hey @MikeLebeau ,

this is not going to be exactly that trivial, since Prism is a multi package and needs a compilation step as well. We're going to need to leverage the MultiStage build feature.

The docker file should resemble something like this:

```Dockerfile
FROM node:12 as compiler

WORKDIR /usr/src/prism

COPY package.json yarn.lock packages/* /usr/src/prism/

RUN yarn
RUN yarn build

FROM node:12 as dependencies

ENV NODE_ENV production
COPY package.json ./package.json
RUN yarn

FROM node:12-alpine

WORKDIR /usr/src/prism
COPY --from=dependencies /usr/src/prism/node_modules /usr/src/prism/node_modules
COPY --from=compiler /usr/src/prism/**/dist /usr/src/prism/**/dist
```

Not exactly a 30 minutes job :)

KillianH:
Hey @MikeLebeau ,

Don't hesitate if you need some help :)
I've modified a little bit the dockerfile provided by @XVincentX 

 I don't understand really well what you try to achieve with 
`COPY --from=compiler /usr/src/prism/**/dist /usr/src/prism/**/dist`

The double asterisk don't work with the copy command. (It doesn't have the same behavior as the cp command)

For the `dependencies` stage there's no direct dependencies in the package.json file so nothing is really happening on this stage. And so this COPY command is useless ? (Or I'm completly mistaken and I don't understand really well how to build prism thats would be pretty likely the case)
`COPY --from=dependencies /usr/src/prism/node_modules /usr/src/prism/node_modules`

```
FROM node:12 as compiler

WORKDIR /usr/src/prism

COPY package.json yarn.lock /usr/src/prism/
COPY packages /usr/src/prism/packages/
RUN yarn
RUN yarn run build

FROM node:12 as dependencies

WORKDIR /usr/src/prism

ENV NODE_ENV production
COPY package.json /usr/src/prism/
RUN yarn

FROM node:12-alpine

WORKDIR /usr/src/prism
COPY --from=dependencies /usr/src/prism/node_modules /usr/src/prism/node_modules/
COPY --from=compiler /usr/src/prism /usr/src/prism/build/
```

XVincentX:
Hey @KillianH,

thanks for looking into this!

> The double asterisk don't work with the copy command.

I know :) As I mentioned in the comment above, that was what the `Dockerfile` should _look like_, it was not a complete and working file. The idea was to show @MikeLebeau the direction and the intentions of the file.

> For the dependencies stage there's no direct dependencies in the package.json file so nothing is really happening on this stage.

I'm not too sure if I understand correctly. When running `yarn` on a workspace it will follow the `yarn.lock` file that has got all the dependencies, included the one from the inner package.json files.

In general the idea is:

1. Compile from TypeScript
2. Install prod dependencies
3. Copy stuff in a cleaner container and make it run

MikeLebeau:
Hey what's up @KillianH ?!
Thanks for your help 😄

Yes I have seen the double asterisk mistake ;D
And for the dependencies stage, Yes, as said @XVincentX yarn will pull all dependencies in `yarn.lock` file

KillianH:
> Hey @KillianH,
> 
> thanks for looking into this!
> 
> > The double asterisk don't work with the copy command.
> 
> I know :) As I mentioned in the comment above, that was what the `Dockerfile` should _look like_, it was not a complete and working file. The idea was to show @MikeLebeau the direction and the intentions of the file.
> 
> > For the dependencies stage there's no direct dependencies in the package.json file so nothing is really happening on this stage.
> 
> I'm not too sure if I understand correctly. When running `yarn` on a workspace it will follow the `yarn.lock` file that has got all the dependencies, included the one from the inner package.json files.
> 
> In general the idea is:
> 
> 1. Compile from TypeScript
> 2. Install prod dependencies
> 3. Copy stuff in a cleaner container and make it run

Oh I was missing some files that's why :)

So something more like this ?

I'll let @MikeLebeau do the
> Copy stuff in a cleaner container and make it run 
 :smile:

```
FROM node:12 as compiler

WORKDIR /usr/src/prism

COPY package.json yarn.lock /usr/src/prism/
COPY packages /usr/src/prism/packages/
RUN yarn
RUN yarn run build

FROM node:12 as dependencies

WORKDIR /usr/src/prism

ENV NODE_ENV production
COPY package.json yarn.lock /usr/src/prism/
COPY packages /usr/src/prism/packages/
RUN yarn

FROM node:12-alpine

WORKDIR /usr/src/prism
COPY --from=dependencies /usr/src/prism/node_modules /usr/src/prism/node_modules/
COPY --from=compiler /usr/src/prism /usr/src/prism/build/
```

> Hey what's up @KillianH ?!

Hey :sunglasses:

MikeLebeau:
Ok! I can build and run this docker image !

I've encountered some strange problems :
    - If I run `npm install tsconfig-paths/register` after `yarn`, `@oclif/command` contains only node_modules folder...

    - I need to install `globby` manually 

Have you some more advises ? Dockerfile improvement ?

XVincentX:
Hey @MikeLebeau and @KillianH,

I went ahead and I made it work. As you can probably see from the new file, it was a little bit more complicated than I was originally thinking, mostly because the project is split into multiple directories and collecting the dist files here and there with their `package.json` is a little bit repetitive :)

KillianH:
I understand a little bit better the prism architecture like this haha.

I'll try this out :)

XVincentX:
Do not make the file scary you though — this is an image built from sources, which probably we're not going to use anyway.

The production docker image will effectively a `scratch` with the binary built in inside, it's going to be much easier :)

PR #361, opened 2019-06-18
feat: expose responseType

P0lip:
## Checklist

Feature

responseType is included

No

### DIFF
--- a/packages/http/src/forwarder/HttpForwarder.ts
+++ b/packages/http/src/forwarder/HttpForwarder.ts
@@ -39,6 +39,7 @@ export class HttpForwarder implements IForwarder<IHttpOperation, IHttpRequest, I
       statusCode: response.status,
       headers: response.headers,
       body: response.data,
+      responseType: (response.request && response.request.responseType) || '',
     };
   }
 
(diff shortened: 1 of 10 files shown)


philsturgeon:
Wait _should_ it be empty for all these instances? Should it not be `"json"`?

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -39,6 +39,7 @@ export class HttpForwarder implements IForwarder<IHttpOperation, IHttpRequest, I
       statusCode: response.status,
       headers: response.headers,
       body: response.data,
+      responseType: (response.request && response.request.responseType) || '',
```
XVincentX: Can really the request be undefined? This is really weird. I'd say it's a problem in the types.

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -72,6 +72,7 @@ export class HttpMocker implements IMocker<IHttpOperation, IHttpRequest, IHttpCo
         'Content-type': negotiationResult.mediaType,
       },
       body: mockedBody,
+      responseType: '', // todo(Vincenzo): to be implemented
```
XVincentX: I have no idea what to put here in a mocked response to be honest.
The spec gives some value that I will probably never going to use; it does not even support XML. 

@philsturgeon

packages/http/src/forwarder/HttpForwarder.ts
P0lip: It's `any`, so well, might be a null.

PR #366, opened 2019-06-19
fix: transfer rootDir and out to build configuration files

XVincentX:
TypeScript's server sometimes goes just nuts and shows errors only after minutes and minutes of usages.

I looked at its source code *yet another time* and I hope this PR will settle the errors from VSCode. Basically VSCode thinks it needs to compile the code but it's not true, so I've transferred as much options as possible to the build files.

### DIFF
--- a/packages/tsconfig.json
+++ b/packages/tsconfig.json
@@ -1,5 +1,4 @@
 {
-  "exclude": ["**/__tests__"],
   "compilerOptions": {
     "baseUrl": ".",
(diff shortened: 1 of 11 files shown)


packages/tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
-  "exclude": ["**/__tests__"],
```
chris-miaskowski: :+1:

PR #368, opened 2019-06-19
feat: remove __contentType query parameter support

XVincentX:
If you want such feature, use the Accept header instead

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,10 +41,9 @@ jobs:
       - image: circleci/node:10
     steps:
       - checkout
-      - restore_cache:
-          name: Restore Yarn Package Cache
-          keys:
-            - yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install Dependencies
+          command: yarn
       - run:
           name: Install dockerize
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -41,10 +41,9 @@ jobs:
       - image: circleci/node:10
     steps:
       - checkout
-      - restore_cache:
```
lag-of-death: hm.. why no longer the cache?

XVincentX: Good point!

So if you look at the file you can see it was currently *only restoring the cache* and nothing else.

My hope was that I could be able to reuse the cache from some previous steps instead of reinstalling the dependencies — but it turns out that since build step and harness are now ran in parallel the thing does not hold anymore.

lag-of-death: get it now @XVincentX, thanks! So we now can rm https://github.com/stoplightio/prism/blob/a777fb05584788def78904bc86626b007cac0930/.circleci/config.yml#L24 ?

XVincentX: Not exactly. The cache can be still be leveraged by the same job that creates it — it simply cannot be used by a different job.

lag-of-death: ah.. I was looking at the wrong line! My bad!

PR #369, opened 2019-06-19
refactor: convert test harness to typescript

XVincentX:
…so that the code is inline with all our stuff we have here around


test-harness/createMasterFiles.ts
```diff
@@ -27,6 +27,4 @@ async function recordMasterFile({ path, method, headers, body }) {
   }
 }
 
-(async function () {
-  await Promise.all(requests.map(recordMasterFile));
-})();
+Promise.all(requests.map(recordMasterFile));
```
lag-of-death: curious, does it not end the execution of `test-harness/createMasterFiles.ts` prematurely now (does it save all the files to be saved)?

package.json
```diff
@@ -66,7 +67,7 @@
     "trailingComma": "es5"
   },
   "lint-staged": {
-    "*.ts": [
+    "*packages/**/.ts": [
```
lag-of-death: yeah, I like it, we have test-harness in `ts`, but not forced linting. We might want to do that in the future though

package.json
```diff
@@ -79,9 +80,5 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  },
-  "publishConfig": {
```
lag-of-death: a bit lost here, why do we remove it now?

package.json
XVincentX: My intention is to enable it — but right now there is a lot the linter would probably complain.

package.json
XVincentX: Because Lerna ignore that thing at all — since it publishes the packages singularly looking at the package.json of the packages — not the main one.

test-harness/createMasterFiles.ts
lag-of-death: so the way it is now (without `async/await`), will `node` wait for the promises to finish?

XVincentX: Correct! Node will always wait for all the open handles before closing the process. 

```javascript
new Promise(resolve => {
  setTimeout(resolve,5000);
})
```

If you put this in a `test.js` and you execute it, you'll see it will wait before exiting the process.

lag-of-death: ha! just tested it :) you learn something new every day

PR #370, opened 2019-06-19
Improve Problem JSON Error Format for Validation

lag-of-death:
## Checklist

- [x] Tests have been adjusted (for bug fixes / features)

Feature.

The current behavior gives validation error details as a JSON with a string (namely: `Your request body is not valid`) concatenated together.
Now, the details (422 validation errors) are just that JSON but under `validation` field. `detail` field is now just a string with a description of the problem. Also, in the JSON, `path` field was renamed to `location` and `severity`s field value is now represented with a string (like `Error`), not a number.

This change might involve some changes in handling Prism-generated 422 responses.


packages/http/src/types.ts
```diff
@@ -90,16 +102,37 @@ export class ProblemJsonError extends Error {
     return error;
   }
 
+  public static asValidationIssue(error: Error & { detail: IPrismDiagnostic[]; status?: number }): any {
```
XVincentX: The spec for the `problem/json` says that effectively additional field apart from the standard one are accepted; I'd suggest, instead of creating a new function to handle the specific case, we could extend the `fromTemplate` function to add an `additional: unknown` parameter that we can add to the payload, in this way the final code should be kind of simplified.

packages/http/src/types.ts
```diff
@@ -90,16 +102,37 @@ export class ProblemJsonError extends Error {
...
       detail: error.detail || '',
     };
   }
 
-  constructor(readonly name: string, readonly message: string, readonly status: number, readonly detail: string) {
+  constructor(
+    readonly name: string,
+    readonly message: string,
+    readonly status: number,
+    readonly detail: string | IPrismDiagnostic[],
```
XVincentX: Is this correct though? According to the code I saw so far it's always a string — and then maybe there might be an additional `validation` property; that one is an `IPrismDiagnostic` one.

test-harness/gold-master-files/__path____no_auth_pets___method___POST___headers____Content-Type___application_json____body____id__1__petId__2__quantity__3__shipDate___2002-10-02T10_00_00-05_00___status___placed___complete__true__.json
```diff
@@ -12,14 +12,34 @@
...
+      "content-length": ["614"],
       "connection": ["close"]
     },
     "body": {
       "type": "https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY",
       "title": "Invalid request body payload",
       "status": 422,
-      "detail": "Your request body is not valid: [{\"path\":[\"body\"],\"code\":\"required\",\"message\":\"should have required property 'name'\",\"severity\":0},{\"path\":[\"body\"],\"code\":\"required\",\"message\":\"should have required property 'photoUrls'\",\"severity\":0},{\"path\":[\"body\",\"status\"],\"code\":\"enum\",\"message\":\"should be equal to one of the allowed values\",\"severity\":0}]"
+      "detail": "Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.",
+      "validation": [
```
XVincentX: The output looks so cool now, fantastic! 👍

philsturgeon: Agreed! Great job.

packages/http-server/src/server.ts
```diff
@@ -123,7 +123,11 @@ const replyHandler = <LoaderInput>(
           .type('application/problem+json')
           .serializer(JSON.stringify)
           .code(status)
-          .send(ProblemJsonError.fromPlainError(e));
+          .send(
+            (e.name.includes('UNPROCESSABLE_ENTITY')
```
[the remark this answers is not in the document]
lag-of-death: hm.. I don't think `ProblemJsonError` has `.type` property

packages/http/src/types.ts
lag-of-death: https://github.com/stoplightio/prism/blob/47d4cad1151b97974e754b0e36de57347fed217e/packages/http/src/mocker/HttpMocker.ts#L62
and then `fromTemplate` does `const error = new ProblemJsonError(...)`

packages/http-server/src/server.ts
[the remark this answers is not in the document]
XVincentX: Ah interesting, you spotted a field that is declared in the type:
https://github.com/stoplightio/prism/blob/master/packages/core/src/types.ts#L106

but it's not propagated in the error.

This is most likely a bug I introduced. The idea would be to extend such field in the error object and use that to recognise the error type.

Does that make sense?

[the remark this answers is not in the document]
lag-of-death: @XVincentX, but you referenced `ProblemJson`, not `ProblemJsonError`. Maybe I'm missing something but I don't see the bug?

packages/http/src/types.ts
lag-of-death: hm.. @XVincentX, did you mean `fromTemplate` or `fromPlainError`?

XVincentX:
@philsturgeon I took over and brought this over the finish line. I've extended the `ProblemJson` handler to accommodate additional fields — the spec allows them.

Also — for the future, the spec mandates that `detail` is always a string; so I had to modify a little bit the code.

packages/http/src/types.ts
```diff
@@ -74,11 +75,22 @@ export type ProblemJson = {
   type: string;
   title: string;
   status: number;
-  detail: string;
+  detail: any;
```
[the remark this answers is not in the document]
philsturgeon: I think this was just a mistake, no reason to change it as far as I can tell.

test-harness/index.ts
```diff
@@ -237,7 +231,9 @@ const createSpec = (specPath) => {
 
             expect(reqRes).toMatchObject(masterFile);
 
-            expect(payload.detail).toContain("should have required property 'name'");
+            expect(payload.detail).toContain(
```
philsturgeon: This line is less interesting than the previous assertion: checking it is moaning about required proper name. Can we assert that this new validation property has the validation we are interested in instead?

packages/http/src/types.ts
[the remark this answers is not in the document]
XVincentX: I fixed it. `detail` must be a string, this is per spec.

test-harness/index.ts
XVincentX: I'll fix it, however all this is going to be removed soon/needs to be converted to the new harness thing — is it worth to spend time on this?

philsturgeon: Ahh yeah sorry i didnt notice it was in there. Ignore.

XVincentX:
I'll give @chris-miaskowski some time to jump into this as well

PR #371, opened 2019-06-20
feat: do not compile test files

XVincentX:
We do not need test files in the final build output, so I removed them

### DIFF
--- a/packages/cli/tsconfig.build.json
+++ b/packages/cli/tsconfig.build.json
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "include": ["src"],
+  "exclude": ["src/**/__tests__/**/*.ts"],
   "compilerOptions": {
     "rootDir": "src",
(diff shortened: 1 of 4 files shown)


packages/cli/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "include": ["src"],
+  "exclude": ["src/**/__tests__/**/*.ts"],
```
philsturgeon: Probably dont need anything in tests right? Not just skipping .ts but... everything in __test__ ?

philsturgeon: OOOOh ignore that its only going to try and compile .ts. right.

XVincentX: I tried. It just does not work unfortunately, and the last thing I want to do is check out the source code of TypeScript _again_.

PR #372, opened 2019-06-20
feat: harness file

XVincentX:
This PR introduces the Harness file format. 

I'm copy/pasting the README I put in the repo as well, it should give a decent overview of what's going on here.

---

# Test Harness

## Prerequisites

* Install the project dependencies with `yarn`
* Generate the binary for your platform with `yarn build.binary`. This will *also* compile the project from TS -> JS

## Running the suite

Run `yarn test.harness` from your terminal

## Adding a new test

* Create a new file in the `./spec` directory. It can have _any_ name and _any_ extension, it does not really matter.
* Use the following template and put your stuff:

```
====test====
Test text, can be multi line.
====spec====
openapi 2/3 document
====server====
command line arguments to run Prism with
====command====
curl command to get the response
====expect====
expected output from curl
```

### A real example?

```
====test====
Hello
====spec====
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: Get Todo Items
          content:
            'application/json':
              example: hello
====server====
mock -p 4010
====command====
curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
====expect====
HTTP/1.1 200 OK
content-length: 7
Date: Fri, 21 Jun 2019 09:25:34 GMT
Connection: keep-alive

"hello"
```

#### Things to keep in mind when creating the files:

* 1 test per file, we do not support multiple splitting.
* Be precise with the separators. They shuold be 4 *before* **AND** *after* the word. `====`
* The 4 keywords are `test,spec,server,command,expect`, nothing else at the moment
* You can run all the tests on the same port `4010`, but you can also choose another one
* The `curl` command does not support piping stuff into other tools; so if you're trying to be cool and do `curl | grep`, well maybe next time.

## Technical details

* A RegExp is used to split the content
* A temporany file with the specification file is stored on your disk
* Prism gets spawn with the specified arguments and waited to be running
* The curl command is performed
* The outputs get converted using `http-string-parser`, a veeery old package transforming CURL output in a consumable format
* Gavel is used to validate the request — it will automagically ignore headers that can change and consider only the "fundamental" one such as content negotiation ones and stuff around.

---
As an example, I wrote two harness tests to make sure it works correctly.

If you see anything blocking/not cool enough/missed requirement, feel free to comment!

P.S: The code is a mess, by definition. The [test part in particular](https://github.com/stoplightio/prism/pull/372/files#diff-d26ad5d0c95e0973686e7fa034e1283bR38)

Handling external processes from a program is flaky and really difficult to handle accordingly. Moreover, in order to ensure that two processes do not try to play on the same port I've been forced to put glueing code in the `it` statements, which is not really cool, but I guess as long it works and since we're not shipping this code with Prism, we can live with it.


test-harness/helpers.ts
```diff
@@ -1,41 +1,19 @@
-import * as  fs from 'fs';
-import { join } from 'path';
-import fetch from 'node-fetch';
+export function parseSpecFile(spec: string): any {
+  const regex = /==(server|test|spec|command|expect)==\n(([^=])*)/gi
```
philsturgeon: Better to do four `====` because == could appear in the content of these things, and [^=] is even more likely to appear in the contents.

Did you try splitting instead of matching?

XVincentX: Changed with 4; trying with splitting…

test-harness/helpers.ts
```diff
@@ -1,41 +1,19 @@
-import * as  fs from 'fs';
-import { join } from 'path';
-import fetch from 'node-fetch';
+export function parseSpecFile(spec: string): any {
+  const regex = /====(server|test|spec|command|expect)====\n(([^=])*)/gi
```
P0lip:
```suggestion
  const regex = /====(server|test|spec|command|expect)====\r?\n(([^=])*)/gi
```
Let's add ca arriage return here for all the Windows folks :)

Moreover, I'd move it outside the function and do the following here
```suggestion
  regex.lastIndex = 0;
```
It's a test though, so you can leave it as is.

test-harness/index.ts
```diff
@@ -1,272 +1,68 @@
...
+
+      prismMockProcessHandle.stdout.pipe(split2()).on('data', (line: string) => {
+        if (line.includes('Prism is listening')) {
+          const clientCommandHandle = cp.spawnSync(command, args, { encoding: 'utf8' });
+          const output: any = parseResponse(clientCommandHandle.stdout.trim());
+          const expected: any = parseResponse(parsed.expect.trim());
+
+          try {
+            expect(validate(expected, output).isValid).toBeTruthy();
+            expect(expected.body).toEqual(output.body)
```
XVincentX: @philsturgeon ^ Static Example validation

test-harness/index.ts
```diff
@@ -1,272 +1,68 @@
...
+      );
+
+      prismMockProcessHandle.stdout.pipe(split2()).on('data', (line: string) => {
+        if (line.includes('Prism is listening')) {
+          const clientCommandHandle = cp.spawnSync(command, args, { encoding: 'utf8' });
+          const output: any = parseResponse(clientCommandHandle.stdout.trim());
+          const expected: any = parseResponse(parsed.expect.trim());
+
+          try {
+            expect(validate(expected, output).isValid).toBeTruthy();
```
XVincentX: @philsturgeon Presence-only-property validation

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
+              example: hello
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
+====expect====
+HTTP/1.1 200 OK
+content-type: application/json; charset=utf-8
+content-length: 7
+Date: Fri, 21 Jun 2019 09:25:34 GMT
```
pytlesk4: - Should be able to blacklist certain headers, why test Date?

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
+"hello"
+```
 
-When doing `yarn test.binary`, the envs are optional, they have defaults.
+#### Things to keep in mind when creating the files:
 
-1. `SPEC` defaults to using `petstore.oas2.json`
-2. `PRISM_PORT` defaults to using `4010`
+* 1 test per file, we do not support multiple splitting.
+* Be precise with the separators. They shuold be 4 *before* **AND** *after* the word. `====`
```
pytlesk4: 😭

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
 
-## To record gold master files:
+## Technical details
 
-1. Run the Prism binary with the target file you want to record
-2. in another terminal window: `node test-harness/createMasterFiles.js` - this will use requests definitions from `requests.js` and save master files under `/gold-master-files`
-
-Gold master files contain data about both request and response.
+* A RegExp is used to split the content
+* A temporany file with the specification file is stored on your disk
```
pytlesk4: - If we just stored the spec on disk as part of the test itself, then none of this needs to happen.

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
 
-1. `yarn test.binary`
-2. `SPEC=./examples/petstore.oas2.json yarn test.binary`
-3. `SPEC=./examples/petstore.oas2.json PRISM_PORT=4011 yarn test.binary`
+### A real example?
 
-> `SPEC` can take comma delimited paths to specs
+```
+====test====
+Hello
```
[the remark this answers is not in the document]
XVincentX: Correct!

marbemac:
Good:

- I think curl is 👍
- I think readability of the test cases is 👍 

Questionable:

- Like tom mentioned, the format as presented is a bit limiting. Some things we will have trouble testing: $refs, multiple requests against the same server

#### Suggestion 1:

Any reason not to use yaml and ditch the proprietary `====expected====` thing? Test might look like this:

```yaml
server: mock -p 4010
tests:
  - name: Hello
    command: curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
    expected: |
        HTTP/1.1 200 OK
        content-length: 7
        Date: Fri, 21 Jun 2019 09:25:34 GMT
        Connection: keep-alive
        
        "hello"
spec:
  openapi: 3.0.2
  paths:
    /todos:
      get:
        responses:
          200:
            description: Get Todo Items
            content:
              'application/json':
                example: hello
```

Note I made tests an array so that one can exercise multiple requests against the same server + spec setup. Will get cumbersome to have to copy entire spec for each little test case.

#### Suggestion 2:

Support `spec` being a string that points to a file, so that we can a) re-use openapi definitions and b) test more complicated $ref situations as needed.

```yaml
server: mock -p 4010
tests:
  - name: Hello
    command: curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
    expected: |
        HTTP/1.1 200 OK
        content-length: 7
        Date: Fri, 21 Jun 2019 09:25:34 GMT
        Connection: keep-alive
        
        "hello"
spec: ./specs/hello.yaml
```

test-harness/README.md
marbemac: Because we're testing e2e - the response should have a date header.

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
+          description: Get Todo Items
+          content:
+            'application/json':
+              example: hello
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
+====expect====
+HTTP/1.1 200 OK
```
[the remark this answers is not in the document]
marbemac: Because we're testing e2e - the response should have this stuff.

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
+      responses:
+        200:
+          description: Get Todo Items
+          content:
+            'application/json':
+              example: hello
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
```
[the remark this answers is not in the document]
marbemac: More people know curl (or can find the docs to learn curl) than would know whatever proprietary object we put together to replace this command. I think this curl approach is really nice actually.

> How do I send multipe requests?

Valid concern - I suggested an option in my main thread comment.

test-harness/README.md
```diff
@@ -1,39 +1,75 @@
...
-2. `SPEC=./examples/petstore.oas2.json yarn test.binary`
-3. `SPEC=./examples/petstore.oas2.json PRISM_PORT=4011 yarn test.binary`
+### A real example?
 
-> `SPEC` can take comma delimited paths to specs
+```
+====test====
+Hello
+====spec====
+openapi: 3.0.2
```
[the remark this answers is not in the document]
marbemac: Valid concern - I suggested an option in my main thread comment.

test-harness/README.md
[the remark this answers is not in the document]
pytlesk4: > More people know curl (or can find the docs to learn curl) than would know whatever proprietary object we put together to replace this command. I think this curl approach is really nice actually.
> 
> > How do I send multiple requests?
> 
> Valid concern - I suggested an option in my main thread comment.

Why not just kill two birds with one stone and test Prism's HTTP client? Fine, users can give us a curl command, but the harness should take a structured request object that can be easily converted to curl command or an axios request, or a Prism request, ect...

test-harness/README.md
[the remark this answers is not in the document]
philsturgeon: If people aren't writing tests their PRs wont be accepted. 🤷‍♂️

test-harness/README.md
pytlesk4: Fine, ignoring the Date header.

Do I have to calculate `content-length` then? If I am creating a test in the harness, the will need to be `content-length` calculated correctly when generating a Raw HTTP Response(that is all the output is from curl). Even if I start out with output from curl, I will need to edit it to fix the errors, which in a lot of cases will require a `content-length` change.

philsturgeon: Yep certain headers will need to be removed for comparison by default, but its fine to have them in the response here. We want minimal mucking about for users/support/business to create these scenarios. It should be copy and pasteable by anyone without technical understanding.

test-harness/README.md
philsturgeon: multipart

test-harness/README.md
philsturgeon: Storing multiple files per scenario leads to confusion, reuse, which overloads the meaning, means people need to ZIP files to share scenarios, etc. 

I'm not sure why we need to mention the tmp files, it seems like an implementation detail @XVincentX? Most software create tmp files somewhere or other.

test-harness/README.md
philsturgeon: I would definitely not expect anyone to be writing these tests by hands no! That would be awful. These are gold master files. https://codeclimate.com/blog/gold-master-testing/

You run it, you store the response, then if it changes you get a diff.

This is not a test suite in the classic sense, we are not writing assertions.

test-harness/README.md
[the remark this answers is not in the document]
pytlesk4: I get the merits of curl, and that it is better at sending requests than Prism. I was just suggesting we structure the request being sent, and internally send a curl request. That gives us the option to add support for sending requests with other clients.

Also, I don't see why cheating is a bad thing in this particular instance. If anything it is pretty a pretty smart solution to be fairly confident prism's code base is very well tested.

> End to end tests are trying to test things that the users are actually doing, so an end to end test for the HTTP server is going to be testing done with common HTTP clients, like curl. Some commands could involve ping, some things might be grepped.

I will admit, didn't think about running grep/ping, that is pretty cool.

test-harness/README.md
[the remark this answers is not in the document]
pytlesk4: [har](http://www.softwareishard.com/blog/har-12-spec), would be better than raw HTTP.

test-harness/README.md
pytlesk4: Just to be clear, I am not advocating that this be a suite in a classic sense.

test-harness/README.md
XVincentX: I thought it was a good idea to underline we're writing stuff on the disk of the user.

philsturgeon: Well it's already scared one person looking at the README. :) I am just wondering if its common to warn people about tmp files because I think all software writes temp files?

philsturgeon: Also do we need to mentioned that a regexp is splitting the content? I am wondering what the benefits here are. Does a test harness contributor need to know this for a reason? Are they meant to be watching out for something? OR being careful to avoid doing something? Or... what? If this is just to explain how some of the current implementation works, maybe it should be a code comment?

XVincentX: I'm ok removing all the technical detail section, I do not care that much.

test-harness/README.md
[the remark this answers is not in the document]
XVincentX: 1. The file goes through Prism's CMDLine that will resolve the references
2. You can't test remote files, you'd need to paste the content here.

test-harness/README.md
XVincentX: Gavel will ignore most of the headers and just evaluate the one that actually make sense (such as content type and some other right now I do not remember)

XVincentX:
@marbemac I do not undertand the ref problem. Prism is using _that thing_, which should be resolving the references, right?

philsturgeon:
Hey! So I know the format seems a little funny at first, but it's how PHP has handled its test files for a long time and its worked pretty well :D 

The exact format is contrived and literally was just the first thing i typed out in a 2 minute sit down, so im not expecting it to be perfect, but it does have advantages over YAML.

1. you dont have to worry about indentation. Anyone can make these files without having tabs or spaces cause meaningful issues, or trying to teach somebody how to tab in multiple lines of content to make it compatible with YAML when they might not be using a proper code editor. 

2. YAML gets really annoying about certain characters if you try to use it without knowing about multi line strings, and when people do know that multi line strings are possible they often pick the [wrong multiline syntax](https://yaml-multiline.info/).

There is nothing saying that only one command can be run in each scenario, if your command section runs two commands then your expected output is bigger. These things are free form so do what you want. That's kinda the point.

There's also nothing saying that these 4 keywords are the only keywords we will ever have, we already recognized that `==== expect ====` will be good for literal matching and another new keyword like `==== expect-fuzzy ====` (or something) will be good for passing off to Gavel, so there's a good chance others will appear. Again we arent trying to build a full assertion system we are just making the simplist framework for checking stuff happens when users run commands.

$ref'ing a file will just involve literally $ref'ing a file, that's fine. 

$ref'ing a URL is a really good point. If this was ruby I would say the scenario should use `thin foo.yaml && prism mock ...` or a magic new keyword. New keyword means we dunno how to get the URL so that one is still a chin scratcher, but the rest of the concerns have simple answers.

marbemac:
> you dont have to worry about indentation. Anyone can make these files without having tabs or spaces cause meaningful issues

🤷‍♂ instead, you have to worry that about how many `====` you are using, if the space `==== foo ====` vs `====foo====` matters (does it?), if space before/after the dividers matters e.g. `==== foo ====\n content` vs `==== foo ==== content` (does it?). 

> teach somebody how to tab in multiple lines of content to make it compatible with YAML when they might not be using a proper code editor

🤷‍♂ instead, nobody will have a proper code editor that can help them with formatting, since it's proprietary.

> YAML gets really annoying about certain characters if you try to use it without knowing about multi line strings

If we really needed to, we could validate test files with JSON schema. More likely we just have an example test object in the readme that somebody can copy paste into a new file and change the values in.

---

I don't really understand what real problems we're solving with this proprietary format. I DO know that it's one more thing for us to build, maintain, and explain, and that it won't play nicely with any other tools e.g. no code editor formatting, linting, or highlighting support like you'd have with JSON or YAML, no option to use JSON schema to document and validate the test file and help the user, etc.

@philsturgeon are there problems other than the couple described above that this `====` solves? I'm still not convinced the pros outweigh the cons.

philsturgeon:
> 🤷‍♂ instead, you have to worry that about how many ==== you are using, if the space ==== foo ==== vs ====foo==== matters (does it?), if space before/after the dividers matters e.g. ==== foo ====\n content vs ==== foo ==== content (does it?).

YAML has meaningful whitespace. Frontmatter has people counting their hyphens. I think this is pretty common.

> 🤷‍♂ instead, nobody will have a proper code editor that can help them with formatting, since it's proprietary.

It's plain text, nobody needs anything special at all for this to work.

> If we really needed to, we could validate test files with JSON schema. More likely we just have an example test object in the readme that somebody can copy paste into a new file and change the values in.

If its complex enough that we need JSON Schema I think we did something extremely wrong? This is meant to be a single file which explodes on some separators to avoid needing to have a folder with multiple specially named files. Should we switch to a folder with specially named files? 

> I don't really understand what real problems we're solving with this proprietary format.

I think the problem is thinking of it as a proprietary format? It's not, its just a plain text file that we're splitting. 

> no code editor formatting, linting, or highlighting support like you'd have with JSON or YAML

I have no expectation of people writing these files by hand. They are copy and paste vehicles.

> no option to use JSON schema to document and validate the test file and help the user, etc.

Again I don't know why we would need JSON Schema for this in any way?

> are there problems other than the couple described above that this ==== solves? I'm still not convinced the pros outweigh the cons.

The options are: 

1. we have one folder per test and we have a `setup.sh` and a `output.txt` and loads of specially named files, which sounds like a pain in the butt to manage, users have to send a ZIP file, etc.
2. we do a YAML file and have people fighting with YAML and you still dont get syntax highlighting, not that I think we need that
3.  we have people writing Jest describe-it syntax like any other test suite but this is meant to be something non tech people can do, and its meant to be as close to real as possible
4. we can do this, where people just copy and paste some things into a file and if it doesnt work we can tweak their PRs in seconds because its just a file with some content 🤷‍♂ 

The test harness started as a table of curl commands and links to specs:

Screen Shot 2019-06-26 at 21 37 49

This is a declarative, portable, human-and-computer replacement for that table. 

It could be a CSV file or an XML file, HAR would not work because its more than just the HTTP messages, but it could be pretty much anything. 

This, in my opinion, would be the least mucking about, for something that the average user would only ever need to do between 0 and 1 times.

pytlesk4:
Is it just me or does anyone one else think the test harness look similarish to Stoplight Scenarios?

marbemac:
> YAML has meaningful whitespace. Frontmatter has people counting their hyphens. I think this is pretty common.

Yes I understand, I was not trying to say that YAML does not require specific formatting. I was responding to "you dont have to worry about indentation. Anyone can make these files without having tabs or spaces cause meaningful issues" with a comment trying to make the point that we're just exchanging one syntax expectation with another.

> I think the problem is thinking of it as a proprietary format? It's not, its just a plain text file that we're splitting.

🤔Wether it's using `====xyz===` as delimiters, or using indentation as delimiters, the text inside of the file needs to be organized in a specific way - one can't toss any old text file in and have it work.

---

Sounds like you'd really like to go with the text splitting, so in the interest of moving forward I guess we can agree to disagree. Let's give the `====` split thing a go 👍.

XVincentX:
@philsturgeon So what do you want to do here? Merge and give it a try?

philsturgeon:
> Wether it's using ====xyz=== as delimiters, or using indentation as delimiters, the text inside of the file needs to be organized in a specific way - one can't toss any old text file in and have it work.

Absolutely, but the point I was failing to make is that teaching somebody people without decent code editors to indent 100 lines of copy and pasted stuff is tricky. Try that in notepad or some other crap editor and you'll have a bad time. 😅 

The split approach does not require people to do any of that "here is a template file chuck the commands in here and the output in here", and its also less of a DSL than this YAML format.

I'm not tied to the single-file split format, I just wanted to talk it out to see what the pros and cons are, and I think we made some progress in general. We seem to have collectively got away from being very concerned about the larger concepts, like confusion over how multiple commands would work, how refs work, etc. 

The only bit we are left on is the format to use. HAR is out, because the expect is just whatever CLI output happens, not always a full HTTP response. So we need something more vague, which could be this YAML DSL, or we stick to the single split file format.

Mini vote on the format? :D 

🚀 = YAML
👀 = Splits

Comments for Other.

chris-miaskowski:
I had a sync up with @philsturgeon and @XVincentX. They gave me a quick overview and I say:
both formats are technically equal and good, but gut tells me to vote on the "flat" format.
The main reason: simplicity.

P0lip:
I didn't participate actively in discussions, but since I was added to reviewers and you are doing some voting, here is my take on the YAML vs `====` format topic.
I'm rather against using a proprietary format that is not backed by any formal specification and tooling - standard AST format, parser, etc.

In general, I believe most of the folks do have some basic code editor installed, but even if they don't, there are plenty of online alternatives out there, including Monaco, so IMHO editing is not a real issue here.

> Absolutely, but the point I was failing to make is that teaching people without decent code editors to indent 100 lines of copy and pasted stuff is tricky. Try that in notepad or some other crap editor and you'll have a bad time. sweat_smile

We could provide a form view in Stoplight Studio :sweat_smile:... but I can't imagine supporting some unclear format. :P
Supporting YAML/JSON or even Markdown would be by far easier since there is tooling available.

Either way, the splits thingy doesn't look bad. I'm just against having ambiguous, unspecified formats (JSON path anyone? :laughing:).

philsturgeon:
@P0lip good points, but again, I think some of us might be overthinking what the split file is:

https://github.com/stoplightio/prism/blob/f1606591cbbfa8a154f71e28645f982d25584932/test-harness/helpers.ts

The entire implementation is 10 lines of code. Arguably just 2. 

We'd need a similar chunk of code for the YAML DSL, because passing around a naked YAML structure in our code without types wouldn't be ideal, so... "complexity of code" should probably removed from the equation. They're both equally simple from a "code we write/maintain" perspective.

XVincentX:
This is blocking the introduction of the Azure pipelines — it is required because at the current state you can't even clone Prism's source code on Windows.

Somebody has to make the call. @marbemac @philsturgeon

philsturgeon:
Yep the vote is a tie so let’s just go with the one we already have.

PR #377, opened 2019-06-21
Build process revisitation

XVincentX:
On tag:

1. Runs *all* the tests on Windows
2. Creates a Windows Binary
3. Uploads it to the current release

Outstanding issues:
- [x] Make sure CircleCI does not produce Windows Binaries anymore
- [ ] Make sure this runs *after* CircleCI

---

Note: there's no way I can properly test this without giving it a run. Bad things could happen, but I'll be there fixing them in case.


XVincentX:
//cc @liguori JFP

liguori:
Good job, man

XVincentX:
…e se lo dice Coraçao, allora è davvero un buon lavoro.

azure-pipelines.yml
```diff
@@ -0,0 +1,38 @@
...
+  - script: yarn build
+    displayName: 'Build the code'
+
+  - script: npx pkg --output ./cli-binaries/prism-cli-linux ./packages/cli/
+    displayName: 'Create binary'
+
+  - script: yarn test.harness
+    displayName: 'Run harness'
+
+  - publish: ./cli-binaries/prism-cli-linux
```
philsturgeon: Do you  mean to be testing the Linux ones on azure?

XVincentX: Well it turns out it's free anyway — I can remove it though, it does not add that much value effectively.

azure-pipelines.yml
```diff
@@ -0,0 +1,52 @@
...
+  - job: test
+    displayName: Test
+    pool:
+      vmImage: vs2017-win2016
+    steps:
+      - task: NodeTool@0
+        inputs:
+          versionSpec: '12.x'
+        displayName: 'Install Node.js'
+      - powershell: | # Not that cool, but it does the job.
```
philsturgeon:
```suggestion
      - powershell: |
```

philsturgeon: Don't pick a fight with windows users. There are a surprising number of them.

test-harness/index.ts
```diff
@@ -36,12 +36,11 @@ describe('harness', () => {
     afterAll(() => tmpFileHandle.removeCallback(undefined, undefined, undefined, undefined));
 
     test(parsed.test, done => {
-      expect.hasAssertions()
       const [command, ...args] = parsed.command.split(' ').map(t => t.trim());
       const serverArgs = [...parsed.server.split(' ').map(t => t.trim()), tmpFileHandle.name];
 
       prismMockProcessHandle = cp.spawn(
-        path.join(__dirname, '../cli-binaries/prism-cli-linux'),
+        path.join(__dirname, '../cli-binaries/prism-cli'),
```
philsturgeon: Am I reading this all right, that we call the windows one `cli-binaries/prism-cli` so this works, but then we build linux and osx at the same time with `npx pkg --out-path ./cli-binaries --targets linux,macos ./packages/cli/`? Is that not gonna be trouble for CircleCI?

azure-pipelines.yml
XVincentX: It's not a fight with the windows users, it's a fight with myself unable to find a better solution. I'll remove it though.

test-harness/index.ts
XVincentX: Very good spot here. Let me see how I can fix this.

XVincentX: @philsturgeon 
Actually this is ok as it is. The multi binary creation `linux,macos` is executed in the "publish" step, when the harness has already ran with linux.

Here you can see the harness uses a binary called `prism-cli`
https://github.com/stoplightio/prism/blob/master/.circleci/config.yml#L39-L57

While the public process is then later creating binary named per platform

https://github.com/stoplightio/prism/blob/master/.circleci/config.yml#L59-L82

These are two separate steps and they do not collide. Does this clarify?

PR #381, opened 2019-06-25
HTTP Server should parse formbody data

### LINKED ISSUE #380: Prism does not allow requests that are not application/json
The OpenAPI spec allows us to define request bodies in different formats, such as application/x-www-form-urlencoded.  However, this line causes prism to return a 415 for any request that is sending data in anything other than application/json:

https://github.com/stoplightio/prism/blob/62c3f45c2ff9a30fcb0a12d70e7017fcf590d033/packages/http-server/src/server.ts#L39

Honestly, I don't care if prism validates my request, but in this case, it should at least return my example response instead of always 415'ing

For reference, here is a spec of my endpoint:
```
 /api/v1/validate-pin:
   post:
     description: Validate a challenge code from user's phone
    requestBody:
       required: true
      content:
             schema:
               type: object
              properties:
                challengeCode:
                   type: string
                   example: 1234
     responses:
       '200':
        headers:
          X-Authorization:
             $ref: '#/components/schemas/authorization'
          X-Entitlements:
            $ref: '#/components/schemas/entitlements'
        content:
           application/json:
             description: A successful request
             schema:
               type: object
               properties:
                userSession:
                   $ref: '#/components/schemas/userSession'
```

XVincentX:
Adds the parser for form body data so that Fastify will not block the request.

This can be only tested from the harness, because the Fastify testing utilities bypass the whole parsing/content type; harness though is being rebuilt in #372 ; so I think we need to coordinate here.

Closes #380

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -56,5 +56,6 @@ module.exports = {
   ],
   collectCoverageFrom: [
     '**/src/**/*.{ts,tsx}',
+    '!**/src/**/__tests__/**/*.{ts,tsx}',
   ],
 };
--- a/packages/http-server/package.json
+++ b/packages/http-server/package.json
@@ -23,6 +23,7 @@
     "@stoplight/prism-http": "^3.0.0-beta.2",
     "fastify": "^2.5.0",
     "fastify-accepts-serializer": "^2.0.1",
+    "fastify-formbody": "^3.1.0",
     "tslib": "^1.10.0",
     "type-is": "^1.6.18"
   }
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -3,6 +3,7 @@ import { createInstance, IHttpMethod, ProblemJsonError, TPrismHttpInstance } fro
 import * as fastify from 'fastify';
 // @ts-ignore
 import * as fastifyAcceptsSerializer from 'fastify-accepts-serializer';
+import * as formbodyParser from 'fastify-formbody';
 import { IncomingMessage, ServerResponse } from 'http';
 import * as typeIs from 'type-is';
 import { getHttpConfigFromRequest } from './getHttpConfigFromRequest';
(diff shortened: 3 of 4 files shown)


philsturgeon:
Seems like this one is handled very differently to incoming JSON. Is that because the fasting formdata plugin is much better, or because we are going to treat formdata differently to JSON?

XVincentX:
Good question.

The pork-around for JSON is required because Fastify, natively, will only recognise `application/json` as something that can serialise as JSON; for this reason we need to use a custom function to detect if the `accept` header is _jsonish_ (as `application/vnd+api.json`) and serialise it using the builtin node function.

The content parser here will effectively recognise and parse `application/x-www-form-urlencoded`

Does that clarify?

LesterTheTester:
I believe you'll still need to modify https://github.com/stoplightio/prism/blob/62c3f45c2ff9a30fcb0a12d70e7017fcf590d033/packages/http-server/src/server.ts#L39, otherwise we will 415, correct?

XVincentX:
That should not be required, that content type parser is called as last resort. If you register one more specific before this one, it'll pick that one

philsturgeon:
We are currently rewriting test harness but it's tough to say we should merge this without having any tests... are you sure fastify cant have test with form data written? It darn well should let you somehow and if they don't support it we should go and talk to them about that. :)

XVincentX:
Fastify's integrated test utility (inject) will go directly to the route handlers, skipping the parsing stage. If that's not enough, we can always spin up an instance of the server ourselves and test the whole thing accordingly, which is what the harness is supposed to do.

We could also make it happen spinning up an instance programmatically.

It's not a Fastify limitation — they're offering an unique byproduct that can't be used to test the whole thing.

philsturgeon:
Let’s manually verify this one and add it on our list of things to write test harness scenarios for. Oooor you could manually test it now and add a scenario in your experimental test harness rewrite branch. :)

--
Phil Sturgeon
@philsturgeon

XVincentX:
Agreed. Who's got the balls to approve it?

Issue #379, opened 2019-06-22
Feature: Improve log message structure


marbemac:
The new logging is already a huge improvement over Prism v2, bravo 👏!

There might be a couple small wins we could consider to improve the readability and usefulness of the logs even more.

Here is the current behavior:

<img width="1337" alt="Screen Shot 2019-06-22 at 1 54 50 PM" src="https://user-images.githubusercontent.com/847542/59967931-3182cf00-94f7-11e9-8acd-7b4bed678156.png">

What if we:

1. Use indentation to visually group related logs together. In the above screenshot the 4 highlighted lines are basically children of the line above them.
2. Use the first ID in the message (in screen above, `[MOCKER]` and `[HTTP SERVER]` is what i'm talking about) to more precisely indicate the stage of the process that is generating the log. Disclaimer: I have not gone deep on the logging implementation so I don't know how easy or not easy this would be.

The example pictured above might look more like this:

```bash
[HTTP SERVER]    ℹ  info     'GET /todos' request received.
    [ROUTER]     ✔  success  Found HTTP operation.
    [VALIDATOR]  ✔  success  The request passed the validation rules.
    [NEGOTIATOR] ✔  success  Found compatible response for 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3'.
    [MOCKER]     ✔  success  Responding with status code 200.
```

IMHO these changes would make the logs easier to understand when there is more than one request in the log stream.

XVincentX:
A possible implementation for this could be:

1. Replace child logger labels with numbers
2. Replace the numbers with labels in the CLI
3. Indent `N` times based on the value passed in the child logger
5. Profit

XVincentX:
@marbemac What do you think of this?

<img width="1550" alt="image" src="https://user-images.githubusercontent.com/1416224/60170651-2c7c9300-9809-11e9-91db-0a05b493910b.png">

pytlesk4:
> @marbemac What do you think of this?
> 
> <img alt="image" width="1550" src="https://user-images.githubusercontent.com/1416224/60170651-2c7c9300-9809-11e9-91db-0a05b493910b.png">
> 
> <img alt="image" width="991" src="https://user-images.githubusercontent.com/1416224/60184289-13cfa580-9828-11e9-87fa-a7802d5988cb.png">
> 
> Help me out with the colours people, I agree they're horrible.

😍

[...]

PR #382, opened 2019-06-26
Indent log messages according to the request status

XVincentX:
Tracking PR for #379


packages/http-server/src/server.ts
```diff
@@ -101,6 +101,7 @@ const replyHandler = <LoaderInput>(
       body,
     };
 
+    request.log.info({ input }, 'Request came in');
```
philsturgeon: Could this be something like 

```suggestion
    request.log.info({ input }, 'Received request GET /foo/bar');
```

Looking at the log output it seems like really funny wording, and it could also continue to be useful if we were to hide negotiator logs behind a verbose switch or something.

XVincentX: The `input` contains the path and the verb already, so you'll see them in the final message.

philsturgeon: Ok, so assuming prism isnt completely broken and failing to talk to the negotiator, maybe we dont need this message?

XVincentX: The reason is to make the indentation look cool, otherwise the user would see negotiator messages directly, which have a "level 1" of indentation, instead of 0. Does that make sense?

XVincentX:
Let's wait for the big boss

marbemac:
60253716-e983f380-98cc-11e9-86e8-a65afa4511aa

Looks great! ONE last comment, I promise :).

In the last screenshot above, it looks like the request resulted in an error, is that correct? I suggest that everything related to a request be nested underneath it, including the final error there. As it stands, it's not clear that that error is the final result of the request above it. Does that make sense?

XVincentX:
@marbemac That isn't really that easy because the error is detected from the HTTP Server which has the informations to display all the data — but it's on a different level. Indenting only *that* part might be difficult. Let me try that though, I'll get back to you.

philsturgeon:
@XVincentX I think we should probably kick this over the line and improve it later if anyone points it out? @marbemac hopefully you're ok with that? I agree this would be nicer but the PR is a big improvement in its own right. We gotta get moving on the e2e test suite.

XVincentX:
Yeah, I'm ok merging this — the reason why I did not it's because it's not blocking anybody, so it can stay here. I'd love to see this merged though!

XVincentX:
image


@philsturgeon @marbemac 

I've implemented the change you requested with an additional property. I have a better idea to implement this in the future but I guess for now let's keep it in this way.

marbemac:
Sorry to hold it up, just seeing this - looks fantastic! Great job @XVincentX 👏 .

PR #383, opened 2019-06-26
Update JSON Schema Faker to latest version

XVincentX:
With this PR we can finally get rid of JSON Schema Faker with the old `jsonpath` dependency.

I've also upgraded some minor packages, and binary now just work without having to keep our fork.


philsturgeon:
AMAZING! Great work, this is top quality open-sourcering.

PR #384, opened 2019-06-27
feat(faker): extend json-schema-faker to enable faker support

vanhoofmaarten:
## Checklist

_feature_

This PR originated from the discussion [here](https://community.stoplight.io/t/prism-dynamic-mocks-with-faker-js/717).

Proposal to extend the JSON-schema-faker instance Prism uses with [faker](https://github.com/marak/Faker.js/) using the `x-faker` keyword which is allowed as an [extension](https://swagger.io/docs/specification/openapi-extensions/) by the OpenAPI spec.

Current behavior ignores the `x-faker`keyword when running Prism with dynamic mocking. 
The new behavior outputs a by faker generated value when running Prism with dynamic mocking.

Not that I am currently aware of.


XVincentX:
I can help @vanhoofmaarten writing a unit test — but I want to make sure we're ok with the feature first. This is going to need some documentation as well I guess.

examples/petstore.oas3.json
```diff
@@ -1260,7 +1260,8 @@
           },
           "name": {
             "type": "string",
-            "example": "doggie"
+            "example": "doggie",
+            "x-faker": "name.firstName"
```
XVincentX: Let's leave the example file as it is

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -6,6 +6,8 @@ import * as jsf from 'json-schema-faker';
 // @ts-ignore
 import * as sampler from 'openapi-sampler';
 
+jsf.extend('faker', () => require('faker'));
```
XVincentX: Shouldn't this be `x-faker`?

XVincentX: Also, can we import faker using `import` and then use the value here directly?

packages/cli/package.json
```diff
@@ -45,7 +45,8 @@
     "prepack": "node -r tsconfig-paths/register ../../node_modules/.bin/oclif-dev manifest",
     "version": "node -r tsconfig-paths/register ../../node_modules/.bin/oclif-dev readme && git add README.md",
     "cli": "node -r tsconfig-paths/register bin/run",
-    "cli:debug": "node --inspect-brk -r tsconfig-paths/register bin/run"
+    "cli:debug": "node --inspect-brk -r tsconfig-paths/register bin/run",
+    "cli:example": "npm run cli -- mock -d ../../examples/petstore.oas3.json"
```
XVincentX: Can we remove this command?

packages/http/src/mocker/generator/JSONSchema.ts
vanhoofmaarten: 1. No. JSON-schema-faker removes all `x-` in the keywords before parsing. [Source](https://github.com/json-schema-faker/json-schema-faker/blob/573317001f7e3154e409f2616ed6bcd62aa03123/src/class/Container.js#L129)
2. https://github.com/stoplightio/prism/pull/384/commits/30f667364e779fe95e9ab4241cf26465750f9715

XVincentX: //cc @philsturgeon Do you see anything potentially risky with this?

packages/cli/package.json
vanhoofmaarten: Sure. I suggest after we've integrated some tests.

packages/http/src/mocker/generator/JSONSchema.ts
philsturgeon: @XVincentX yeah seeing as this is all just part of the output process this seems fine to me.

vanhoofmaarten:
Because of the rapid evolution of this PR, I'll try to be transparent in terms of hours I can commit to this PR in short term so I won't create false expectations. 

I'm able to squeeze in 2 hours tomorrow for writing a few suggested tests. 
For the rest of the weekend, I'm offline. In the coming week I'll be able to resume where we left off.

If you'd like to commit to this pull request in the meanwhile, I've added @XVincentX & @philsturgeon as collaborators to my fork.

packages/cli/package.json
vanhoofmaarten: https://github.com/stoplightio/prism/pull/384/commits/7e238f9a9ebf9ad6c28594e6ce1abe28108e45ee

XVincentX:
@vanhoofmaarten I see you've pushed some more stuff — that's great and I also think you're in the right direction. Can you ping me when you think it's done/can't work on this anymore so I can review it and bring it over the finish line?

vanhoofmaarten:
What telepathy! I just finished up. I would very much like your feedback on my changes.

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -9,15 +9,37 @@ describe('JSONSchema generator', () => {
...
+      };
+
+      const instance = generate(schema);
+      expect(instance).toHaveProperty('name');
+      expect(instance).toHaveProperty('email');
+      expect(instance).toHaveProperty('ip');
+      // @ts-ignore
+      expect(instance.ip).not.toBeUndefined();
+      // @ts-ignore
+      expect(instance.ip).not.toEqual('internet.ip');
```
philsturgeon: I have some questions :D 

What errors were you getting that made you add the ts.ignore? I'm certainly no TS expert but I'd like to see if we can avoid it. 

Also, this test is proving something that is never gonna happen anyway right?

I think what would be excellent here, is running this schema but just `ip: { type: 'string' }` then find some sort of "is IP address" assertion which will fail because json schema faker is going to put "cheese" "monkies" "lorum" etc into the value. 

Then you slide the `x-faker` keyword in:

```
schema.properties.ip['x-faker'] = 'internet.ip';
```

Now run it again and assert that it now _is_ an IP address.

This will prove that json schema faker is indeed leveraging faker to do its job and its spotting the keyword. A perfect integration test.

packages/cli/README.md
```diff
@@ -69,6 +69,53 @@ OPTIONS
...
+Fork the http server from the CLI
+
+#### `-p, --port=port`
+**Required**
+
+Port that prism will run on.
+
+Default: `4010`
+
+
```
XVincentX: Unfortunately this is not the right section — this gets generated automatically from oclif when we're publishing so this would just get lost. We need to write it in another section; this should probably go to the main README? @philsturgeon

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -1,3 +1,5 @@
+// @ts-ignore
```
XVincentX: This is not required if you install `@types/faker` in the repository root.

`yarn add -D -W @types/faker`

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
XVincentX: @philsturgeon The reason for that lines is that the responses are not typed; if you want them typed, you need to validate them appropriately, thing that we're not doing now. We *could* use io-ts, but not now. I want to introduce it one day though.

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -9,15 +9,37 @@ describe('JSONSchema generator', () => {
...
+        },
+        required: ['name', 'email', 'ip'],
+      };
+
+      const instance = generate(schema);
+      expect(instance).toHaveProperty('name');
+      expect(instance).toHaveProperty('email');
+      expect(instance).toHaveProperty('ip');
+      // @ts-ignore
+      expect(instance.ip).not.toBeUndefined();
```
XVincentX: I would say the first expect is not required — the previous one (`https://github.com/stoplightio/prism/pull/384/files#diff-8e6fad9f7c42e5f1bdb26a87b73283b0R36`) is implying the second one

packages/cli/README.md
philsturgeon: Time to start a docs/ folder using Stoplight Studio to write Markdown docs, just like Spectral has. 

We need an "Examples and Response Generation" section explaining: 

1. Static Examples
2. Example Selector
3. Dynamic Examples
4. Dynamic Examples (with Faker)

Bigger job though so you know the drill, make an issue and shove it in next sprint :)

packages/http/src/mocker/generator/JSONSchema.ts
vanhoofmaarten: 395fd7af5f3ff9305bf89b9969b31debd2771c29

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
vanhoofmaarten: 950e6347233b712d6d74d36c3c86b4537102cece

vanhoofmaarten:
I made the requested changes. I believe the 2 unresolved requests are something you folks have to work out internally.

packages/cli/README.md
XVincentX: I'm resolving this since we have a card for documentation.

XVincentX:
@vanhoofmaarten I finished this up for you. Thanks a lot for your work.

vanhoofmaarten:
My pleasure! Thank you for helping me finishing up. I learned quite a bit of Typescript and Lerna working on this PR. The best of luck with v3 of Prism!

XVincentX:
@vanhoofmaarten This is a double win — we got a cool feature, you've learned something.

To the next PR!

PR #385, opened 2019-06-27
docs: add decision flow diagram

chris-miaskowski:
The following change adds to the Prism's documentation. 

Specifically it adds the decision flow diagram so that users and contributors can grasp the nuances of Prism's decisions quicker and with less fuss.


packages/http/README.md
```diff
@@ -400,6 +400,12 @@ You would get this in response
 ]
 ```
 
+## Prism Decision Flow Diagram
+
+The below diagram represents all logical decision we make to figure out the best HTTP response to server. Note that this diagram is valid only when mocking is turned on.
+
+![Decision Flow Diagram](docs/images/pdfd.png)
```
chris-miaskowski: Not sure why github is not showing that image in the preview... seems that the file isn't uploaded to the right location. I expect that to be resolved once merged.

XVincentX: What does `pdfd` stand for?

chris-miaskowski: prism decision flow diagram

packages/http/README.md
```diff
@@ -400,6 +400,12 @@ You would get this in response
 ]
 ```
 
+## Prism Decision Flow Diagram
+
+The below diagram represents all logical decision we make to figure out the best HTTP response to server. Note that this diagram is valid only when mocking is turned on.
```
XVincentX:
```suggestion
The below diagram represents all logical decision we make to figure out the best HTTP response to the specific request. Note that this diagram is valid only when mocking is turned on.
```

packages/http/README.md
philsturgeon: haha could we maybe call it mock-server-dfd?

packages/http/README.md
philsturgeon: Mocking is always turned on! 

Remember, we have only the `$ prism mock` CLI command, and we have an experimental JS API we're not really advertising or pushing. Prism is a mock HTTP server and having a mock HTTP server which you need to "turn on mocking" for is going to be very confusing to the reader.

packages/http/README.md
philsturgeon: maybe it needs a ./ at the front?

philsturgeon:
This looks really really good! 

One trivial piece of feedback, it could just be my eyes are tired from the sunlight and whatnot, but im struggling to read the orange / grey combo. Can we go for darker grey closer to black for anything thts on a dark color? If I am struggling with the contrast it means other people (especially some color blind folks) will struggle.

philsturgeon:
Do you think we could bump things up on the left, so that we can tuck the second diagram below, instead of below and over to the right? That would mean wherever we display this, it will be narrower and taller, meaning people can scroll down but not have to scroll left to right so much. I'm having to zoom in then pan.

philsturgeon:
oooo wait I found one incorrect thing. Only one though!

After negotiator block it says "Success?" then the No goes up to "takes lowest 2xx" but 2xx is for success. What are we trying to say here?

packages/http/README.md
chris-miaskowski: That's not **entirely** true. I'm adding to the `http` package, not the `cli` package. To be technically valid I must say that this depends on mocking (it'll be more confusing if someone discovers that it does and that I lied that it doesn't ;p).

packages/http/README.md
chris-miaskowski: > oooo wait I found one incorrect thing. Only one though!
> 
> After negotiator block it says "Success?" then the No goes up to "takes lowest 2xx" but 2xx is for success. What are we trying to say here?

I meant: if the negotiator block resulted in "success" message then X otherwise "go up and... take lowest 2xx". 

And this is actually a correct behaviour. We first try to match to a given code and if we fail we try to match to lowest 2xx. This part of the diagram represents a recurrence in the code.

chris-miaskowski: @philsturgeon ^

chris-miaskowski:
> 1. The diagram has your name on the top right; just for your information. If it's intentional then it's fine.

Intentional. I tried to lessen the confusion :P

> 2. The diagram says 0.1.0 as version, I guess this should be changed to 3.0.0?

Oh yes because it's the diagram's version. I though about editing it to be 3.0.0 but probably that's not really semantically correct (?). E.g. the diagram may change faster than Prism code and vice-versa.

chris-miaskowski:
@philsturgeon @XVincentX done

packages/http/README.md
philsturgeon: The goal was for this to explain the mock server, the diagram title says Prism CLI, but your words are describing Prism HTTP which is meant to just be the client, and we're not meant to be talking about the client. 

So there's a lot of WellActually in here. Can we decide what we're trying to do and be consistent? 

If we are going to pick one thing here I would like it to be the "Prism Mock Server" and only talk about mock server functionality because we're marketing this as a mock server and the only command is mock server. :)

chris-miaskowski: Ok, kicked that confusing line out.

PR #386, opened 2019-06-28
Docs/so-133/dfd

chris-miaskowski:
moved DFD to main readme


marbemac:
This is cool! Couple of suggestions:

- Remove date from image
- Change v1.0.0 to the prism major or minor version? Otherwise kind of confusing. E.g. `v3`. Or remove version altogether
- Add a link below it for somebody to edit it? E.g. if there is an inconsistency now or in the future that somebody notices, how are they going to put in a PR to update it?

marbemac:
Orr not lol. At least add a link somewhere so that folks can edit that diagram? Re-creating it would be a bummer.

chris-miaskowski:
> Orr not lol. At least add a link somewhere so that folks can edit that diagram? Re-creating it would be a bummer.

:D I think Phil didn't notice your comments, I created another PR with the link to original

PR #394, opened 2019-07-04
feat(JSONSchema): Programmatically extend JSON Schema Faker

vanhoofmaarten:
## Checklist

_Feature_

Programatically extend JSON Schema Faker options, formats and keywords.

The provided solution implemented in #384 is not sufficient for certain use-cases.

Eg. Generated HTML, localized Faker support, custom functions, etc.

The goal is to be as **randomly realistic** to the production API as possible.

The new behaviour allows custom JSON Schema Faker options, formats and keywords to be added when using the Prism mocking server programmatically.

The OpenAPI spec allow custom formats for the properties with the [type `string`](https://swagger.io/docs/specification/data-models/data-types/#string) and custom keywords using [extensions](https://swagger.io/docs/specification/openapi-extensions/)

**Example**: mocking-server.js
```js
const { createServer } = require("@stoplight/prism-http-server");
const { createLogger } = require("@stoplight/prism-core");
const minimist = require("minimist");
const { loremIpsum } = require("lorem-ipsum");
const Chance = require('chance');


const { port = 3001, schema: path } = minimist(process.argv);

const JSONSchemaFakerConfig = {
  options: {
    // Available options: https://github.com/json-schema-faker/json-schema-faker/tree/master/docs#available-options
    maxItems: 100,
  },

  // https://github.com/json-schema-faker/json-schema-faker/blob/master/docs/USAGE.md#custom-formats
  customFormats: {
    html: loremIpsum({
      count: 5,
      format: "plain",
      units: "paragraphs"
    });
  },

  // https://github.com/json-schema-faker/json-schema-faker/blob/master/docs/USAGE.md#extending-dependencies
  extensions: {
    chance: () => {
      const chance = new Chance();

      chance.mixin({
        user: () => {
          const first = chance.first();
          const last = chance.last();
          const email = chance.email();
          return `${first} ${last} – ${email}`;
        }
      });

      return chance;
    }
  }
}

const server = createServer(
  { path },
  {
    config: { mock: { dynamic: JSONSchemaFakerConfig } },
    components: {
      logger: createLogger()
    }
  }
);

server.listen(port).then(() => {
  console.log(`🤖 Mocking server enabled on port ${port}`);
});
```

No.


philsturgeon:
@vanhoofmaarten this looks really cool, but just to set expectations, we are in feature freeze right now while we get v3.0 stable. In a week or two we can come back to this. 

On the plus side, that will give you a chance to iron out any kinks and play around with it. Maybe write a few more tests so we can just click that green button without fear. :)

vanhoofmaarten:
@philsturgeon 

Fully understandable. I the meantime I could indeed write a few extra tests. `HttpMocker.spec.ts` could use a few extra I think.

It seems that [the failing test](https://circleci.com/gh/stoplightio/prism/2243?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) is not because of the code.

`Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag`

Any assistance here would be much appreciated.

XVincentX:
@vanhoofmaarten You can ignore that error. It happens because we set CircleCI to not pass env variables to external PRs and so the CodeClimate coverage upload fails. It's all good on your side, this is up to me to review.

philsturgeon:
@XVincentX we can throw this in there to early exit instead of fail:

```
  # Command based on instructions from CircleCI blog for avoiding fork secret leakage or failures
  # https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors/
  early-return-forks:
    description: >-
      If this build is from a fork, stop executing the current job and return success.
      This is useful to avoid steps that will fail due to missing credentials.
    steps:
      - run:
          name: Early return if this build is from a forked PR
          command: |
            if [ -n "$CIRCLE_PR_NUMBER" ]; then
                echo "Nothing to do for forked PRs, so marking this step successful"
                circleci step halt
            fi
```

XVincentX:
Totally! I’ll make that happen somewhere in another PR in an hour or so.

XVincentX:
Done — tests are now green.

vanhoofmaarten:
Hi there. I found the time to add some more thorough testing. I also split up the extensions part of the original PR to 2 separate properties, external generators and custom generators, as JSON Schema Faker also uses 2 separate methods.

If you could review these changes and provide feedback, I'll write up some documentation afterwards.

Thanks!

philsturgeon:
@vanhoofmaarten hey, thank you for the update and improved testing! We've had a few sizeable changes in master lately as we're trying to get the HTTP Client (#583) sorted out. Could you try and sort out the conflicts, and we'll aim to get this merged? It went into the Icebox for a little bit whilst we focused on tech debt and general cleanup, but it's becoming clear we need more extensibility, and sooner rather than later.

vanhoofmaarten:
Will do!

packages/http/src/types.ts
```diff
@@ -9,11 +9,57 @@ export type PrismHttpComponents = IPrismComponents<IHttpOperation, IHttpRequest,
 // TODO: should be complete | and in the @stoplight/types repo
 export type IHttpMethod = HttpMethod | 'trace';
 
+// TODO: Replace with official JSON Schema Faker types
```
philsturgeon: Please can you take care of this bit

philsturgeon:
@vanhoofmaarten thank you. Two more things:

1. how does this work for CLI people? How can we register these extensions without needing to use the JS API.

2. When 1 is figured out, can we add something to `docs/`? Maybe a guide on the concept in general with examples for both, and link to that from the existing JS and CLI docs?

vanhoofmaarten:
@philsturgeon 

## CLI
Good question. I don't think there's escaping plugging straigth in the Javascript internals of Prism. However, we could circumvent the writing of a custom server implementation by making Prism CLI watch a predefined set of files or folder in the directory it is running.

The folder structure could look something like this:
```
folder/prism/is/ran/in
└── prism
    ├── custom-extensions.js
    ├── external-generators.js
    └── custom-generators.js
```

```
folder/prism/is/ran/in
└── prism
    ├── custom-extensions
    │   ├── extensions-1.js
    │   ├── extensions-2.js
    │   └── extensions-3.js
    ├── external-generators.js
    │   ├── external-generators-1.js
    │   ├── external-generators-2.js
    │   └── external-generators-3.js
    └── custom-generators.js
        ├── custom-generators-1.js
        ├── custom-generators-2.js
        └── custom-generators-3.js
```
Every file exports a module which Prism picks up and plugs in the correct JSON-Schema-Faker configuration property.

Perhaps an extra file `json-schema-faker-options.js` which on it's turn configures the IJsonSchemaFakerOptions?

This way of working also allows custom functionalities with the Prism Docker image by simply mounting the Prism folder in the Docker container.

Thinking a bit further, we could even tap in the Fastify server by making Prism pick up JS modules from a `/hooks` folder which contain files that tap in to the Fastify [Hooks](https://github.com/fastify/fastify/blob/master/docs/Hooks.md) API.

## Docs
Of course. I'll write out the technical possibilities.

philsturgeon:
Those options all sound interesting, but maybe a good thing to model this after might be custom functions. I'm not sure if you noticed them appear in rulesets in v4.1.0? 

https://stoplight.io/p/docs/gh/stoplightio/spectral/docs/guides/custom-functions.md

I feel like this might be a good model for description extensions.

philsturgeon:
That said we do not have the existing "ruleset/config file" for Prism, so our options might indeed be restricted to A) cli option, B) look for a hooks directory on load. 

I'm ok with conventions being used and configuration appearing later.

vanhoofmaarten:
>Those options all sound interesting, but maybe a good thing to model this after might be custom functions. I'm not sure if you noticed them appear in Spectral v4.1.0 https://stoplight.io/p/docs/gh/stoplightio/spectral/docs/guides/custom-functions.md I feel like this might be a good model for description extensions.

I don't have any experience with that feature of Spectral, I'll look in to it.

> I'm ok with conventions being used and configuration appearing later.

Agreed

packages/http/src/mocker/negotiator/types.ts
```diff
@@ -14,6 +14,6 @@ export type NegotiationOptions = IHttpOperationConfig;
 
 export type NegotiatePartialOptions = {
   code: string;
-  dynamic: boolean;
+  dynamic: boolean | IHttpOperationDynamicConfig;
```
philsturgeon: Hmm what is happening here? I generally do not like boolean or something else, but... maybe. 

Also, is this a generic extension system, or a "dynamic mode extension system". I think maybe the former would be more useful, and we pass in things like "mode: dynamic" to that extension system.

vanhoofmaarten: Well my suggestion here is to build further upon the functionality you've already provided.

3 scenario's:
- Boolean
  - False: Mock with examples.
  - True: Dynamic mocking with file-based extensions from the default directories.
  - IHttpOperationDynamicConfig: Dynamic mocking with programmatically defined extensions +  file-based extensions from the default or reconfigured directories.

The way the code is build now, it's only a "dynamic mode extension system". This suggested extension system plugs in to JsonSchemaFaker's `extend` & `define` methods. 

Does this answer your question?

prism
```diff
@@ -0,0 +1 @@
+./packages/http/src/__tests__/fixtures/extensions/
```
vanhoofmaarten: I've put a relative symlink in the root of the project folder to test this use case. https://github.com/stoplightio/prism/blob/529171ec862891f798c4a94781fb63ea099891ad/packages/http/src/mocker/__tests__/HttpMocker.spec.ts#L226

I'm not all that sure of this being a good solution for this problem.

packages/http/src/getMockExtensions.ts
```diff
@@ -0,0 +1,153 @@
...
+const validateExtensions = <T>(extensions: T, validator: JsonSchemaFakerExtensionValidator): T =>
+  Object.keys(extensions).reduce((acc, extensionKey) => {
+    const extension = extensions[extensionKey];
+    if (validator(extension)) {
+      return {
+        ...acc,
+        [extensionKey]: extension,
+      };
+    } else {
+      // TODO: Use Prism logging to warn use about the invalid module
```
vanhoofmaarten: @philsturgeon 

I'd like to replace this warning with a message outputted by the Logger Prism uses, only, I'm far from experienced with the `fp-ts` paradigm you are using in this project. Some help here would be much appreciated.

XVincentX:
@vanhoofmaarten I guess you'll let me know when you're done so I can take a look at this?

This PR is starting to become gigantic

packages/http/src/types.ts
vanhoofmaarten: Took care of it.

vanhoofmaarten:
@XVincentX I believe I touched all possible use cases regarding dynamic mocking. So be my guest.

vanhoofmaarten:
@XVincentX 

Thank you for the thorough update. It's pretty clear now, knowing Prism has a roadmap to be a hosted service, the use cases this PR tries to solve don't suffice the long-term requirements.

How do we continue from here on?

XVincentX:
I would say I need to read the issue again :trollface: and understand to which extent we can aim without compromising these requirements. I'll definitely keep you posted.

XVincentX:
This PR is stale — I'll close it and re-evaluate when the time comes.

PR #400, opened 2019-07-05
Make the validation process sync

XVincentX:
I've checked the code multiple times and there is really no reason the validation process should be marked as async — it's indeed not.

This PR removes any doubts about it so it's going to be easier to reason about it


XVincentX:
@philsturgeon Get ready for another monster soon :)

philsturgeon:
This is good work, but I think we should focus on writing test harness files so we can be more confident in e2e continuing to work through these big changes. Now is a time for stabilisation not clearing tech debt. :)

--
Phil Sturgeon
@philsturgeon

PR #402, opened 2019-07-06
Expand fs-ts in Prism's router 

XVincentX:
The following PR extends the usage of fp-ts to the routing package as well by basically making sure it does not throw exceptions anymore, but rather use the `Either` object to express an error object.

With this — the router and the mocker finally compose because the type match (they both return an `Either<Error, T>`.

Extend the `Either` usage to the router was indeed the easy part: https://github.com/stoplightio/prism/pull/402/files#diff-f9a10b37616fb5669ecd5218fc8535c9L16

The problem started when I started to integrate and try to compose the new function in the mega-file-to-split:

1. The whole flow is sync apart from the edge case when we need to employ the forwarder, and this requires an additional abstraction layer
https://github.com/stoplightio/prism/pull/402/files#diff-47c5dc2d65fd624c869f5f08d0cfb56aR45

2. What's really preventing from having a clean and functional flow is the validation process that's basically creating an empty array, giving it to the mocker and expecting to receive a filled array. This forces me to keep some stuff here and some stuff there; if the mocker could just return the validations, that'd improve the code a lot.

3. In order to keep the API compatible with what we have, I have to do some wrapping I'd like to avoid
https://github.com/stoplightio/prism/pull/402/files#diff-47c5dc2d65fd624c869f5f08d0cfb56aR98

---

That said, the funny thing is that, although this Pull Request is meant to be an improvement, you can argue that the code is effectively uglier than it is. (Well I do not think it is, but you mileage may vary)

The good news though is that — I'm not sure if you remember, we were discussing on how to refactor this part and nobody (me included) really came up with good ideas.

By trying to extend the functional parts to the router — I now know exactly what needs to be done and how to move forward. This is freaking awesome, to be honest.

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -35,7 +35,7 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         }
       },
 
-      process: (input: Input, c?: Config) => {
+      process: async (input: Input, c?: Config) => {
         // build the config for this request
         const configMerger = configMergerFactory(defaultConfig, customConfig, c);
         const configObj: Config | undefined = configMerger(input);
@@ -145,14 +145,14 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
             );
         }
 
-        return Promise.resolve({
+        return {
           input,
           output: undefined,
           validations: {
             input: [],
             output: [],
           },
-        });
+        };
       },
     };
   };
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -1,6 +1,6 @@
 import { DiagnosticSeverity } from '@stoplight/types';
-import { right, toError } from 'fp-ts/lib/Either';
-import { fromEither, left2v, tryCatch } from 'fp-ts/lib/TaskEither';
+import { toError } from 'fp-ts/lib/Either';
+import { fromEither, left2v, right2v, tryCatch } from 'fp-ts/lib/TaskEither';
 import { configMergerFactory, PartialPrismConfig, PrismConfig } from '.';
 import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, PickRequired, ProblemJsonError } from './types';
 
@@ -42,31 +42,28 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         const inputValidations: IPrismDiagnostic[] = [];
 
         if (components.router) {
-          return fromEither(
-            components.router
-              .route({ resources, input, config: configObj }, defaultComponents.router)
-              .map(r => {
-                if (r) return r;
-                return undefined;
-              })
-              .orElse(error => {
-                // rethrow error we if we're attempting to mock
-                if ((configObj as IPrismConfig).mock) {
-                  throw error;
-                }
-                const { message, name, status } = error as ProblemJsonError;
-                // otherwise let's just stack it on the inputValidations
-                // when someone simply wants to hit an URL, don't block them
-                inputValidations.push({
-                  message,
-                  source: name,
-                  code: status,
-                  severity: DiagnosticSeverity.Warning,
-                });
+          return fromEither(components.router.route({ resources, input, config: configObj }, defaultComponents.router))
+            .map(r => {
+              if (r) return r;
+              return undefined;
+            })
+            .orElse(error => {
+              // rethrow error we if we're attempting to mock
+              if ((configObj as IPrismConfig).mock) {
+                throw error;
+              }
+              const { message, name, status } = error as ProblemJsonError;
+              // otherwise let's just stack it on the inputValidations
+              // when someone simply wants to hit an URL, don't block them
+              inputValidations.push({
+                message,
+                source: name,
+                code: status,
+                severity: DiagnosticSeverity.Warning,
+              });
 
-                return right(undefined);
-              }),
-          )
+              return right2v<Error, Resource | undefined>(undefined);
+            })
             .chain(resource => {
               // validate input
               if (resource && components.validator && components.validator.validateInput) {
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -43,27 +43,26 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
 
         if (components.router) {
           return fromEither(components.router.route({ resources, input, config: configObj }, defaultComponents.router))
-            .map(r => {
-              if (r) return r;
-              return undefined;
-            })
-            .orElse(error => {
-              // rethrow error we if we're attempting to mock
-              if ((configObj as IPrismConfig).mock) {
-                throw error;
-              }
-              const { message, name, status } = error as ProblemJsonError;
-              // otherwise let's just stack it on the inputValidations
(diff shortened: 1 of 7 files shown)


packages/core/src/factory.ts
```diff
@@ -33,101 +35,124 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
-                config: configObj,
-              },
-              defaultComponents.validator,
-            )),
-          );
-        }
+                const { message, name, status } = error as ProblemJsonError;
+                // otherwise let's just stack it on the inputValidations
+                // when someone simply wants to hit an URL, don't block them
+                inputValidations.push({
```
StefanDywersant: Pushing things to array which is defined outside function chains feels not right. How about using validations? (https://dev.to/gcanti/getting-started-with-fp-ts-either-vs-validation-5eja - Validation section in the middle)

packages/core/src/factory.ts
```diff
@@ -33,101 +35,124 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
-          output = components.mocker
-            .mock(
-              {
-                resource,
-                input: { validations: { input: inputValidations }, data: input },
-                config: configObj,
+                return right2v<Error, Resource | undefined>(undefined);
               },
-              defaultComponents.mocker,
+              value => right2v(value),
```
StefanDywersant: I had a problem finding documentation for `right2v`. Isn't it deprecated?

packages/core/src/factory.ts
```diff
@@ -33,101 +35,124 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
         }
       },
 
-      process: async (input: Input, c?: Config) => {
+      process: (input: Input, c?: Config) => {
```
StefanDywersant: You could keep `async` here, you are calling `Promise.resolve` at the end anyway (`components.router.route()` is also async). Without `async` someone could think that is dealing with sync block.

XVincentX: Fixed in d19d15e

packages/core/src/factory.ts
XVincentX: It is deprecated in 2.x — Prism is currently on 1.9. I'll migrate it soon though.

packages/core/src/factory.ts
XVincentX: Yes! That's exactly what I wanted to do. However I'm proceeding with baby steps here as — you've probably noticed — it's hard to digest all in one place; moreover we're close to the release and I didn't want to change that many things at the time.

…but yes it's exactly the direction I want to take!

XVincentX:
> One general note: the code looks complicated because it is long and deeply nested. How about enclosing parts into descriptive functions, so a human can see the overall flow and can dig into details if needed.

Yeah I will probably try again — I gave up because every function was taking something like 3000 parameters complicating the thing anyway. I'm willing to give it another try though.

Thanks for the review!

packages/core/src/types.ts
```diff
@@ -43,14 +44,18 @@ export interface IRouter<Resource, Input, Config> {
...
-  ) => Resource;
+  ) => Either<Error, Resource>;
 }
 
 export interface IForwarder<Resource, Input, Config, Output> {
   forward: (
     opts: { resource?: Resource; input: IPrismInput<Input>; config?: Config },
     defaultForwarder?: IForwarder<Resource, Input, Config, Output>,
   ) => Promise<Output>;
+  fforward: (
```
chris-miaskowski: why `fforward`? it's a breaking change for internal packages you realize?

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -188,14 +188,9 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
...
 
-  it.skip('should automagically provide the parameters when not provided in the query string and a default is defined', async () => {
-    const response = await server.fastify.inject({
-      method: 'GET',
-      url: '/pets/findByStatus',
-    });
-
-    expect(response.statusCode).toBe(200);
-  });
+  test.todo(
```
chris-miaskowski: Such a decent test :( why remove it?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -269,24 +259,7 @@ describe('Http Client .process', () => {
...
-            },
-            body: {
-              id: 'should not be a string',
-              status: 'should be one of "placed", "approved", "delivered"',
-              complete: 'should be a boolean',
-            },
-          }),
-        ).rejects.toThrowError(ProblemJsonError.fromTemplate(UNPROCESSABLE_ENTITY));
-      });
+      test.todo('with invalid body returns validation errors');
```
chris-miaskowski: :fire: :fire: :fire:  You're removing my lovely tests :( :fire:

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: It was skipped — I'm fine putting it back but it's not getting executed anyway. I'd rather see a todo that says "Hey write this" instead of a skip that you do not easily see

packages/http/src/__tests__/http-prism-instance.spec.ts
XVincentX: It was skipped — I'm fine putting it back but it's not getting executed anyway. I'd rather see a todo that says "Hey write this" instead of a skip that you do not easily see

packages/core/src/factory.ts
```diff
@@ -39,93 +41,114 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
-        let output: Output | undefined;
-        if (resource && components.mocker && (configObj as IPrismConfig).mock) {
-          // generate the response
-          output = components.mocker
-            .mock(
-              {
-                resource,
-                input: { validations: { input: inputValidations }, data: input },
-                config: configObj,
+                return right2v<Error, Resource | undefined>(undefined);
```
chris-miaskowski: `right2v`, whoa what a lovely interface :P I have a suggestion 
`right2lolz` because why not? Ok enough sarcazm for today

XVincentX: It's actually a good point — it has been fixed in 2.x but I didn't want to deal with breaking changes right now.

StefanDywersant: We'll get less PR from random people. Entry threshold is set high ;)

philsturgeon:
We should keep skipped tests in place and try to mark them. Please add notes explaining why they’re skipped and make stories to finish implementing them. It’s tech debt we need to resolve. If we just delete them then we’re ignoring technical debt and that’s not ideal!

--
Phil Sturgeon
@philsturgeon

XVincentX:
@philsturgeon 

Just to clarify — the tests haven't been removed; I've just moved them from `test.skip` to ´test.todo´. I also left the comments.

The current implementation there wasn't useful anyway and just gave the impression that the test was done. Now we can clearly see that the test is not done, and we'll fix that in the future.

philsturgeon:
Sounds awesome, good work! 

--
Phil Sturgeon
@philsturgeon

PR #442, opened 2019-07-08
test: run test suite on macos on master

XVincentX:
Run harness test suite on macOS when we merge something in master and when we're releasing stuff. Closes #442 

SO-149

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,11 +68,11 @@ jobs:
           command: yarn
       - run:
           name: Install dockerize
-          command: curl -OL https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-$(uname)-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-$(uname)-amd64-$DOCKERIZE_VERSION.tar.gz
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -51,28 +57,29 @@ jobs:
...
+    executor: << parameters.executor >>
     steps:
       - checkout
       - run:
           name: Install Dependencies
           command: yarn
       - run:
           name: Install dockerize
-          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+          command: curl -OL https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-$(uname)-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-$(uname)-amd64-$DOCKERIZE_VERSION.tar.gz
```
P0lip: Keep in mind `uname` returns `Linux` than `linux` when executed under Linux. I guess this doesn't matter, though.

P0lip: I'd download it to `/opt` or even better to `/tmp`, unpack there and move afterwards.

.circleci/config.yml
```diff
@@ -116,9 +123,18 @@ workflows:
...
+          name: harness-macos
+          executor: macos
           filters:
             tags:
               only: /.*/
+      - harness:
+          executor: docker-node
+          filters:
+            branches:
+              only: master
```
P0lip: I guess you might want to include develop here later on.

XVincentX: I'll do whatever I can to *not* use git flow in Prism. It's solving problems that Prism does not have.

.circleci/config.yml
XVincentX: Apparently works, so I'll not touch it :trollface:

PR #446, opened 2019-07-10
Remove the body validation registry

### LINKED ISSUE #420: Prism does not even attempt to validate a body if "consumes" not defined.
Prerequisites

The following spec:
```
{
  "swagger": "2.0",
  "paths": {
    "/pet/findByStatus": {
      "post": {
        "produces": ["application/json"],
        "parameters": [
          {
            "name": "status",
            "in": "query",
            "required": true,
            "type": "array",
            "items": {
              "type": "string",
              "enum": ["available", "pending", "sold"],
              "default": "available"
            },
            "collectionFormat": "multi"
          },
          {
            "in": "body",
            "name": "body",
            "schema": {
              "type": "object",
              "properties": {
                "id": {
                  "type": "integer",
                  "format": "int64"
                },
                "status": {
                  "type": "string",
                  "description": "Order Status",
                  "enum": ["placed", "approved", "delivered"]
                },
                "complete": {
                  "type": "boolean",
                  "default": false
                }
              }
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```
**Steps to reproduce**

1. Run Prism with the provided spec2. Make a request to PRISM `curl -X POST http://127.0.0.1:4010/pet/findByStatus?status=sold -H 'Content-Type: application/json' --data '{ "id": "should be a number" }'`3. Observe the output

**Expected**

{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid: [{\"path\":[\"body\",\"id\"],\"code\":\"type\",\"message\":\"should be integer\",\"severity\":0}]"}

**Actual**

A lorem ipsum response similar to "exercitation reprehenderit dolore quis Lorem"

Context & Discussion

The problem stems for the fact that the OAS file lacks `"consumes": ["application/json"],`.When we attempt to validate the body  we first look at the expected media type of that body (as defined by OAS). Since the OAS does not specify the media type we can't find a validator and we don't even try to validate resulting in no validation errors.

 suggested that this is not a correct behavior and we should distinct between serialization and validation. In other words we should try to validate everything that can be serialized to a js object.

Read the Slack discussion for more contextAnd this one

Like  said:

... we Schema validate everything that has a Schema, and transform specific mine types that have known transformations (like xml and so far nothing else that I know of)

XVincentX:
This PR removes the validation registry concept for the body parameters, since it was based on the assumption that a validator is associated with the serialised type — something we won't be doing anymore.

Deserialisation and validation will be two different things and so the validation is always possible, given a "schema" or something.

I've removed the code that was building up such infrastructure; by doing this I've discovered some mocking problems in certain parts of the code (basically missed cleanups) that were faking some test results. I've modified these as well to behave correctly.

I've added two harness tests to check that it's behaving correctly.

Closes #420


packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -112,10 +112,10 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
 
   it('should not mock a verb that is not defined on a path', async () => {
     const response = await server.fastify.inject({
-      method: 'POST',
```
XVincentX: In case you're asking yourself: yes, this test has been wrong since forever. It's now fixed.

packages/http/src/validator/validators/__tests__/__snapshots__/body.spec.ts.snap
```diff
@@ -1,13 +1,12 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`HttpBodyValidator validate() body schema is provided validator for given media type exists return validation errors 1`] = `
+exports[`HttpBodyValidator validate() body schema is provided return validation errors 1`] = `
 Array [
   Object {
-    "code": "testName",
-    "message": "testMessage",
+    "code": "type",
```
XVincentX: This was caused because the mocked functions apparently were not restored correctly, thous returning a not correct error

packages/http/src/__tests__/fixtures/no-refs-petstore-minimal.oas2.json
```diff
@@ -16,28 +16,6 @@
               "default": "available"
             },
             "collectionFormat": "multi"
-          },
-          {
-            "in": "body",
```
XVincentX: By not mocking the validation process anymore everywhere there was a test checking a completely different thing now starting to fail because of the body validation (which is now happening for real) — therefore I removed the body parameter at all so that the request goes through correctly as it should.

XVincentX:
> have you checked what happens if we send a text/plain

Fastify currently is not configured to handle such content type, so you will receive a 415 and Prism won't be invoked at all.

PR #450, opened 2019-07-11
fix: respect the required parameter in the body

### LINKED ISSUE #419: Prism does not take the "required" property into account when validating body param
Prerequisites
```
{
  "swagger": "2.0",
  "paths": {
    "/pet": {
      "get": {
        "produces": ["application/json"],
        "consumes": ["application/json"],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "required": false,
            "schema": {
              "type": "object",
              "properties": {
                "id": {
                  "type": "integer",
                  "format": "int64"
                }
              }
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```
**Steps to reproduce**

1. Start prism with the provided endpoint2. `curl -X POST http://127.0.0.1:4010/pet`3. Observe output

**Expected**

Should return 200 and a lorem ipsum response

**Actual**

Returns an error

`{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid: [{\"path\":[\"body\"],\"code\":\"type\",\"message\":\"should be object\",\"severity\":0}]"}`

**Context**

We currently don't check the required field when validating body. The proposed solutions is that `HttpBodyValidator ` should extend `HttpParamsValidator` class.

XVincentX:
Closes #419
Requires #446


packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -53,30 +53,61 @@ describe('HttpValidator', () => {
...
         });
 
         describe('request.body is set', () => {
-          it(
-            'validates body',
-            test({
-              request: { body: { contents: [] }, path: [], query: [], headers: [], cookie: [] },
-            }),
-          );
+          describe('request body is not required', () => {
```
philsturgeon: This is a good use of describe but I think you could improve the reliability and readability of these tests. It looks like your setup was copied and pasted from the setup in the test() method, which means that method could change, and then the test is not actually confirming the difference between required true/false m, but confirming the difference between test and this copied and pasted code. 

Can we either pass in some arguments for request object; or find a less hidden approach? 

Often setting up an object with the default structure, then overriding specific parts in a single test, is a good way to make sure you’re really testing the difference behaviour related to that change.

XVincentX: Yeah I think this is a good point, I'll refactor this and make sure it's better

XVincentX: Done in ecee804

packages/http-server/src/__tests__/fixtures/getOperationWithBody.oas2.json
```diff
@@ -2,7 +2,7 @@
   "swagger": "2.0",
   "paths": {
     "/pet": {
-      "get": {
+      "post": {
```
XVincentX: In case you're wondering, get should not really have body parameters.

packages/http/src/validator/index.ts
```diff
@@ -33,10 +33,15 @@ export class HttpValidator implements IValidator<IHttpOperation, IHttpRequest, I
 
     if (config.body) {
       const { body } = input;
-
-      this.bodyValidator
-        .validate(body, (request && request.body && request.body.contents) || [], mediaType)
-        .forEach(validationResult => results.push(validationResult));
+      if (request && request.body) {
```
XVincentX: This will be way easier to read when we'll introduce `Option<T>` in fp-ts.

PR #454, opened 2019-07-12
Release beta.6

XVincentX:
This might be the 3.0 final!

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@types/type-is": "^1.6.2",
     "chance": "^1.0.18",
     "gavel": "^6.1.3",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "@types/type-is": "^1.6.2",
     "chance": "^1.0.18",
     "gavel": "^6.1.3",
-    "globby": "^10.0.1",
+    "globby": "^9.2.0",
```
XVincentX: This was not an intended update — breaking changes not really, not right now. I need to read the changeling and then I'll schedule the upgrade.

PR #455, opened 2019-07-15
docs: remove banners

XVincentX:


tbarn:
Now that I think about it... I'm not entirely sure this is what is causing the issue. But if you want to try this, I'm down.

XVincentX:
These informations are useless in any case, I think.

PR #458, opened 2019-07-16
test: add 'accept content negotiation' e2e tests

### LINKED ISSUE #421: E2E Accept / Content Negotiation
AC

come up with e2e tests covering the content negotiation (see  if you need examples, also see below link)

add those tests to the test harness

record failure messages in gold master file (if the result is nonsense - e.g. it's clearly not correct - then mark the test as skipped and create Jira bugs for each one. Link this story to those jira bugs and assign to  - he'll schedule it)

test both oas3 and oas2

try to time box this story to ~2 days (e.g. don't write 1000 tests )

if you run out of test cases  and  may be able to give you suggestions

Make sure to add the actual tests to TestRails

Examples content negotiation issues to be tested

When I send a request to an operation with Accept header containing X (ie. application/json)
And this operation can produce a response that matches this X format
Then I should get back a response matching that format 
And this response should have Content-Type header matching X format
And this response should have 200 status code 

When I send a request to an operation with Accept header containing application/xml,application/json
And this operation can produce a response with content-type of application/xml
Then I should get back a response with content-type of application/xml
And this response should have 200 status code 

When I send a request to an operation with Accept header containing text/plain,application/json
And this operation can produce a response with content-type of text/plain
Then I should get back a response with content-type of text/plain
And this response should have 200 status code 

When I send a request to an operation with Accept header containing application/xml,application/json
And this operation cannot produce a response with content-type of application/xml
And this operation can produce a response with content-type of application/json
Then I should get back a response with content-type of application/json
And this response should have 200 status code 

When I send a request to an operation with Accept header containing application/*
And this operation can produce a response with content-type of application/*
Then I should get back a response with the first content-type that matched application/*
And this response should have 200 status code 

When I send a request to an operation with Accept header containing X (ie. application/json)
And this operation cannot produce a response that matches this X format
Then I should get back a response with Content-Type of application/problem+json
And this response should have 406 status code 
And this response status text should be Not Acceptable

When I send a request to an operation with Accept header containing `*/*`
And this operation produces a response with content-type of `*/*`
Then I should get back a response with Content-Type of `*/*`

When I send a request to an operation with Accept header containing `*/*`
And this operation does not produce a response with content-type of `*/*`
Then I should get back a response with the first content-type from the list

Notes

Please test / use some more advanced Accept header examples - multiple types, weighted types (;q= (q-factor weighting)), stars (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept, https://en.wikipedia.org/wiki/Media_type).

Related: 

Accept = undefined

find out what should happen when Accept header is not defined (check out this bug report, we don't know how to solve at the type of writing this story )

lag-of-death:
Harness tests for content negotiation

Closes #421

Failing tests in https://github.com/stoplightio/prism/pull/470


lag-of-death:
@XVincentX, can I mark some tests as "it's OK that these fail for now" somehow?

XVincentX:
@lag-of-death No, we will have to fix the bug in this PR :trollface:

philsturgeon:
The "it's OK that these fail for now" stance was while we were trying to harden prism v3.0 for the last few months. At this point if something fails we need to fix it.

XVincentX:
Yeah, just keep the PR here, I'll jump on in a while, fix the bug and add the necessary other internal tests.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -197,10 +197,11 @@ const helpers = {
               ...contentNegotiationResult,
             }));
         } else {
-          logger.trace(`Unable to find a content for ${mediaTypes}, returning an empty text/plain response.`);
+          logger.trace(`Unable to find a content for ${mediaTypes}, returning application/json+problem response.`);
+
           return right({
```
philsturgeon: This is probably making your test pass but this is providing the functionality that what we need. 

Assuming this spec is being used:

```
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: Get Todo Items
          content:
            'text/plain':
              example: hello
```

This bit of code should be triggered when a matching status code is defined, but no schema or example can be found. 

Seeing as this does not have a 406 defined, Prism should not default with a 200 OK text/plain, by which I mean it should not be hitting this point in the code. This code is correct, but Prism should respond with a 406 Not Acceptable application/problem+json instead. 


This change fixes the symptom but not the cause.

XVincentX: I agree, we should probably return a `left` saying "we cannot generate anything here" and then somewhere outside of this piece of code we'd be creating such response.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -197,12 +198,8 @@ const helpers = {
...
             }));
         } else {
-          logger.trace(`Unable to find a content for ${mediaTypes}, returning an empty text/plain response.`);
-          return right({
-            code,
-            mediaType: 'text/plain',
-            headers: headers || [],
-          });
+          logger.warn(`Unable to find a content for ${mediaTypes}`);
+          return left(ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find a content for ${mediaTypes}`));
```
lag-of-death: is it ok to say `a content`? ain't that uncountable?

XVincentX: Ah, I do not know :) — what'd the be correct sentence?

XVincentX:
I've also added an unit test and two changelog lines so that making the changelog will be way easier.

XVincentX:
Closes #421

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
philsturgeon: just content is ok, no need for a.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -197,12 +142,8 @@ const helpers = {
               ...contentNegotiationResult,
             }));
         } else {
-          logger.trace(`Unable to find a content for ${mediaTypes}, returning an empty text/plain response.`);
```
philsturgeon: Hey I posted a comment on this before but its gone, and I talked about it on Slack, this old behavior was correct. Let mew see if I can dig it out to avoid writing again.

XVincentX: Ok, I guess it was lost because of the rebases!

philsturgeon: Here it is https://github.com/stoplightio/prism/pull/458#discussion_r303908700

I see a reply on there @XVincentX im not really sure what left and right are all about, but I'll be happy to learn.

Regardless of that, the code is still changing. Being more functional was not the issue. 

The issue is the removal of the old text/plan message. This behavior _is wanted_ for when X status is requested and X status is defined in responses but it just so happens to not have any content (no example(s) or schema).  Not having content **must not** cause a 406 error, this change has now become a bug.

XVincentX: @philsturgeon The code is correct. This is *still* happening [here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L158-L167)

This part of the code happens when *there's a content*, but not the one you're requesting.

Does that clarify?

XVincentX: `left`: something went wrong
`right`: I was able to do what you asked me to do

philsturgeon: Awesome, thank you!

CHANGELOG.md
```diff
@@ -1,7 +1,15 @@
 # Change Log
 
 All notable changes to this project will be documented in this file.
-See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to Semantic Versioning.
+
+# Unreleased
+
+## Bug Fixes
```
philsturgeon:
```suggestion
## Fixed
```

chris-miaskowski:
> The "it's OK that these fail for now" stance was while we were trying to harden prism v3.0 for the last few months. At this point if something fails we need to fix it.

Btw I'm totally cool with that if things are easy to but bear in mind that this is called "scope creep" ;)

CHANGELOG.md
```diff
@@ -1,11 +1,20 @@
...
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to Semantic Versioning.
+
+# Unreleased
+
+## Fixed
+
+- Prism is now returning a `406` error instead of an empty response in case it is not able to find a response whose content type satisfies the provided `Accept` Header
+- Prism now respects the `q` value in the `Accept` header to specify the content type preference
+- Prism is now correctly returning `text/plain` when the document does *not* specify any Content Type for the examples
```
XVincentX: This is in #465 — but yeah.

XVincentX:
@philsturgeon I've deleted the failing test and put them in #470 — they're all happening because the OAS2 adapter is not correctly creating one example per content type, but rather a single example with multiple keys inside, and this is not correct.

philsturgeon:
@chris-miaskowski it’s only scope creep if we expected everything to work perfectly first time. Hopefully we don’t discover any bugs, but if we do we should not be sweeping them under the carpet for the sake of closing this issue quickly. These stories  are the stories where we implement tests and fix bugs. If it takes a long time because there’s a lot of bugs that’s ok, because we need to fix the bugs. 🤷🏻‍♂️

Issue #413, opened 2019-07-06
Test Harness: Default Responses


brianmrock:
**User Story**

As a developer I want to get back a default response when no other response codes can be returned.

**ACs**
- add test cases to test harness
- add at least the cases described in the examples section
- if any of the test fails then mark that it is an invalid result (e.g. mark as skipped) and report a Jira bug
- link that bug to this card 
- Make sure to add the actual tests to TestRails

**Additional resources**

https://stoplightio.atlassian.net/wiki/spaces/EN/pages/52887637/Prism#Prism-MockServerFeatures (please find relevant row)

**Example tests**

When I send a request to an operation
And this operation is defined
And this operation has only a default response defined
Then I should get back the default response

When I send a request to an operation
And I force the response by adding __code to the path (ie http://localhost:4010/pets/123?__code=200)
And this operation has no response with such code defined
And this operation has a default response defined
Then I should get back the default response

When I send a request to an operation
And I force the response by adding __code to the path (ie http://localhost:4010/pets/123?__code=200)
And this operation has no response with such code defined
And this operation does not have default response defined
Then I should get back a response with 500 status code
And the response's payload should describe the error

When I send a request to an operation
And this operation has no response with a matching status code
And this operation does not have default response defined
Then I should get back a response with 500 status code
And the response's payload should describe the error

When I send a request to an operation
And in the path I specify ?_code=some-code (ie ?_code=200)
And the operation has such status code response defined
Then I get back a response with a status code of 200

When I send a request to an operation
And that operation has any 2xx defined
And the request was valid
Then I get back the lowest 2xx response (e.g. if 204 and 210 are defined, return 204)

When I send a request to an operation
And in the path I specify ?_code=404
And the operation has no 404 status code response defined
And the operation has 400 status code defined, but no schema or example defined in the response
Then I get back a response with a blank body with Content-Type of text/plain

When I send a request to an operation
And in the path I specify ?_code=404
And the operation has 404 status code response defined
And in this operation there is no schema or example defined in the response
Then I get back a response with Content-Type of application/json

XVincentX:
Blocked on product decision.

philsturgeon:
@XVincentX could you add context to what decion is blocking so folks know what’s up when they read this? I don’t remember myself. If this is the lowest 400 thing then could you mention that?

XVincentX:
@philsturgeon Comment updated.

philsturgeon:
Cool thanks! 

[...]

PR #465, opened 2019-07-16
test: add e2e tests for default responses

lag-of-death:
Adds test examples (passing ones) for #413 
Failing examples are in PR https://github.com/stoplightio/prism/pull/468


test-harness/specs/default_responses-AC-1.oas2.txt
```diff
@@ -0,0 +1,30 @@
...
+          schema:
+            type: string
+            example: test
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos
+====expect====
+HTTP/1.1 200 OK
+content-type: */*
```
XVincentX: @philsturgeon I think this response is not correct

test-harness/specs/default_responses-AC-5.oas2.txt
```diff
@@ -0,0 +1,26 @@
...
+          schema:
+            type: string
+            example: ok
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos?_code=200
+====expect====
+HTTP/1.1 200 OK
+content-type: */*
```
XVincentX: @philsturgeon Same here, this needs an investigation

test-harness/specs/default_responses-AC-6.oas2.txt
```diff
@@ -0,0 +1,30 @@
...
+          description: test1
+        210:
+          description: test2
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos
+====expect====
+HTTP/1.1 200 OK
+content-type: */*
```
XVincentX: dtto

test-harness/specs/default_responses-AC-1.oas2.txt
philsturgeon: oh damn, yeah, is this gavel just being too friendly? or is gavel not used here? 

Did we ever decide on how we'd implement the gavel-based expect, and the more strict expect?

test-harness/specs/default_responses-AC-1.oas3.txt
```diff
@@ -0,0 +1,26 @@
...
+      responses:
+        default:
+          description: default response
+          content:
+            application/json:
+              example: test
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos
```
philsturgeon: Out of interest, why do we always have -i when we dont actually care about response headers? Sometimes we care about headers, but sometimes we dont, should we maybe remove it from tests where we dont care?

test-harness/specs/default_responses-AC-3.oas3.txt
```diff
@@ -0,0 +1,27 @@
...
+    get:
+      responses:
+        200:
+          content:
+            application/json:
+              example: ok
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos?__code=304
```
philsturgeon:
```suggestion
curl -i http://localhost:4010/todos?__code=304
```

GET is the default header so we never need to specify it.

test-harness/specs/default_responses-AC-1.oas3.txt
XVincentX: Without the `-i` flag the library we're using to parse and compare the output (http-string-parser) fails. It's outlined in the README of the harness

test-harness/specs/default_responses-AC-5.oas2.txt
XVincentX: @philsturgeon I'm struggling here. What would be the content type for this response?

XVincentX:
@philsturgeon I think the response in this case should be `text/plain`, since we have *no* idea what to return.

I've modified the tests and the code, this should now be good to go.

test-harness/specs/default_responses-AC-1.oas3.txt
philsturgeon: The question [here](https://github.com/stoplightio/prism/pull/465#discussion_r304365627) also applies here 

We talked about having one gavel based expect and one literal string match. We wouldn’t need to http parse if it’s a literal string match. Remember that not all expect will be an entire HTTP Response, because not every single command is going to be a pure curl request.

XVincentX: I think I was wrong and we can simply *not* use the `-i` flag

philsturgeon: Is prism actually returning that literally, or was this written in incorrectly and gavel is comparing the two things and evaluating as true? 

I am asking how we can have literal expect and gavel based expect because I want o use literal expect in most instances to avoid this sort of confusion.

XVincentX: Prism is literally returning that value — I'd say it's a bug indeed, since such header (I think) is not valid.

I've fixed it in 974aab0 — but if so we might need to revisit some other tests here because two of them are really expecting */* as content type

XVincentX: > I am asking how we can have literal expect and gavel based expect because I want o use literal expect in most instances to avoid this sort of confusion.

`expect` = gavel validation + body matching (key and value)
`expect-loose`= gavel validation only

philsturgeon: Oh ok, thats a bit different from what I thought. I thought there was expect = strict and expect-loose would be gavel. Nevermind, another topic for another time. 

No test should expect `*/*` so lets delete them or change them.

XVincentX: They've been fixed — the idea is that if we have such situation we return text/plain

PR #468, opened 2019-07-17
test: additional examples for default responses

### LINKED ISSUE #413: Test Harness: Default Responses

lag-of-death:
This adds failing examples for #413. The passing ones are in https://github.com/stoplightio/prism/pull/465.

Closes #413


XVincentX:
@philsturgeon 

Comment on these failing tests: according to the code this is failing “on purpose”. The ability to look for the closest response is implemented only for successful responses; so you can do 2XX, and you’ll get the first one. Do you want to make the change and expand it to other responses as well, such as 4XX (which is the part of this test?)

packages/http/package.json
```diff
@@ -24,7 +24,7 @@
     "caseless": "^0.12.0",
     "faker": "^4.1.0",
     "fp-ts": "^1.19.3",
-    "json-schema-faker": "json-schema-faker/json-schema-faker",
+    "json-schema-faker": "XVincentX/json-schema-faker",
```
philsturgeon: Is this intentional?

XVincentX: Hmm, this is a leftover. Will fix.

philsturgeon:
@lag-of-death I had a go at fixing the conflicts but there is still a yarn.lock change floating around which I don't think should be there. These tests look great, if you can fix that I'll thumbs up and we're done here! 🎉

lag-of-death:
@phil, `yarn.lock` updated now

lag-of-death:
BLOCKED: https://github.com/stoplightio/prism/issues/413#issuecomment-513702133

test-harness/specs/default_responses-AC-8.oas3.txt
```diff
@@ -0,0 +1,22 @@
...
+      responses:
+        404:
+          description: no resource
+====server====
+mock -p 4010
+====command====
+curl -i -X GET http://localhost:4010/todos?__code=404
+====expect====
+HTTP/1.1 404 Not Found
+content-type: application/problem+json
```
XVincentX: @philsturgeon I think this is incorrect. If there's a response with no content, we should return an empty text/plain, right?

PR #470, opened 2019-07-17
Restore AC failing tests

XVincentX:
The tests failing here are all on OpenAPI 2.0 — the problem is that our http spec translator is not correctly creating an example per each content type, but rather a single example with multiple keys inside.


XVincentX:
This is the fix we need: https://github.com/stoplightio/http-spec/pull/2

packages/http/src/mocker/negotiator/__tests__/InternalHelpers.unit.ts
```diff
@@ -0,0 +1,27 @@
...
+        ],
+      };
+
+      it('should respect the q parameter', () => {
+        const response = findBestHttpContentByMediaType(avaiableResponses, [
+          'application/json;q=0.8',
+          'application/xml;q=1',
+        ]);
+
+        expect(response).toBeDefined();
```
philsturgeon: This seems redundant? undefined will not have properties so the next line will cover it?

packages/http/src/mocker/negotiator/__tests__/InternalHelpers.unit.ts
```diff
@@ -0,0 +1,27 @@
...
+describe('InternalHelpers', () => {
+  describe('findBestHttpContentByMediaType', () => {
+    describe('with multiple content types for a response', () => {
+      const avaiableResponses = {
+        code: '200',
+        contents: [
+          {
+            mediaType: 'application/xml',
+          },
+          { mediaType: 'application/json' },
```
philsturgeon: Funny formatting

XVincentX:
@philsturgeon @lag-of-death Boom! Graphite updated, and tests are now green!

PR #475, opened 2019-07-18
test: add query/headers/path parameters e2e tests

### LINKED ISSUE #428: E2E OAS2/3 Query, Header & Path Parameters
**User Story**

As a developer I want to have confidence that Prism validates query / path / header parameters.

**ACs**
- link that bug to this card 

**Additional resources**

**Resources**

Go through the “OpenAPI is Weird” talk and take every single example and use it as a test harness. 

 

**Acceptance Criteria**
- We should make sure request validation works for every type of parameter: query, path, header
- Missing required header/path item/query param = error
- Missing optional header / query string = ok
- Type mismatch header / path item / query string = error
- Format mismatch header / path item / query string = error
- Complex validation rules for header / path item / query string = error
- We should make sure response mocking works for headers, with every type of example/examples in every location they can be (consult the OpenAPI slideshow for all examples) (will be handled in SO-283)

**Example tests**

And the operation has a required path segment 
And in the request I sent that segment is missingI should get a 400 response back
And the payload of the response should be in application/problem+json content-type

And the operation has a required query param
And in the request I sent that query param is missing
Then I should get a 400 response back
And the payload of the response should be in application/problem+json content-type

And the operation has a required query param 
And in the request I sent that query param is provided
And the provided query param is of different data type then required (described in its schema)
Then I should get a 422 response back
And the payload of the response should be in application/problem+json content-type

And the operation has a required header
And in the request I sent that header is missing
Then I should get a 400 response back
And the payload of the response should be in application/problem+json content-type

And the operation has a required header
And in the request I sent that header is provided
And the provided header is of different data type then required (described in its schema)
Then I should get a 422 response back
And the payload of the response should be in application/problem+json content-type

**OAS3**

And the operation has a required query param
And in the request I sent that param is present
And the param I sent was not valid style-wise (ie form instead of pipeDelimited)
Then I should get a 422 response back
And the payload of the response should be in application/problem+json content-type

lag-of-death:
This PR adds tests to cover scenarios from #428 

Closes #428

### DIFF
--- a/test-harness/specs/parameters-ac1.oas2.txt
+++ b/test-harness/specs/parameters-ac1.oas2.txt
@@ -1,0 +1,28 @@
+====test====
+When I send a request to an operation
+And the operation has a required path segment
+And in the request I sent that segment is missing
+I should get a 400 response back
+And the payload of the response should be in application/problem+json content-type
+====spec====
+swagger: '2.0'
+paths:
+  /pet/a/{b}:
+    get:
+     parameters:
+       - in: path
+         name: b
+         required: true
+         type: number
+     responses:
+       '200':
+         description: ok
+         schema:
+           type: string
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet/a
+====expect====
+HTTP/1.1 400 Bad Request
+content-type: application/problem+json
--- a/test-harness/specs/parameters-ac1.oas2.txt
+++ b/test-harness/specs/parameters-ac1.oas2.txt
@@ -2,7 +2,7 @@
 When I send a request to an operation
 And the operation has a required path segment
 And in the request I sent that segment is missing
-I should get a 400 response back
+I should get a 404 response back
 And the payload of the response should be in application/problem+json content-type
 ====spec====
 swagger: '2.0'
@@ -24,5 +24,5 @@ mock -p 4010
 ====command====
 curl -sIXGET http://127.0.0.1:4010/pet/a
 ====expect====
-HTTP/1.1 400 Bad Request
+HTTP/1.1 404 Not Found
 content-type: application/problem+json
--- a/test-harness/specs/parameters-ac1.oas2.txt
+++ b/test-harness/specs/parameters-ac1.oas2.txt
@@ -9,15 +9,16 @@ swagger: '2.0'
 paths:
   /pet/a/{b}:
     get:
-     parameters:
-       - in: path
-         name: b
-         required: true
-     responses:
-       '200':
-         description: ok
-         schema:
-           type: string
+      parameters:
+        - in: path
+          name: b
+          required: true
+          type: number
+      responses:
+        '200':
+          description: ok
+          schema:
+            type: string
 ====server====
 mock -p 4010
 ====command====
--- a/test-harness/specs/parameters-ac2.oas2.txt
+++ b/test-harness/specs/parameters-ac2.oas2.txt
@@ -1,0 +1,28 @@
+====test====
+When I send a request to an operation
+And the operation has a required query param
+And in the request I sent that query param is missing
+Then I should get a 400 response back
+And the payload of the response should be in application/problem+json content-type
+====spec====
+swagger: '2.0'
+paths:
+  /pet:
+    get:
+      parameters:
+        - in: query
+          name: age
+          required: true
+          schema:
+            type: number
+      responses:
+        '200':
+          schema:
+            type: string
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet
+====expect====
+HTTP/1.1 400 Bad Request
+content-type: application/problem+json
--- a/test-harness/specs/parameters-ac2.oas2.txt
+++ b/test-harness/specs/parameters-ac2.oas2.txt
@@ -13,8 +13,10 @@ paths:
         - in: query
           name: age
           required: true
+          type: number
       responses:
-        '200':
+        200:
+          description: ok
           schema:
             type: string
 ====server====
--- a/test-harness/specs/parameters-ac2.oas2.txt
+++ b/test-harness/specs/parameters-ac2.oas2.txt
@@ -2,7 +2,7 @@
 When I send a request to an operation
 And the operation has a required query param
 And in the request I sent that query param is missing
-Then I should get a 400 response back
+Then I should get a 422 response back
 And the payload of the response should be in application/problem+json content-type
 ====spec====
 swagger: '2.0'
@@ -24,5 +24,5 @@ mock -p 4010
 ====command====
 curl -sIXGET http://127.0.0.1:4010/pet
 ====expect====
-HTTP/1.1 400 Bad Request
+HTTP/1.1 422 Unprocessable Entity
 content-type: application/problem+json
--- a/test-harness/specs/parameters-ac2.oas2.txt
+++ b/test-harness/specs/parameters-ac2.oas2.txt
@@ -13,8 +13,6 @@ paths:
         - in: query
           name: age
           required: true
-          schema:
-            type: number
       responses:
         '200':
           schema:
--- a/test-harness/specs/parameters-ac4.oas2.txt
+++ b/test-harness/specs/parameters-ac4.oas2.txt
@@ -13,8 +13,10 @@ paths:
         - in: header
           name: x-age
           required: true
+          type: number
       responses:
-        '200':
+        200:
+          description: ok
           schema:
             type: string
 ====server====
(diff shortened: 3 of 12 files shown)


lag-of-death:
@XVincentX, wanna take a look? Please mind that, the tests don't do any response body comparison. cc @philsturgeon

test-harness/specs/parameters-ac1.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+          type: number
+      responses:
+        '200':
+          description: ok
+          schema:
+            type: string
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet/a
```
philsturgeon: What is `-sIXGET ` doing? This looks like a mistake?

XVincentX: I think is:

`-I`: show *only* the headers, and not the body
`-s`: do not show progress messages
`-X` specify the verb

test-harness/specs/parameters-ac4.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+        200:
+          description: ok
+          schema:
+            type: string
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet
+====expect====
+HTTP/1.1 400 Bad Request
```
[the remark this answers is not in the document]
lag-of-death: @XVincentX, seems like 400 to me since the requests was not syntactically correct (it was missing the required header). So it was lacking in the format required for the request. But I'm not so sure. Basically, for the status codes in these tests I followed what was established in ACs (and what I also thought were the correct ones to use). Having said that, If you believe it should be 412 I will change it, since I don't have firm belief it should be 400/422

test-harness/specs/parameters-ac1.oas2.txt
lag-of-death: basically I could just send a `HEAD`, but thought it would be more natural to keep `get` in the spec

test-harness/specs/parameters-ac4.oas2.txt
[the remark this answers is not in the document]
XVincentX: Yeah fair enough, I'm fine with 400 as well, I was just wondering. @philsturgeon — got any thoughts on this?

test-harness/specs/parameters-ac1.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+====expect====
+HTTP/1.1 400 Bad Request
```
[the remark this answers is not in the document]
lag-of-death: thinking out-loud here: if the path segment is required, and it is not provided in the request, does it still make the request correct syntax wise? Not so sure. Or maybe it should be 404?

test-harness/specs/parameters-ac1.oas2.txt
philsturgeon: I am impressed that curl is able to parse that GET out of there, and not confuse it for another switch. We talked yesterday about how we don't need `-X GET` ever because it is default.

test-harness/specs/parameters-ac4.oas2.txt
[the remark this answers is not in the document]
philsturgeon: @XVincentX i think this is a misunderstanding of what 412 is about. Precondition failed means something like, for example, you tried to use a `If-Unmodified-Since` or `If-None-Match` on a POST, and the server told you to get lost because that's daft.

test-harness/specs/parameters-ac1.oas2.txt
XVincentX: BTW a lot of UNIX software is able to understand "glued" parameters

`-I test` = `-Itest`. I'm curious about why such thing.

test-harness/specs/parameters-ac1.oas2.txt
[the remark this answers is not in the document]
philsturgeon: > @philsturgeon I guess the reason we're not returning 422 is because this is not related to the payload, but to the query/headers, right?

This is a false distinction I think that is caused by OpenAPI treating a lot of different types of thing as a parameter. Path parameters are part of the path, and if you enter a path which does not exist, you should get a 404. 

So, if you request `/a` but the path was expecting `/a/b` then that should be a 404. 

If you request `/a/foo` but the operation parameters have validation rules saying foo it should have been `format: uuid`, then that should either be 422 (validation failure) or 404 (you entered a nonsense path), but there is nothing here that relates to syntax or 400.

test-harness/specs/parameters-ac1.oas2.txt
philsturgeon: Yep my last message was saying two things:

1. cool and I learned a thing
2. we dont need -X GET ever because it is the default

test-harness/specs/parameters-ac1.oas2.txt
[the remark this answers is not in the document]
XVincentX: @philsturgeon So if you look at this specific test — 404 is the status code that should be returned, thous the harness expectation is wrong. Is that correct?

test-harness/specs/parameters-ac6.only_oas3.txt
```diff
@@ -0,0 +1,34 @@
...
+      responses:
+        '200':
+          description: ok
+          content:
+            application/json:
+              example: "hello"
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet?name=abc,xyz
```
XVincentX: @philsturgeon @lag-of-death Why is this supposed to fail? The `pipeDelimited` serializer can still correctly serialize this query string.

`"abc,xyz".split('|') = abc,xyz`

Since the schema is just requesting an array of string, the final result is `['abc,xyz']` which is still a valid array of strings.

test-harness/specs/parameters-ac1.oas2.txt
XVincentX: Apparently when using -s the verb is mandatory, otherwise curl sends a POST. I've restored the command.

test-harness/specs/parameters-ac2.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+        200:
+          description: ok
+          schema:
+            type: string
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://127.0.0.1:4010/pet
+====expect====
+HTTP/1.1 400 Bad Request
```
XVincentX: In order to make this happen some work is going to be required @philsturgeon 

Fundamentally now our validation system returns 422 — for whatever error it encounters. We will need to pop error informations from inside instead of catching all of them and then return just 422. It should be viable in a day (or two, if something goes wrong).

The problem _could be_ the returned error message highly depends on the validation order.

I am assuming we'll do

Path Validation (Router) — Headers Validation — Body Validation.

test-harness/specs/parameters-ac1.oas2.txt
philsturgeon: What an interface 🤦🏻‍♂️ Thank you :)

test-harness/specs/parameters-ac6.only_oas3.txt
philsturgeon: Ooo good point. Could we add "minItems: 2" to the schema so we get an error if there is only one item?

test-harness/specs/parameters-ac1.oas2.txt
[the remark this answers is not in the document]
philsturgeon: 404 if the URL is missing segments because that would be a non existent URL and that’s what 404 is all about. 

If for now everything is doing 422 that’s ok for now. I’m not sure where the 400 came from that’s not right.

test-harness/specs/parameters-ac2.oas2.txt
philsturgeon: Thread glue: https://github.com/stoplightio/prism/pull/475#discussion_r305303199

It’s ok for now if everything is consistently a 422, I’ll accept that. I will make an issue to have certain cases be a 404

test-harness/specs/parameters-ac1.oas2.txt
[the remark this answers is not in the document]
XVincentX: It's not always 422 — we'll get 404 in cause of a missing path parameter, because it's an error thrown from the router component, which only emits 404 at the moment.

philsturgeon:
That sounds great then. 404 if missing and 422 if invalid is exactly what we want. Ship it!

--
Phil Sturgeon
@philsturgeon

Issue #407, opened 2019-07-06
Public JS API Improvements


brianmrock:
Prism currently depends on Graphite, which is public only because it has to be. We'd like to make that private again, and free Prism from its Graphite reliance. 

**Background**

There was interest in removing the loader, but now that we are about to release v3.0 we cannot just rip it out. Let’s add an other way to do things, which supports passing in an array of http-spec objects, and keep the loader in place as a convenience method. 

We can deprecate it later and remove, but we do not need to make decisions about that now.

This is a great chance for us to make a new and improved `createInstance` function in the core factory. Our instance has a whole lot of options, and if anyone has ideas on how to make it a little nicer for actual end users, now would be the time. Maybe we can seek inspiration from other mocking tools, or other popular TypeScript pacakges.

**Details**

1. prism reads the target openapi file itself
2. prism calls `@stoplight/json` or `@stoplight/yaml` parse to parse file contents
3. prism passes parsed content into `@stoplight/json-ref-resolver` to create a resolved version
4. prism passes that resolved version into `@stoplight/http-spec` transformOperations func to get a list of IHttpOperation objects

XVincentX:
Note: as part of the improvement, make sure to display a nice error message in case the target file or URL fails the fetching operation.

chris-miaskowski:
Comment on the implementation details @XVincentX . If we're removing the whole loading logic from Graphite we should wrap it up nicely in a lib so that UI utils like Request Maker don't have to repeat the same code.

philsturgeon:
Removing Graphite was only one aspect of this, we were going to improve the JS API in general, not _just_ remove Graphite.

XVincentX:
Ok, but in that case the issue is lacking the details.

philsturgeon:
I'll start a new issue with specific implementation ideas. Leaving it open ended didn't work out.

PR #476, opened 2019-07-18
adios oclif

### LINKED ISSUE #461: Switch from oclif to something else
**User story.**
As a Stoplight developer, I do not want to keep fighting agianst oclif odities with TS and packaging. 

One of many problems:

```
error TS5055: Cannot write file 'packages/http/dist/index.d.ts' because it would overwrite input file.
```

Here is another issue from spectral.

```
$ spectral --version
(node:30001) [MODULE_NOT_FOUND] Error Plugin: @stoplight/spectral: Cannot find module 'globby'
module: @oclif/config@1.13.0
task: not loading commands, globby not found
plugin: @stoplight/spectral
root: /Users/phil/.asdf/installs/nodejs/10.15.3/.npm/lib/node_modules/@stoplight/spectral
See more details with DEBUG=*
@stoplight/spectral/4.0.0-beta1 darwin-x64 node-v10.15.3
```

Use [commander.js](https://github.com/tj/commander.js) for the command framework.

- No more oclif
- No breaking changes to the CLI (e.g: its ok if help page looks a bit different)
- Lint command is still tested, coverage going up would be nice

**Additional context**

Spectral is also planning to do the same. https://github.com/stoplightio/spectral/issues/325

XVincentX:
Removes oclif in favour of Yargs

* No more tsnode and tslint dependencies in our published package
* No more 2000 dependencies to install
* Faster in loading because there's no auto mechanism of understanding if it's javascript or typescript project
* Remove hacks in tsconfig that was causing sometimes VSCode to go crazy yelling you to change your job
* Supports argument validation (if port is a string you get an error message and we do not have to validate it on our own)
* Supports the `involve` flag, which is going to be really useful for validation

Closes #461


packages/cli/README.md
```diff
@@ -1,68 +1,8 @@
...
-ARGUMENTS
-  SPEC  Path to a spec file. Can be both a file or a fetchable resource on the web
-
-OPTIONS
-  -d, --dynamic       Dynamically generate examples.
-  -h, --host=host     [default: 127.0.0.1] Host that Prism will listen to.
-  -m, --multiprocess  Fork the http server from the CLI
-  -p, --port=port     (required) [default: 4010] Port that Prism will run on.
-```
-<!-- commandsstop -->
```
XVincentX: In case you're wondering why — this (redundant) part was generated from oclif, I guess we can simply not have it anymore and send the user back to the CLI help.

packages/cli/bin/run.cmd
```diff
@@ -1,3 +0,0 @@
-@echo off
-
-node "%~dp0\run" %*
```
XVincentX: In case you're wondering, npm automatically generates the cmd version for Windows when you install the software.

P0lip:
I'm all in for using `yargs`.
I used it on a number of occasions and it never disappointed me.
It might not be the most feature-complete tool on the market yet it does come with everything you need to implement a robust command-line interface. We don't need more than that, do we?

marbemac:
Re auto lookup commands in the directory - can we please not do this? This sounds like oclif, and is too much magic IMHO. It also breaks certain common patterns, like bundling.

What it's "solving" is just us having to add a few lines to the `cli/src/index.ts` file right?

XVincentX:
@marbemac I'm all for it! I intentionally tried to stay close as possible to the old oclif mechanism but if you're ok getting rid of it — sure!

philsturgeon:
Is it meant to show this output? 

Screen Shot 2019-07-23 at 13 58 20

philsturgeon:
Same for subcommands 

image

philsturgeon:
Some of our param descriptions have full stops and some dont, lets pick a convention and stick to it

philsturgeon:
Loading a non-existant description document (nope.yaml) causes a whole lot of unexpected output. 

Screen Shot 2019-07-23 at 14 00 39

XVincentX:
1 & 2: This is part of the yargs output and it's intentional, it's telling that the positional and required arg is missing.
3: good point, I'll add the period everywhere
4. This has nothing to do with oclif/yargs — you can see the thrown exception is from the file loader which is going to die soon anyway

philsturgeon:
1 & 2: can we get rid of it? People pretty regularly type in the subcommand to get help on how the thing works, i dont think we want to shout at them?

4. could is that #407 and if so could you add notes about this in there? if not could we make an issue to track? 

I found more odd things its tough to know whats broken now and what was already broken, I guess I should make new issues for all new things?

XVincentX:
1 & 2: removed in 1dd0f4a
4. Issue updated.

> I found more odd things its tough to know whats broken now and what was already broken, I guess I should make new issues for all new things?

Just drop the bombs as issues and I'll go through them.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,12 @@
+import * as yargs from 'yargs';
+
+import mockCommand from './commands/mock';
+
+const _v = yargs
+  .scriptName('prism')
+  .version(false)
```
philsturgeon: Does this mean prism --version will not be an option? 

Currently we have: 

```prism --version
@stoplight/prism-cli/3.0.0-alpha.1 darwin-x64 node-v10.15.3```

(ha im on a very old version).

XVincentX: Very good catch. Fixed in 982af0c

XVincentX:
@philsturgeon Can you give it another try when you've got some time?

PR #477, opened 2019-07-18
test: add e2e tests for static examples

### LINKED ISSUE #414: E2E Static Examples
**User Story**

As a developer/Prism user I want to have control over how a response is generated (either with a static example or a dynamically generated payload).

**ACs**
- link that bug to this card 

**Additional resources**

**Example tests**

*NOTE: Prism2 doesn't have --dynamic CLI option**Running Prism with --dynamic only refers to Prism3*

JSON Schema Example
```
{
              type: 'object',
              properties: {
                pet: {
                  type: 'object',
                  properties: {
                    name: { type: 'string', examples: ['Clark'] },
                    middlename: { type: 'string', examples: ['J'], default: 'JJ' },
                  },
                },
              },
            }
```
**Example Tests**

And Prism was run with --dynamic
And this operation has an example
And I include ?__dynamic=false in the path
Then the static example should be returned in the payload

**Specific example**

Consider the following spec (example.oas3.yaml)
```
openapi: 3.0.2
paths:
  /todos:
    get:
      parameters:
        - name: title
          in: query
          style: form
          schema:
            enum:
              - eat
              - drink
      responses:
        200:
          description: Get Todo Items
          content:
            "application/json":
              schema:
                type: object
                properties:
                  title:
                    type: string
                    examples:
                      - Work
                      - Rest
                  description:
                    type: string
                  priority:
                    type: number
                    default: 0
```
Given 

prism mock example.oas3.yaml

When 

curl localhost:4010/todos

Expect output to be 

{"title":"Work","description":"string","priority":0}

Given 

prism mock basic.oas3.yaml --dynamic

When 

curl localhost:4010/todos?__dynamic=false

Expect 

{"title":"Work","description":"string","priority":0}

lag-of-death:
Add tests for static examples - closes #414

### DIFF
--- a/test-harness/specs/static_examples-AC-1.oas2.txt
+++ b/test-harness/specs/static_examples-AC-1.oas2.txt
@@ -1,0 +1,42 @@
+====test====
+Given
+
+Prism is started without --dynamic flag.
+
+When
+
+curl localhost:4010/todos
+
+Expect output to be
+
+{"title":"Work","description":"string","priority":0}
+====spec====
+swagger: "2.0"
+info:
+  title: 'title'
+  version: '1.0'
+paths:
+  /todos:
+    get:
+      produces:
+        - application/json
+      responses:
+        200:
+          description: Get Todo Items
+          schema:
+            type: object
+            properties:
+              title:
+                type: string
+                example: Work
+              description:
+                type: string
+              priority:
+                type: number
+                default: 0
+====server====
+mock -p 4010
+====command====
+curl http://localhost:4010/todos
+====expect====
+{"title":"Work","description":"string","priority":0}
--- a/test-harness/specs/static_examples-AC-1.oas2.txt
+++ b/test-harness/specs/static_examples-AC-1.oas2.txt
@@ -12,9 +12,6 @@ Expect output to be
 {"title":"Work","description":"string","priority":0}
 ====spec====
 swagger: "2.0"
-info:
-  title: 'title'
-  version: '1.0'
 paths:
   /todos:
     get:
--- a/test-harness/specs/static_examples-AC-1.oas2.txt
+++ b/test-harness/specs/static_examples-AC-1.oas2.txt
@@ -1,15 +1,6 @@
 ====test====
-Given
-
-Prism is started without --dynamic flag.
-
-When
-
-curl localhost:4010/todos
-
-Expect output to be
-
-{"title":"Work","description":"string","priority":0}
+When prism is started without the --dynamic flag,
+responses should be built with examples from the spec.
 ====spec====
 swagger: "2.0"
 paths:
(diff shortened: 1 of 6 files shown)


test-harness/specs/static_examples-AC-1.oas3.txt
```diff
@@ -0,0 +1,52 @@
...
+
+curl localhost:4010/todos
+
+Expect output to be
+
+{"title":"Work","description":"string","priority":0}
+====spec====
+info:
+  title: Sample API
+  version: 1.0.0
```
XVincentX: You've been skipping this in the other documents, right? Maybe we can remove it from here too?

test-harness/specs/static_examples-AC-1.oas3.txt
```diff
@@ -0,0 +1,52 @@
...
+openapi: 3.0.2
+paths:
+  /todos:
+    get:
+      parameters:
```
XVincentX: I think the parameter here is not required, we can remove it.

test-harness/specs/static_examples-AC-2.oas3.txt
```diff
@@ -0,0 +1,52 @@
...
+====spec====
+info:
+  title: Sample API
+  version: 1.0.0
+openapi: 3.0.2
+paths:
+  /todos:
+    get:
+      parameters:
+        - name: title
```
XVincentX: Dtto, can we remove this part?

test-harness/specs/static_examples-AC-2.oas3.txt
```diff
@@ -0,0 +1,52 @@
...
+
+curl localhost:4010/todos?__dynamic=false
+
+Expect
+
+{"title":"Work","description":"string","priority":0}
+====spec====
+info:
+  title: Sample API
+  version: 1.0.0
```
XVincentX: Yeah I think we can remove this

test-harness/specs/static_examples-AC-1.oas2.txt
```diff
@@ -0,0 +1,42 @@
+====test====
+Given
```
philsturgeon: I don't understand why we would do this? This seems to be duplicating information from the rest of the file, and I am not sure what benefit it serves? 

Yesterday we talked about simplifying the `==test==` content by making it more human, less gherkin, and avoiding duplication with the test file. This seems to have made it less human, more gherkin, increased duplication, and added some blank lines in for unknown reasons. 

> When prism is started with the --dynamic flag, responses should continue to match expected document structure

Done :)

lag-of-death:
@XVincentX, I cleaned up the specs and also fixed some false positives. Turns out, that when there was no `-i` in `curl something`, then, the bodies from https://github.com/stoplightio/prism/blob/21ba780a85ef20a9313286dd511af18965ff591d/test-harness/index.ts#L62 were empty

XVincentX:
@lag-of-death Ok that's the part I should check out in the harness, apparently when no -i the body is…gone. In meantime I've added the flag to all the missing tests so we tests the things :)

PR #478, opened 2019-07-18
docs: better demo

XVincentX:
Still not great, but at least this one has a border


philsturgeon:
I cant tell what the first few commands are, everything is scrolling by way too quickly. Maybe we could remove some paths so the output is shoved off the top, then add the ?__example manually so people can see the typing.

XVincentX:
Yeah as said I agree is bad — this is just the same with the overlay border. I'll probably re-record it while on the plane which I'm about to take :)

PR #479, opened 2019-07-18
test: add e2e tests for oas2 dynamic json responses

### LINKED ISSUE #416: Test Harness: Dynamic JSON Responses
**User Story**

As a Prism user, I can request a dynamic JSON response, so that I don't have to add static examples to a spec.

**ACs**
- link that bug to this card

**Additional resources**

**Test examples**

 NOTE: It's probably not only about application/json, but also about its derivatives too (like application/vnd.api+json). 

Given there are no examples defined for an operation
And this operation can produce application/json response
And this operation response defines a JSON Schema (schema field)
And I make a request to this operation
And the request has Accept header of application/json
Then I should get a response with dynamically generated JSON as payload

Given there are examples defined for an operation
And this operation can produce application/json response
And this operation response defines a JSON Schema (schema field)
And I make a request to this operationAnd the request has Accept header of application/json
And I add __dynamic=true to the query path
Then I should get a response with dynamically generated JSON as payload

Note this last test is an example of testing dynamically generated headers

Given there is a dynamic JSON response generated for an operation
And the operation has headers defined
When I make a request to this operation
Then I should get back a JSON response having these headers generated and included in the response

lag-of-death:
Closes #416


lag-of-death:
@XVincentX, this is a **WIP** and the code quality shouldn't be really considered here (so don't get paranoid about these added `if`s etc, and, honestly, these `if`s as they are for now, might just give false positives), it's just about the concept. So the thing is, when there are dynamic responses, I look at filenames. If the files have `.dynamic-response.` or `.dynamic-headers` in their names, an appropriate checker/validator (`dynamicHandlers` in the code) is run. This was a low hanging fruit to have a POC, but maybe we should do it some other way? For example, we could have `====expect====` section be a command too (like `====command====`), so we can parse the response somehow and maybe even do the assertions there, not in `test-harness/index.ts`?

test-harness/index.ts
```diff
@@ -1,15 +1,39 @@
...
+import * as toJsonSchema from 'to-json-schema';
+import { parseSpecFile } from './helpers';
 
 jest.setTimeout(60000);
 
+const dynamicHandlers = [
+  {
+    shouldHandle: value => value.includes('.dynamic-response.'),
+    handle: (output, expected) => {
+      const schema = toJsonSchema(JSON.parse(expected.body));
```
XVincentX: Hmm, I see two problems here:

1. The generated JSON Schema will never be precise as the original one. If I'm returning

```json
{ "test": "hello" }
```

1. The library will probably generate a `type: "string"` and most likely miss other informations, such as this value been an enum or something like that
2. Since the schema is generated *from the payload* the check will pass, by definition. Therefore this check is kind of useless.

test-harness/specs/dynamic_json_responses-AC-1.dynamic-response.oas2.txt
```diff
@@ -0,0 +1,39 @@
...
+            type: object
+            properties:
+              title:
+                type: string
+              description:
+                type: string
+              priority:
+                type: integer
+====server====
+mock -p 4010
```
XVincentX: The dynamic example should be explicitly enabled
```suggestion
mock -d -p 4010
```

This was not catch because of the problem I exposed above: the payload will always pass the schema validation because the schema is generated from the payload.

philsturgeon:
In my head I thought that we’d have expect literal and expect gavel, and gavel would be doing what it does in dress where the body JSON is just matched up with being roughly the right data type, no more. 

So for static examples we’d be using the expect static, and for dynamic examples we’d be using expect gavel, then all of this would be solved? 

Checking that “is the dynamic property a valid IP address” or any other format based logic is too fine grained for these end to end tests and is best covered with unit and integration tests, so this should be ok?

--
Phil Sturgeon
@philsturgeon

XVincentX:
@philsturgeon  All right, if you're willing to loose a little bit the requirements we can make it work.

@lag-of-death We basically just need to leave the framework as it is and provide a decent object from the JSON schema, and that'll be it. We can then use `==expect-loose==` instead of `expect` so that only gavel checks will be performed, and not the strict checks.

test-harness/specs/dynamic_json_responses-AC-1.dynamic-response.oas2.txt
lag-of-death: @XVincentX, isn't `dynamic generation` enabled by default?

XVincentX: No, by default is not enabled.

philsturgeon:
Check diagram or docs when uncertain. 

--
Phil Sturgeon
@philsturgeon

test-harness/specs/dynamic_json_responses-AC-1.dynamic-response.oas2.txt
```diff
@@ -0,0 +1,39 @@
+====test====
+Prism generates dynamic values for schema when no static examples are provided.
+
+Given there are no examples defined for an operation
+And this operation can produce application/json response
+And this operation response defines a JSON Schema (schema field)
```
philsturgeon:
```suggestion
And this operation response defines a Schema Object
```

Let's use OpenAPI terminology here and call it a Schema Object. It's not JSON Schema so we should not say it is.

PR #480, opened 2019-07-19
test: add examples for manual example selection

### LINKED ISSUE #411: Test Harness: Manually Select Example
**User Story**

As a Prism CLI user, I want to specify an example in a request's path, so that I can get this example in the response's payload. 

**ACs**
- link that bug to this card 

**Additional resources**

**Test case examples**

Given that an operation has no examples
When I send a request like ?__example=examplename ie http://localhost:4010/pets/123?__example=dog
Then the payload I get back should describe the error
And I should get back 404 status code

**OAS3**:

Given that an operation defines named response examples (ie cat, dog from *SPEC1* below)
And one of the examples defined is examplename 
When I send a request like ?__example=examplename (ie http://localhost:4010/pets/123?__example=dog)
Then I should get back this example as the payload 
And I should get back 200 status code

Given that an operation defines one example (with example keyword)
When I send a request like ?__example=default (ie http://localhost:4010/pets/123?__example=default)
Then I should get back the example from example field as the payload
And I should get back 200 status code

Given that an operation does not define an example (with example keyword)
And this operation does not define examples neither
When I send a request like ?__example=mantis-shrimp (ie http://localhost:4010/pets/123?__example=mantis-shrimp)
Then the payload I get back should describe the error
And I should get back 404 status code

Given that an operation defines both named response examples and example
And there is examplename example defined
When I send a request like ?__example=examplename (ie http://localhost:4010/pets/123?__example=dog)
Then I should get back this example as the payload 
And I should get back 200 status code

**Example oas3 spec**
```
"content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Pet"
                },
                "examples": {
                  "cat": {
                    "summary": "An example of a cat",
                    "value": {
                      "id": 2,
                      "category": {
                        "id": 1,
                        "name": "Felis"
                      },
                      "tags": [
                        {
                          "id": 1,
                          "name": "pet"
                        }
                      ],
                      "name": "Fluffy",
                      "status": "available",
                      "photoUrls": []
                    }
                  },
                  "dog": {
                    "summary": "An example of a dog",
                    "value": {
                      "id": 1,
                      "category": {
                        "id": 2,
                        "name": "Canis"
                      },
                      "tags": [
                        {
                          "id": 2,
                          "name": "dog"
                        }
                      ],
                      "name": "Sharik",
                      "status": "available",
                      "photoUrls": []
                    }
                  }
                }
              }
            }
          },
```
**OAS2:**

Given that there is an operation in that provides application/json
And this operation provides examples field with an example for application/json
When I send a request like ? (ie http://localhost:4010/pets/123)  to this operation

Then I should get back this example as the payload
And I should get back 200 status code

Given that there is an operation in that provides application/json
And this operation provides examples field with an example for application/json
When I send a request like ?__example=default (ie http://localhost:4010/pets/123?__example=default)  to this operation
Then I should get back this example as the payload 
And I should get back 200 status code

Given that there is an operation in that provides application/json and application/xml
And this operation provides examples field with an example for application/xml
When I send a request like ?__example=default (ie http://localhost:4010/pets/123?__example=default)  to this operation
Then the payload I get back should describe the error
And I should get back 404 status code

**Notes**

Manually selecting an example should be already implemented for both Prism2 and Prism3, but OAS2 does not support naming examples and is therefore limited to default only. If you find a bug, please create a dedicated card for it and link that to SO-276.

lag-of-death:
Add failing examples for #411 
Successful ones are in https://github.com/stoplightio/prism/pull/481

Closes #411


XVincentX:
@philsturgeon There are two tests that are mutually exclusive — if one passes, the other will effectively fail; we need to sort out from a product perspective what we want to do with non existent `default` examples.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -39,8 +39,19 @@ const helpers = {
           mediaType,
           bodyExample: example,
         });
+      } else if (exampleKey === 'default' && httpContent.examples) {
+        return right({
```
philsturgeon: In an OAS3 context somebody might define a bunch of examples and the last one is default. This code would pick the first example instead of default, that’s not what we want. 

The feature is to support folks picking default regardless of OAS version. Http spec should be abstracting the differences, so that it doesn’t matter which version is coming in. 

We discussed this as one of two options to try but it looks like it won’t work. Instead of looking for “the first”, we probably want to look for one named default. If http spec needs to start labelling OAS2 examples with default then that’s an ok change with me.

XVincentX: > In an OAS3 context somebody might define a bunch of examples and the last one is default. This code would pick the first example instead of default, that’s not what we want.

This is incorrect. This code will trigger only if there's not an example named `default`; it's a fallback.

The point is:

1. We have a test that's checking out that if there's no default example and I ask for `default` we should return 404
2. Another test that's checking that if there's no default example and I ask for a `default` we should return the first example.

philsturgeon: Here's an example of what I mean:

```
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: Get Todo Items
          "content":
            application/json:
              example:
                id: 1
                name: turtle
              examples:
                cat:
                  summary: An example of a cat
                  value:
                    id: 2
                    name: cat
```

This should be identical to:

```
swagger: "2.0"
paths:
  /todos:
    get:
      produces:
        - application/json
      responses:
        200:
          description: Get Todo Items
          examples:
            application/json: 
              id: 1
              name: turtle
            cat: # this is the awkward hack discussed elsewhere
              id: 2
              name: cat
```

Ignoring the hack which we've employed in older tooling to give OAS2 named examples despite the spec not actually supporting it, the example named `application/json` should be treated the same as an in OAS3 under application/json and using the `example` keyword.

Does that make sense?

philsturgeon: > Another test that's checking that if there's no default example and I ask for a default we should return the first example.

Where is this requirement coming from? And what is the use case where asking for default should return first?

XVincentX: I do not know :trollface: — I found it there and it's not passing, so I was wondering why's that happening


```
====test====
	Requesting a default example results in sending back the first and only example for an operation.
	====spec====
	swagger: "2.0"
	paths:
	  /todos:
	    get:
	      produces:
	        - application/json
	      responses:
	        200:
	          description: Get Todo Items
	          examples:
	            application/json: "hello"
	====server====
	mock -p 4010
	====command====
	curl -i -X GET http://localhost:4010/todos?__example=default
	====expect====
	HTTP/1.1 200 OK
	
	"hello"
```

XVincentX: There are so many things here 😢 

1. The OAS3 document you posted is actually invalid.

> The example object is mutually exclusive of the examples object.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#mediaTypeObject

Assuming this is not the issue

```yaml
swagger: "2.0"
paths:
  /todos:
    get:
      produces:
        - application/json
        - application/xml
      responses:
        200:
          description: Get Todo Items
          examples:
            application/json: 
              id: 1
              name: turtle
            application/xml: 
              id: 1
              name: turtle
            cat: # this is the awkward hack discussed elsewhere
              id: 2
              name: cat
```

@philsturgeon What would be the returned example, in this case? This would be equivalent, in the OAS3 example, to have *two* `example` — which we treat as the "default" one

philsturgeon: Ok yep its not valid OAS3 syntax I'm trying to highlight a bigger point. Let's do it without the named bits:

```
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: Get Todo Items
          content:
            application/json:
              example:
                id: 1
                name: turtle
```

This should be identical to:

```
swagger: "2.0"
paths:
  /todos:
    get:
      produces:
        - application/json
      responses:
        200:
          description: Get Todo Items
          examples:
            application/json: 
              id: 1
              name: turtle
```

These should both be named "default", so ?__example=default should work the same in either.

The whole "just look for the first one" might get to the same result in this instance but it feels a bit like a hack to me, no?

what I'm getting at is the following should definitely not let ?__example=default return cat, it should be _the_ example for the mime type in question.

```
      responses:
        200:
          description: Get Todo Items
          examples:
            cat: 
              id: 2
              name: cat
            application/json: 
              id: 1
              name: turtle
```

OAS2 confuses names and mime types, and we kinda do with our hack to allow named examples, but we should not mess up default over that right?

philsturgeon: Or we just delete `?__example=default`. This whole conversation is making me wonder what the point of this feature even is?

XVincentX: I struggle to see a scenario where I would explicitly ask for a `default` example — the default (the way I see it) is a fallback when I can't find anything suitable.

We can keep the functionality if you think it's useful; I will need to modify the http spec to "rename" the examples matching the `produces` into `defaults` and it should be good to go.

Your call!

test-harness/specs/manually-select-examples-AC-0.oas2.txt
```diff
@@ -0,0 +1,23 @@
+====test====
+Trying to access a default example when no example is defined, fails with 404.
+====spec====
```
philsturgeon: If we’re removing default let’s remove default? Also let’s not have a test which requests default?

XVincentX: Right. This is a leftover, removed in 238e060

Issue #456, opened 2019-07-15
XML Keyword Support


philsturgeon:
**User story.**
As a OpenAPI v3.0 user I can use the `xml` keyword in all its glory, to support wrapped arrays, namespaces, prefixes, and everything else outlined in [Representing XML](https://swagger.io/docs/specification/data-models/representing-xml/).

**Is your feature request related to a problem?**
Prism is currently very JSON-centric and we should widen support for folks who use OpenAPI to describe XML APIs. 

**Describe the solution you'd like**

See if OpenAPI Sampler can handle [the `xml`openapi object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#xmlObject), and if not, look into adding support to it.

chohmann:
@philsturgeon going through to try to get a handle on open Prism issues. Is this one still valid?

philsturgeon:
Technically yes but nobody seems to be interested in it. It's not been mentioned since I wrote it, and it's not got any thumbs here, so let's close it. 

If anyone wants this feature or any other features, please make a feature request over here: https://roadmap.stoplight.io/tabs/7-under-consideration/submit-idea

XVincentX:
Yeah I agree. It seems like XML is really not a thing anymore.

PR #482, opened 2019-07-22
Tests/#408/oas23 body param validation

### LINKED ISSUE #408: Test Harness: OAS2/3 Body Param Validation
**User Story**

As a developer I want to have confidence that Prism's validates request payload (body).

**ACs**
- add at least the cases described in the examples section- 
- link that bug to this card 

**Additional resources**

**Stuff to test oas2/oas3**
- [x] required x-www-form-urlencoded param
- [x] required x-www-form-urlencoded body
- [x] required application/json body
- [x] invalid form-data type 
- [x] invalid `enum`
- [x] form-data array (csv, pipes, multi)
- [x] form-data formats (e.g. byte)
- [x] allowEmptyValue

**oas3 related**
- [x] form-data + schema
- [ ] encoding/contentType
- [ ] encoding/headers
- [ ] encoding/allowReserved
- [ ] encoding/style
- [ ] encoding/explode

**Not in scope**

multipart/form-data

**Example tests**

Given there is an operation, that can consume requests of Content-Type CTAnd it is required that the payload has X fieldWhen I send a POST request (with Content-Type of CT) to this operationAnd the request doesn't have his X required field in the payloadThen I should get back a 400 response back

Given there is an operation, that can consume requests of Content-Type CTAnd it is required that the payload provides X fieldWhen I send a POST request (with Content-Type of CT) to this operationAnd the request does have this X required fieldThen I should get back a 2xx response back

Given there is an operation, that can consume requests of Content-Type CTAnd the operation has no required field for the payloadWhen I send a POST request (with Content-Type of CT) to this operationThen I should get back a 2xx response back

When I send a POST request (with Content-Type of CT) to an operationAnd this operation does not consume requests of Content-Type CTThen I should get back a 415 response back

**Example Specs**

**Open API 2**

Consider this rather minimal oas2 spec. It has

- POST /path operation
    - that operation takes a body
    - that body is application/json (see consumes)
    - that body requires 'x' param of string type
- PUT /path operation
   - that operation takes body
   - that body is of formData type
   - it has a required X field that can take only one of the values "one" or "two"

This is just an example.
```
swagger: '2.0'
info:
  title: Open API 2 Sink
  version: '1.0'
host: 'localhost:3000'
schemes:
  - http
produces:
  - application/json
consumes:
  - application/json
paths:
  /path:
    post:
      summary: create path
      tags: []
      responses: {}
      parameters:
        - in: body
          schema:
            type: object
            properties:
              x:
                type: string
            required:
              - x
    put:
      summary: update
      tags: []
      responses: {}
      parameters:
        - in: formData
          type: string
          name: x
          required: true
          description: ''
          enum:
            - one
            - two
definitions: {}
```
**Open API 3 **

Similarly to the previous example here is OAS3 (note the difference in formData)
```
openapi: 3.0.0
info:
  title: Open API 3 Sink
  version: '1.0'
servers:
paths:
  /path:
    post:
      summary: create
      tags: []
      responses: {}
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                x:
                  type: string
              required:
                - x
    put:
      summary: update
      tags: []
      responses: {}
      requestBody:
        content:
            schema:
              type: object,
              properties:
                  name:
                    description: Updated name of the pet
                    type: string
                  status:
                    description: Updated status of the pet,
                    type: string
              required:
                - status
components:
  schemas: {}
```

chris-miaskowski:
## Checklist

_Tests_

## Notes

Some tests fail and I will create issues for them shortly.

Closes #408


packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+async function instantiatePrism(fixture: string) {
+  const server = createServer({}, { components: { logger }, config: { mock: { dynamic: false } } });
+  await server.prism.load({
+    path: relative(process.cwd(), resolve(__dirname, 'fixtures', fixture)),
+  });
+  return server;
+}
+
+const expectPayload = (response: fastify.HTTPInjectResponse) => {
+  const parsed = JSON.parse(response.payload);
```
lag-of-death: seems like `expectPayload` only depends on `response.payload`, so there's no need to pass the whole `response` object most likely

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+  return expect(parsed);
+};
+
+describe('body params validation', () => {
```
lag-of-death: hm.. these don't look like test-harness to me? maybe we could put these into another PR?

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+      server = await instantiatePrism(`operations-with-formdata-param.${oas}.yaml`);
+    });
+
+    describe('required parameter not in body', () => {
+      test('returns 422', async () => {
+        const response = await server.fastify.inject({
+          method: 'POST',
+          url: '/path',
+          payload: {},
+          // headers: {
```
lag-of-death: should it be removed now?

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+    describe('parameter does not match enum criteria', () => {
+      test('returns 422 & proper validation message', async () => {
+        const response = await server.fastify.inject({
+          method: 'POST',
+          url: '/path',
+          payload: {
+            id: 'not integer',
+            status: 'somerundomestuff',
+          },
+          // headers: {
```
lag-of-death: should it be removed now?

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+    describe('valid parameter provided', () => {
+      test('returns 200', async () => {
+        const response = await server.fastify.inject({
+          method: 'POST',
+          url: '/path',
+          payload: {
+            id: 123,
+            status: 'open',
+          },
+          // headers: {
```
lag-of-death: should it be removed now?

packages/http-server/src/__tests__/body-params-validation.spec.ts
philsturgeon: I think its ok, no need for more paperwork so long as all the tests are good.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+  afterAll(async () => {
+    await server.fastify.close();
+  });
+
+  describe.each([['oas2'], ['oas3']])('%s with body param', oas => {
+    beforeEach(async () => {
+      server = await instantiatePrism(`operations-with-body-param.${oas}.yaml`);
+    });
+
+    describe('operation with no consumes', () => {
```
philsturgeon: Can we talk about this in a more generic way? oas3 does not have "consumes"

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+            id: 'not integer',
+            status: 'somerundomestuff',
+          },
+          // headers: {
+          // 'content-type': 'application/x-www-form-urlencoded',
+          // },
+        });
+
+        expect(response.statusCode).toBe(422);
+        const parsed = JSON.parse(response.payload);
```
philsturgeon: expectPayload ?

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+
+async function instantiatePrism(fixture: string) {
+  const server = createServer({}, { components: { logger }, config: { mock: { dynamic: false } } });
+  await server.prism.load({
+    path: relative(process.cwd(), resolve(__dirname, 'fixtures', fixture)),
+  });
+  return server;
+}
+
+const expectPayload = (response: fastify.HTTPInjectResponse) => {
```
philsturgeon: I have some concerns here.

1. Payload is a generic term which often means request body but could mean request or response, but here it always means response.
2. Should this be an expect wrapper? could it be a custom matcher? 
3. this is always JSON only, we are meant to be JSON and XML at least, especially with #456 coming up this/next sprint

chris-miaskowski:
Thanks for reviews folks. I moved this PR to https://github.com/stoplightio/prism/pull/501 and merged with @XVincentX's graphite removal.

PR #484, opened 2019-07-22
Security scheme validation

### LINKED ISSUE #422: Test Harness: Security Scheme Validation
**User Story**

As a Prism CLI user, I want to secure my endpoints, so that I know the API is only used by authorized users.

**ACs**

add test cases to test harness

add at least the cases described in the examples section

if any of the test fails then mark that it is an invalid result (e.g. mark as skipped) and report a Jira bug

link that bug to this card 

Make sure to add the actual tests to TestRails

**Additional resources**

When operations have security schemes, how do they respond when no authentication information is being sent?

 

Try and write tests for every single authentication scheme in OpenAPI 2/3, and write tests for every different location those schemes security information can be in. 

API Key in header

API key in query

API key in cookie (ugh why)

HTTP in Header

HTTP in query…

etc

Try all OpenAPI v2 specs against Prism v2 to get a general idea for how it is meant to work, then try that stuff in Prism v3. 

**Acceptance Criteria and Scope**

Prism should throw errors if security criteria are not matched

If Prism does not respond the way it should (based on v2 or common sense) create a story for investigation and move on

### Example tests

Given a spec has no global security scheme
And an operation has no security requirements defined
When I send a request to this operation
Then I get a response with 200 status code

Given a spec has a global security scheme
And an operation has no security requirements defined
When I send a request to this operation
And I don't include any credentials in the request
Then I get back a response with 401 status code
And the response should have WWW-Authenticate header

Given a spec has multiple global security schemes
And an operation has no security requirements defined
When I send a request to this operation
And I don't include any credentials in the request
Then I get back a response with 401 status code
And the response should have WWW-Authenticate header

Given a spec has multiple global security schemes
And an operation has no security requirements defined
When I send a request to this operation
And I include valid credentials for one of the global security schemes
Then I get back a response with 200 status code

Given a spec has multiple global security schemes
And an operation has no security requirements defined
When I send a request to this operation
And I include invalid credentials for one of the global security schemes
Then I get back a response with 403 status code

Given a spec has one or multiple global security schemes
And an operation defines its security requirements by specifying one of the security schemes
And I send a request to this operation
And I include invalid credentials for this operation security requirements
Then I get back a response with 403 status code

Given a spec has one or multiple global security schemes
And an operation overrides the global security scheme by overwriting them (making the operation not secured, having its security requirements as [])
And I send a request to this operation
And I don't include any credentials in the request
Then I get back a response with 200 status code

### LINKED ISSUE #425: Security validation
Validate a incoming requests against security schemes, and react accordingly. If security validation fails, then try to mock back a 401 if defined, or throw a Prism error with a 401 explaining the situation.

Security validation is basically a case of looking at security scheme (check the OpenAPI specification for more information).

There are multiple schemes: "apiKey", "http", "oauth2", "openIdConnect". We can support only apiKey, http for now if oauth2 and openid look too complex.

Check in to see where to start looking, then check the name to see specifically what the header/query/body param is called. If that item does not exist, or is empty, then 401.

A secondary check should be added for bearerFormat if type == http, to see what the Authentication scheme is. OpenAPI defers to RFC 7235 (https://tools.ietf.org/html/rfc7235#section-5.1) which defers to IANA (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) for the list of supported schemes. We should not stop people adding custom schemes, but if a spec file has one of those schemes, we should make sure that the request is using that scheme.

So, if spec file has:
```
  "type": "http",
  "scheme": "bearer",
```
then Prism should look for 
```
Authorization: Bearer some-token
```
and that is valid, but if they send

```
Authorization: Digest some-token
```

Then that is a mismatch, and a 401 should be thrown.

The Prism error here should mention “Invalid security scheme used” in the error message.

lag-of-death:
Closes #425 
Closes #422 

## Info

This PR is a huge diff, but the implementation is just a minor part of it. What makes the diff huge is:

1. test-harness files
2. packages/http-server/src/__tests__/fixtures/petstore.no-auth.oas2.yaml
3. packages/http-server/src/__tests__/fixtures/petstore.no-auth.oas3.yaml
4. https://github.com/stoplightio/prism/blob/7374ffae85c3a5b3c93a89194248e49c024bf381/packages/core/src/utils/__tests__/security.spec.ts is unit tests.

`2.` and `3.` were added to have the already existing tests pass with minimal footprint. We might remove `2.` and `3.` and just have `petstore.oas2.yaml` and `petstore.oas3.yaml` adjusted instead.

### Feature implementation changes are in files:

- packages/core/src/factory.ts 
- packages/core/src/types.ts 
- packages/core/src/utils/security/handlers.ts
- packages/core/src/utils/security/index.ts
- packages/http-server/src/server.ts
- packages/http/src/mocker/HttpMocker.ts
- packages/http/src/mocker/negotiator/NegotiatorHelpers.ts


test-harness/specs/security-scheme-validation-AC-10.oas2.txt
```diff
@@ -0,0 +1,34 @@
+====test====
+Resource is returned when a bearer token is provided as a header.
```
lag-of-death: Assumed: it's OK to send any token to pass the security.
So no checking if the passed token is valid.

test-harness/specs/security-scheme-validation-AC-10.oas3.txt
```diff
@@ -0,0 +1,35 @@
+====test====
+Resource is returned when a bearer token is provided as a header.
```
lag-of-death: Assumed: it's OK to send any token to pass the security.
So no checking if the passed token is valid.

test-harness/specs/security-scheme-validation-AC-11.only_oas3.txt
```diff
@@ -0,0 +1,29 @@
+====test====
+Resource is returned when a bearer token is provided as a header.
```
lag-of-death: Assumed: it's OK to send any token to pass the security.
So no checking if the passed token is valid.

test-harness/specs/security-scheme-validation-AC-12.only_oas3.txt
```diff
@@ -0,0 +1,58 @@
+====test====
+After a user obtains a cookie from /login, s/he can then use the cookie to access protected endpoint /todos.
```
lag-of-death: Assumed: since it's relatively easy to create an endpoint that returns *Set-Cookie* header,
it would be OK to actually compare *Set-Cookie* content with what is sent later. So, basically, if
SESSIONID from /login response (*Set-Cookie*) and SESSIONID from a request to /todos (*Cookie*) match.

test-harness/specs/security-scheme-validation-AC-13_failing.only_oas3.txt
```diff
@@ -0,0 +1,56 @@
+====test====
+No access is granted when the provided session cookie is wrong.
```
lag-of-death: Assumed: since it's relatively easy to create an endpoint that returns *Set-Cookie* header,
it would be OK to actually compare *Set-Cookie* content with what is sent later. So, basically, if
SESSIONID from /login response (*Set-Cookie*) and SESSIONID from a request to /todos (*Cookie*) match.

test-harness/specs/security-scheme-validation-AC-5_failing.oas2.txt
```diff
@@ -0,0 +1,30 @@
+====test====
+Making a request (and including invalid credentials) to an endpoint,
```
lag-of-death: Assumed: since 403 examples ended up in the reviewed ACs, I came up with this. But not sure if should do so (?).
How would we know that dGVzdDp0ZXN0 is valid?

test-harness/specs/security-scheme-validation-AC-5_failing.oas3.txt
```diff
@@ -0,0 +1,33 @@
+====test====
+Making a request (and including invalid credentials) to an endpoint,
```
lag-of-death: Assumed: since 403 examples ended up in the reviewed ACs, I came up with this. But not sure if should do so (?).
How would we know that dGVzdDp0ZXN0 is valid?

test-harness/specs/security-scheme-validation-AC-6_failing.oas2.txt
```diff
@@ -0,0 +1,34 @@
+====test====
+Making a request (and not including valid credentials) to an endpoint,
```
lag-of-death: Assumed: since 403 examples ended up in the reviewed ACs, I came up with this. But not sure if should do so (?).
How would we know that dGVzdDp0ZXN0 is valid?

test-harness/specs/security-scheme-validation-AC-6_failing.oas3.txt
```diff
@@ -0,0 +1,35 @@
+====test====
+Making a request (and not including valid credentials) to an endpoint,
```
lag-of-death: Assumed: since 403 examples ended up in the reviewed ACs, I came up with this. But not sure if should do so (?).
How would we know that dGVzdDp0ZXN0 is valid?

test-harness/specs/security-scheme-validation-AC-8.oas2.txt
```diff
@@ -0,0 +1,29 @@
+====test====
+Resource is returned when an api-key is provided as a header.
```
lag-of-death: Assumed: no validation needed, just api key presence is enough for the security to pass.
Maybe the api-key should actually be validated somehow?

test-harness/specs/security-scheme-validation-AC-8.oas3.txt
```diff
@@ -0,0 +1,29 @@
+====test====
+Resource is returned when an api-key is provided as a header.
```
lag-of-death: Assumed: no validation needed, just api key presence is enough for the security to pass.
Maybe the api-key should actually be validated somehow?

test-harness/specs/security-scheme-validation-AC-9.only_oas3.txt
```diff
@@ -0,0 +1,28 @@
+====test====
+Resource is returned when a bearer token is provided as a header.
```
lag-of-death: Assumed: no validation needed, just token presence is enough for the security to pass.
Maybe the api-key should actually be validated somehow?

lag-of-death:
@XVincentX, @philsturgeon, @chris-miaskowski 
I've had some doubts about the tests and added some comments. Some of them are duplicated (mostly because of oas2/oas3 examples). The comments should be considered in the context of the whole file, that the comment is added to.

test-harness/specs/security-scheme-validation-AC-10.oas2.txt
philsturgeon: Yes

test-harness/specs/security-scheme-validation-AC-5_failing.oas2.txt
philsturgeon: Invalid can mean a lot of things I guess, this might have been lost in translation from notes to examples, and now be an impossible test. We want to make sure there’s tests for missing, or wrong type of token, or if it wants bearer and you don’t specify the bearer keyword, etc, but no the actual token is opaque and we don’t know what is valid or not.

test-harness/specs/security-scheme-validation-AC-2_failing.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+    scheme: basic
+security:
+  - BasicAuth: []
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://localhost:4010/todos
+====expect====
+HTTP/1.1 401 Unauthorized
+WWW-Authenticate: Basic realm="todos"
```
lag-of-death: I'm also not sure about `realm` thing. Maybe we could just always send back `realm="*"` for such cases? Here I just took the name of the operation

lag-of-death:
thanks @philsturgeon, I've adjusted the tests and they should be quite review-able at this point. One thing though I'm still not sure about is https://github.com/stoplightio/prism/pull/484#discussion_r306186564
@XVincentX, @chris-miaskowski, could you guys take a look? thanks! :)

test-harness/specs/security-scheme-validation-AC-12.only_oas3.txt
```diff
@@ -0,0 +1,39 @@
...
+    get:
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              example: test
+components:
+  schemas:
+    LoginRequest:
```
chris-miaskowski: Why is this necessary? Looks like you're not using it, correct?

test-harness/specs/security-scheme-validation-AC-13_failing.only_oas3.txt
```diff
@@ -0,0 +1,37 @@
...
+    get:
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              example: test
+components:
+  schemas:
+    LoginRequest:
```
chris-miaskowski: Probably can remove `schemas`

test-harness/specs/security-scheme-validation-AC-10.oas2.txt
```diff
@@ -0,0 +1,34 @@
+====test====
+Adding a required header of "Authorization: Bearer <<BEARER_TOKEN>>" results in granted access.
```
chris-miaskowski: Would be nice to have a test checking 'denied' access for oauth2. What do you think?

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
```diff
@@ -0,0 +1,31 @@
...
+security:
+  - BasicAuth: []
+  - ApiKeyAuth: []
+====server====
+mock -p 4010
+====command====
+curl -sIXGET http://localhost:4010/todos
+====expect====
+HTTP/1.1 401 Unauthorized
+WWW-Authenticate: Basic
```
chris-miaskowski: I'm not sure about that expectation. Reading
https://tools.ietf.org/html/rfc7235#section-4.1 looks like we may want to return more than one schema since we define two ways to authenticate (?)

test-harness/specs/security-scheme-validation-AC-2_failing.oas2.txt
chris-miaskowski: Let's do `"*"` or `"prism"`, this parameter is arbitrary and depends on how server implements it. If we ever support a plugin system for Prism this is a good candidate for a plugin.

test-harness/specs/security-scheme-validation-AC-6_failing.oas3.txt
```diff
@@ -0,0 +1,34 @@
...
+          description: OK
+          content:
+            application/json:
+              example: test
+components:
+  securitySchemes:
+    BasicAuth:
+      type: http
+      scheme: basic
+    ApiKeyAuth:
```
chris-miaskowski: Would be nice to test the `ApiKeyAuth` validation as well.

test-harness/specs/security-scheme-validation-AC-8.oas3.txt
```diff
@@ -0,0 +1,29 @@
+====test====
+Resource is returned when *X-API-Key* required header is provided.
```
chris-miaskowski: We should check for denied as well.

chris-miaskowski:
@lag-of-death Most of this looks legit, I would add a couple more tests but other than that great job!
If @philsturgeon doesn't have any other comments feel free to start implementing the actual validation :)

test-harness/specs/security-scheme-validation-AC-12.only_oas3.txt
lag-of-death: right! thanks, this is a leftover (when I had a login endpoint), will remove

test-harness/specs/security-scheme-validation-AC-10.oas2.txt
lag-of-death: it sure would :) will add :+1:

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
lag-of-death: @chris-miaskowski, what about:
```
WWW-Authenticate: Basic realm="*", X-API-Key realm="*"
```
cc @XVincentX, @philsturgeon

test-harness/specs/security-scheme-validation-AC-6_failing.oas3.txt
lag-of-death: @chris-miaskowski, I've added presence validation in https://github.com/stoplightio/prism/pull/484/commits/4a0fd4990a575d7a42d26ec28cc5e62dba5c0719#diff-0b83a74c0a2f17482bdada0815512f6f
don't know if we can other type of validation here since api key is free-form (?)

test-harness/specs/security-scheme-validation-AC-8.oas3.txt
lag-of-death: added in https://github.com/stoplightio/prism/pull/484/commits/4a0fd4990a575d7a42d26ec28cc5e62dba5c0719#diff-5fe41ed9481abb1f32aa1142c6054827

test-harness/specs/security-scheme-validation-AC-10.oas2.txt
lag-of-death: added in https://github.com/stoplightio/prism/pull/484/commits/4a0fd4990a575d7a42d26ec28cc5e62dba5c0719#diff-ddb070efafba3257e4c1ff0884ad6463

lag-of-death:
done with the changes for now, can be reviewed

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
lag-of-death: @chris-miaskowski, that's why I ask, couldn't find anything that would tell for sure. https://tools.ietf.org/html/rfc7235 doesn't have answers. Do you know where to look?

philsturgeon: X-API-Key is an arbitrary header name. It could be Chunkey-Monkey or anything else. You won’t find any standards for it. 

The question is: why is real there in the first place? All of these tests should come from examples in the OpenAPI spec and/or HTTP spec but we can’t guess at how these security schemas work.

test-harness/specs/security-scheme-validation-AC-10.oas3.txt
```diff
@@ -0,0 +1,35 @@
...
+            read: Grants read access
+            write: Grants write access
+            admin: Grants read and write access to administrative information
+security:
+  - OAuth2: [read, write]
+====server====
+mock -p 4010
+====command====
+curl -i http://localhost:4010/todos -H "Authorization: Bearer abc123"
+====expect====
```
philsturgeon: Why are we sending the same authorization headers for every different type of security Schema? They all work in very different ways: where did these examples come from?

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
lag-of-death: hmm.. @philsturgeon, I'm not sure I follow. `realm` is required for `Basic` authentication scheme:
```
The Basic authentication scheme utilizes the Authentication Framework
   as follows.

   In challenges:

   o  The scheme name is "Basic".

   o  The authentication parameter 'realm' is REQUIRED ([RFC7235],
      Section 2.2).
```
that's why it's there(?). And about `X-API-Key`, from https://tools.ietf.org/html/rfc7235:
```
 WWW-Authenticate: Newauth realm="apps", type=1,
                       title="Login to \"apps\"", Basic realm="simple"
```
`Newauth` is clearly custom and it still has `realm` (although this might not be necessary) so I added `realm` for `X-API-Key` here. And why is this here? Well, I thought before implementing the security, it would be good if the tests would test the right thing (?) So I get that `realm` is not to be added for custom authentication schemes

test-harness/specs/security-scheme-validation-AC-10.oas3.txt
lag-of-death: Hm.. but don't they all (the three tests) depend on spending `-H "Authorization: Bearer abc123"`? The source of these examples would be me trying to meet the ACs.

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
chris-miaskowski: @lag-of-death you are right about the Basic as far I can tell. Not sure what @philsturgeon means. 

I would ignore the `realm` for anything else than `Basic`.

philsturgeon: I’m not sure what is causing confusion here? I asked what the source of these examples were, then Mateus responded with his examples. That’s what I asked for. 

I think the Newauth is a bit different from the sort of usage being expected in X-RANDOM-KEY and that is different again from OpenID, so as Chris said, probably ditch realm for these things. As I said this seems like some guesswork was involved, educated guesswork but still potentially not correct.

philsturgeon: To provide more context, there are a few potential sources you could be getting these examples from. The OpenAPI spec, the HTTP spec (which might cover things that OpenAPI cannot), or Swagger.io documentation: https://swagger.io/docs/specification/authentication/

Extra context: there are "authentication schemes" which are official specified things which go in the `Authorization` header, and those usually have a corresponding keyword which goes into the content of the `Authorization` header value. So the Basic scheme looks like `Authorization: Basic foo` and the Bearer scheme goes into `Authorization: Bearer foo`. That's all part of the Authorization header. Some good docs on that are over here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

API keys could use the Authorization header and have an Authorization scheme, _or_ they could be totally arbitrary headers, query string parameters, or even a body param! 

A header example:

```
X-API-Key: foo
Some-Nonsense-Key: foo
Feet: foo
```

These are all valid.

```
components:
  securitySchemes:
    BasicAuth:
      type: http
      scheme: basic
    BearerAuth:
      type: http
      scheme: bearer
    NonsenseKeyAuth:
      type: apiKey
      in: header
      name: Feet
    ApiKeyAuth:
      type: apiKey
      in: header
      name: Feet
    FeetAuth:
      type: apiKey
      in: header
      name: Feet
```

It might not be a header at all:

```
http://example.org/foo?someNonsenseKey=foo 
```

This would be defined like this:

```
    SomeNonsenseAuth:
      type: apiKey
      in: query
      name: someNonsenseKey
```

There are a lot of options here. apiKey type params will not have a real. 

Understanding this stuff is a fair bit of work, I do not remember everything off the top of my head but I do have a fair bit of context and can help dig in. I am here for questions.

test-harness/specs/security-scheme-validation-AC-10.oas3.txt
philsturgeon: Bearer is one of many authorization schemes available. I believe the GitHub issue for this links you to the appropriate part of the OpenAPI specification, which points you to https://tools.ietf.org/html/rfc7235#section-5.1

philsturgeon: For example I don't see the word 'Digest' in this PR, but it's in #425

lag-of-death: hi @philsturgeon, so there already was an example of `Basic`: https://github.com/stoplightio/prism/blob/271588bc86be61144387ee4e24b39b758f5acec3/test-harness/specs/security-scheme-validation-AC-16_failing.oas2.txt
and I added 'Digest' example: https://github.com/stoplightio/prism/blob/271588bc86be61144387ee4e24b39b758f5acec3/test-harness/specs/security-scheme-validation-AC-17_failing.oas2.txt

test-harness/specs/security-scheme-validation-AC-3_failing.oas2.txt
lag-of-death: `realms`s from `X-API-Key` removed. I've left `realm` here though: https://github.com/stoplightio/prism/blob/266e90c700e12529be507489682c3681c53f70b3/test-harness/specs/security-scheme-validation-AC-13_failing.only_oas3.txt#L28
what do you @philsturgeon think? should this cookie `realm` also be removed?

philsturgeon: That's fine, check it out: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate 

WWW-Authenticate is allowed to have a real, its in the spec there, so you're all good.

lag-of-death:
@chris-miaskowski, @philsturgeon, @XVincentX, if there are no more concerns (?), then the implementation could be started

philsturgeon:
@lag-of-death it looks good enough to get started! Anything else we missed will come out in the wash. Great work!

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -64,7 +64,7 @@ describe('GET /pet?__server', () => {
   }
 });
 
-describe.each([['petstore.oas2.yaml'], ['petstore.oas3.yaml']])('server %s', file => {
+describe.each([['petstore.no-auth.oas2.yaml', 'petstore.no-auth.oas3.yaml']])('server %s', file => {
```
lag-of-death: Had to remove security from the specs to have these tests be still useful.

packages/core/src/utils/security/handlers.ts
```diff
@@ -0,0 +1,130 @@
...
+      ? checkHeader<R>(authorizationHeader, resource)
+      : Either.left(genUnauthorisedErr('Basic realm="*"'));
+  },
+};
+
+function checkHeader<R>(authorizationHeader: string, resource?: R) {
+  const [schema, token] = authorizationHeader.split(' ');
+
+  const isBasicTokenGiven = token && isBasicToken(token);
+  const isBasicSchema = schema === 'Basic';
```
philsturgeon: The variable name here should probably be `authScheme` just to disambiguate it from a lot of other schemes (and its not a schema). 

We should make it case insensitive too as I checked the HTTP Basic RFC and they mention this:
 
> Note that both scheme and parameter names are matched case-insensitively.

https://tools.ietf.org/html/rfc7617

Make sure the code being written for specific auth types match what the spec says is ok:

https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

packages/core/src/utils/security/handlers.ts
```diff
@@ -0,0 +1,130 @@
...
+  test: ({ type, scheme }: SecurityScheme) => scheme === 'bearer' && type === 'http',
+  handle: bearerHandler.bind({}, 'Bearer'),
+};
+
+const oauth2 = {
+  test: ({ type }: SecurityScheme) => type === 'oauth2',
+  handle: bearerHandler.bind({}, 'OAuth2'),
+};
+
+const apiKeyInCookie = {
```
philsturgeon: There are lots of different ways to sort methods around, some folks alphabetize everything, or group public and private then alpha those, or they just group similar stuff together. 

I like to group and alphabetize within the group, but whatever approach you decide on, this method seems to a long way from its friends. :)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -235,27 +235,49 @@ const helpers = {
...
+        if (response) logger.success(`Created a ${response.code} from a default response`);
+        return response;
+      }
+    }
+
+    logger.success(`Found response ${result.code}. I'll try with it.`);
+    return result;
+  },
+
+  findByStatusCode(statusCode: string, httpResponses: IHttpOperationResponse[]) {
```
philsturgeon: It looks like this method is just aliasing `findResponseByStatusCode` and sapping the order of arguments? Is this needed?

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+      ]);
+    });
+  });
+
+  describe('basic auth', () => {
+    describe('valid', () => {
+      test('passes when the correct header is provided', () => {
+        expect(
+          validateSecurity<any, any>(
+            { headers: { authorization: 'Basic dGVzdDp0ZXN0' } },
```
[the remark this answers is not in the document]
lag-of-death: Thanks, have not considered it

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+            validateSecurity<any, any>(
+              { headers: { 'x-api-key': 'abc123' } },
+              { security: [{ in: 'header', type: 'apiKey', name: 'x-api-key' }] },
+            ),
+          ).toStrictEqual([]);
+        });
+      });
+
+      describe('invalid', () => {
+        test('returns an error object with 401 code', () => {
```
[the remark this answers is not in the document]
lag-of-death: @philsturgeon, there's a couple of `test('returns an error object with 401 code` there. Would you rather see them all changed to `it('fails with XYZ`?

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+          expect(
+            validateSecurity<any, any>(
+              { headers: {} },
+              { security: [{ in: 'header', type: 'apiKey', name: 'x-api-key' }] },
+            ),
+          ).toStrictEqual([
+            {
+              headers: {
+                'WWW-Authenticate': 'x-api-key',
```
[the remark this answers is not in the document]
lag-of-death: @philsturgeon, I think I understand. Will adjust the code

packages/core/src/utils/security/handlers.ts
lag-of-death: @philsturgeon, are all schemes from https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml to be supported?

philsturgeon: No don't worry, I don't know what many of them are. We have only discussed Basic, Digest, and Bearer.

packages/core/src/utils/__tests__/security.spec.ts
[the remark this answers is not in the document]
philsturgeon: A better way to state that ruke, if `type === 'http'` then `WWW-Authenticate: ${authSchema}` otherwise no www-authenticate.

packages/core/src/utils/security/handlers.ts
lag-of-death: Digest isn't supported for now. Didn't implement it since it isn't (?) openapi supported. Will add it :+1:

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
lag-of-death: ha! thanks :) not needed indeed :+1:

packages/core/src/utils/security/handlers.ts
philsturgeon: OpenAPI "supports" anything in the http-authschemes.xhtml list through type === http, but we dont have to write tests for it all. 

The most common are Basic, Digest and Bearer, so having tests for that wouldn't hurt, but if we dont have digest im fine with that. Not important.

lag-of-death: I meant that there is no implementation for `Digest` (it's not handled) and also no tests for it. I will add it. For now, basic / apiKey / bearer (for the three of: scheme: `bearer` / `type`: oauth2/openIdConnect) are supported.

lag-of-death:
@philsturgeon, I have rearranged symbols - put them in different files and ordered them by private/public membership and then alphabetically.
Also, the added `digest auth` support only checks for required parameters to `Authorization` request header and the presence of `Digest` word. It does not for example check if `response` is indeed a MD5 (or any other algorithm) hash. Do you think we should add this check? For example, something like: `isMD5hash(response)` ? Please also take a look at https://github.com/stoplightio/prism/pull/484#discussion_r311599219 many thanks! :)

packages/core/src/utils/__tests__/security.spec.ts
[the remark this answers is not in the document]
philsturgeon: Yeah generally we use the `it` stuff because we are following describe-it, so might as well. 

Check this awesome site out http://www.betterspecs.org/ I was a big fan when I was over in ruby land, and we recommended it for everyone trying to figure testing naming and practices. It doesnt all apply because ruby, but things like `let(:foo)` is just `const foo` and a lot of it does work. 

Using more `describe` (because we dont have `context`) for setup and description of the state, then the it just describes the outline of the test without getting too stuck into exactly what assertions are.

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+      expect(validateSecurity<any, any>({}, { security: [{}] })).toStrictEqual([
+        {
+          message: 'No handler for the security scheme found.',
+        },
+      ]);
+    });
+  });
+
+  describe('basic auth', () => {
+    describe('valid', () => {
```
philsturgeon: this context does not seem to be adding anything?

packages/core/src/utils/__tests__/security.spec.ts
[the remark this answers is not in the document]
lag-of-death: @philsturgeon, the tests are now adjusted, hopefully adjusted correctly ;)

packages/core/src/factory.ts
```diff
@@ -73,7 +74,13 @@ export function factory<Resource, Input, Output, Config>(
                     components.mocker.mock(
                       {
                         resource,
-                        input: { validations: { input: inputValidations }, data: input },
+                        input: {
+                          validations: {
+                            input: inputValidations,
+                            security: validateSecurity<Resource, Input>(input, resource),
```
chris-miaskowski:
```suggestion
                            security: validateSecurity(input, resource),
```

TS should infer the types, right?

packages/core/src/types.ts
```diff
@@ -80,6 +81,7 @@ export interface IPrismInput<I> {
   data: I;
   validations: {
     input: IPrismDiagnostic[];
+    security: AuthErr[];
```
chris-miaskowski: Hey @lag-of-death, curious about that, was `validations.input` not possible/easy to reuse?

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+    it('passes the validation', () => {
+      expect(validateSecurity<any, any>({}, { security: [] })).toStrictEqual([]);
+    });
+  });
+
+  describe('when given an unknown security scheme type', () => {
+    it('fails with a message explaining the issue', () => {
+      expect(validateSecurity<any, any>({}, { security: [{}] })).toStrictEqual([
+        {
+          message: 'No handler for the security scheme found.',
```
chris-miaskowski: Hey, what does 'no handler found' mean? Is that what we're sending to the user?
WDYT about rephrasing to something like:

```suggestion
          message: 'Security scheme type is not defined.',
```

Or maybe it should silently pass if we provide invalid security scheme?

packages/core/src/factory.ts
lag-of-death: Hm.. don't know really, was not inferring them for me

packages/core/src/types.ts
lag-of-death: @chris-miaskowski, I don't know. I didn't want to reuse input. If you'd see `security` to be part of `input` I think it won't be hard to put it there.

packages/core/src/utils/__tests__/security.spec.ts
lag-of-death: Well, only some security schemes are handled (like Basic/Digest/Bearer). It would be impossible/hard to handle all types of authorization schemes (@philsturgeon mentioned this link before https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml). So, for example, if a user specifies in a spec:
```
components:
  securitySchemes:
    VapidAuth:
      type: http
      scheme: vapid
```
then security scheme (and its type) is defined, but we don't have a handler for it.

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+          validateSecurity<any, any>(
+            { headers: { authorization: 'Basic abc123' } },
+            { security: [{ scheme: 'basic', type: 'http' }] },
+          ),
+        ).toStrictEqual([
+          {
+            headers: {},
+            message: 'Invalid credentials used',
+            name: 'Forbidden',
+            status: 403,
```
chris-miaskowski: Are you sure it should be 403 not 422?

chris-miaskowski: Hmm yeah you're right I think. :)

packages/core/src/utils/security/handlers/apiKey.ts
```diff
@@ -0,0 +1,38 @@
+import { fromNullable, getOrElse, mapNullable } from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get } from 'lodash';
+import { SecurityScheme } from './types';
+import { when } from './utils';
+
+export const apiKeyInCookie = {
+  test: ({ type, in: where }: SecurityScheme) => where === 'cookie' && type === 'apiKey',
+  handle: <R, I>(someInput: I, name: string, resource?: R) => {
+    const probablyCookie = get(someInput, ['headers', 'cookie']);
```
chris-miaskowski: I like this "probably" naming convention :)

packages/core/src/utils/security/handlers/apiKey.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const isApiKeyInCookie = pipe(
+      fromNullable(probablyCookie),
+      mapNullable(cookie => new RegExp(`${name}=.+`).test(cookie)),
+      getOrElse(() => false),
```
chris-miaskowski: Neat stuff, @XVincentX will like it! :)

packages/core/src/utils/security/handlers/apiKey.ts
lag-of-death: well, you know, "life was like a box of chocolates. You never know what you're gonna get." ;)

packages/core/src/utils/security/handlers/apiKey.ts
```diff
@@ -0,0 +1,38 @@
...
+import { when } from './utils';
+
+export const apiKeyInCookie = {
+  test: ({ type, in: where }: SecurityScheme) => where === 'cookie' && type === 'apiKey',
+  handle: <R, I>(someInput: I, name: string, resource?: R) => {
+    const probablyCookie = get(someInput, ['headers', 'cookie']);
+
+    const isApiKeyInCookie = pipe(
+      fromNullable(probablyCookie),
+      mapNullable(cookie => new RegExp(`${name}=.+`).test(cookie)),
```
chris-miaskowski: Not sure about this regexp. Have you tried some weird edge cases? 
`Set-Cookie: id=a3fWa; Expires=Wed, 21 Oct 2015 07:28:00 GMT;`
Or
`Set-Cookie: id=bla=bla;`

packages/core/src/utils/security/handlers/apiKey.ts
```diff
@@ -0,0 +1,38 @@
...
+    );
+
+    return when<R>(isApiKeyInCookie, '', resource);
+  },
+};
+
+export const apiKeyInHeader = {
+  test: ({ type, in: where }: SecurityScheme) => where === 'header' && type === 'apiKey',
+  handle: <R, I>(someInput: I, name: string, resource?: R) => {
+    const isAPIKeyProvided = get(someInput, ['headers', name.toLowerCase()]);
```
chris-miaskowski: What was the reason of making those functions generic (<R, I> stuff)?
More specific types could benefit from static type checking and you could avoid `get(...)`

lag-of-death: @chris-miaskowski, it's just that `factory` function is generic and I had no choice (?)

packages/core/src/utils/security/handlers/apiKey.ts
lag-of-death: ha! I haven't, thanks, will look into that

packages/core/src/utils/__tests__/security.spec.ts
chris-miaskowski: So maybe "we currently do not support this type of security scheme"? Or something along those lines? Less cryptic imo.

packages/core/src/utils/security/handlers/apiKey.ts
lag-of-death: @chris-miaskowski, after giving it a bit of thought, I am inclined to say that this regex is OK.
Also, I assumed that `Set-Cookie` was mistaken for `Cookie` (?). If not, please LMK :)

In the end what's happening is this:
```
new RegExp('SOME_COOKIE=.+').test('SOME_COOKIE=<<some_value>>)
```
So, I'm just checking if there is this SOME_COOKIE with some (any) value. The value itself is not important, it's only about the presence of it (where `it` is a cookie named `name`) . So, the examples you provided, would pass the test and so would any other value for `id` cookie. 

Also, I think that `Expires` is kinda irrelevant, since it wouldn't be present in `Cookie` header (?). And even it it was present in `Cookie` header it would be (in the context of the example) considered another cookie named `Expires` I think.

I don't know if my answer addresses your concerns, LMK :)

packages/core/src/utils/security/handlers/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+  message: 'Invalid security scheme used',
+  status: 401,
+  headers: msg ? { 'WWW-Authenticate': msg } : {},
+});
+
+export function isScheme(authScheme: string, shouldBeScheme: string) {
+  return (authScheme || '').toLowerCase() === shouldBeScheme;
+}
+
+export function when<R>(isOk: boolean, msg: string, resource?: R) {
```
XVincentX: Is it really worth to make this a separate function? This function is a little bit confusing to me.

```suggestion
export function when<R>(condition: boolean, errorMessage: string, resource?: R) {
```

packages/core/src/utils/security/handlers/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+    {
+      test: () => isSchemeProper,
+      handle: () => invalidCredsErr,
+    },
+  ].find(possibleHandler => possibleHandler.test());
+
+  return handler ? handler.handle() : Either.left(genUnauthorisedErr(msg));
+}
+
+export const genUnauthorisedErr = (msg: string): AuthErr => ({
```
XVincentX: In general all that’s a `Left` should be at least an `Error` so that error handling is easier

packages/core/src/utils/security/handlers/utils.ts
```diff
@@ -0,0 +1,39 @@
+import * as Either from 'fp-ts/lib/Either';
+import { AuthErr } from './types';
+
+const invalidCredsErr = Either.left({
```
XVincentX: I think you should use a `ProblemJson` class instance here

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    : (() => {
+        const allWWWAuthHeaders = invalidSecuritySchemas.reduce((accumulator: string, currentValue) => {
+          return pipe(
+            currentValue,
+            Either.mapLeft(err => [accumulator, get(err, pathToHeader)].filter(identity).join(', ')),
+            Either.fold(authHeader => authHeader || '', () => ''),
+          );
+        }, '');
+
+        return set(pathToHeader, allWWWAuthHeaders, firstLeftValue);
```
XVincentX: Hmm, can you quickly explain what’s going on here?

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      })();
```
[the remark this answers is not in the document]
lag-of-death: It could be implemented another way I guess, but is `IIFE` a bad choice here?

packages/core/src/utils/security/handlers/utils.ts
lag-of-death: It's used a couple of times and I find it quite useful in terms of tidying things up. But if you'd rather see it removed, LMK :)

packages/core/src/utils/security/index.ts
lag-of-death: Sure :) so, first, all `WWW-Authenticate` headers were gathered from schemes that did not pass the validation: `allWWWAuthHeaders` (https://github.com/stoplightio/prism/blob/f68e4abe51d9c7284dc30ee28974acfcbdeaa810/packages/core/src/utils/security/index.ts#L21).

Then, I've taken the first scheme that failed - `firstLeftValue` (the value of the first Left) and "extended" its `['headers', 'WWW-Authenticate']` to be `allWWWAuthHeaders`. So the response will have that header with info on ways to authenticate.

lag-of-death:
hi @XVincentX, I think I've addressed the feedback. LMK if there's more or I have missed/misunderstood something :+1:

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function validateSecurity<R, I>(someInput: I, resource?: R) {
+  const securitySchemas = get(resource, 'security', []);
+
+  return !securitySchemas.length
+    ? []
+    : (() => {
```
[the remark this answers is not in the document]
XVincentX: Same as above, it's not really worth to use the IIFE here for the ternary operator.

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function validateSecurity<R, I>(someInput: I, resource?: R): IPrismDiagnostic[] {
+  const securitySchemes = get(resource, 'security', []);
+
+  return !securitySchemes.length
+    ? []
+    : (() => {
+        const validatedSecuritySchemes = securitySchemes.map((definedSecScheme: SecurityScheme) => {
+          const schemeHandler = securitySchemeHandlers.find(handler => {
+            return handler.test(definedSecScheme);
```
XVincentX: const schemeHandler = securitySchemeHandlers.find(handler => handler.test(definedSecScheme))

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -235,27 +235,45 @@ const helpers = {
     return helpers.negotiateOptionsForDefaultCode(httpOperation, desiredOptions);
   },
 
+  find400Or422(httpResponses: IHttpOperationResponse[], logger: Logger) {
```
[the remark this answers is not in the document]
XVincentX: Ping

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -235,27 +235,45 @@ const helpers = {
...
+
   negotiateOptionsForInvalidRequest(
     httpResponses: IHttpOperationResponse[],
   ): ReaderEither<Logger, Error, IHttpNegotiationResult> {
-    return pipe(
-      withLogger(logger => {
-        let result = findResponseByStatusCode(httpResponses, '422');
-        if (!result) {
-          logger.trace('Unable to find a 422 response definition');
+    return this.genNegotiateOptionsForInvalidRequest(httpResponses, this.find400Or422);
```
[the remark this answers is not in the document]
XVincentX: Ping

lag-of-death:
@XVincentX, I'm handing this to you. No need to bother with a PR to this branch, can just push to this branch :+1:

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -55,34 +55,57 @@ export class HttpMocker
...
-                })),
-              },
-            ),
-          ),
-        ),
-      ),
-    );
+  if (get(input, ['validations', 'security', 'length'])) {
+    return handleSecurityValidation(input, resource);
+  } else if (input.validations.input.length > 0) {
```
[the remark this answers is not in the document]
XVincentX: @philsturgeon Ping

XVincentX:
@philsturgeon @chris-miaskowski 

I've added a bunch of commits to modify some of the behaviours and make sure we still stick with and reuse the current validation system (although it's starting to show off its limits and might need to be refactored kind of soon).

It took me some time to make this work, unfortunately some of the technical debt here in Prism is kind of a ticking bomb. Anyway, for now we can get away with it.

XVincentX:
@philsturgeon I also recall we kind of decided that security validation would be enabled by default, but they could be disabled using a CLI flag. Is this still the case? Shall I implement this?

philsturgeon:
Yes security validation should be enabled by default, just like other aspects of request validation.

packages/core/src/utils/security/index.ts
```diff
@@ -29,10 +29,10 @@ function getAllInvalidSec<R>(invalidSecuritySchemes: Array<Left<IPrismDiagnostic
   }
 }
 
-export function validateSecurity<R, I>(someInput: I, resource?: R): IPrismDiagnostic[] {
+export function validateSecurity<R, I>(someInput: I, resource?: R): IPrismDiagnostic | undefined {
   const securitySchemes = get(resource, 'security', []);
 
-  if (!securitySchemes.length) return [];
+  if (!securitySchemes.length) return;
```
lag-of-death: Why make `validateSecurity` return `IPrismDiagnostic | undefined` if it could be simpler to just have `IPrismDiagnostic[]`? What's wrong with an empty array? If we do `return this type or this type`, then we stray away from functional principles. If we really want to have `undefined` here, at least let's put it inside `Either`, so it's either `Right<IPrismDiagnostic>` or `Left<undefined>`.

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const allWWWAuthHeaders = invalidSecuritySchemes.reduce((accumulator, currentValue) => {
+      return pipe(
+        currentValue,
+        mapLeft(err => [accumulator, get(err, ['headers', 'WWW-Authenticate'])].filter(identity).join(', ')),
+        fold(authHeader => authHeader || '', () => ''),
+      );
+    }, '');
+
+    if (allWWWAuthHeaders !== '')
+      firstLeftValue.tags ? firstLeftValue.tags.push(allWWWAuthHeaders) : (firstLeftValue.tags = [allWWWAuthHeaders]);
```
lag-of-death: `if` statement without `{}` is just a shoot in the foot. Also, why `push`? Let's construct it without mutations.

packages/core/src/utils/security/handlers/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const genUnauthorisedErr = (msg: string): IPrismDiagnostic => ({
+  severity: DiagnosticSeverity.Error,
+  message: 'Invalid security scheme used',
+  code: 401,
+  tags: msg ? [msg] : [],
+});
+
+export function isScheme(authScheme: string, shouldBeScheme: string) {
+  return (authScheme || '').toLowerCase() === shouldBeScheme;
```
lag-of-death: let's switch the parameters to be `shouldBeScheme: string, authScheme: string = ''`

packages/core/src/utils/security/handlers/apiKey.ts
```diff
@@ -0,0 +1,38 @@
+import { fromNullable, getOrElse, map } from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get } from 'lodash';
+import { SecurityScheme } from './types';
+import { when } from './utils';
+
+export const apiKeyInCookie = {
+  test: ({ type, in: where }: SecurityScheme) => where === 'cookie' && type === 'apiKey',
+  handle: (someInput: unknown, name: string, resource?: unknown) => {
```
lag-of-death: why is it `unknown` now? I remember these having concrete types

packages/core/src/utils/security/handlers/bearer.ts
```diff
@@ -0,0 +1,32 @@
...
+  return pipe(
+    fromNullable(get(inputHeaders, 'authorization')),
+    map(authorization => !!authorization.match(/^Bearer\s.+$/)),
+    getOrElse(() => false),
+  );
+}
+
+export const bearer = {
+  test: ({ type, scheme }: SecurityScheme) => scheme === 'bearer' && type === 'http',
+  handle: bearerHandler.bind(undefined, 'Bearer'),
```
lag-of-death: what's safer? `(undefined).abc` or `({}).abc`?

packages/http-server/src/server.ts
```diff
@@ -105,8 +105,11 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
           reply
             .type('application/problem+json')
             .serializer(JSON.stringify)
-            .code(status)
-            .send(ProblemJsonError.fromPlainError(e));
+            .code(status);
+
+          if (e.additional && e.additional.headers) reply.headers(e.additional.headers);
```
lag-of-death: let's add `{}`

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -55,34 +54,47 @@ export class HttpMocker
...
+        helpers.negotiateOptionsForInvalidRequest(resource.responses),
+        mapLeft(() => {
+          const securityValidation = input.validations.input.find(i => i.code === 401 || i.code === 403);
+
+          return securityValidation
+            ? ProblemJsonError.fromTemplate(securityValidation.code === 401 ? UNAUTHORIZED : FORBIDDEN, '', {
+                headers:
+                  securityValidation.tags && securityValidation.tags.length
+                    ? { 'WWW-Authenticate': securityValidation.tags.join(',') }
+                    : undefined,
```
lag-of-death: let's not create headers if we have no need for them

packages/core/src/utils/security/handlers/apiKey.ts
XVincentX: They had not real concrete types. They were using generic arguments, true, but we had no real informations about then — effectively behaving like an `unknown` and using lodash functions to access its properties.

I'd rather make this explicit so that one day I can refactor this instead of pretending we're using the generics correctly but we cannot.

packages/core/src/utils/security/handlers/bearer.ts
XVincentX: `bind` with undefined does not mean that the `thisArg` for the function is undefined. 

> If no arguments are provided to bind, the this of the executing scope is treated as the thisArg for the new function.

Basically it will propagate the current this scope instead of creating a new one. It's like doing `handle: () => bearerHandler('Bearer')`

packages/core/src/utils/security/index.ts
XVincentX: There might be other value in the array, and I didn't want to cancel these.

packages/core/src/utils/security/index.ts
XVincentX: An array should represent the repetition of something; this is clearly not happening here. The ideal would be to return an `Option<IPrismDiagnostic>` — which is something I really want to start doing here in the codebase.

I will try to do this now but it might be painful and this needs to be rewritten anyway, the validation system is starting to kill us.

lag-of-death: @XVincentX, if `An array should represent the repetition of something`, then why it is allowed in some languages for an array to have misc types? Is `[1, 'abc', {}]` a repetition of what? A box of oranges (even if empty) is still a box of oranges. And a function's domain should never be a box of oranges and apples, just one of these

XVincentX: @lag-of-death I've refactored the function to make it return an Option 796d18508bfa4d1279bd52498dd827ead2eeb4f7 — this should resolve this issue

XVincentX:
@lag-of-death Can we give it another round?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -235,37 +234,43 @@ const helpers = {
...
+        logger.trace('Unable to find a 422 response definition');
+
+        result = findResponseByStatusCode(httpResponses, '400');
+        if (!result) {
+          logger.trace('Unable to find a 400 response definition.');
+          const response =
+            findResponseByStatusCode(httpResponses, '401') ||
+            findResponseByStatusCode(httpResponses, '403') ||
+            createResponseFromDefault(httpResponses, '422');
+          if (response) logger.success(`Created a ${response.code} from a default response`);
```
lag-of-death: lacks `{}`

packages/core/src/utils/security/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    return firstLeftValue;
+  }
+}
+
+export function validateSecurity(someInput: unknown, resource?: unknown): Option<IPrismDiagnostic> {
+  const securitySchemes = get(resource, 'security', []);
+
+  if (!securitySchemes.length) return none;
```
lag-of-death: lacks `{}`

lag-of-death:
:+1:

lag-of-death:
@philsturgeon, do we have a green light to merge?

rubencaro:
> @philsturgeon I also recall we kind of decided that security validation would be enabled by default, but they could be disabled using a CLI flag. Is this still the case? Shall I implement this?

@XVincentX  The CLI flag to disable it was actually not implemented, right? Maybe I missed it.

XVincentX:
@rubencaro Correcto, en este momento Prism no ofrece esta posibilidad.

PR #485, opened 2019-07-23
feat: add prism user-agent header for http requests made with http client

### LINKED ISSUE #440: Set a custom user-agent when making HTTP requests
**User Story**

As a Prism HTTP Client user, I want to see the user-agent header set to something like `Prism v3.0.1` so I know where the request came from.

Prism HTTP Client is making calls to real servers. That **was** part of “forwarding” (which is remaking requests as they come in, like a proxy) but right now we have that experimental HTTP Client JS API remember? That should set user agent.

**Acceptance Criteria**

Any requests going through the Prism HTTP client have a user agent set with the version number.

lag-of-death:
## Checklist

Feature.

Currently, when a HTTP request is made with HTTP Client, no `user-agent` header is set.
The new behavior is to set that header to `Prism/<<PRISM_VERSION>>` value.

No.

Closes #440


packages/http/README.md
```diff
@@ -237,6 +237,7 @@ const config = {
 ```
 
 With the above configuration the http client will proxy your requests to a server and will validate the response with the OpenAPI you loaded. However, it will not validate the input (e.g. will not check whether the provided query param is valid).
+Please also note, that the proxied requests will have 'user-agent' header set to Prism/<<PRISM_VERSION>>.
```
philsturgeon: I don't think we need to mention this. I doubt anyone will care, and its very obvious if they do. Most HTTP clients set user agent already, so this is just changing from `User-Agent: axios whatever` to `User-Agent: prism whatever`.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -16,6 +16,8 @@ const staticExamplesOas2Path = fixturePath('static-examples.oas2.json');
 const serverValidationOas2Path = fixturePath('server-validation.oas2.json');
 const serverValidationOas3Path = fixturePath('server-validation.oas3.json');
 
+const { version: prismVersion } = require('../../package.json');
```
philsturgeon: @XVincentX is this ok?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -172,6 +174,55 @@ describe('Http Client .process', () => {
...
+              method: 'get',
+              url: {
+                path: '/pet',
+              },
+            },
+            config,
+          );
+        });
+
+        it('specifying user-agent in the options parameter has no impact on user-agent header to be sent', async () => {
```
philsturgeon: This user-agent should just be a default, folks can override if they want. That's how most HTTP clients handle it anyhow.

packages/http/README.md
lag-of-death: Hm.. I don't really see why delete this.

XVincentX: I agree with @philsturgeon here — this information does not really bring any value to the consumer; I'd remove it as well.

packages/http/src/__tests__/http-prism-instance.spec.ts
XVincentX: This should be ok — I'd just not use require and import it.

```suggestion
import { version as prismVersion } from '../../package.json';
```

This will probably give us some type checking as well

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -172,6 +174,55 @@ describe('Http Client .process', () => {
...
       });
+
+      describe('Prism user-agent header', () => {
+        it('adds user-agent header in the form of Prism/<<PRISM_VERSION>>', async () => {
+          const oasBaseUrl = 'http://example.com/api';
+
+          nock(oasBaseUrl)
+            .get('/pet')
+            .reply(function() {
+              expect(this.req.headers['user-agent']).toBe(`Prism/${prismVersion}`);
```
XVincentX: Instead of doing the assertion here I would rather save the value in a variable and then asset it in a following and explicit `it` block — this should make the things easier to red.

```ts
describe("Prism user agent header", () => {
  describe("when the defaults are used", () => {
    it("shuold have Prism/version", () => {
      expect(savedvalue).toBe("Prism/version3.10");
    });
  });
});
```

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -172,6 +174,55 @@ describe('Http Client .process', () => {
...
+
+          nock(oasBaseUrl)
+            .get('/pet')
+            .reply(function() {
+              expect(this.req.headers['user-agent']).toBe('Other_Agent/1.0.0');
+
+              return [200, 'reply', {}];
+            });
+
+          await prism.process(
```
XVincentX: Same as above.

packages/http/src/forwarder/HttpForwarder.ts
```diff
@@ -63,20 +65,17 @@ export class HttpForwarder implements IForwarder<IHttpOperation, IHttpRequest, I
...
-
-    // host header provided? override with actual hostname
-    if (headers.hasOwnProperty('host')) {
-      return {
-        ...headers,
-        host: new URL(baseUrl).host,
-        forwarded: `host=${headers.host}`,
-      };
-    }
+  private static updateHostHeaders(baseUrl: string, headers: IHttpNameValue = {}) {
```
XVincentX: I'm ok with not having this as an instance method — but I think instead of making it static let's just put it out of the way of the class and leave it in the file somewhere else so nobody uses it outside of this file (which is good)

packages/http/README.md
philsturgeon: We should do things for a reason, not come up with reasons to not do things. :)

Documentation is important, but keeping a high ratio of signal to noise is important s the more stuff we add to docs which isn’t particularly useful, the more users have to scroll, the higher their chance of missing useful information, which means they’re posting issues, asking for help in support, which is time people could have spent doing other things. 

So let’s delete this. It adds no value. :)

lag-of-death: okey-dokey :+1:

lag-of-death:
@XVincentX, @philsturgeon, this is now ready for re-review

XVincentX:
Nice. I’ll jump on this after lunch!

Enviado do meu iPhone

Em 24 de jul de 2019, à(s) 14:02, lag-of-death <notifications@github.com> escreveu:

PR #490, opened 2019-07-25
Remove loader and graphite

### LINKED ISSUE #390: Using $ref "You must provide specification in the third parameter"
I have a relative file path in my `$ref`, the error message is not explicit

```
    [HTTP SERVER] get /map/default ✖  error     Request terminated with error: Error: Your schema contains $ref. You must provide specification in the third parameter.
```

My spec:

```yaml
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: map_model.yaml#/components/schemas/MapFull
```

Thanks for info

_Originally posted by @fgravin in https://github.com/stoplightio/prism/issues/116#issuecomment-507601171_

### LINKED ISSUE #407: Public JS API Improvements

XVincentX:
This PR will modify Prism to not use Graphite anymore, but rather load the files directly. This will give us support for refs since they're not coupled with Graphite and removes *tons* of dependencies that Prism was not using — Spectral in primis.

I've left some comments around the code to show the relevant parts. Let's discuss it there.

Closes #407 
Closes #390


packages/cli/package.json
```diff
@@ -7,8 +7,14 @@
...
   "bugs": "https://github.com/stoplightio/prism/issues",
   "dependencies": {
+    "@stoplight/http-spec": "^1.0.4",
+    "@stoplight/json": "^3.0.1",
+    "@stoplight/json-ref-resolver": "^2.0.3",
+    "@stoplight/path": "^1.1.0",
     "@stoplight/prism-core": "^3.0.3",
     "@stoplight/prism-http-server": "^3.0.3",
+    "@stoplight/yaml": "^2.8.0",
+    "axios": "^0.19.0",
```
philsturgeon: Do we need axios or could we get away with node-fetch?

packages/cli/src/resolvers/http-and-file.ts
```diff
@@ -0,0 +1,37 @@
+import { Resolver } from '@stoplight/json-ref-resolver';
```
philsturgeon: Is this a copy and paste from Spectral? Could we hoist this up somewhere and have it shared?

packages/cli/src/util/getHttpOperations.ts
```diff
@@ -0,0 +1,34 @@
+import { transformOas2Operation, transformOas3Operation } from '@stoplight/http-spec';
+import { parse } from '@stoplight/yaml';
+import * as fs from 'fs';
+import { defaults, flatten, get, keys, map } from 'lodash';
+import { httpAndFileResolver } from '../resolvers/http-and-file';
+
+export default async function getHttpOperations(spec: string) {
```
philsturgeon: This looks like it assumes the provided spec is always a local file and not a URL?

packages/cli/package.json
XVincentX: We do not need axios specifically, but it's already installed because it's the library being used by the prism http client. I would refrain to add a second http client at the moment.

packages/cli/src/resolvers/http-and-file.ts
XVincentX: This is a copy paste from spectral, that's correct. I think the `json-ref-resolver` could have such utility out of the box @P0lip @marbemac; would you be open to that?

packages/cli/src/util/getHttpOperations.ts
XVincentX: You're right, there's a missing piece that I'm adding right now :)

packages/cli/src/resolvers/http-and-file.ts
P0lip: I don't think `json-ref-resolver` should provide out-of-the-box support for that.
That said, we could export such a resolver from `json-ref-resolver` pkg, but I'm not the owner of it, so it's not my call.

XVincentX: > we could export such a resolver from json-ref-resolver

Yeah sorry that was what I meant. Precanned resolver.

packages/cli/src/resolvers/http-and-file.ts
```diff
@@ -0,0 +1,37 @@
...
+          });
+        });
+      },
+    },
+  },
+
+  parseResolveResult: async opts => {
+    const ext = extname(opts.targetAuthority.toString());
+
+    if (ext === '.yml' || ext === '.yaml') {
```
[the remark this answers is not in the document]
XVincentX: I do not need nor the AST nor the metadata here.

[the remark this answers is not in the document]
P0lip: Then, you can take the sample above. Make sure to cover no yaml/json given case.

packages/cli/src/resolvers/http-and-file.ts
philsturgeon: Right, making them available from the package is what we are advocating for here, not changing the default resolver. The goal is to move it somewhere accessible for both Spectral and Prism so we don’t have to copy and paste.

P0lip: `json-ref-resolver` sounds like the right place.
@XVincentX mind opening the PR?

packages/cli/src/resolvers/http-and-file.ts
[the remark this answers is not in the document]
philsturgeon: Yeah the URL might be: http://foo.com/openapi and it should still work.

packages/cli/package.json
philsturgeon: I thought prism switched to fetch 😭 it should, then both should. 

There’s high level http clients and low level http clients. Fetch is low level, and prism should be high level - an alternative to axios. If we can fetch here then switch to fetch in the http client that would be excellent.

packages/cli/src/resolvers/http-and-file.ts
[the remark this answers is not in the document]
P0lip: > Yeah the URL might be: http://foo.com/openapi and it should still work.

Yeah, exactly. We mostly focused on file paths and they all usually contain extensions.
URLs on the other hand, may not contain any extension and we should still support it.
Using `parse` here, as shown above, is a safe bet.
If a different format than JSON/YAML is provided, I suspect it's safe to simply ignore the data?
We don't support other formats.

packages/cli/package.json
philsturgeon: The client change can be done while we’re refactoring

packages/cli/src/resolvers/http-and-file.ts
XVincentX: Hmm not too sure it's the right place, now that I think about it. Its job is really to resolve references — adding this capability would require to add an http client *and* the yaml parsing library to the project which I'd say clearly do not belong there. We should probably put this stuff somewhere else.

packages/cli/src/commands/mock.ts
```diff
@@ -10,8 +12,13 @@ const mockCommand: CommandModule = {
         description: 'Path to a spec file. Can be both a file or a fetchable resource on the web.',
         type: 'string',
       })
-      .fail(() => {
-        yargs.showHelp();
+      .middleware(async argv => {
+        argv.operations = await getHttpOperations(argv.spec!);
+      })
+      .fail((msg, err) => {
+        if (msg) yargs.showHelp();
```
XVincentX: This is basically to make sure that it will show help in case you're missing the argument (as @philsturgeon asked) and forward the error in case it's happening for any other reason (a thrown exception or something)

packages/cli/src/commands/mock.ts
```diff
@@ -46,15 +53,15 @@ const mockCommand: CommandModule = {
...
       }),
   handler: parsedArgs => {
-    const { multiprocess, dynamic, port, host, spec } = (parsedArgs as unknown) as CreatePrismOptions & {
+    const { multiprocess, dynamic, port, host, operations } = (parsedArgs as unknown) as CreatePrismOptions & {
       multiprocess: boolean;
     };
 
     if (multiprocess) {
-      return createMultiProcessPrism({ dynamic, port, host, spec });
+      return (parsedArgs.p = createMultiProcessPrism({ dynamic, port, host, operations }));
```
XVincentX: # Yes
## You're right!
### This is exactly what you think it is!

This is yet another test that has been passing forever incorrectly and I discovered it because I've removed 30192930189031283 layers that Prism was not using. We need a reference here to correctly wait the command to be processed before start doing assertions

packages/cli/src/util/createServer.ts
```diff
@@ -96,13 +91,9 @@ function logCLIMessage(message: string) {
...
-function isHttp(spec: string) {
-  return !!spec.match(/^https?:\/\//);
-}
-
 export type CreatePrismOptions = {
   dynamic: boolean;
   host?: string;
   port: number;
-  spec: string;
+  operations: IHttpOperation[];
```
XVincentX: You can see from here that Prism does not care about the source of the operations. You need to supply them somehow.

packages/core/src/factory.ts
```diff
@@ -5,38 +5,23 @@ import * as TaskEither from 'fp-ts/lib/TaskEither';
 import { configMergerFactory, PartialPrismConfig, PrismConfig } from '.';
 import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, PickRequired, ProblemJsonError } from './types';
 
-export function factory<Resource, Input, Output, Config, LoadOpts>(
+export function factory<Resource, Input, Output, Config>(
```
XVincentX: Since the loader is not a thing anymore in Prism, we have one less template argument to take care of.

packages/core/src/factory.ts
```diff
@@ -5,38 +5,23 @@ import * as TaskEither from 'fp-ts/lib/TaskEither';
...
-
-      load: async (opts?: LoadOpts): Promise<void> => {
-        const { loader } = components;
-        if (opts && loader) {
-          resources = await loader.load(opts, defaultComponents.loader);
-        }
-      },
-
-      process: async (input: Input, c?: Config) => {
+      process: async (input: Input, resources: Resource[], c?: Config) => {
```
XVincentX: Yes — Prism is now completely stateless and you need to supply the available resources every time you want to mock something.

The state will be held (if really necessary) by some external software (CLI, Studio, something else)

packages/http/src/__tests__/fixtures/server-validation.oas3.json
```diff
@@ -1,5 +1,5 @@
 {
-  "swagger": "3.0.0",
+  "openapi": "3.0.0",
```
XVincentX: Yeah, this test was broken because this is an invalid OpenAPI document. I fixed it.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -225,25 +241,21 @@ describe('Http Client .process', () => {
   describe('given no-refs-petstore-minimal.oas2.json', () => {
     beforeAll(async () => {
       prism = createInstance({ mock: { dynamic: false } }, { logger });
-      await prism.load({ path: noRefsPetstoreMinimalOas2Path });
-    });
-
-    it('keeps the instances separate', async () => {
```
XVincentX: We do not need to test this anymore! Prism is stateless! This can't happen!

packages/cli/src/util/createServer.ts
```diff
@@ -50,18 +44,19 @@ export function createSingleProcessPrism(options: CreatePrismOptions) {
 }
 
 async function createPrismServerWithLogger(options: CreatePrismOptions, logInstance: Logger) {
-  const server = createServer(
-    options.spec,
-    { mock: { dynamic: options.dynamic } },
-    logInstance.child({ name: 'HTTP SERVER' }),
-  );
+  if (options.operations.length === 0) {
```
XVincentX: This will be refactored in the future; I can force TypeScript to make sure that the operations is *at least* an array with an element. Not now though.

packages/cli/src/util/getHttpOperations.ts
```diff
@@ -0,0 +1,43 @@
...
+        .filter(k => methods.includes(k))
+        .map(method =>
+          defaults(
+            transformOperationFn({
+              document: resolvedContent,
+              path,
+              method,
+            }),
+            { servers: resolvedContent.servers },
+          ),
```
XVincentX: @marbemac Is this ok?

packages/cli/src/resolvers/http-and-file.ts
philsturgeon: It depends what we think resolving is. Resolving for many people is literally just the lookup, not the loader. JSON Ref Resolver right now is a resolve + file loader. 

File loading and URL loading is both loading, so maybe we should make a new package for some common loaders we have, and let people define their own? 

So they could use our resolver with their own loader, or their own resolver entirely... this is a bigger job but if we think that direction is viable, we can out this copy paste in now with a story on the backlog to pay off the tech debt later. 

Also interested in @marc and @P0lip opinions here

XVincentX: > File loading and URL loading is both loading, so maybe we should make a new package for some common loaders we have, and let people define their own?

Correct, this is the direction I think we should go. Can we leave it as a next step?

philsturgeon:
absolutely, I’m eating lunch but would you make a story for before we forget on prism here and we can do the work on tech debt week. 

--
Phil Sturgeon
@philsturgeon

XVincentX:
You got it.

packages/cli/src/resolvers/http-and-file.ts
chris-miaskowski: Hey @XVincentX I think it makes sense to extract this to its own package. Can you create an issue for that?

XVincentX: We did it already, #493

P0lip: @XVincentX @philsturgeon 
> File loading and URL loading is both loading, so maybe we should make a new package for some common loaders we have, and let people define their own?

Yeah, a new package might be in fact a better choice.
Bear in mind that Spectral comes with its **own** `parseResolveResult` method, so it won't benefit as much as it might seem.
That said, I'm still leaning towards extracting those resolvers somewhere.

packages/cli/src/util/getHttpOperations.ts
```diff
@@ -0,0 +1,43 @@
+import { transformOas2Operation, transformOas3Operation } from '@stoplight/http-spec';
```
chris-miaskowski: Logic in this file should be extracted to the same repo all that ref resolving stuff will go. Doesn't make sense to keep it here imo.

chris-miaskowski: Curious, are you removing that logic from Graphite as well?

chris-miaskowski: If you don't want to move this code to individual repo then may consider moving to a different package? The JS API users will want to use this function too. Keeping this in `CLI` probably isn't the best place.

packages/cli/src/util/getHttpOperations.ts
chris-miaskowski: @XVincentX why are you providing `servers`? `transformOperationFn` should handle it for you IIRC.

packages/cli/src/util/getHttpOperations.ts
```diff
@@ -0,0 +1,43 @@
...
+              document: resolvedContent,
+              path,
+              method,
+            }),
+            { servers: resolvedContent.servers },
+          ),
+        ),
+    ),
+  );
+}
```
chris-miaskowski: I see you removed loader tests but haven't covered this file yet. Are you planning to do it?

packages/http-server/src/server.ts
```diff
@@ -9,10 +10,7 @@ import * as typeIs from 'type-is';
 import { getHttpConfigFromRequest } from './getHttpConfigFromRequest';
 import { IPrismHttpServer, IPrismHttpServerOpts } from './types';
 
-export const createServer = <LoaderInput>(
-  loaderInput: LoaderInput,
-  opts: IPrismHttpServerOpts<LoaderInput>,
-): IPrismHttpServer<LoaderInput> => {
+export const createServer = (loaderInput: IHttpOperation[], opts: IPrismHttpServerOpts): IPrismHttpServer => {
```
chris-miaskowski:
```suggestion
export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServerOpts): IPrismHttpServer => {
```

packages/cli/src/util/getHttpOperations.ts
XVincentX: It's not happening; I agree that's weird, but I've been checking the code and effectively such function is not doing anything with regards of the `servers`. I'll try to dig into more — in meantime maybe @marbemac knows what's going on here?

packages/cli/src/util/getHttpOperations.ts
XVincentX: 1. I'm not removing this from Graphite — I think this glue code is still used there somewhere. Also, this is not an exact copy and paste of what's in Graphite.
2. Probably you're right, this is not the right place. I *could* bring it in `core` but I'd like to avoid it if possible; in general I want people to reason about Prism with arrays of `HttpOperation`, and not with files, web resources and stuff like that.

packages/cli/src/util/getHttpOperations.ts
XVincentX: Yes, I should definitely write some test for this function. I just wanted at least another set of eyes confirming that the implementation is correct, since I'm not 100% sure. I mean harness tests pass which is great but still…

packages/http-server/src/server.ts
XVincentX: Good point. Done!

dinfyru:
Approve it :)

packages/cli/src/util/getHttpOperations.ts
XVincentX: You were right, the server transformation is happening. I removed the code. @chris-miaskowski

packages/cli/package.json
```diff
@@ -7,8 +7,12 @@
   },
   "bugs": "https://github.com/stoplightio/prism/issues",
   "dependencies": {
+    "@stoplight/http-spec": "^1.0.4",
+    "@stoplight/json-ref-resolver": "^2.0.3",
     "@stoplight/prism-core": "^3.0.3",
     "@stoplight/prism-http-server": "^3.0.3",
+    "@stoplight/yaml": "^2.8.0",
+    "axios": "^0.18.0",
```
philsturgeon: Seems like we're removing node-fetch and adding axios, but last time we talked we were going to remove axios and keep node-fetch? Was that a miscommunication or were there some new reasons?

XVincentX: Ah, I thought this was going to be a separate PR — do you want to couple the node-fetch replacement here?

Wouldn't it be better to have a dedicated PR?

packages/cli/src/util/getHttpOperations.ts
chris-miaskowski: Sure, it's not a big deal!

packages/cli/package.json
philsturgeon: Ok if you make a story for it I’ll approve this.

XVincentX:
KABOOM!

dinfyru:
wait f4 new version :)

XVincentX:
@dinfyru I do not think we're going to release this that soon. If you're really in haste you should either download the source code or download out "binaries" done from the master branch

https://circleci.com/gh/stoplightio/prism/3277#artifacts/containers/0

PR #491, opened 2019-07-25
feat: respond with xml

### LINKED ISSUE #424: Basic XML Rendering / Validation
**User Story**

As a Prism HTTP Server user, I want to be able to send `Content-Type: application/xml` and/or `Accept: application/xml` and see XML in the response, for any HTTP operation which supports XML.

**Details**

OpenAPI v2 and v3 support basic XML shapes through the same `schema` object as JSON, so you can reference the same models in different requests/responses.

Screen Shot 2019-07-15 at 20 12 58

We need to make sure this works in Prism, for content type validation, for response generation, everything. This story can stop short of supporting [the `xml` keyword](https://swagger.io/docs/specification/data-models/representing-xml/) though, as this will require a whole lot more work. That further work for the `xml` object will be done in #456

### LINKED ISSUE #492: Prism is not able to handle correctly the */* Accept Header
The petstore example from https://petstore.swagger.io/v2/swagger.json is not working with prism v3.

**To Reproduce**

1. Start prism
```
$ prism mock https://petstore.swagger.io/v2/swagger.json
[CLI] …  awaiting  Starting Prism…
[HTTP SERVER] ℹ  info      Server listening at http://127.0.0.1:4010
[CLI] ●  note      POST       http://127.0.0.1:4010/pet
[CLI] ●  note      PUT        http://127.0.0.1:4010/pet
[CLI] ●  note      GET        http://127.0.0.1:4010/pet/findByStatus
[CLI] ●  note      GET        http://127.0.0.1:4010/pet/findByTags
[CLI] ●  note      GET        http://127.0.0.1:4010/pet/{petId}
[CLI] ●  note      POST       http://127.0.0.1:4010/pet/{petId}
[CLI] ●  note      DELETE     http://127.0.0.1:4010/pet/{petId}
[CLI] ●  note      POST       http://127.0.0.1:4010/pet/{petId}/uploadImage
[CLI] ●  note      GET        http://127.0.0.1:4010/store/inventory
[CLI] ●  note      POST       http://127.0.0.1:4010/store/order
[CLI] ●  note      GET        http://127.0.0.1:4010/store/order/{orderId}
[CLI] ●  note      DELETE     http://127.0.0.1:4010/store/order/{orderId}
[CLI] ●  note      POST       http://127.0.0.1:4010/user
[CLI] ●  note      POST       http://127.0.0.1:4010/user/createWithArray
[CLI] ●  note      POST       http://127.0.0.1:4010/user/createWithList
[CLI] ●  note      GET        http://127.0.0.1:4010/user/login
[CLI] ●  note      GET        http://127.0.0.1:4010/user/logout
[CLI] ●  note      GET        http://127.0.0.1:4010/user/{username}
[CLI] ●  note      PUT        http://127.0.0.1:4010/user/{username}
[CLI] ●  note      DELETE     http://127.0.0.1:4010/user/{username}
[CLI] ▶  start     Prism is listening on http://127.0.0.1:4010
```

2. accessing prism via httpie returns an empty response
```
$ http GET http://127.0.0.1:4010/pet/123
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 0
Date: Fri, 26 Jul 2019 09:45:11 GMT



```
3. this error is shown in the prism log
```
[HTTP SERVER] get /pet/123 ℹ  info      Request received
    [NEGOTIATOR] ℹ  info      Request contains an accept header: */*
    [VALIDATOR] ✔  success   The request passed the validation rules. Looking for the best response
    [NEGOTIATOR] ✔  success   Found a compatible content for */*
    [NEGOTIATOR] ✔  success   Responding with the requested status code 200
    [HTTP SERVER] get /pet/123 ✖  error     Request terminated with error: FastifyError [FST_ERR_REP_INVALID_PAYLOAD_TYPE]: FST_ERR_REP_INVALID_PAYLOAD_TYPE: Attempted to send payload of invalid type 'object'. Expected a string or Buffer.

```
**Expected behavior**
I would have expected to get a mocked response.

**Environment:**
 - prism version: 3.0.3
 - OS: Ubuntu Ubuntu 19.04

lag-of-death:
Closes #492 
Closes #424


packages/http-server/src/server.ts
```diff
@@ -19,24 +55,7 @@ export const createServer = <LoaderInput>(
     logger: (components && components.logger) || createLogger('HTTP SERVER'),
     disableRequestLogging: true,
     modifyCoreObjects: false,
-  })
```
lag-of-death: If `application/json` was first in `content-type` from request (like `application/json,application/xml`), it was ignored if there was a manually added serializer for `application/xml`.
Since the serializer for `JSON`, was the default one, I decided to follow that path and not add a serializer for JSON to `serializers`. Instead I moved out serializers out of `fastify`.

packages/http-server/src/server.ts
```diff
@@ -19,24 +55,7 @@ export const createServer = <LoaderInput>(
...
-    .register(formbodyParser)
-    .register(fastifyAcceptsSerializer, {
```
lag-of-death: If `application/json` was first in `content-type` from request (like `application/json,application/xml`), it was ignored if there was a serializer for `application/xml`.
Since the serializer for `JSON`, was the default one, I decided to follow that path and not add a serializer for JSON to `serializers`. Instead I moved out serializers out of `fastify`.

lag-of-death:
@XVincentX, could you take a look at this [wip] PR? I'm not sure if the validation is going in the right direction and if it is indeed needed for #424.

packages/http/src/validator/validators/body.ts
```diff
@@ -16,7 +41,7 @@ export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent>
     }
 
     if (!content.schema) {
-      return [];
+      return checkExamples(target, mediaType);
     }
```
philsturgeon: I’m not sure I understand this bit. Why are we changing how things work when no examples are found?

lag-of-death: @philsturgeon, I'm not sure I understand. The above says: `if there is no schema in the content, then please check *target*, which is response body, which is a user provided example`. Or at least, this is my understanding. And also, as I understand, there was no validation on outputs, only inputs were validated. Output validations were always passed as a `Right` (`Either`).

packages/http-server/src/server.ts
XVincentX: Ok I'm going to check out because this is really weird. The serialisers should work according to the header preference and not the order of installation. I'll check it out and write it back.

packages/core/src/factory.ts
```diff
@@ -113,19 +113,24 @@ export function factory<Resource, Input, Output, Config, LoadOpts>(
...
+                        resource,
+                        output,
+                        config: configObj,
+                      },
+                      defaultComponents.validator,
+                    )
+                  : [];
 
+              return outputValidations.length
+                ? TaskEither.left(outputValidations[0] as any)
```
XVincentX: Nice stuff! You're getting into it!

1. Why are you returning only the first element?
2. We should really avoid the `as any` and convert the output validation in a real error somehow, otherwise this is probably not going to work.

packages/http/src/validator/validators/body.ts
XVincentX: I think @lag-of-death is right, although this is probably not the correct PR to make this happen.

If I understand correctly — the idea is that the example specified by the user shall be validated before being sent to the client.

For instance

```yaml
example:
  application/json: apodkosakdospakd
  application/xml: aisjaiw1023
```

These are probably *not* valid; we're still returning them and send them with their respective headers. I think @lag-of-death was trying to fix this by validating the examples; I think this is something that Spectral should do.

@lag-of-death Did I understand your intention?
@philsturgeon Does it clarify? If so, do you think this is something that Prism should do?

In any case maybe we could create a separate issue to work on this instead of melting the changes together.

packages/core/src/factory.ts
lag-of-death: @XVincentX, I think that when input is validated, it also returns one (first?) element. So that, at the end, there is `throw e` in that last `Left` handling function. So I thought this must be intentional (?).

lag-of-death:
@XVincentX, about https://github.com/stoplightio/prism/pull/491#discussion_r307684244, yes, that was my intention :) :+1:

packages/http/src/validator/validators/body.ts
philsturgeon: > I think @lag-of-death is right, although this is probably not the correct PR to make this happen.

That explains it! I was trying to work out how it was related to the feature we are implementing, and its not. Okie doke! 

@lag-of-death could you make a story outlining what is going wrong, what this aims to address, the whole thing, then we can decide if we'd like to work on it later.

lag-of-death: not so lengthy, but I think to the point issue created: https://github.com/stoplightio/prism/issues/495 LMK if there's still some confusion/anything :+1:
cc @XVincentX, @philsturgeon

PR #501, opened 2019-07-29
Fix/oas23 body param validation nographite

### LINKED ISSUE #408: Test Harness: OAS2/3 Body Param Validation

### LINKED ISSUE #483: OAS2 invalid form data not validated correctly
**Describe the bug**

Prism fails to handle an invalid OAS2 form data parameters.

**To Reproduce**

(see https://github.com/stoplightio/prism/pull/482)

```
====test====
Given form data expected in request body then return 422 when that body is invalid.
====spec====
swagger: '2.0'
paths:
  /path:
    post:
      consumes:
      responses:
        200:
          schema:
            type: string
      parameters:
        - in: formData
          type: integer
          name: id
          required: true
        - in: formData
          type: string
          name: status
          required: true
          enum:
            - open
            - close
====server====
mock -p 4010
====command====
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data-urlencode "status=ooopsie!"
====expect====
Connection: keep-alive

{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"code":"required","message":"should have required property 'id'","severity":"Error"},{"code":"enum","location":["body","status"],"message":"should be equal to one of the allowed values","severity":"Error"}]}
```

**Actual**

```
 "{\"type\":\"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY\",\"title\":\"Invalid request body payload\",\"status\":422,\"detail\":\"Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.\",\"validation\":[{\"location\":[\"body\"],\"severity\":\"Error\",\"code\":\"type\",\"message\":\"should be integer\"}]}"
```

 - Library version: [e.g. 3.0.1]

**Additional context**

Most likely the issues must be fixed in Graphite

Failing integration tests:

```
oas2 with form data param
      required parameter not in body
        ✕ returns 422 (21ms)
      parameter does not match enum criteria
        ✕ returns 422 & proper validation message (17ms)
      valid parameter provided
        ✕ returns 200 (13ms)
```

PR: https://github.com/stoplightio/http-spec/pull/9

chris-miaskowski:
Closes #408
Closes #483


packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -257,8 +201,8 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
     // accorging to the schema
 
     const expectedValues = {
-      'x-rate-limit': file === 'petstore.oas3.json' ? 1000 : expect.any(Number),
-      'x-stats': file === 'petstore.oas3.json' ? 1500 : expect.any(Number),
+      'x-rate-limit': file === 'petstore.oas3.yaml' ? 1000 : expect.any(Number),
```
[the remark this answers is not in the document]
chris-miaskowski: Yup

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -194,40 +172,6 @@ describe.each([['petstore.oas2.json'], ['petstore.oas3.json']])('server %s', fil
     expect(payload).toHaveProperty('userStatus');
   });
 
-  it('should validate body params', async () => {
```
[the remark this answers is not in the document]
chris-miaskowski: Yes, exactly, I moved all to the `body-param-validation-etc` file, makes more sense imo to group everything related to body validation in one place.

test-harness/index.ts
```diff
@@ -1,20 +1,24 @@
...
+import * as tmp from 'tmp';
+import { parseSpecFile } from './helpers';
 
-jest.setTimeout(60000);
+const glob = globFs({ gitignore: true });
+jest.setTimeout(5000);
 
 describe('harness', () => {
-  const files = fs.readdirSync(path.join(__dirname, './specs/'));
+  const files = process.env.TESTS
```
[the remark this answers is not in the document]
chris-miaskowski: Yes, or multiple test files separated by a comma.

jest.config.js
```diff
@@ -9,6 +9,9 @@ const projectDefault = {
   transform: {
     '^.+\\.(ts)$': 'ts-jest',
   },
+  setupFilesAfterEnv: [
+    '<rootDir>/test-utils/index.js'
```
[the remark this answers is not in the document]
chris-miaskowski: Good spot!

test-harness/index.ts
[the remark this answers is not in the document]
XVincentX: Ok nice, can you update the documentation? Test Harness has a readme.md file — it would be cool to put both the env variable as well as the glob rules so we know how to use this stuff.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+            ],
+          });
+        });
+      });
+    });
+
+    describe('operation with required property', () => {
+      const operation: fastify.HTTPInjectOptions = {
+        method: 'POST',
+        url: '/json-body-property-required',
```
XVincentX: I'm afraid this test is wrong. The `required` property is false by default, and according to the definition file

```yaml
 "/json-body-property-required":
    post:
      consumes:
      - application/json
      parameters:
      - in: body
        name: body
        schema:
          type: object
          properties:
            id:
              type: integer
              format: int64
          required:
          - id
      responses:
        '200':
          schema:
            type: string
```

This should not return a validation error. Is this a problem in `http-spec`?

packages/http-server/src/__tests__/fixtures/encodings.oas3.yaml
```diff
@@ -0,0 +1,21 @@
...
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              properties:
+                reserved:
+                  type: string
+            encoding:
+              reserved:
+                contentType: text/plain
+                allowReserved: false
```
XVincentX: Missing newline

test-harness/specs/validate-body-params/form-byte-format-fail.oas2.txt
```diff
@@ -0,0 +1,27 @@
...
+            type: string
+      parameters:
+        - in: formData
+          type: string
+          format: byte
+          name: id
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "id=!#$!%"
```
XVincentX: Can we simplify the example?

`%` should just do 
```suggestion
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "id=%"
```

test-harness/specs/validate-body-params/form-data-encoding-contenttype-fail.oas3.txt.unsupported
```diff
@@ -0,0 +1,33 @@
...
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "json=123"
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+content-type: application/problem+json
+Connection: keep-alive
+
+{"tobe": "done"}
```
XVincentX: New line

```suggestion
{"tobe": "done"}

```

test-harness/specs/validate-body-params/form-data-invalid-request.oas3.txt
```diff
@@ -0,0 +1,39 @@
+====test====
+Given form data expected in request body 
+but invalid data sent then return 422
```
XVincentX: I think we should be more specific instead of just saying invalid data; what is the invalid part that we're sending?

test-harness/specs/validate-body-params/form-dont-allow-empty-values.oas2.txt.unsupported
```diff
@@ -0,0 +1,30 @@
...
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "id"
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+content-type: application/problem+json
+Connection: keep-alive
+
+{ "tobe": "done" }
```
XVincentX: New line

test-harness/specs/validate-body-params/form-dont-allow-empty-values.oas2.txt.unsupported
```diff
@@ -0,0 +1,30 @@
...
+        - application/x-www-form-urlencoded
+      responses:
+        200:
+          schema:
+            type: string
+      parameters:
+        - in: formData
+          type: string
+          name: id
+          allowEmptyValue: false
```
XVincentX: Why didn't we fix this in https://github.com/stoplightio/http-spec/pull/9?

test-harness/specs/validate-body-params/form-data-encoding-contenttype-ok.oas3.txt.unsupported
```diff
@@ -0,0 +1,29 @@
...
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              properties:
+                json:
+                  type: object
+            encoding:
+              json:
+                contentType: application/json
```
XVincentX: Man, these quirks of OpenAPI 3.0 really make me feel so sad.

test-harness/specs/validate-body-params/form-data-encoding-contenttype-fail.oas3.txt.unsupported
XVincentX: What is this payload supposed to be anyway?

test-harness/specs/validate-body-params/form-data-allow-reserved-fail.oas3.txt.unsupported
```diff
@@ -0,0 +1,35 @@
...
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "reserved=:/?#[]@!$&'()*+,;"
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+content-type: application/problem+json
+Connection: keep-alive
+
+{ "to-be": "done" }
```
XVincentX: new line

test-harness/specs/validate-body-params/form-allow-empty-values.oas2.txt.unsupported
```diff
@@ -0,0 +1,37 @@
...
+        - in: formData
+          type: string
+          name: id
+          allowEmptyValue: true
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "id"
+====expect====
+HTTP/1.1 200 OK
```
XVincentX: New line

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+        payload: {
+          reserved: ":/?#[]@!$&'()*+,;",
+        },
+      });
+
+      expect(response.statusCode).toEqual(422);
+    });
+  });
+
+  describe.each([['oas2'], ['oas3']])('%s with body param', oas => {
```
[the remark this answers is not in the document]
philsturgeon: Integration tests are for making sure values are being passed around correctly and the boundaries between code work properly. E2E is more overview “does the whole thing work”, but struggles with intricate assertions or multiple similar-but-slightly-tweaked scenarios like this.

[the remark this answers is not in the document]
philsturgeon: Aaaactually yeah this is a bit tough to see what the test is doing because it’s all coming from a file. I see what you mean about feeling like an E2E test @XVincentX, this might be a bit too zoomed out.

packages/http-server/src/__tests__/body-params-validation.spec.ts
philsturgeon: @XVincentX not quite, the body itself is optional, but of present it must have the id property

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
chris-miaskowski: I did this for one particular reason to be honest: debugging :) It's a PAIN to debug test harness but with these tests I was able to spot the issue in a matter of minutes. 

Yup it's all coming from a file because of how the loader works (loads stuff from files :) )
It would be great to have this stuff inline I agree, maybe we'll refactor at some point later when we support 'passing pure specs'.

E2E means testing, like Phil said, is testing if the **whole thing works** from a user perspective. Since Prism is both an `exe` file and and API e2e is a thin line. I would like to keep those tests in though because they test more than what I ended up with in test-harness.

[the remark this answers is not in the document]
philsturgeon: @chris-miaskowski I am not advocating for removing these tests, I am advocating for making them more readable by doing less. Instead of using the loader surely we can just pass in operations, right? that's the benefit of not requiring graphite, we dont have to load a file we can pass in operations. 

Then this becomes more of.an integration test, because we are checking to see how operations can effect the various classes involved which result in a HTTP response being generated. And the readability of each single test will be greatly improved because you dont have to open up another file to read what its trying to do. win win win.

test-harness/index.ts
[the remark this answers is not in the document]
chris-miaskowski: done

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,277 @@
+import getHttpOperations from '@stoplight/prism-cli/src/util/getHttpOperations';
+import { createLogger } from '@stoplight/prism-core';
+import fastify = require('fastify');
```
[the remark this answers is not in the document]
chris-miaskowski: No idea :) doesnt look like autocomplete tho. Will change that, thanks!

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+  const path = relative(process.cwd(), resolve(__dirname, 'fixtures', fixture));
+  const operations = await getHttpOperations(path);
+  const server = createServer(operations, { components: { logger }, config: { mock: { dynamic: false } } });
+  return server;
+}
+
+describe('body params validation', () => {
+  let server: IPrismHttpServer;
+
+  afterAll(async () => {
```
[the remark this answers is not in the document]
chris-miaskowski: good point, fixed

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
chris-miaskowski: @philsturgeon ok gotcha. Sounds like a good idea, **but...**! The problem with using pure http-operations is that we'll not spot issues related to transforming oas2/3 to http-operations.

These tests were intended to test the actual specs (vs the abstracted http-operation). I agree that doing the latter will match more to the definition of integration test but at the same time will cover less stuff.

packages/http-server/src/__tests__/body-params-validation.spec.ts
chris-miaskowski: @XVincentX like @philsturgeon said. This test is checking if (when optional body is passed) we verify that a required **property** is passed.

I pass empty body `{}` but spec expects `{id:123}`. Therefore we get validation error - which is correct.

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
philsturgeon: @chris-miaskowski I understand. This might sound like the concern is about purity of tests to a textbook definition, but its more that these conventions exist for a reason, a reason that we would benefit from.  "Covering more stuff" is one way to look at it, but another way would be "less targeted and therefore less clear about what we're actually trying to test."

E2E is for catching overall problems, integration tests should be trying to put code into the appropriate state to trigger the reaction expected. More tests can be written to target the other specific things, but trying to do it all at once just means we are making confusing tests that are spooky to touch because what are we asserting. 

Accidental coverage should not be our goal here.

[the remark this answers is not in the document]
chris-miaskowski: @philsturgeon let's not forget about debuggability. We can ditch the files, convert all specs to http-operations and call it a day. We loose the subtle differences between oas2 and oas3 this way which we're hoping to get covered by e2e. In reality though it's much slower to cover them with e2e than this "semi-integration semi whatever" test. 

I'm not going to try to defend this though :) If you folks think it makes more sense to loose some coverage and ease of development in favor of readability then let it be.

@XVincentX @philsturgeon if you both give thumbs up I will convert :)

test-harness/specs/validate-body-params/form-data-encoding-contenttype-fail.oas3.txt.unsupported
chris-miaskowski: Good question :) Not sure that's why I left if as TODO. I guess it would be a new type of error? E.g. something along the lines of "expected json received string"?

XVincentX:
@chris-miaskowski also — do not forget to update the CHANGELOG.md

test-harness/specs/validate-body-params/form-dont-allow-empty-values.oas2.txt.unsupported
chris-miaskowski: I don't know unfortunately. I hoped it would get fixed but it didn't there must be some other quirk here but I decided not to follow up for now. I created a bug report IIRC so that I or someone else can look at it later.

test-harness/specs/validate-body-params/form-data-encoding-contenttype-ok.oas3.txt.unsupported
chris-miaskowski: :)

test-harness/specs/validate-body-params/form-data-invalid-request.oas3.txt
chris-miaskowski: I tried to be less verbose as per @philsturgeon suggestions :) I will add some context.

test-harness/specs/validate-body-params/form-dont-allow-empty-values.oas2.txt.unsupported
XVincentX: Yeah do not worry — it is effectively better to handle it as a separate task.

test-harness/specs/validate-body-params/form-data-encoding-contenttype-fail.oas3.txt.unsupported
XVincentX: OOOOOOOOH MAN I UNDERSTOOD IT NOW

TO BE DONE :trollface: :trollface: I didn't get it at the beginning.

This is ok.

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
philsturgeon: @chris-miaskowski unit tests on transformers, integration tests that are aimed at that specific goal, and the end to end tests cover this. Again, accidental coverage should not be our goal here.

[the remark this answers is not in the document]
XVincentX: I stand with @philsturgeon here — if the observability/debug experience is the main issue of writing them here, then we might need to address this somehow.

One of the things that I think would be beneficial would be a flag where you can either run the Harness on the final executable **OR** on the "source code" directly. This should improve the debuggbility significantly. Does that address your concern?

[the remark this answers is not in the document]
chris-miaskowski: @XVincentX @philsturgeon updated that test. Looks good now?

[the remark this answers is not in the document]
chris-miaskowski: > One of the things that I think would be beneficial would be a flag where you can either run the Harness on the final executable OR on the "source code" directly. This should improve the debuggbility significantly. Does that address your concern?

I don't think we should do that. E2E should always run on binary. We have `jest` to run on source code.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -0,0 +1,520 @@
...
+                },
+              ],
+            });
+          });
+        });
+      });
+    });
+  });
+
+  describe.each([['oas2'], ['oas3']])('%s with form data param', oas => {
```
philsturgeon: Do we  see special oas 2 and 3 variances now that we’re just using operations directly?

chris-miaskowski: Omg I freaking missed it :P fixing asap

chris-miaskowski: @philsturgeon done

packages/http-server/src/__tests__/fixtures/operations-with-body-param.oas2.yaml
```diff
@@ -0,0 +1,80 @@
+swagger: '2.0'
+paths:
+  "/json-body-no-request-content-type":
+    post:
+      parameters:
+      - in: body
```
philsturgeon: Are these fixtures all still required?

chris-miaskowski: deleted :+1:

Issue #509, opened 2019-08-03
Prism is not respecting AND relation in security schemes


lag-of-death:
**Describe the bug**

`transformOperationFn` (https://github.com/stoplightio/prism/blob/master/packages/cli/src/util/getHttpOperations.ts#L38) returns flattened security schemes. For example, if a spec contains:
```yaml
security:
  - ApiKeyAuth: []
    ApiKeyID: []
```

then `transformOperationFn` will give back `security` that is an array of 2 elements, not an array of one element containing these two security schemes.

Hence, the result says it is `ApiKeyAuth OR ApiKeyID` and not (what was intended) `ApiKeyAuth AND ApiKeyID`.

**To Reproduce**

Since security scheme validation is not yet in place, the bug is not visible to the end user.
To reproduce take a spec from one of the provided e2e examples (down below) and take a look at the return value here https://github.com/stoplightio/prism/blob/master/packages/cli/src/util/getHttpOperations.ts#L33

**Expected behavior**
`transformOperationFn` shouldn't flatten security schemes. It should be clear from `transformOperationFn` returned value that some schemes are connected with `AND` relation.

***Possibly required changes include modules such as***
`types, http-spec, prism, graphite`

***Code to investigate***
https://github.com/stoplightio/http-spec/blob/master/src/oas3/accessors.ts#L11

**Environment (remove any that are not applicable):**
 - Prism version: 3.0.0
 - OS: Linux Ubuntu 18.04 KDE Neon 5.16

**Additional context**
E2E example tests for the bug:
***OAS2***
```
====test====
Providing two required API KEYS grants access to a resource.
====spec====
swagger: '2.0'
paths:
  /todos:
    get:
      produces:
        - application/json
      responses:
        200:
          description: OK
          examples:
            application/json: test
securityDefinitions:
  ApiKeyOne:
    type: apiKey
    in: header
    name: X-API-Key
  ApiKeyTwo:
    type: apiKey
    in: query
    name: apiKey
security:
  - ApiKeyOne: []
    ApiKeyTwo: []
====server====
mock -p 4010
====command====
curl -i http://localhost:4010/todos?apiKey=abc123 -H "X-API-Key: xyz987"
====expect====
HTTP/1.1 200 OK

"test"
```
***OAS3***
```
====test====
Providing two required API KEYS grants access to a resource.
====spec====
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: OK
          content:
            application/json:
              example: test
[...]

PR #514, opened 2019-08-07
[WIP] AND relations in security schemes

lag-of-death:
## What does this PR solve?
https://github.com/stoplightio/prism/issues/509

## Dependencies

This PR depends on:

https://github.com/stoplightio/prism/blob/9cdb9471ff8eaacf2c4157c7ae2b83fb34b206b8/packages/cli/package.json#L10

https://github.com/stoplightio/prism/blob/55d4f32e32c613834006f2773c6494711b062053/package.json#L27

## Changes

Temporarily (untill `http-spec` branch is merged and built) depends on dist from `http-spec`:
https://github.com/stoplightio/prism/blob/c4bbe4380c2be2733bb5464f9c7551e3ca090328/packages/cli/src/util/getHttpOperations.ts#L1

`types` branch also has `dist` folder checked in, but `prism` does not depend on it.

**Why github branches and no `yalc`? I was trying `yalc` first, but were having issues with nested `.yalc` folders (like `.yalc/@stoplight/something/.yalc/@stoplight/something_else`). Yarn, for some reason, couldn't have resolved dependencies.**

## Order of merging:
1. `@stoplight/types` (https://github.com/stoplightio/types/pull/50)
2. `@stoplight/json-ref-resolver` (https://github.com/stoplightio/json-ref-resolver/pull/86)
3. `@stoplight/http-spec` (https://github.com/stoplightio/http-spec/pull/15) and `@stoplight/spectral` (https://github.com/stoplightio/spectral/pull/444)
4. `@stoplight/graphite` (https://github.com/stoplightio/graphite/pull/211) and `@stoplight/prism` (https://github.com/stoplightio/prism/pull/514)


packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -329,4 +329,65 @@ describe('validateSecurity', () => {
       });
     });
   });
+
+  describe('AND relation between security schemes', () => {
+    describe('when given only one key', () => {
```
philsturgeon: I see you're trying to use describe for "context" and it's good that you're trying, but this seems a bit off. Generally the context is to provide meaningful setup for multiple tests. The action is not in the context, so "when given one key" is kinda in the wrong place.

I would do it more like this:

```
  describe('AND relation between security schemes', () => {
    describe('when security scheme expects two keys', () => {
      const securityScheme = [
        [
          {
            in: 'header',
            type: 'apiKey',
            name: 'x-api-key',
          },
          {
            in: 'query',
            type: 'apiKey',
            name: 'apiKey',
          },
        ],
      ];

      it('fails with an invalid security scheme error', () => {
        expect(
          validateSecurity<any, any>(
            { headers: { 'x-api-key': 'abc123' } },
            { security: securityScheme },
          ),
        ).toStrictEqual([
          {
            headers: {},
            message: 'Invalid security scheme used',
            name: 'Unauthorised',
            status: 401,
          },
        ]);
      });
    });
```

This has a few benefits. 

1. setup is seperate from assertion, so its easier to read what is going on, 
2. people can easily add similar tests/assertions using the same setup, reducing accidental false positives of badly copied and pasted code
3. the describe is describing itself :)

Can you have a go at this?

packages/core/src/utils/__tests__/security.spec.ts
```diff
@@ -329,4 +329,65 @@ describe('validateSecurity', () => {
...
+      test.only('fails with an invalid security scheme error', () => {
+        expect(
+          validateSecurity<any, any>(
+            { headers: { 'x-api-key': 'abc123' } },
+            {
+              security: [
```
philsturgeon: This could use the same `const securtyScheme` as above, using contexts to reduce duplication! 🙌

packages/core/src/utils/security/index.ts
```diff
@@ -1,54 +1,86 @@
...
-            return handler.test(definedSecSchema);
+        const authResults = securitySchemes.map((securityScheme: SecurityScheme[]) => {
+          const authResult = securityScheme.map((definedSecSchema: SecurityScheme) => {
+            const schemeHandler = securitySchemaHandlers.find(handler => {
+              return handler.test(definedSecSchema);
+            });
+
+            return schemeHandler
+              ? schemeHandler.handle(someInput, definedSecSchema.name, resource)
+              : left({ message: 'We currently do not support this type of security scheme.' });
```
philsturgeon: @XVincentX hey, do we want to hard error here or let it pass? I think we'd rather keep out the way for things we dont know how to do, and only shout at users when we know they are wrong.

test-harness/specs/security-scheme-validation-AC-24.oas2.txt
```diff
@@ -0,0 +1,34 @@
...
+  - ApiKeyOne: []
+    ApiKeyTwo: []
+====server====
+mock -p 4010
+====command====
+curl -i http://localhost:4010/todos?apiKey=abc123 -H "X-API-Key: xyz987"
+====expect====
+HTTP/1.1 200 OK
+
+"test"
```
philsturgeon: we dont always need to have a json string output at the end, a empty 200 OK should be fine. If you remove the JSON and the example it'll flop our a text/plan 200 OK right?

lag-of-death: @Phil, yes, but I wanted to keep `oas2` test and `oas3` test as close as possible. And since doing what you described for `oas3` results in `406` response, I went with an example JSON response for the both tests. Hope it makes sense

philsturgeon: I don't follow. I'm saying this test is about security scheme validation, not about JSON examples, so im not sure why we've added JSON examples?

lag-of-death: Hm, so you're saying that if the response is the default `text/plain`, then all of a sudden, it only is about security scheme validation and not about `text/plain`? How is it different that having `text/plain` default response is not testing this part of Prism and only testing security scheme validation? But, having said that... as I said, I wanted to have `test-harness/specs/security-scheme-validation-AC-24.oas2.txt` and `test-harness/specs/security-scheme-validation-AC-24.oas3.txt` (and all other `oas2`/`oas3` pairs) as close as possible.

So, how do I make `oas3` example returning `text/plain` default answer? If I adjust `oas3` file the was I adjust it for `oas2`, I get `406`.

philsturgeon: I'm saying we dont need to add extra lines to the test in order to create a JSON example which we dont need. By removing irrelevant things we slim the test down to only relevant things... I'm not sure I understand your concerns here, or what you're getting at. I was just asking if we could remove irrelevant things.

lag-of-death: @philsturgeon, here's the change https://github.com/stoplightio/prism/pull/514/commits/dff5f9ee5b05afccde19e5fbae64b0e23cd8eda6

chris-miaskowski:
> @stoplight/types (stoplightio/types#50)
> @stoplight/json-ref-resolver (stoplightio/json-ref-resolver#86)
> @stoplight/http-spec (stoplightio/http-spec#15) and @stoplight/spectral (stoplightio/spectral#444)
> @stoplight/graphite (stoplightio/graphite#211) and @stoplight/prism (#315 

Ha I can see you opened a can of worms with this PR :) let me know when the PR is ready I'm happy to review as well

lag-of-death:
work is being done under https://github.com/stoplightio/prism/pull/546

PR #515, opened 2019-08-08
test: execute harness tests on source code

XVincentX:
…I had nothing better to do in the airport.


test-harness/index.ts
```diff
@@ -44,10 +44,30 @@ describe('harness', () => {
...
+                NODE_OPTIONS: '--inspect',
+              }
+            : undefined,
+          stdio: 'pipe',
+        }
+      );
 
       prismMockProcessHandle.stdout.pipe(split2()).on('data', (line: string) => {
         if (line.includes('Prism is listening')) {
+          console.log(parsed.test);
```
philsturgeon: Leftover?

test-harness/README.md
```diff
@@ -24,6 +24,16 @@ This means that you can:
...
 
+## Debugging
+
+In case some tests are giving trouble, it's possible to run the harness on the source code directly instead of the built binary. To make this happen, set `FROM_SOURCES` variable to a truthy value when running the test suite.
+
+` FROM_SOURCES=1 yarn test.harness`
+
+Keep in mind that the whole source code gets recompiled for **every** harness test. For this reason, it's recommended to use this flag with the `TESTS` env variable to narrow down the runned tests.
+
+The test is also run
```
chris-miaskowski: `also run...`
@XVincentX master of dramatic suspense :)

XVincentX:
@chris-miaskowski Try it out :)

test-harness/index.ts
XVincentX: This is kind of on purpose. Unfortunately the way the harness is run is "all or nothing" (there are technical reasons for this) and so jest can't really display what particular test is running right now.

With this hack, you can see the executed test.

test-harness/README.md
XVincentX: I do not remember what was this for. Removed.

XVincentX:
Unfortunately I have to close this. There's no way I can make this work; the process spawning chain is too long — yarn which spawns node that uses ts-node — it's really too complicated.

We can revisit this in the future maybe, but at the current state this is not a viable alternative. Sorry.

Issue #511, opened 2019-08-06
Prism as a HTTP Client


philsturgeon:
**User story.**
As a Prism user writing JS/TS, I want to make HTTP requests with Prism easily, using something that feels a lot like a common HTTP client, and it should be easy and seamless to talk to real servers or a mock from OpenAPI. 

I should be able to create an instance of the client, and make multiple requests, without needing to create new instances.

**Describe the solution you'd like**

I'd like to deprecate or remove the `http` package, create a new `http-client` package, and work on a nice interface for using Prism as a HTTP client. If keeping the `http` package and making this a thin layer over the top is easier, then we can do that. If renaming, rehoming, and refactoring code in one big change is easier, that is also ok.

Axios is a great example of a HTTP client which is clear to understand:

```js
const axios = require('axios');

axios.get('/user', {
    params: {
      ID: 12345
    }
  })
  .then(function (response) {
    console.log(response);
  })
  .catch(function (error) {
    console.log(error);
  });
 
// Want to use async/await? Add the `async` keyword to your outer function/method.
async function getUser() {
  try {
    const response = await axios.get('/user?ID=12345');
    console.log(response);
  } catch (error) {
    console.error(error);
  }
}
```

They have a bunch of aliases:

```js
axios.request(config)
axios.get(url[, config])
axios.delete(url[, config])
axios.head(url[, config])
axios.options(url[, config])
axios.post(url[, data[, config]])
axios.put(url[, data[, config]])
axios.patch(url[, data[, config]])
```

Config can be created for a specific instance which means the requests later on can be made with less information:

```js
const instance = axios.create({
  baseURL: 'https://some-domain.com/api/',
  timeout: 1000,
  headers: {'X-Custom-Header': 'foobar'}
});
```

they have a very useful response object:

```js
{
  // `data` is the response that was provided by the server
  data: {},
 
  // `status` is the HTTP status code from the server response
  status: 200,
 
  // `statusText` is the HTTP status message from the server response
  statusText: 'OK',
 
  // `headers` the headers that the server responded with
  // All header names are lower cased
  headers: {},
 
  // `config` is the config that was provided to `axios` for the request
  config: {},
 
  // `request` is the request that generated this response
  // It is the last ClientRequest instance in node.js (in redirects)
  // and an XMLHttpRequest instance the browser
  request: {}
}
```

I think Prism could benefit from using axios as inspiration for a new and improved HTTP client package. 

_Current API_

```js
[...]

PR #518, opened 2019-08-09
docs: rewrite with getting started

philsturgeon:


README.md
```diff
@@ -51,14 +37,6 @@ docker run -P stoplight/prism:3 mock -h 0.0.0.0 api.oas2.yml
 - [ ] Recording / "Learning" mode to create spec files
 - [ ] Data Persistence (allow Prism act like a sandbox)
 
-## Experimental programmatic APIs (advanced topic)
```
mallachari: We won't expose this API any more?

docs/guides/cli.md
```diff
@@ -51,5 +52,8 @@ Connection: keep-alive
 
 The body, headers, etc. for this response will be taken from the API description document.
 
+## Running in Production
```
pytlesk4: Why would I run prism in production?

docs/getting-started/installation.md
```diff
@@ -0,0 +1,34 @@
+
+# Installation
```
pytlesk4: Can you add an update prism section?

docs/guides/mock-responses.md
```diff
@@ -0,0 +1,159 @@
...
+    "name": "doggie",
+    "photoUrls": [
+        "string"
+    ]
+}
+```
+
+Notice that `name` had an `example` with a value so Prism used it, but `photoUrls` did not, so it just returned `"string"`. 🤷‍♂️
+
+### Dynamic Response Generation
```
pytlesk4: Not rendering correctly

image

docs/guides/workflow.md
```diff
@@ -0,0 +1,16 @@
+# Workflow
```
pytlesk4: Feels like an introduction?

docs/guides/cli.md
philsturgeon: People do all sorts of things and I'd rather have an example for them than not. I'm moving this from a nested README to a smarter place, and we can improve on its explanation another time.

docs/getting-started/installation.md
philsturgeon: Update is just installing again right?

docs/guides/workflow.md
philsturgeon: It's introductory in the sense that it teaches people specifically about where it fits into the grand scheme of things, which you might not need to be told if you are already very aware of the API design-first workflow. I put it in as a guide for people who are not.

docs/guides/mock-responses.md
philsturgeon: Fixed! Great spot, this is a really weird whitespace bug which I'll make an issue for.

README.md
philsturgeon: It was a huge oversight on my part that this API ever made it in. We were not meant to be advertising the JS API until it was stable, and we were meant to be explaining it was experimental in the `packages/http/README.md` not the _main README_. Putting this here before it is ready (#511) is an advertisement we don't want.

Hopefully not too many people have started building anything in production around it.

docs/getting-started/installation.md
philsturgeon: > Note, the binaries do _not_ auto-update, so you will need to run it again to install new versions.

We got it covered :)

docs/guides/cli.md
XVincentX: Prims will run in production on the hosted platform (Stoplight), for example. I want *you* to know how to use it properly :trollface:

PR #525, opened 2019-08-16
CORS support

### LINKED ISSUE #412: Add CORS Support
**User Story**

As a frontend developer integrating my Prism HTTP instance with a browser-based application, I want to be able to configure CORS on Prism, so my browser doesn't complain about same origin policy failures, etc.

**ACs**
- [HTTP headers relating to CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#The_HTTP_response_headers) can be controlled via CLI flags / options
- add at least the following test harness scenarios described in the examples section

**Test Harness Scenarios**

CORS enabled 
- Access-Control-Allow-Origin equal to "*" or the domain Prism is running on matches Origin header sent from a client CORS request - a request made from one domain to another domain (Origin header and URL on different domains)

Given I run Prism v3 with --cors-enabled=false
When I send a CORS request to an operation
Then I should get a 401 response back

When I send a request to an endpoint
And Origin header in the request is the same domain as Prism
Then I should get back a 2xx response back

Given I run Prism v3 with --cors-enabled or I omit this flag

When I send a CORS request to an endpoint
And I don't include Origin header in the request
Then I should get a 400 response back

When I send a CORS request to an endpoint
And Origin header in the request is not in the same domain as Prism 
Then I should get back a 401 response

Given no Access-Control-Allow-Origin was configured for an operation
When I send a request to this operation
And Origin header is not in the same domain as Prism
Then Prism should response with a 401 status code
And there should be no Access-Control-Allow-Origin in the response

Given an operation is configured to answer with Access-Control-Expose-Headers header
When I send a request to this operation
Then Access-Control-Expose-Headers should be included in the response

Given an operation is configured to answer with Access-Control-Max-Age header
When I send a request to this operation
Then Access-Control-Max-Age should be included in the response

Given an operation is configured to answer with Access-Control-Allow-Credentials header
When I send a request to this operation
Then Access-Control-Allow-Credentials should be included in the response

Given an operation is configured to answer with Access-Control-Allow-Methods header
When I send an OPTIONS request to this operation
And this request has Access-Control-Request-Method set to X (where X is a HTTP verb like DELETE)
Then Access-Control-Allow-Methods should be included in the response

Given an operation is configured to answer with Access-Control-Allow-Headers header
When I send an OPTIONS request to this operation
Then Access-Control-Allow-Headers should be included in the response

XVincentX:
This PR installs the CORS middleware to the Prism instance when the `--cors` flag is enabled (which is enabled by default).

Docs and changeling added as well

Closes #412

### DIFF
--- a/docs/guides/cors.md
+++ b/docs/guides/cors.md
@@ -2,7 +2,7 @@
 
 By default, Prism will handle all the [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) responding with a `204` and with headers that will allow all the methods and all the origins.
 
-This is regardless of your OpenAPI Specification. A a request to `OPTIONS /api/todo` will receive a `204` as a response even though your OpenAPI document does not specify the `OPTIONS` verb for that endpoint, **and** even if your OpenAPI document does **not** have an `/api/todo` path at all.
+This is regardless of your OpenAPI description. A request to `OPTIONS /api/todo` will receive a `204` as a response even though your description document does not specify the `OPTIONS` verb for that endpoint, **and** even if your description document does **not** have an `/api/todo` path at all.
 
 Moreover, in case your OpenAPI Document specifies an `OPTIONS` handler (returning a custom status code or different headers) it will be **skipped** in favour of the default one.
 
--- a/docs/guides/cors.md
+++ b/docs/guides/cors.md
@@ -1,0 +1,9 @@
+# CORS
+
+By default, Prism will handle all the [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) responding with a `204` and with headers that will allow all the methods and all the origins.
+
+This is regardless of your OpenAPI Specification. A a request to `OPTIONS /api/todo` will receive a `204` as a response even though your OpenAPI document does not specify the `OPTIONS` verb for that endpoint, **and** even if your OpenAPI document does **not** have an `/api/todo` path at all.
+
+Moreover, in case your OpenAPI Document speicifes an `OPTIONS` handler (returning a custom status code or different headers) it will be **skipped** in favour of the default one.
+
+You can disable the default CORS handler with a command line flag (`prism mock document.yml --cors=false`). In that case, you are in complete control of the response for all the preflight request, and you can decide which header you want to return.
--- a/docs/guides/cors.md
+++ b/docs/guides/cors.md
@@ -4,6 +4,6 @@ By default, Prism will handle all the [preflight requests](https://developer.moz
 
 This is regardless of your OpenAPI Specification. A a request to `OPTIONS /api/todo` will receive a `204` as a response even though your OpenAPI document does not specify the `OPTIONS` verb for that endpoint, **and** even if your OpenAPI document does **not** have an `/api/todo` path at all.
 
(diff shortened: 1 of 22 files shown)


philsturgeon:
Reviewing on mobile so I might just have missed it but I couldn’t see a test that shows CORS actually working? A test harness to show tyre and false do things, and maybe some unit or integration with more specifics.

XVincentX:
@philsturgeon See #412 — I added some comments there. We need to take a "product" decision before bringing this over the finish line.

XVincentX:
@philsturgeon I've added two tests checking out the response, the headers and what happens if the CORS is disabled. Let me know if you want to test some particular scenarios.

XVincentX:
>I think we need one more e2e test where both OPTIONS is defined in a spec and CORS is enabled (so that we can see that OPTIONS from spec is ignored).

Very good point, I'll add one!

packages/cli/src/util/createServer.ts
```diff
@@ -15,10 +15,6 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
 
     signale.await({ prefix: chalk.bgWhiteBright.black('[CLI]'), message: 'Starting Prism…' });
 
-    if (options.dynamic) {
```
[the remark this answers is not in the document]
XVincentX: I figured that the value added here is almost 0 and it was repeating the code twice, and I killed it.

docs/guides/cors.md
```diff
@@ -0,0 +1,9 @@
+# CORS
+
+By default, Prism will handle all the [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) responding with a `204` and with headers that will allow all the methods and all the origins.
+
+This is regardless of your OpenAPI Specification. A a request to `OPTIONS /api/todo` will receive a `204` as a response even though your OpenAPI document does not specify the `OPTIONS` verb for that endpoint, **and** even if your OpenAPI document does **not** have an `/api/todo` path at all.
```
philsturgeon:
```suggestion
This is regardless of your OpenAPI description. A request to `OPTIONS /api/todo` will receive a `204` as a response even though your description document does not specify the `OPTIONS` verb for that endpoint, **and** even if your description document does **not** have an `/api/todo` path at all.
```

packages/http-server/src/server.ts
```diff
@@ -36,11 +39,17 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
-  const mergedConfig = configMergerFactory({ mock: { dynamic: false } }, config, getHttpConfigFromRequest);
+  const mergedConfig = configMergerFactory({ cors: false, mock: { dynamic: false } }, config, getHttpConfigFromRequest);
 
   const prism = createInstance(mergedConfig, components);
 
-  server.all('*', {}, replyHandler(prism));
+  opts.config.cors
+    ? server.route({
+        url: '*',
+        method: ['GET', 'DELETE', 'HEAD', 'PATCH', 'POST', 'PUT'],
```
philsturgeon: if we are going to whitelist methods we should at least have all the ones OpenAPI supports. 

Should we switch to `method: '*'` instead?

XVincentX: Unfortunately we cannot use `*` because otherwise the CORS plugin would complain (you need to leave the `OPTIONS` verb free); so I put all the other verbs that Fastify supports. Any other will make the framework scream at me.

PR #529, opened 2019-08-20
Auto-Approve Dependabot

philsturgeon:
Let's find out how good our test suite is! 😈


XVincentX:
@philsturgeon 

How is this going to help us? Basically this auto approves, but somebody will need to merge this anyway. I would say — if we want to go in such way, we can configure DependaBot to auto merge the PRs 

image


This will effectively save me some time.

philsturgeon:
We chatted about this on Slack but to move the thread forwards, approval and merging are two different things, approval is required for merging. Dependabot wil only be able to merge once the PR has been approved, so you want to enable both.

philsturgeon:
To get the full benefit, uncheck the security updates only. We only turned that on to temporarily reduce the number of PRs coming in, because we were manually having to update them.

XVincentX:
Yeah we can merge it — I just want to check out the dependabot settings so we do not make mistakes and stuff with it. I'll approve/merge later today.

philsturgeon:
No need to put this in docs because users do not send dependabot PRs and therefore have no interaction with this logic.

--
Phil Sturgeon
@philsturgeon

PR #544, opened 2019-08-22
refactor: simplify validation configuration

### LINKED ISSUE #543: Simplify the Validation configuration
Currently in Prism you can specify, for both validation in input and output, if you want to validate query, headers, body, parameters and such.

This complicates the code a lot with defaults, merging, and single booleans meaning different things. We're going to simplify and kill this part with a single boolean. Either you validate the request, entirely, or you do not.

Part of #511

XVincentX:
This PR will remove the ability to selectively enable validations for request and response.

Either you validate the whole request/response, or you do not. Thanks to this change I've been able to remove a good chunk of code and tests that were checking what validations were triggered and which not.

There's still a lot of work to do on validation, but this is a start.

Closes #543 
Part of #511


packages/core/src/factory.ts
```diff
@@ -35,7 +35,7 @@ export function factory<Resource, Input, Output, Config>(
             Either.fold(
               error => {
                 // rethrow error we if we're attempting to mock
-                if ((configObj as IPrismConfig).mock) {
+                if (((configObj as unknown) as IPrismConfig).mock) {
```
XVincentX: I know this is not ok, but the amount of templating in Prism is ridiculous and just drives me crazy. For now we will have to stick with these horrible things, as we proceed I will try to remove them all.

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -226,7 +226,12 @@ describe('HttpMocker', () => {
             const response = httpMocker.mock({
               input: mockInput,
               resource: mockResource,
-              config: { cors: false, mock: { dynamic: false, exampleKey: 'test key' } },
+              config: {
+                cors: false,
+                mock: { dynamic: false, exampleKey: 'test key' },
+                validateRequest: true,
+                validateResponse: true,
```
philsturgeon: Is it required to pass this in? We have it in all the tests, and that along with cors makes this all look rather tricky.

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
+            extendResource,
+          ),
+          input: { method: 'get', url: { path: '/' } },
+          config: { cors: false, mock: { dynamic: false }, validateRequest: true, validateResponse: true },
+        }),
+      ).toHaveLength(length);
+    };
+
+    describe('request is not set', () => {
+      it('validates headers', test(undefined, 2));
```
lag-of-death: this is a bit enigmatic, can we add some info here? Like, what's the reason for an undefined request to have 2 not-passed checks?

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
 
-        expect(resolveValidationConfigModule.resolveRequestValidationConfig).toHaveBeenCalled();
-        expect(httpBodyValidator.validate).not.toHaveBeenCalled();
-        expect(httpHeadersValidator.validate).not.toHaveBeenCalled();
-        expect(httpQueryValidator.validate).toHaveBeenCalledWith({}, [], undefined);
-      };
+  describe('query validation in enabled', () => {
+    const test = (
+      extendResource?: Partial<IHttpOperation>,
+      extendInput?: Partial<IHttpRequest>,
```
lag-of-death: `extendInput` -> `inputToExtend`

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
+        }),
+      ).toHaveLength(lenght);
+
+      expect(httpBodyValidator.validate).not.toHaveBeenCalled();
+      expect(httpHeadersValidator.validate).toHaveBeenCalled();
+      expect(httpQueryValidator.validate).toHaveBeenCalledWith({}, [], undefined);
+    };
+
+    describe('request is not set', () => {
+      it('validates query', test(undefined, undefined, 2));
```
lag-of-death: hm.. maybe in addition to `toHaveLength(length)`, we could also check for the result itself? Not only its `length`?

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
+              id: '1',
+              request: {},
+              responses: [{ code: '200' }],
+            },
+            extendResource,
+          ),
+          input: Object.assign({ method: 'get', url: { path: '/', query: {} } }, extendInput),
+          config: { cors: false, mock: { dynamic: false }, validateRequest: true, validateResponse: true },
+        }),
+      ).toHaveLength(lenght);
```
lag-of-death: `lenght` -> `length`

XVincentX:
> I think having all of these properties that we are required to pass in is getting a bit verbose. I don't want triple booleans, but there must be a way to say "this can be boolean but optional" and only enable the things if they are true and disable if they are not true. Or something.

If you're talking about the *public/user* API, I agree with you and we will find the best way to have the less verbosity possibile.

However this is a change in the internal API (that I will make surface soon somehow) and I explicitly want every single configuration not to be optional. 

Having optional stuff is the reason why we have a non sense cyclical configuration defaulter/provider/constructor that I really want to get rid of because it's confusing even TypeScript itself (it takes 2-3 seconds to realise what's going on in that file). Does that clarify?

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
XVincentX: I can probably DRY this up and reuse the object but yes they are mandatory on purpose. See the comment below for an explanation.

packages/http/src/validator/__tests__/index.spec.ts
XVincentX: I can, but it's not going to be that useful, since the result is an array of the same object `[mockResult, mockResult]` — the validation functions here are mocked and always returning the same stuff.

philsturgeon:
Yeah fair enough, our internal API can be verbose. We can simplify our tests with setting objects at the describe level and modifying them, and the public HTTP Client can be a bunch nicer so its fine.

packages/http/README.md
```diff
@@ -180,24 +180,9 @@ The actual interface looks like this (but rather than explain each property we'r
...
-        };
-
-    response?:
-      | boolean
-      | {
-          headers?: boolean;
-          body?: boolean;
-        };
-  };
+  cors: boolean;
```
chris-miaskowski: Curious how did `cors` got into this PR?

XVincentX: I forgot about it in the previous PR that was making the cors feature happen.

packages/http/src/validator/__tests__/index.spec.ts
XVincentX: It's not really the input being extended, this object is the extension to the default input that's backed in the function.

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
-                responses: [{ code: '200' }],
-              },
-              extendResource,
-            ),
-            input: Object.assign({ method: 'get', url: { path: '/', query: {} } }, extendInput),
-            config: { cors: false, mock: { dynamic: false }, validate: { request: { query: true } } },
-          }),
-        ).toEqual([mockError]);
+  describe('headers validation in enabled', () => {
+    const test = (extendResource?: Partial<IHttpOperation>, length: number = 1) => () => {
```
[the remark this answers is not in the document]
XVincentX: It's not really the resource being extended, this object is the extension to the default resource that's backed in the function. We might need to find a different name

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -76,55 +66,19 @@ describe('HttpValidator', () => {
...
               },
             },
-            config: { cors: false, mock: false, validate: { request: { headers: true, query: false, body: false } } },
+            config: { cors: false, mock: false, validateRequest: true, validateResponse: false },
           }),
         ).toEqual([]);
       });
     });
 
-    describe('only query validation is turned on', () => {
```
chris-miaskowski: A lot of this stuff got removed but I can't see from the diff that we're still covering query validation, are we?

packages/http/src/validator/__tests__/functional.spec.ts
```diff
@@ -76,55 +66,19 @@ describe('HttpValidator', () => {
...
-        expect(
-          await validator.validateInput({
-            resource: httpOperations[2],
-            input: BAD_INPUT,
-            config: { cors: false, mock: false, validate: { request: { headers: false, query: true, body: false } } },
-          }),
-        ).toMatchSnapshot();
-      });
-
-      describe('when all required params are provided', () => {
```
chris-miaskowski: ditto

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -51,150 +43,112 @@ describe('HttpValidator', () => {
...
-      });
-
-      describe('request is set', () => {
-        describe('request.body is not set', () => {
+      describe('request.body is set', () => {
+        describe('request body is not required', () => {
           it(
             'does not try to validate the body',
-            test({ request: { path: [], headers: [], query: [], cookie: [] } }, []),
+            test(
```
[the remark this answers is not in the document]
chris-miaskowski: Good point @lag-of-death, how about
```suggestion
            validate(
```

packages/http/src/validator/__tests__/index.spec.ts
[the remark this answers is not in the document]
chris-miaskowski: Looks like a typo to me, maybe
```suggestion
    const test = (extendedResource?: Partial<IHttpOperation>, length: number = 1) => () => {
```

packages/http/src/validator/__tests__/index.spec.ts
chris-miaskowski: or 
```suggestion
      extendedInput?: Partial<IHttpRequest>,
```

XVincentX: It's not the extended neither — I've gone with `inputExtension`, because it's what really is.

packages/http/src/validator/__tests__/index.spec.ts
chris-miaskowski: Nothing to do with this PR @XVincentX but I think we should compare to what the actual tests are not a length. Could be a good one to refactor later.

packages/http/src/validator/__tests__/index.spec.ts
chris-miaskowski: oooh, didn't see this comment :) never mind!

packages/http/src/validator/__tests__/functional.spec.ts
XVincentX: Yeah I need to check these again, I might have been too hasty in the removal

packages/http/src/validator/__tests__/index.spec.ts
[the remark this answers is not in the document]
lag-of-death: hm.. `extendingResource`?

PR #545, opened 2019-08-22
docs: add errors page

### LINKED ISSUE #527: Create a document for errors
Prism's error types are URLs that point nowhere at the moment. Create an error file that lists all the errors and then make sure to create the appropriate redirects in the platform so that people do not a 404 when they're trying to understand what's going on.

XVincentX:
Closes #527


test-harness/specs/security-scheme-validation-AC-2.oas3.txt
```diff
@@ -29,4 +29,4 @@ curl -i http://localhost:4010/todos
 HTTP/1.1 401 Unauthorized
 www-authenticate: Basic realm="*"
 
-{"type":"https://stoplight.io/prism/errors#UNAUTHORIZED","title":"Invalid security scheme used","status":401,"detail":"","headers":{"WWW-Authenticate":"Basic realm=\"*\""}}
+{"type":"https://stoplight.io/prism/errors#UNAUTHORIZED","title":"Invalid security scheme used.","status":401,"detail":"","headers":{"WWW-Authenticate":"Basic realm=\"*\""}}
```
lag-of-death: @XVincentX, do you think we are OK to leave `headers` in the response body? Started wondering... I could possibly remove that in https://github.com/stoplightio/prism/pull/514 I don't think having `headers` in the response body violates https://tools.ietf.org/html/rfc7807, but now I'm thinking: do we need this if we already have this info in the headers?

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
+# Errors
+
+All the errors that Prism returns conform to the [RFC7807](https://tools.ietf.org/html/rfc7807). In a nutshell, this means that you'll always get a **JSON object** with the following properties:
+
+* `type` (string)
+* `title` (string)
+* `status` (number)
+* `detail` (string)
+
+This document enumerates all the possible errors returned by Prism and provides some further informations on how to solve them.
```
lag-of-death: `informations` -> `information`

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Routing errors
+
+This class of error is returned when Prism is trying to identify the right resource to use to respond to the provided HTTP Request.
```
lag-of-death: `error` -> `errors`

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+### NO_BASE_URL_ERROR
+#### Message: Attempted to make a request to a server but neither baseUrl param provided nor servers defined in the spec.
```
lag-of-death: `but neither baseUrl param was provided nor servers were`?

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+---
+
+### NO_SERVER_CONFIGURATION_PROVIDED_ERROR
+#### Message: Route not resolved, no server configuration provided.
+#### Returned Status Code: `404`
+**Explanation:**
+
+## Validation errors
+
+This class of error is returned when Prism is validating the request/response against the provided OpenAPI file.
```
lag-of-death: `errors`(?)

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
+# Errors
+
+All the errors that Prism returns conform to the [RFC7807](https://tools.ietf.org/html/rfc7807). In a nutshell, this means that you'll always get a **JSON object** with the following properties:
```
philsturgeon:
```suggestion
All the errors that Prism returns conform to [RFC7807](https://tools.ietf.org/html/rfc7807) - Problem Details for HTTP APIs, which means you'll always get a **JSON object** with the following properties:
```

docs/guides/errors.md
philsturgeon:
```suggestion
This document enumerates all the possible errors returned by Prism and provides some further information on how to solve them.
```

docs/guides/errors.md
philsturgeon: Can we avoid long titles? This could just be bold.

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+* `status` (number)
+* `detail` (string)
+
+This document enumerates all the possible errors returned by Prism and provides some further informations on how to solve them.
+
+## Routing errors
+
+This class of error is returned when Prism is trying to identify the right resource to use to respond to the provided HTTP Request.
+
+### NO_BASE_URL_ERROR
```
philsturgeon: Does this generate an automatic ID for us that we can link to? `#NO_BASE_URL_ERROR` or does docs make it all pretty like `#no-base-url-error`? 

If this does not work as expected, we can do:

```
### <a id="NO_BASE_URL_ERROR"></a>No Base URL Error
```

test-harness/specs/E2E-Accept-Content-Negotiation-AC-6.oas2.txt
```diff
@@ -24,4 +24,4 @@ content-type: application/problem+json
 content-length: 0
 Connection: keep-alive
 
-{"type":"https://stoplight.io/prism/errors#NOT_ACCEPTABLE","title":"The server cannot produce a representation for your accept header","status":406,"detail":"Unable to find content for application/json"}
+{"type":"https://stoplight.io/prism/errors#NOT_ACCEPTABLE","title":"The server cannot produce a representation for your accept header.","status":406,"detail":"Unable to find content for application/json"}
```
philsturgeon: I'm not sure if a title should have a full stop?

XVincentX: Yeah I think you're right. I'll delete it everywhere.

docs/guides/errors.md
XVincentX: It will generate an anchor link we can use.

docs/guides/errors.md
XVincentX: @philsturgeon Help!

test-harness/specs/security-scheme-validation-AC-2.oas3.txt
XVincentX: Good question! I was wondering the same and after a while I came up with these conclusions:

1. The RFC has indeed space for some other fields, so we're still in line with it
2. Effectively the mapping between the `header` field and HTTP Headers is the same, but in theory another protocol could use the same information to return something different (I'm thinking about GraphQL or something). Moreover, The HTTP Client could use these informations to show something to the UI (or further elaboration) The point is, this is not directly related to the HTTP Server.

CONTRIBUTING.md
```diff
@@ -77,6 +77,24 @@ yarn cli:debug mock file.oas.yml
 
 The application will wait for a debugger to be attached and break on the first line; from there, you can put your breakpoint here and there and help us debug the software!
 
+#### Every single function I use is stepping into the `fp-ts` code, and this is boring. What do I do?
```
philsturgeon:
```suggestion
**Every single function I use is stepping into the `fp-ts` code, and this is boring. What do I do?**
```

philsturgeon: Or `#### What is fp-ts all about?` but let’s avoid sentence headers.

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+
+This document enumerates all the possible errors returned by Prism and provides some further information on how to solve them.
+
+## Routing errors
+
+This class of errors is returned when Prism is trying to identify the right resource to use to respond to the provided HTTP Request.
+
+### NO_BASE_URL_ERROR
+**Message: Attempted to make a request to a server but neither baseUrl param provided nor servers defined in the spec**
+#### Returned Status Code: `400`
```
philsturgeon: This doesn’t seem like it’s a header it’s just some more information.

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+paths:
+```
+
+`curl -X POST http://localhost:4010/pets`
+
+---
+
+### NO_PATH_MATCHED_ERROR
+**Message: Route not resolved, no path matched**
+#### Returned Status Code: `404`
```
philsturgeon: This error could do with an example as server validation is a little not confusing.

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+`curl -X POST http://localhost:4010/hello`
+
+---
+
+### NO_SERVER_MATCHED_ERROR
+**Message: Route not resolved, no server matched**
+#### Returned Status Code: `404`
+**Explanation:** This error occurs when the server validation is enabled, and the current request has **not** sent the current server or the provided one is not among the defined in the relative array in the file.
```
philsturgeon: This one could do with an example because server validation is a little confusing.

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+* `detail` (string)
+
+This document enumerates all the possible errors returned by Prism and provides some further information on how to solve them.
+
+## Routing errors
+
+This class of errors is returned when Prism is trying to identify the right resource to use to respond to the provided HTTP Request.
+
+### NO_BASE_URL_ERROR
+**Message: Attempted to make a request to a server but neither baseUrl param provided nor servers defined in the spec**
```
lag-of-death: Hm.. I still feel like this line might need `was` and `were` (?)

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+`curl -X POST http://localhost:4010/pets`
+
+---
+
+### NO_SERVER_CONFIGURATION_PROVIDED_ERROR
+**Message: Route not resolved, no server configuration provided**
+#### Returned Status Code: `404`
+**Explanation:**
```
lag-of-death: this might need `Explanation` filled

docs/guides/errors.md
```diff
@@ -0,0 +1,156 @@
...
+**Message: Invalid security scheme used**
+#### Returned Status Code: `401`
+**Explanation:** This error occurs when the security scheme for the current resource does not match the one that the one that the request being processed has provided.
+
+---
+
+### FORBIDDEN
+**Message: Invalid credentials used**
+#### Returned Status Code: `403`
+**Explanation:** This error occurs when the current request is using the correct security scheme, but its content results invalid. This might be an error in the decoding process (such as a not valid base64 payload)
```
lag-of-death:
```suggestion
**Explanation:** This error occurs when the current request is using the correct security scheme, but its content was invalid. This might be an error in the decoding process (such as a not valid base64 payload)
```

docs/guides/errors.md
philsturgeon: Yeah that’s what I meant. This is content which does not need to be a h4 or hAnything.

PR #555, opened 2019-08-26
Remove the config merger strategy to simplify the http client

### LINKED ISSUE #554: Remove the config merger strategy
Prism right now has a merging strategy for both configurations and components (mocker, forwarder) in name of a flexibility we do not need.

Let's remove all this code we do not need so that the readability improves and we can move forward to expose a good HTTP Client API

XVincentX:
Closes #554


XVincentX:
Lost the context, I'm re-trying from scratch.

PR #557, opened 2019-08-27
[WIP] Return the defined 204 response regardless of content-type / accept header

jameshalsall:
## TODO:

- [ ] implement the fix for provided failing tests

## Checklist

Bug fix.

**Currently:**

1. Define a route in your spec with a 204 response, e.g.

```yaml
paths:
  /foo:
   post:
     summary: No response body, 204 code
     operationId: 204_no_response_body
     responses:
       204:
         description: No Content
```

2. Send a request to that route with a content-type of `application/json`, and you receive a "406 Not Acceptable" response instead.

**New Behaviour:**

If a 204 response is defined in the spec, then it should be returned regardless of the request's `Accept` header. 

The swagger 2.0 spec states (https://swagger.io/docs/specification/2-0/describing-responses/):

> Some responses, such as 204 No Content, have no body. To indicate the response body is empty, do not specify a schema for the response. Swagger treats no schema as a response without a body.

And for OAS 3.0+ (https://swagger.io/docs/specification/describing-responses/):

> Some responses, such as 204 No Content, have no body. To indicate the response body is empty, do not specify a content for the response:

No.

### DIFF
--- a/test-harness/specs/204-no-content.oas2.txt
+++ b/test-harness/specs/204-no-content.oas2.txt
@@ -1,0 +1,21 @@
+====test====
+When I send a request to an operation
+And this operation has only a 204 response defined
+Then I should get back the 204 response code
+====spec====
+swagger: "2.0"
+paths:
+  /foo:
+    post:
+      summary: No response body, 204 code
+      operationId: 204_no_response_body
+      responses:
+        204:
+          description: No Content
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/foo -H "accept: application/json"
+====expect====
+HTTP/1.1 204 No Content
+content-type: application/json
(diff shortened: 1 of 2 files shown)


jameshalsall:
If there is anyone who knows more about prism than I do who can point me in the right direction of a fix then that would be great 👍

philsturgeon:
A great place to start is with the e2e testing system:

https://github.com/stoplightio/prism/tree/master/test-harness

You can put a test in there with expected behavior and try to flesh out how it should look, then fill in the gaps with code, knowing that things aren't breaking around it. 

I expect @XVincentX can give better input.

jameshalsall:
Yeah I already have the tests that reproduce the error (they're in this PR). Will investigate the solution a bit more tomorrow but a point in the right direction if someone knows would be great - I expect it's something in the `http-server` package on initial look

jameshalsall:
Ok, I figured out you can do this by just being more explicit in your spec file:

```
paths:
  /foo:
   post:
     summary: No response body, 204 code
     operationId: 204_no_response_body
     responses:
       204:
         description: No Content
         content:
           application/json:
             examples:
               default:
                 value: ""
```

PR #560, opened 2019-08-27
fix: text/plain string are allowed in Prism

### LINKED ISSUE #558: [Bug] text/plain response is not working
Does prism not support plain text? Why am I getting this error for a simple GET endpoint?

OpenAPI 3:
```
  /api/ping:
    get:
      tags:
        - Public
      description: Connectivity test
      responses:
        200:
          description: OK
          content:
            text/plain:
              schema:
                type: string
                example: pong
```

```
curl -X GET "http://127.0.0.1:4010/api/ping" -H "accept: text/plain"
```
Response

```
{
  "statusCode": 406,
  "error": "Not Acceptable",
  "message": "Allowed: application/*+json,application/*+xml,application/json"
}
```

XVincentX:
Closes #558

### DIFF
--- a/packages/http-server/src/serializers.ts
+++ b/packages/http-server/src/serializers.ts
@@ -25,13 +25,12 @@ export default [
       if (typeof data === 'string') {
         return data;
       }
-      throw ProblemJsonError.fromPlainError(
-        Object.assign(new Error('Internal Server Error'), {
-          detail: 'Cannot serialise complex objects as text/plain',
-          status: 500,
-          name: 'https://stoplight.io/prism/errors#NO_COMPLEX_OBJECT_TEXT_PLAIN',
-        }),
-      );
+
+      throw Object.assign(new Error('Cannot serialise complex objects as text/plain'), {
+        detail: 'Cannot serialise complex objects as text/plain',
+        status: 500,
+        name: 'https://stoplight.io/prism/errors#NO_COMPLEX_OBJECT_TEXT_PLAIN',
+      });
     },
   },
 ];
--- a/test-harness/specs/E2E-text-plain-complex-serialisation.oas3.txt
+++ b/test-harness/specs/E2E-text-plain-complex-serialisation.oas3.txt
@@ -28,4 +28,4 @@ curl -i http://localhost:4010/ -H 'accept: text/plain'
 ====expect====
 HTTP/1.1 500 Internal Server Error
 
-{"type":"https://stoplight.io/prism/errors#INTERNAL_SERVER_ERROR","status":500,"detail":"Cannot serialise complex objects as text/plain"}
+{"type":"https://stoplight.io/prism/errors#NO_COMPLEX_OBJECT_TEXT_PLAIN","title":"Cannot serialise complex objects as text/plain","status":500,"detail":"Cannot serialise complex objects as text/plain"}


packages/http-server/src/serializers.ts
```diff
@@ -18,4 +20,11 @@ export default [
     },
     serializer: (data: unknown) => (typeof data === 'string' ? data : xmlSerializer.parse(data)),
   },
+  {
+    regex: /text\/plain/,
+    serializer: (data: unknown) => {
+      if (typeof data === 'string') return data;
+      throw ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, 'Cannot serialise complex objects as text/plain');
```
philsturgeon: It’s late here and I can’t think of which error it should be, but I don’t think 406 is appropriate. The server is failing to create a representation for something defined in the document, and that puts it more towards a 5xx error than something the user did wrong, no?

CHANGELOG.md
```diff
@@ -7,16 +7,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 
-## Added
+## Added
+
 - Prism is now able to validate the security specification of the loaded document #484
 
 ## Fixed
 
 - Prism is not crashing anymore when referencing the same model multiple times in the specification document #552
+- Prism won't crash anymore in case you're trying to return a primitive type as a `text/plain` #560
```
philsturgeon: Let's save the term "crash" for the server falling over and throwing awkward exceptions. 
```suggestion
- Prism won't return 406 when users request a simple `text/plain` #560
```

packages/http-server/src/serializers.ts
philsturgeon: The client is requesting content that is not possible, but that doesn't mean it is the clients fault. 

A 406 means "you cannot accept this, you should accept something that we support". This API (according to the OpenAPI document) supports text/plain, but in trying to request this thing the client is being told it is not supported.

Surly the fact that the client cannot possibly change anything on their end to make this situation work means it is not a client error?

philsturgeon: Absolutely, 500 is a good 🤷‍♂ which I think is the closest status code to this rather odd situation.

test-harness/specs/E2E-text-plain-serialisation.oas3.txt
```diff
@@ -0,0 +1,25 @@
...
+                example: pong
+====server====
+mock -p 4010
+====command====
+curl -i http://localhost:4010/ -H 'accept: text/plain'
+====expect====
+HTTP/1.1 200 OK
+content-type: text/plain
+
+string
```
philsturgeon: Should this not be `pong`?

packages/http-server/src/serializers.ts
```diff
@@ -18,4 +19,17 @@ export default [
     },
     serializer: (data: unknown) => (typeof data === 'string' ? data : xmlSerializer.parse(data)),
   },
+  {
+    regex: /text\/plain/,
+    serializer: (data: unknown) => {
+      if (typeof data === 'string') return data;
```
lag-of-death: needs them sweet `{}`
also... I know we talked about, but maybe:
`if (condition) {} else {}` 
I strongly believe, that this way we show our reasoning better. But this is just a non-blocking suggestion :)

philsturgeon: Fine with the {} suggestion but please no unnecessary conditionals: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

test-harness/specs/E2E-text-plain-complex-serialisation.oas3.txt
```diff
@@ -0,0 +1,31 @@
+====test====
+When I send a request to an operation
+And this operation has a complex schema (with no primitives)
+and the returned content type is a text/plain string
```
lag-of-death:
```suggestion
And the returned content type is a text/plain string
```

lag-of-death:
@XVincentX, how about adding a test to check if `text/plain` is returned by default (when the operation does not specify `text/plain` as its returned content-type) ?

packages/http-server/src/serializers.ts
lag-of-death: Ha! Martin Fowler trick ;> yeah, I don't agree with this rule personally. So, yeah, as I said, just a suggestion and personal preference.

philsturgeon:
Great work on this everyone!

PR #576, opened 2019-08-30
add e2e tests for dynamic xml responses

### LINKED ISSUE #415: Test Harness: Dynamic XML Responses
**User Story**

As a Prism user, I can request a dynamic XML response, so that I don't have to add static examples to a spec.

Note this feature is currently not implemented so expect tests to fail 

When dealing with OAS3 refer to this doc: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#xml-object

One example of more advanced use case is for instance:
```
Person:
  type: object
  properties:
    id:
      type: integer
      format: int32
      xml:
        attribute: true
    name:
      type: string
      xml:
        namespace: http://example.com/schema/sample
        prefix: sample
```
converts to:
```
<Person id="123">
    <sample:name xmlns:sample="http://example.com/schema/sample">example</sample:name>
</Person>
```
But honestly just check out the documentation, it has plenty of simple and advanced use cases  

**ACs**
- link that bug to this card

**Additional resources**

**Test examples**

Given there are no examples defined for an operationAnd this operation can produce application/xml responseAnd that operation's response schema has XML propertyAnd I make a request to this operationAnd the request has Accept header of application/xmlThen I should get a response with dynamically generated XML as payload

Given there are examples defined for an operationAnd this operation can produce application/xml responseAnd that operation's response schema has XML propertyAnd I make a request to this operationAnd the request has Accept header of application/xmlAnd I add __dynamic=true to the query pathThen I should get a response with dynamically generated XML as payload

Given there is a dynamic XML response generated for an operationAnd the operation has headers definedWhen I make a request to this operationThen I should get back a XML response having these headers generated and included in the response

lag-of-death:
Closes #415

## Checklist

Adds e2e tests.

No.


test-harness/helpers.ts
```diff
@@ -1,3 +1,31 @@
+import { Dictionary } from '@stoplight/types/dist';
+import * as xmlDiff from 'diff-js-xml';
+import * as parser from 'fast-xml-parser';
+import { get, omit } from 'lodash';
+
+type Result = { body: string; headers: Dictionary<string, string> };
+
+const xmlValidator = {
+  test: (contentType: string, content: string) => {
+    const doesContentTypeMatch = !!contentType.match(/application\/.*xml/);
```
philsturgeon: Did you know, `text/xml` is equally correct to describe XML. An interesting chat on the topic: https://stackoverflow.com/questions/4832357/whats-the-difference-between-text-xml-vs-application-xml-for-webservice-respons

philsturgeon:
Could you walk me through this one a bit? Why are we creating special loose validation logic and diffing the XML instead of comparing the generated XML to an expected XML?

test-harness/helpers.ts
chris-miaskowski: Out of curiosity, can package like https://www.npmjs.com/package/accepts help with this type of stuff? There must be a package that checks `package.isMediaType(content, 'xml')`.

test-harness/specs/dynamic_xml_responses-AC-1.dynamic-response.oas2.txt
```diff
@@ -0,0 +1,31 @@
...
+              priority:
+                type: integer
+====server====
+mock -d -p 4010
+====command====
+curl -i http://localhost:4010/todos -H "accept: application/xml"
+====expect-loose====
+HTTP/1.1 200 OK
+
+<title>enim</title><description>id dolore</description><priority>90439782</priority>
```
chris-miaskowski: Hey, correct me if I'm wrong but I think this is not a valid XML. XMLs need a root element, right?
Should we wrap it with empty root? `<></>` or `<response></response>`?

XVincentX: This is fixed in Prism 3.1 — rebase it with the current master and you should get this.

philsturgeon: `<></>` might be valid JSX but that's not a thing in XML.

test-harness/helpers.ts
philsturgeon: Yep that's exactly the sort of thing accepts is designed to handle, and it is used for this elsewhere in Prism.

I'd rather step back a level, and find out why we are even trying to detect XML and do something special for it. I don't understand the premise. https://github.com/stoplightio/prism/pull/576#issuecomment-527118433

philsturgeon:
> Do we want to implement the more "advanced" scenarios? For example see the OAS2 XML properties

@chris-miaskowski Can't test what we dont have! Prism only covers basic XML right now, we have no support for the `xml` keyword (#456).

chris-miaskowski:
> > Do we want to implement the more "advanced" scenarios? For example see the OAS2 XML properties
> 
> @chris-miaskowski Can't test what we dont have! Prism only covers basic XML right now, we have no support for the `xml` keyword (#456).

Ok, that makes sense. Do we have a follow up "feature" issue so that we don't loose that of sight? Do we want to implement it?

philsturgeon:
@chris-miaskowski that's what #456 is.

chris-miaskowski:
> @chris-miaskowski that's what #456 is.

Thanks!

lag-of-death:
@philsturgeon, so:
we use `parseResponse` from `http-string-parser`:
https://github.com/stoplightio/prism/blob/93af5877c2fee65fb428dd3092f5c10fc6e54873/test-harness/index.ts#L56-L57
this gives us two objects: `output` and `expected`. These two objects are `valid`ated with `gavel`: https://github.com/stoplightio/prism/blob/93af5877c2fee65fb428dd3092f5c10fc6e54873/test-harness/index.ts#L60 If the validation with `gavel` fails, there's a fallback validation:
```
expect(output).toMatchObject(expected)
```

`output` and `expected` objects both have `.body`, property. This property is a type of `string`.
Neither `gavel` not `.toMatchObject` works as expected for when `.body` holds XML data. Also, it doesn't work for JSON either (it gives false positives). If you take a look at https://github.com/stoplightio/prism/blob/9d10471e3cc3f78b00056332c553ea7a8909888a/test-harness/specs/dynamic_json_responses-AC-1.dynamic-response.oas2.txt#L39
and change that line to let's say `{"title": "officia voluptate amet ut"}` or even `{}`, it would still pass. 
So we should not only have a validator for XML, but fix these false positives we already have in the test harness.

This PR tries to add XML validation, we should also have another for JSON or any other type as what we have ATM isn't working properly IMHO.

test-harness/index.ts
```diff
@@ -58,10 +58,13 @@ describe('harness', () => {
 
           try {
             const isValid = validate(expected, output).valid;
-            if (!!isValid) expect(validate(expected, output).valid).toBeTruthy();
-            else {
-              expect(output).toMatchObject(expected);
+
+            if (!!isValid) {
+              expect(validate(expected, output).valid).toBeTruthy();
```
chris-miaskowski:
```suggestion
              expect(isValid).toBeTruthy();
```

chris-miaskowski: This statement says 
`if (true) then expect(true).toBeTrue()` :) 
Anyway knows whats happening? ;) @XVincentX @lag-of-death

XVincentX: This will make the test fail on purpose — because `isValid` in this condition, is going to be  always `false`

test-harness/helpers.ts
```diff
@@ -18,3 +44,15 @@ export function parseSpecFile(spec: string) {
     expectLoose: splitted[1 + expectLooseIndex],
   };
 }
+
+export const validateLoosely = (expected: Result, output: Result) => {
+  const foundValidator = validators.find(validator => {
+    return validator.test(get(output, ['header', 'content-type'], ''), expected.body);
+  });
+
+  if (!!foundValidator) {
```
chris-miaskowski:
```suggestion
  if (!!validator.test(get(output, ['header', 'content-type'], ''), expected.body)) {
```

test-harness/helpers.ts
```diff
@@ -18,3 +44,15 @@ export function parseSpecFile(spec: string) {
...
+    foundValidator.validate(expected, output);
```
chris-miaskowski:
```suggestion
    xmlValidator.validate(expected, output);
```

chris-miaskowski:
```suggestion
    foundValidator.validate(expected, output);
    expect(omit(output, 'body')).toMatchObject(omit(expected, 'body'));
```

test-harness/helpers.ts
```diff
@@ -1,3 +1,29 @@
...
+    const isContentXML = parser.validate(content) === true;
+
+    return doesContentTypeMatch || isContentXML;
+  },
+  validate: (expected: Result, output: Result) => {
+    xmlDiff.diffAsXml(expected.body, output.body, {}, { compareElementValues: false }, result =>
+      expect(result).toStrictEqual([]),
+    );
+
+    expect(omit(output, 'body')).toMatchObject(omit(expected, 'body'));
```
chris-miaskowski:
```suggestion
```

see comment below. this is the same for every validator so could be extracted. up to you

test-harness/index.ts
chris-miaskowski: `false`? it will be `true`

```
if (!!isValid) {
```
will pass if isValid is truthy.

XVincentX:
I need to take a look at this as well. Please hold on. I promise tomorrow.

philsturgeon:
@chris-miaskowski do you think you could summarize here what the XML validaiton and diffing is all about because I am still pretty confused after the above writeup. 

expect-loose uses gavel which is mostly checking type of fields, so yeah a `foo: "bar"` and `foo: "baz"` match because they are both strings. This is not an issue it is a known outcome of using gavel. If that is not what is required then expect should be used. 

I'm still a bit confused by the expect keyword in prism, the spectral test harness is not doing any magic with the output keyword you just grep the output down if you dont want as much.

```
====server====
mock -d -p 4010
====command====
curl -i http://localhost:4010/todos -H "accept: application/xml" | grep "<xml>" # 👈
====expect====
<xml><title>enim</title><description>id dolore</description><priority>90439782</priority></xml>
```

This is how I thought test harness would work. Is it not?

XVincentX:
I'm sorry for being late, I'm jumping on this now.

test-harness/helpers.ts
```diff
@@ -1,3 +1,27 @@
...
+import * as parser from 'fast-xml-parser';
+import { get, omit } from 'lodash';
+import * as typeIs from 'type-is';
+
+type Result = { body: string; headers: Dictionary<string, string> };
+
+const xmlValidator = {
+  test: (contentType: string, content: string) => {
+    const doesContentTypeMatch = !!typeIs.is(contentType, ['application/xml', 'application/*+xml', 'text/xml']);
+    const isContentXML = parser.validate(content) === true;
```
XVincentX: In theory we could skip this check. Since we're the people writing the tests if we're not providing a correct xml payload we might want the test to just fail.

lag-of-death: @XVincentX, this is for when `get(output, ['header', 'content-type'], ''` returns `''`. Don't remember know exactly when it happens, but it must have since I added `const isContentXML = parser.validate(content) === true;` ;)

lag-of-death:
@XVincentX, why would we want to do XML -> JSON, since "and change that line to let's say {"title": "officia voluptate amet ut"} or even {}, it would still pass." ?

XVincentX:
This is a weird thing that I’m now checking (and fixing if required)

XVincentX:
@lag-of-death 

Ok I've checked out why such test is passing and according to what gavel does — it's correct.


When doing `validate(expected, output)` Gavel is basically trying to derive a schema from the `expected` object and comparing it with the `output`; since the schema for `{}` is basically empty, any object will effectively pass the test.

This does not really concern me though, since we're always putting the real objects there.

PR #577, opened 2019-08-30
fix: accept and handle csv collection format

### LINKED ISSUE #498: OAS2 does not support sending arrays encoded as CSV
Prism 3.0.3

The following test can is already available on `master` or `fix/oas23-body-param-validation-nographite` branch in `test-harness/specs/validate-body-params/form-data-array-csv.oas2.txt` file.

> Hint: search by `TO BE FIXED BY: https://github.com/stoplightio/prism/issues/498` in project

```
====test====
Request form-data body should be an array of strings (comma separated values)
====spec====
swagger: '2.0'
paths:
  /path:
    post:
      consumes:
      responses:
        200:
          schema:
            type: string
      parameters:
        - in: formData
          type: array
          name: arr
          items:
            type: string
          collectionFormat: csv
====server====
mock -p 4010
====command====
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "arr=a,b,c"
====expect====
HTTP/1.1 200 OK
content-type: text/plain

string
```

**Actual**

```
content-length: 355
Date: Fri, 26 Jul 2019 16:35:18 GMT
Connection: keep-alive

{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["body"],"severity":"Error","code":"type","message":"should be array"}]}
```

Note: same happens with other type of encodings like "pipes"
It also doesn't work with "exploded/form" style.

XVincentX:
Closes #498

This PR updates the http spec with the version that supports the CommaDelimited enum value, adds a serialiser for it and uses it for the payload in case it's necessary.

Note — I know this is not exactly a great piece of code, but unfortunately there's no other way to make this right unless we dismantle the whole validation system, which won't happen in this PR.

### DIFF
--- a/test-harness/specs/parameters-ac7.oas2.txt
+++ b/test-harness/specs/parameters-ac7.oas2.txt
@@ -1,0 +1,29 @@
+====test====
+Request form-data body should be an array of strings (comma separated values)
+====spec====
+swagger: '2.0'
+paths:
+  /path:
+    post:
+      consumes:
+        - application/x-www-form-urlencoded
+      responses:
+        200:
+          schema:
+            type: string
+      parameters:
+        - in: formData
+          type: array
+          name: arr
+          items:
+            type: string
+          collectionFormat: csv
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "arr=a,b,c"
+====expect====
+HTTP/1.1 200 OK
+content-type: text/plain
+
+string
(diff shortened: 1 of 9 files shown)


test-harness/specs/parameters-ac7.oas2.txt
```diff
@@ -0,0 +1,29 @@
+====test====
+Request form-data body should be an array of strings (comma separated values)
```
philsturgeon: This test only proves there is not an error when `--data "arr=a,b,c"` but it doesn't prove there is an error when --data "arr=a/b/c" or something else. If we can try to show a positive and a negative it gives better confidence that the code is doing what it is meant to, otherwise it could just be a coincidence. "Not blowing up" != "works correctly".

XVincentX: Good point, adding this now!

PR #582, opened 2019-09-03
Remove the Http Client Merger mechanism

### LINKED ISSUE #554: Remove the config merger strategy

XVincentX:
The following PR is a mid-size refactor that's going to be needed for the new HTTP Client API.

Hopefully this code removal will simplify a lot of the complicated process to create the config object which had load of features nobody was using and nobody will use.

*Note:* The code was so articulated it took me two days to figure it out what was going on.

In particular:

* Removal of the Prism's factory of the factory (which was used to change the configurations and the components on too many places). This was not used anywhere.
* Removed the weird API that was allowing to do stuff like `router.route(data, alternativeRouter)`
* Removal of the `IPrismConfigurationProvider` function concept that was used to get the informations from the HTTP Request — we're using a simpler approach which is just baked in the http-server — that is where the logic should really be
* Removal of all tests and stuff we're not using anymore
* Removal of the forwarder logic that was not used anywhere and was not even configurable

Closes #554


packages/core/src/factory.ts
```diff
@@ -3,165 +3,139 @@ import * as Either from 'fp-ts/lib/Either';
...
+                    },
                     config: configObj,
+                  })(components.logger.child({ name: 'NEGOTIATOR' })),
+                ),
+                TaskEither.map(output => ({ output, resource })),
+              );
+            } else if (components.forwarder) {
+              // forward request and set output from response
+              return pipe(
+                components.forwarder.fforward({
```
philsturgeon: Is this forwarder logic used somewhere? We’re meant to be removing the forwarder as it exists in favour of clean and obvious separation of HTTP server and HTTP client. Anything that’s forwarder related which is not directly being used to power the HTTP Client should be removed and/or refactored as we go. Doesn’t have to be done in this PR though.

packages/http-server/src/server.ts
```diff
@@ -52,11 +53,12 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
     return done(undefined, payload);
   });
 
-  const mergedConfig = configMergerFactory(
-    { cors: false, mock: { dynamic: false }, validateRequest: true, validateResponse: true },
-    config,
-    getHttpConfigFromRequest,
-  );
+  const mergedConfig = defaults<Partial<IHttpConfig>, IHttpConfig>(config, {
+    cors: true,
```
philsturgeon: Are we intentionally changing cors from false to true here?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -168,3 +168,5 @@ function computeBody(
   }
   return undefined;
 }
+
+export const mocker = new HttpMocker();
```
philsturgeon: Do we want a singleton here? Is there any weird state stuff which can happen with this class?

XVincentX: All the other components were singletons as well — I've just unified the code. The mocker is stateless, so singleton does not change the game. Ultimately I want to remove this as a class — the mocker should be nothing more than a function. Baby steps!

packages/http-server/src/server.ts
XVincentX: Yes! `cors` is enabled by default but the whole merger mechanism was hiding this automatically injecting stuff from 300 different places and functions executed here and there. By removing this and providing an explicit path, these errors popped up quickly and have been solved.

packages/core/src/factory.ts
XVincentX: Yes, I'm actually super ok removing the whole part and it should be really easy.

philsturgeon:
Great feedback and love the direction this is going in. Let me know when you want a review!

packages/http-server/src/server.ts
XVincentX: Done in c4cfd24fcaac0e58a92bda907d519e5c1bf77064

packages/core/src/factory.ts
XVincentX: Done in c4cfd24fcaac0e58a92bda907d519e5c1bf77064

PR #583, opened 2019-09-04
Http Client

### LINKED ISSUE #511: Prism as a HTTP Client

XVincentX:
Closes #511

This pull request is the final one closing the issue #511 (there were other PRs that are basically back ports of some of these commits back to the master branch to make sure this PR was not a nightmare to review).

In particular this one:

* Rename the `process` method of prims to `request`, that gives a better idea of what's going on.
* Splits the current `getHttpOperations` function (that was fetching the content from file or from an http resource) into `getHttpOperations` and `getHttpOperationsFromResource`; the first one creates the resource from a string, the second one reads it from a file/web and then calls the `getHttpOperations` internally. This is required to make sure the web version of Prism will still work (you can't read from file in the web without using shims; and this is something we'd like to avoid)
* Propagates the `security` enablement from `IPrismConfig` to `IHttpConfig`
* Introduces the "new" HttpClient. Almost all the meat is in https://github.com/stoplightio/prism/pull/583/files#diff-02f82f5596b8f010bef15dd4663b1af8

### Create from filename or http resource

```ts
const client = await createClientFromResource("examples/petstore.oas2.yaml", {
  mock: false,
  validateRequest: true,
  validateResponse: true
});
```

### Create from OAS spec in string format

```ts
const oasSpec = `
openapi: 3.0.2
paths:
  /hello:
    get:
      responses:
        200:
          description: hello
`;

const client = await createClientFromString(oasSpec, {
  mock: false,
  validateRequest: true,
  validateResponse: true
});
```

### Create from manual Http Operations (this is going to be useful for Studio)

```ts
const client = createClientFromOperations(
  [
    {
      method: "get",
      path: "/hello",
      id: "n1",
      responses: [{ code: "200" }]
    }
  ],
  { mock: false, validateRequest: true, validateResponse: true }
);
```

---

Once you've got a client instance:

1. You can perform the request using the generic method:

```ts
client.request('https://google.it', { method: 'get' }).then(response => console.log(response));
```

The response object has all the informations you need, including the configuration object.

image

2. You can override the configuration object on a request level if you prefer

```ts
client
  .request('https://google.it', { method: 'get' }, { validateResponse: false })
  .then(response => console.log(response));
```

This disables the validation response *only for the current request*

3. You can do the same thing using the shortcut methods

```ts
client
  .get('https://google.it', { mock: false })
  .then(response => console.log(response));
```

You can't provide a method if you are using a shortcut

image

For the shortcut methods (since the only mandatory option is intrinsic in the function name) the option parameter can be omitted

```ts
client
  .get('https://google.it', { validateRequest: false })
  .then(response => console.log(response));
```

I've tried to put great details on the Typings so you will be able to use all the overloads in TypeScript (and Javascript as well if you're using the TypeScript server anyway)

image

Do you want to play more? Why don't you open a PR to `client.spec.ts` and add some tests? :trollface: 

### Outstanding parts

~Base url handling~
~Security Option (this might require a change in Prism itself)~


packages/http/src/client.ts
```diff
@@ -0,0 +1,104 @@
...
+      };
+    }
+
+    return data;
+  };
+
+  return {
+    request,
+    get: (url, input, resources, config) => request(url, { method: 'get', ...input }, resources, config),
+    put: (url, input, resources, config) => request(url, { method: 'put', ...input }, resources, config),
```
XVincentX: @liguori JFP?

liguori: I totally agree with this implementation! @XVincentX  seigie

XVincentX: WOFA.

liguori: FRAATM!

packages/http/src/__tests__/client.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+              name: 'MIT License',
+              spdx_id: 'MIT',
+              url: 'https://api.github.com/licenses/mit',
+              node_id: 'MDc6TGljZW5zZTEz',
+            },
+          });
+
+        const client = await createNewClientInstance(
+          { mock: false, validateRequest: true, validateResponse: true },
+          require.resolve('../../../../examples/petstore.oas2.yaml'),
```
philsturgeon: Can we avoid using examples for tests? Let’s duplicate the file and make a fixture instead.

packages/http/src/client.ts
```diff
@@ -0,0 +1,131 @@
...
+        ...input,
+        url: httpUrl,
+      },
+      resources,
+      config,
+    );
+
+    const o: PrismOutput = {
+      status: data.output.statusCode,
+      statusText: getStatusText(data.output.statusCode),
```
philsturgeon: Can we see if there is an actual status text being returned? People can override them so `200 Monkeys` Is actually perfectly valid. We’d want to see statusText as Monkeys and this code would make it “OK”

packages/http/src/client.ts
```diff
@@ -0,0 +1,131 @@
...
+  (url: string, body: unknown, input: Pick<IHttpRequest, 'headers'>, config?: IHttpConfig): Promise<PrismOutput>;
+  (url: string, body: unknown, config?: IHttpConfig): Promise<PrismOutput>;
+}
+
+export type PrismHttp = {
+  request: RequestFunction;
+  get: IRequestFunctionWithVerb;
+  put: IRequestFunctionWithVerbWithBody;
+  post: IRequestFunctionWithVerbWithBody;
+  delete: IRequestFunctionWithVerb;
```
philsturgeon: Can we avoid verb and say method instead? It should never be verb, we should change any instance of verb we see.

XVincentX: Done!

XVincentX: Done

packages/http/src/__tests__/client.spec.ts
XVincentX: Definitely — this is just to show how the Client/API will look like. I'll finalise this when we're good with it.

packages/http/src/client.ts
XVincentX: Good point. Currently when Prism is mocking, basically there's no status text returned. That's why I used this function. However in case we're hitting a real server it actually make sense to just return what the server returned back to us. I'll look into this.

packages/http/src/client.ts
```diff
@@ -0,0 +1,141 @@
...
+        ? request(url, { method: 'options', ...input }, config)
+        : request(url, { method: 'options' }, input),
+    head: (url: string, input?: Pick<IHttpRequest, 'headers'> | IHttpConfig, config?: IHttpConfig) =>
+      isInput(input) ? request(url, { method: 'head', ...input }, config) : request(url, { method: 'head' }, input),
+    patch: (url: string, body: unknown, input?: Pick<IHttpRequest, 'headers'> | IHttpConfig, config?: IHttpConfig) =>
+      isInput(input)
+        ? request(url, { method: 'patch', ...input, body }, config)
+        : request(url, { method: 'patch', body }, input),
+    trace: (url: string, input?: Pick<IHttpRequest, 'headers'> | IHttpConfig, config?: IHttpConfig) =>
+      isInput(input) ? request(url, { method: 'trace', ...input }, config) : request(url, { method: 'trace' }, input),
```
XVincentX: If you're wondering why this — this is to offer the overloads in TypeScript. It's a little bit of code but it's worth for the developer experience, I'd say.

packages/http/src/__tests__/__snapshots__/http-prism-instance.spec.ts.snap
```diff
@@ -20,6 +20,9 @@ Object {
       "Content-type": "application/json",
     },
     "statusCode": 200,
+    "statusText": Object {
+      "_tag": "None",
```
lag-of-death: wait, what? ;)

packages/http/src/__tests__/client.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+        response = await client.get('/pet/10');
+      });
+
+      test('should not call the mocker', () => expect(mocker.mock).not.toHaveBeenCalled());
+      test('should call the forwarder', () => expect(forwarder.fforward).toHaveBeenCalled());
+      test('should receive a response', () => expect(response).toBeDefined());
+      test('should have output validations errors', () =>
+        expect(response.validations.output.length).toBeGreaterThan(0));
+
+      describe('when setting the validateResopnse false on request level', () => {
```
lag-of-death:
```suggestion
      describe('when setting the validateResponse false on request level', () => {
```

packages/http/src/__tests__/__snapshots__/http-prism-instance.spec.ts.snap
lag-of-death: ah, nvm, it's just `Option.None` in a snapshot :+1:

packages/http/src/client.ts
```diff
@@ -0,0 +1,141 @@
...
+import { parse as parseUrl } from 'url';
+import { createInstance } from '.';
+import { forwarder } from './forwarder';
+import getHttpOperations, { getHttpOperationsFromResource } from './getHttpOperations';
+import { mocker } from './mocker';
+import { router } from './router';
+import { IHttpConfig, IHttpRequest, IHttpResponse, IHttpUrl } from './types';
+import { validator } from './validator';
+
+async function createClientFromResource(specFile: string, defaultConfig: IHttpConfig): Promise<PrismHttp> {
```
lag-of-death: `createClientFromResource` and `createClientFromString` are almost identical, what about some partial application ?
```typescript
async function createClientFrom(getResource: Function, spec: string, defaultConfig: IHttpConfig): Promise<PrismHttp> {
  const resources = await getResource(spec);
  return createClientFromOperations(resources, defaultConfig);
}
```
```
createClientFrom.bind({}, getHttpOperationsFromResource)
createClientFrom.bind({}, getHttpOperations)
```

packages/http/src/client.ts
```diff
@@ -0,0 +1,141 @@
...
+function createClientFromOperations(resources: IHttpOperation[], defaultConfig: IHttpConfig) {
+  const obj = createInstance(defaultConfig, {
+    logger,
+    router,
+    forwarder,
+    validator,
+    mocker,
+  });
+
+  const request: RequestFunction = async (url, input, config) => {
```
lag-of-death: could `request` be moved out of `createClientFromOperations`? Or, is there any reason to keep it here?

packages/http/src/client.ts
```diff
@@ -0,0 +1,141 @@
...
+      data: data.output.body || {},
+      config: defaults(config, defaultConfig),
+      request: { ...input, url: httpUrl },
+      validations: data.validations,
+    };
+
+    return output;
+  };
+
+  function isInput(input?: Pick<IHttpRequest, 'headers'> | IHttpConfig): input is Pick<IHttpRequest, 'headers'> {
```
lag-of-death: if there's no reason to keep it inside `createClientFromOperations`, then let's move it out

lag-of-death:
@XVincentX, I have a feeling that test coverage for `client.ts` ain't that great. `client.spec.ts` only has examples for `createClientFromResource`. I've left some suggestions, but it would be great to add more tests :+1:

packages/http/src/client.ts
XVincentX: Yeah, I guess that will work, good point!

packages/http/src/client.ts
XVincentX: The idea is to make sure nobody will use this function, if not `createClientFromOperations`

The rationale is that no matter how you want to effectively create a client; all the ways MUST eventually call this function. By making the `request` function private here, nobody can create a client in another way. Does that make sense?

packages/http/src/client.ts
XVincentX: Answered in the comment above.

XVincentX:
@lag-of-death You're totally right, I didn't finish to write the tests since I still have to handle some cases; the idea is to give you folks an overview of how the client looks like. I'll add more as I go; the idea is to bring this over the finish line between today and tomorrow.

lag-of-death:
> @lag-of-death You're totally right, I didn't finish to write the tests since I still have to handle some cases; the idea is to give you folks an overview of how the client looks like. I'll add more as I go; the idea is to bring this over the finish line between today and tomorrow.

great :+1: waiting for the updates :)

packages/http/src/client.ts
XVincentX: 81d5630

packages/http/src/client.ts
```diff
@@ -0,0 +1,142 @@
...
+      status: data.output.statusCode,
+      statusText: pipe(
+        data.output.statusText,
+        getOrElse(() => getStatusText(data.output.statusCode)),
+      ),
+      headers: data.output.headers || {},
+      data: data.output.body || {},
+      config: defaults(config, defaultConfig),
+      request: { ...input, url: httpUrl },
+      validations: data.validations,
```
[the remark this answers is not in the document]
XVincentX: Ok, fair enough. I'll use `securityChecks` and `violations`

packages/http/src/client.ts
```diff
@@ -0,0 +1,142 @@
...
+        url: httpUrl,
+      },
+      resources,
+      config,
+    );
+
+    const output: PrismOutput = {
+      status: data.output.statusCode,
+      statusText: pipe(
+        data.output.statusText,
```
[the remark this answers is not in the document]
XVincentX: This is not a thing anymore. You can ignore it.

To explain why this was needed — Axios will return the statusText if provided; but the Prism mocker won't. For this reason, we needed a default in case the response was coming from the mocker. Does this clarify?

XVincentX:
@philsturgeon 

There's a point we should think about.

Currently if an input validation fails in Prism, it will look for 422, then 400, then a `default` (and creating a 422 out of it) and in worst case, create a custom `422` with the validation errors.

The day we'll introduce the proxy strategy I guess we won't do the same, but we will rather proxy the request to the upstream server anyway, and then log somehow all the errors (input/output). 

This would differentiate a little bit the behaviours I guess; are you ok with that? Do you have any other plan for it?

philsturgeon:
@XVincentX that’s an interesting question but not what I was getting at. To answer that:

If mock: true, it’s business as usual.
If mock: false, it’s making requests to the real server. 

The proxy is just always making HTTP Client requests with mock: false.

What I was getting at before is wether the validateRequest is going to be used for controlling the Prism Error (422, 400, or our own HTTP Problems), or if it will be used for contract testing, or... both? 

We probably pretty much always want the contract violati

XVincentX:
> If mock: true, it’s business as usual, negotiator kicks in and it’s looking for a mock.
>If mock: false, it’s making requests to whatever server is mentioned, negotiator and mocker are sidestepped

I'm assuming that in *both* cases we will still log all the errors to the stdout, right? The fact you're getting a 422 from Prism will not preclude the logging actions.

> The proxy is always making HTTP Client requests with mock: false, but there is a use case for the proxy where it needs to be able to reject requests that are invalid, so it’s still able to use validateRequest or similar.

Ok so that means that Prism, when run when `--proxy` will have an additional option to say whether to refuse invalid requests or not. This is something that's missing in the client right now (and in Prism in general). Noted.

> What I was getting at before is wether the validateRequest is going to be used for controlling the Prism Error (422, 400, or our own HTTP Problems), or if it will be used for contract testing, or... both?

Right now this flag is *not* used to controlling the Prims error (422, 400, or our own HTTP Problems). This is fixed in Prism and it's not configurable at the moment. What that flag is doing right now is deciding whether to validate the request (headers and body).

> If we think about validation as “showing the user errors relating to their request” and violations as the result of contract testing on request response, does it make sense to have validateRequest/validateResponse or do we need some other controls?

The *current* idea (and code) on `validateResponse` is whether the response shall be validated against the OpenAPI spec (are all the declared headers present? Is the body satisfying the JSON Schema?)

The result of that is an array of `violations.response`

Does that clarify?

packages/core/src/__tests__/factory.test.ts
```diff
@@ -12,7 +12,7 @@ describe('validation', () => {
   };
 
   const prismInstance = factory<string, string, string, IPrismConfig>(
-    { mock: { dynamic: true }, validateRequest: false, validateResponse: false },
+    { mock: { dynamic: false }, validateRequest: false, validateResponse: false, securityChecks: true },
```
lag-of-death: just something to think about:
`validateResponse` vs `securityChecks`, both validate and validations are even in the same array, but we use different forms for the flags. What about `securityChecks` -> `checkSecurity` ?

packages/core/src/utils/security/handlers/bearer.ts
```diff
@@ -18,15 +18,15 @@ function isBearerToken(inputHeaders: Headers) {
 
 export const bearer = {
   test: ({ type, scheme }: SecurityScheme) => scheme === 'bearer' && type === 'http',
-  handle: bearerHandler.bind(undefined, 'Bearer'),
+  handle: partial(bearerHandler, 'Bearer'),
```
lag-of-death: hm.. why do we care for `this`? it should burn in hell

packages/http/src/__tests__/client.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      const config: IHttpConfig = {
+        mock: { dynamic: false },
+        validateRequest: true,
+        validateResponse: true,
+        securityChecks: true,
+      };
+
+      beforeAll(async () => {
+        client = await createClientFromOperations(
```
lag-of-death: still seems like the coverage might be low? 
`export { createClientFromResource, createClientFromString, createClientFromOperations };` vs only `createClientFromOperations` tested

packages/http/src/client.ts
lag-of-death: @XVincentX, hm.. why making sure of that? Isn't only exporting what you are exporting enough? Moving out the inner functions and keeping them private (by not exporting them), gives us slimmer functions and I believe better performance (no hard facts on this though).

lag-of-death:
and also https://github.com/stoplightio/prism/pull/583#discussion_r326696558 ;)

packages/core/src/__tests__/factory.test.ts
XVincentX: Yes, that makes a lot of sense.

packages/core/src/utils/security/handlers/bearer.ts
XVincentX: We do not — `partial` will indeed make sure to not touch the `this`; it's basically a nicer `bearerHandler.bind(undefined, 'Bearer')`

packages/http/src/__tests__/client.spec.ts
XVincentX: Clearly we're not there yet. I am trying to find a decent coverage to test this stuff.

packages/http/src/index.ts
```diff
@@ -5,40 +5,17 @@ import mock from './mocker';
...
-  ProblemJsonError,
-} from './types';
+export * from './getHttpOperations';
 
-const createInstance = (config: IHttpConfig, components?: PickRequired<Partial<PrismHttpComponents>, 'logger'>) => {
-  return factory<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig>(
-    config,
+import { IHttpConfig, IHttpRequest, IHttpResponse, PickRequired, PrismHttpComponents } from './types';
+
+const createInstance = (
```
StefanDywersant: It's so minor in comparison with the whole PR that I doubt I should comment ;)
```suggestion
export const createInstance = (
```

XVincentX:
Please put any comment! Don’t hold yourself!

packages/http/src/getHttpOperations.ts
```diff
@@ -6,15 +6,20 @@ import * as fs from 'fs';
...
+    ? (await axios.get(file, { transformResponse: res => res })).data
+    : fs.readFileSync(file, { encoding: 'utf8' });
+
+  return getHttpOperations(fileContent);
+}
+
+export default async function getHttpOperations(specContent: string): Promise<IHttpOperation[]> {
+  const parsedContent = parse(specContent);
   const { result: resolvedContent, errors } = await httpAndFileResolver.resolve(parsedContent, {
-    baseUri: resolve(spec),
```
philsturgeon: Trying to avoid the word spec it’s a description document or just a document or doc

PR #587, opened 2019-09-05
Backport client stuff in Prism

XVincentX:
This PR back ports some of the changes I'm doing to the HttpClient to Prism master branch

1. Rename the `configObj` variable name so that we can use object auto-assigning (or whatever that feature is)
2. All configuration now must at least be an `IPrismConfig` — we were casting to it anyway so it's probably better be explicit and stop pretending to be super generic
3. Use `map + getOrElse` instead of `fold` — is semantically more correct.
4. More the `cors` option away from the `IHttpConfig` object

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -8,22 +8,22 @@ import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, PickRequired,
 import { validateSecurity } from './utils/security';
 
 export function factory<Resource, Input, Output, Config>(
-  config: Config,
+  defaultConfig: Config,
   components: PickRequired<Partial<IPrismComponents<Resource, Input, Output, Config>>, 'logger'>,
 ): IPrism<Resource, Input, Output, Config> {
   return {
     process: async (input: Input, resources: Resource[], c?: Config) => {
       // build the config for this request
-      const configObj = defaults(c, config);
+      const config = defaults(c, defaultConfig);
       const inputValidations: IPrismDiagnostic[] = [];
 
       if (components.router) {
         return pipe(
-          components.router.route({ resources, input, config: configObj }),
+          components.router.route({ resources, input, config }),
           Either.fold(
             error => {
               // rethrow error we if we're attempting to mock
-              if (((configObj as unknown) as IPrismConfig).mock) {
+              if (((config as unknown) as IPrismConfig).mock) {
                 return TaskEither.left(error);
               }
 
@@ -48,7 +48,7 @@ export function factory<Resource, Input, Output, Config>(
                 ...components.validator.validateInput({
                   resource,
                   input,
-                  config: configObj,
+                  config,
                 }),
               );
             }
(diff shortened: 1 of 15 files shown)


XVincentX:
@pytlesk4 Do you want to take a look and give me your blessing?

packages/core/src/factory.ts
```diff
@@ -48,19 +48,20 @@ export function factory<Resource, Input, Output, Config>(
...
               pipe(
                 validateSecurity(input, resource),
-                fold<IPrismDiagnostic, IPrismDiagnostic[]>(() => [], value => [value]),
+                map(sec => [sec]),
+                getOrElse<IPrismDiagnostic[]>(() => []),
               ),
             );
 
-            if (resource && components.mocker && ((configObj as unknown) as IPrismConfig).mock) {
+            if (resource && components.mocker && config.mock) {
```
pytlesk4: Way better!

packages/core/src/factory.ts
```diff
@@ -1,29 +1,29 @@
...
 
       if (components.router) {
         return pipe(
-          components.router.route({ resources, input, config: configObj }),
+          components.router.route({ resources, input, config }),
           Either.fold(
             error => {
               // rethrow error we if we're attempting to mock
-              if (((configObj as unknown) as IPrismConfig).mock) {
+              if (config.mock) {
```
pytlesk4: Way better!

XVincentX:
@pytlesk4 @philsturgeon 

Please re-review this. I've back ported another commit which probably is way more important.

Fundamentally I made the wrong choice of putting the `cors` option in the `IHttpConfig` type — which is wrong. The `cors` option is an `HttpServer` option, it has nothing to do with the mocker itself. I've removed the field from there. Let me know if you need another clarification.

pytlesk4:
Let me know when you are done pushing changes and I can take a look again!

XVincentX:
I’m done here boss. Do not hesitate to ask clarifications if something is unclear.

packages/http-server/src/types.ts
```diff
@@ -4,6 +4,7 @@ import { FastifyInstance } from 'fastify';
 export interface IPrismHttpServerOpts {
   components?: PickRequired<TPrismHttpComponents, 'logger'>;
   config: IHttpConfig;
+  cors: boolean;
```
pytlesk4: In theory couldn't dynamic mocking be a server config as well?

pytlesk4: And validation for that matter?

XVincentX: Not exactly — Prism has basically the `Http` and the `HttpServer` packages

The Http package is able to run the mocking and forwarding functionalities without having to require any server. It's basically this:

```ts

const prism = createPrism(operations, config); // config is IHttpConfig
prism.process(inputRequest) // get an answer
```

For this, it makes sense to know whether you want the input validation (which is in the `inputRequest`) or the output validation (which could be come from the mocker or the forwarded response from the upstream server).

In this case, `cors` does not make any sense; it should be a configuration parameter of the `http-server` — which is used by Studio and the CLI package.

I'm merging this, but in meantime we can keep the conversation going here.

pytlesk4: makes sense

PR #593, opened 2019-09-06
refactor: make Prism's Input and Output mandatory

XVincentX:
This PR is another back port from the work I've been doing on the HTTP Client that I thought it would be a good idea  to bring in master as soon as possible.

This is still to make sure the final PR it's not a 🐲 dragon 🐲 to review.

* Rename `TSomething` to `Something` — the Hungarian era is over
* Stop pretending that components can just be undefined and skip entire parts of the code. Currently if the `router` component is not provided, Prism will just *not* do anything. I've modified this behavior to make the `router` component mandatory. Either you give me what I need to do my job or get out of here
* Make sure that, by definition, either the mocker or the forwarder is passed as a component

For an easier review, disable whitespaces from the diff:

➡️ https://github.com/stoplightio/prism/pull/593/files?w=1 ⬅️

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -4,7 +4,7 @@ import { getOrElse, map } from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/lib/pipeable';
 import * as TaskEither from 'fp-ts/lib/TaskEither';
 import { defaults } from 'lodash';
-import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, PickRequired, ProblemJsonError } from './types';
+import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, ProblemJsonError } from './types';
 import { validateSecurity } from './utils/security';
 
 export function factory<Resource, Input, Output, Config extends IPrismConfig>(
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -9,7 +9,7 @@ import { validateSecurity } from './utils/security';
 
 export function factory<Resource, Input, Output, Config extends IPrismConfig>(
   defaultConfig: Config,
-  components: PickRequired<Partial<IPrismComponents<Resource, Input, Output, Config>>, 'logger'>,
+  components: IPrismComponents<Resource, Input, Output, Config>,
 ): IPrism<Resource, Input, Output, Config> {
   return {
     process: async (input: Input, resources: Resource[], c?: Config) => {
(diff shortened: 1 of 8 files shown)


packages/core/src/types.ts
```diff
@@ -40,13 +40,21 @@ export interface IValidator<Resource, Input, Config, Output> {
   validateOutput?: (opts: { resource: Resource; output?: Output; config?: Config }) => IPrismDiagnostic[];
 }
 
-export interface IPrismComponents<Resource, Input, Output, Config extends IPrismConfig> {
+type MockerOrForwarder<Resource, Input, Output, Config extends IPrismConfig> =
+  | {
+      forwarder?: IForwarder<Resource, Input, Config, Output>;
```
chris-miaskowski:
```suggestion
```

If it's mocker OR forwarder then why still make forwarder an option?

packages/core/src/types.ts
```diff
@@ -40,13 +40,21 @@ export interface IValidator<Resource, Input, Config, Output> {
...
+      mocker: IMocker<Resource, Input, Config, Reader<Logger, Either<Error, Output>>>;
+    }
+  | {
+      forwarder: IForwarder<Resource, Input, Config, Output>;
+      mocker?: IMocker<Resource, Input, Config, Reader<Logger, Either<Error, Output>>>;
```
chris-miaskowski: ditto

packages/core/src/types.ts
XVincentX: That's the way you do this in TypeScript. You need to make two types and union them where each of them is 

See an example:

https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAhg+gewE5wEZQLxQN6wPwBcUAzsEgJYB2A5gNxSpGUCuAtqhElAL5QA+OWEVIUa9VISgt2nHrQBQ8gMYJKpKAgCMReMjSYc8qEKgAiABYQQp+dwXLV6hACYdiFOizYjDIpoAMtvYA9MFQAKJISMgANFKq0OTEUGDRAG7kACYQmQ5qwBoAzG56njh28kA

However your comment made me realise there's probably a more clear way that I pushed in 5c3edf9

packages/core/src/factory.ts
```diff
@@ -90,53 +73,59 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
                 config,
-              });
-            }
+              }),
+              TaskEither.map(output => ({ output, resource })),
+            );
+          }
 
-            return {
-              input,
```
philsturgeon: Neither mock or forwarder

Not doesn’t work quite like that :)

packages/core/src/types.ts
```diff
@@ -40,13 +40,21 @@ export interface IValidator<Resource, Input, Config, Output> {
...
+    };
+
+export type IPrismComponents<Resource, Input, Output, Config extends IPrismConfig> = {
   router: IRouter<Resource, Input, Config>;
-  forwarder: IForwarder<Resource, Input, Config, Output>;
-  mocker: IMocker<Resource, Input, Config, Reader<Logger, Either<Error, Output>>>;
   validator: IValidator<Resource, Input, Config, Output>;
   logger: Logger;
-}
+} & MockerOrForwarder<Resource, Input, Output, Config>;
```
philsturgeon: I don’t want us adding anything that says “forwarder” if we can avoid it. We should be talking about “http client” and “mocks” but forwarder needs to be chipped away at not increased.

packages/core/src/factory.ts
XVincentX: I do not understand what you're trying to say

philsturgeon: new Error('Neither forwarder or mocker have been selected. Something is wrong'));

PR #594, opened 2019-09-09
Move the validator enablement out of the validator

XVincentX:
This PR is yet another back port from the HTTP Client changes were doing. We're still trying to make sure the final PR is not a 🐲 dragon 🐲 

* Move the logic to understand whether the validator should be ran or not on the caller, and not the validator itself. This allows to remove the need of passing the whole configuration object
* The output parameter for the `validateOutput` function is now mandatory. If you do not have it, you've got nothing to validate.
* I've moved the enablement tests from the validator function to the factory function, which is where the real logic is happening.

### DIFF
--- a/packages/core/src/__tests__/factory.test.ts
+++ b/packages/core/src/__tests__/factory.test.ts
@@ -1,0 +1,47 @@
+// @ts-ignore
+import logger from 'abstract-logging';
+import { right } from 'fp-ts/lib/Either';
+import { asks } from 'fp-ts/lib/ReaderEither';
+import { Logger } from 'pino';
+import { factory, IPrismConfig } from '..';
+
+describe('validation', () => {
+  const validator = {
+    validateInput: jest.fn().mockReturnValue(['something']),
+    validateOutput: jest.fn().mockReturnValue(['something']),
+  };
+
+  const prismInstance = factory<string, string, string, IPrismConfig>(
+    { mock: true, validateRequest: false, validateResponse: false },
+    {
+      validator,
+      router: {
+        route: jest.fn().mockReturnValue(right('hey')),
+      },
+      logger: { ...logger, child: jest.fn().mockReturnValue(logger) },
+      mocker: {
+        mock: jest.fn().mockReturnValue(asks<Logger, Error, string>(r => 'hey')),
+      },
+    },
+  );
+
+  describe.each([
+    ['input', 'validateRequest', 'validateInput', 'validateOutput'],
+    ['output', 'validateResponse', 'validateOutput', 'validateInput'],
+  ])('%s', (_type, fieldType, fnName, reverseFnName) => {
+    describe('when enabled', () => {
+      beforeAll(async () => {
+        const obj: any = {};
+        obj[fieldType] = true;
+        await prismInstance.request('', [], obj);
+      });
+
+      afterEach(() => jest.clearAllMocks());
+      afterAll(() => jest.restoreAllMocks());
+
+      test('should call the relative validate function', () => expect(validator[fnName]).toHaveBeenCalled());
+      test('should not call the relative other function', () =>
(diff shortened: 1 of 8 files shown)


packages/core/src/__tests__/factory.test.ts
```diff
@@ -0,0 +1,47 @@
+// @ts-ignore
+import logger from 'abstract-logging';
+import { right } from 'fp-ts/lib/Either';
+import { asks } from 'fp-ts/lib/ReaderEither';
+import { Logger } from 'pino';
+import { factory, IPrismConfig } from '..';
+
+describe('validation', () => {
+  const validator = {
+    validateInput: jest.fn().mockReturnValue(['something']),
```
lag-of-death: `jest.fn().mockReturnValue(['something'])` can be put in a variable and reused below

packages/http/src/validator/index.ts
```diff
@@ -2,88 +2,61 @@ import { IPrismDiagnostic, IValidator } from '@stoplight/prism-core';
...
+    const results: IPrismDiagnostic[] = [];
+    const mediaType = caseless(output.headers || {}).get('content-type');
+    const responseSpec = resource.responses && findOperationResponse(resource.responses, output.statusCode);
 
-      this.headersValidator
-        .validate(output.headers || {}, (responseSpec && responseSpec.headers) || [], mediaType)
-        .forEach(validationResult => results.push(validationResult));
+    this.bodyValidator
+      .validate(output.body, (responseSpec && responseSpec.contents) || [], mediaType)
+      .forEach(validationResult => results.push(validationResult));
```
lag-of-death: any chance for a map/reduce instead of `forEach` and getting rid of `.push`ing?

packages/http/src/validator/index.ts
```diff
@@ -2,88 +2,61 @@ import { IPrismDiagnostic, IValidator } from '@stoplight/prism-core';
...
 
-      return results;
-    }
+    this.headersValidator
+      .validate(output.headers || {}, (responseSpec && responseSpec.headers) || [], mediaType)
+      .forEach(validationResult => results.push(validationResult));
```
lag-of-death: ditto

packages/core/src/__tests__/factory.test.ts
XVincentX: I cannot do this — the returned mocked function has an internal counter that tracks the number of calls and they need to be separate.

packages/http/src/validator/index.ts
XVincentX: Yes, but it's not worth it since this is going dismantled very soon for a more generic approach. Spending time here is indeed worthless right now.

As next step when we'll be writing the contract testing piece we'll be refactoring most of this code.

packages/http/src/validator/index.ts
XVincentX: Yes, but it's not worth it since this is going dismantled very soon for a more generic approach. Spending time here is indeed worthless right now.

As next step when we'll be writing the contract testing piece we'll be refactoring most of this code.

lag-of-death: cool :+1: good to hear :)

packages/core/src/__tests__/factory.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        mock: jest.fn().mockReturnValue(asks<Logger, Error, string>(r => 'hey')),
+      },
+    },
+  );
+
+  describe.each([
+    ['input', 'validateRequest', 'validateInput', 'validateOutput'],
+    ['output', 'validateResponse', 'validateOutput', 'validateInput'],
+  ])('%s', (_type, fieldType, fnName, reverseFnName) => {
+    describe('when enabled', () => {
```
lag-of-death: @XVincentX, should we have `when disabled` too?

XVincentX:
Well that won’t hurt for sure. Added

PR #601, opened 2019-09-10
Find 401/403 responses before 422/400

### LINKED ISSUE #591: When request doesn't contain proper security scheme, return the user-defined 401 first
When a request to an endpoint does not meet the security scheme (i.e, lacking bearer token), Prism currently looks for a 422 definition, followed by a 400 definition, followed by 401.  This seems odd, if we have our own 401 defined, it should be the first one tried by Prism.

**To Reproduce**

1. Given an endpoint defined like so:

```
/users:
    get:
      summary: "Retrieves the authenticated user"
      description: "Returns the `user` object associated with the bearer token in the Authorization header."
      responses:
        200:
        400:
          description: "The provided user data was malformed, or attempted to modify a read-only field (such as `balance`)."
        401:
```

2. Curl the endpoint without an "Authorization: Bearer" token
3. Prism log:

```
    [VALIDATOR] ⚠  warning   Request did not pass the validation rules
    [NEGOTIATOR] ●  note      Unable to find a 422 response definition
    [NEGOTIATOR] ✔  success   Found response 400. I'll try with it.
    [NEGOTIATOR] ●  note      Unable to find a content with an example defined for the response 400
    [NEGOTIATOR] ✔  success   The response 400 has a schema. I'll keep going with this one
    [NEGOTIATOR] ✔  success   Responding with the requested status code 400
```

**Expected behavior**
That the 401 we defined would be used first, before trying 422/400 since this is explicitly an authorization failure.

lag-of-death:
Closes #591

## Checklist

Bug fix.

The order of finding a response for an invalid request is now changed 
**from:**
422, 400, 401, 403, 422 (`.default`)
**to:**
401, 403, 422, 400, 422 (`.default`)
401 and 403 are only checked for if there were security issues

No.


philsturgeon:
Agreed! 

1. 401 and 403 are not interchangeable 
2. security checks (lets avoid calling it validation) are different to request validation in general 

request validation should look for 422 then 400
missing auth information should look for a 401 

a 403 would only be triggered with invalid auth information, like an expired token, but we don't have any way of knowing if things are real so probably dont worry about looking for a 403?

lag-of-death:
> Agreed!
> 
> 1. 401 and 403 are not interchangeable
> 2. security checks (lets avoid calling it validation) are different to request validation in general
> 
> request validation should look for 422 then 400
> missing auth information should look for a 401
> 
> a 403 would only be triggered with invalid auth information, like an expired token, but we don't have any way of knowing if things are real so probably dont worry about looking for a 403?

@philsturgeon, we do have e2e tests that check for 403s. Except that, the implementation does what you said. I have to resolve the conflicts now that #600 is in master, but could you take a look at the code and (dis)approve?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Prism is correctly responding with the correct status code when security fails
```
philsturgeon: As mentioned in another PR this changelog item is not as descriptive as it could be,. 

```suggestion
- Requests lacking security credentials will return a 401 instead of the usual 422/400 validation error
```

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -59,7 +59,17 @@ function handleInputValidation(input: IPrismInput<IHttpRequest>, resource: IHttp
     withLogger(logger => logger.warn({ name: 'VALIDATOR' }, 'Request did not pass the validation rules')),
     chain(() =>
       pipe(
-        helpers.negotiateOptionsForInvalidRequest(resource.responses),
+        helpers.negotiateOptionsForInvalidRequest(
+          resource.responses,
+          [
+            input.validations.input.find(
+              inputValidation => inputValidation.code === 401 || inputValidation.code === 403,
```
philsturgeon:
```suggestion
              inputValidation => inputValidation.code === 401,
```

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -59,7 +59,17 @@ function handleInputValidation(input: IPrismInput<IHttpRequest>, resource: IHttp
...
+            ),
+          ]
+            .filter(identity)
+            .map(securityValidation => (securityValidation as { code: number }).code)
+            .concat([422, 400]),
+        ),
         mapLeft(() => {
           const securityValidation = input.validations.input.find(i => i.code === 401 || i.code === 403);
```
philsturgeon:
```suggestion
          const securityValidation = input.validations.input.find(i => i.code === 401);
```

philsturgeon:
@lag-of-death looking at this implementation suggests the feedback from @XVincentX and myself was not clear. 

This code specifically: `export const matchingOrder = [401, 403, 422, 400];`

This seems like we are still confusing security checks and general request validation? 

1. Security Check returns a 401
2. Request Validation returns 422 or 400
3. We do not need to worry about 403 at all because we dont know if the tokens are valid or not.

lag-of-death:
@philsturgeon, I'm trying to wrap my head around this, so.. you're saying that e2e tests that check for 403 should be rewritten to check fo 401? And, the order you pointed out is just a default one, what's more important is how the concatenation looks like:
https://github.com/stoplightio/prism/blob/77fec6ed8900c6ac75f7144f25940a6ab56dd491/packages/http/src/mocker/HttpMocker.ts#L64-L71

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Requests lacking security credentials will return 401/403 error instead of the usual 422/400 validation error
```
philsturgeon:
```suggestion
- Requests lacking security credentials will return 401 error instead of the usual 422/400 validation error
```

lag-of-death:
> Ok I think there's some confusion from both sides, I'll try to clarify and see if we can move this forward.
> 
> > We do not need to worry about 403 at all because we dont know if the tokens are valid or not.
> 
> @philsturgeon That is not 100% true, there are some cases where we can effectively say for sure that a token is not valid.
> 
> An example is when for example we're asking for a Basic security scheme
> 
> ```yaml
>   securitySchemes:
>     BasicAuth:
>       type: http
>       scheme: basic
> ```
> 
> …but we're sending a token that's not base64 encoded. In such case, no matter the content, the credential used is invalid. There are other (limited) cases. `403` looks the right response to send back in such case. @lag-of-death You're good there, there is nothing to change.
> 
> On the code side, I think we should probably go in a different direction.
> 
> The best thing to handle this is probably to expand the current definition for `negotiateOptionsForInvalidRequest` to accept as well the validation error; with such information we can make a meaningful decision and call `findResponse` which should be as well modified to get an "array" of choices.
> 
> Does that make sense @lag-of-death ?

@XVincentX, we could do that, but I'd rather split `input validation` and `security validation` like https://github.com/stoplightio/prism/pull/484#discussion_r314377512 originally proposed

XVincentX:
@lag-of-death Can you elaborate a little bit more on the advantages? My main point is that the security validation *is* an input validation and it should leverage 99% of the code that's there already. The only part where the path "forks" is when to look for a status code to return back.

lag-of-death:
@XVincentX, the advantage is that we don't put every input validation in the same box. But nevermind, could you explain better what you had in mind? Because I kind of did what you're asking for (?)

philsturgeon:
> That is not 100% true, there are some cases where we can effectively say for sure that a token is not valid.

I think we are semantically slightly misaligned on the validity of a token. 

- 401: I do not know who you are
- 403: I know who you are and i am saying you cannot do this thing

If Prism asks for a JWT and the client give me something else, I do not know who you are.

If Prism asks for a basic token and it gets `nonsense` then it still does not know who you are.

Prism should only return 401 for security checks that fail.

philsturgeon:
> 401 Unauthorized
> 
> The request has not been applied because it lacks valid authentication credentials for the target resource.
> 
> The server generating a 401 response MUST send a WWW-Authenticate header field1 containing at least one challenge applicable to the target resource.
> 
> If the request included authentication credentials, then the 401 response indicates that authorization has been refused for those credentials. The user agent MAY repeat the request with a new or replaced Authorization header field2. If the 401 response contains the same challenge as the prior response, and the user agent has already attempted authentication at least once, then the user agent SHOULD present the enclosed representation to the user, since it usually contains relevant diagnostic information.

lag-of-death:
@philsturgeon, even if
> Prism should only return 401 for security checks that fail.

then it will not be done in this PR

philsturgeon:
@lag-of-death I am telling you what the end goal needs to be, I don't mind how you get there. If there is an existing mistake around security checks then now is a chance for us to fix that, in this PR or another one, but we certainly do not want to extend that mistake with more code that we'd have to remove later.

lag-of-death:
@philsturgeon, sure, could you create an issue for that 403 -> 401 change?

XVincentX:
I'll take care of the issue.

philsturgeon:
@lag-of-death it seems like you have more context on this, i do not know what code is producing 403's and I do not have the time to go and look. Can you use the context you have, plus the information provided about what 401 is for, and create the issue yourself?

XVincentX:
We're cool.

@lag-of-death https://github.com/stoplightio/prism/issues/624 — can you jump on this one?

lag-of-death:
> We're cool.
> 
> @lag-of-death #624 — can you jump on this one?

yeah, sure, but please explain what solution for this PR you had in mind

XVincentX:
Sure, I'm just jumping between 32098430938410928319028490128 PRs and comments. I'll get to this :)

XVincentX:
We need #625 merged to move this over the finish line.

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const matchingOrder = [401, 403, 422, 400];
+
+export default function matchResponse(
+  httpResponses: IHttpOperationResponse[],
+  logger: Logger,
+  order: number[] = matchingOrder,
+): Option<IHttpOperationResponse> {
+  return pipe(
+    head(order),
```
XVincentX: This is guaranteed to succeed if we make sure that the `order` is not empty

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
+import { IHttpOperationResponse } from '@stoplight/types';
+import { head } from 'fp-ts/lib/Array';
+import { flatten, fold, none, Option, some } from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { tail } from 'lodash';
+import { Logger } from 'pino';
+import { createResponseFromDefault, findResponseByStatusCode } from './InternalHelpers';
+
+const msg = (code: number) => `Unable to find a ${code} response definition`;
+const defaultMsg = (code: number) => `Created a ${code} from a default response`;
```
XVincentX: If not used at least 2-3 times, maybe let's not rush in create an external function.

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
+import { IHttpOperationResponse } from '@stoplight/types';
+import { head } from 'fp-ts/lib/Array';
+import { flatten, fold, none, Option, some } from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { tail } from 'lodash';
+import { Logger } from 'pino';
+import { createResponseFromDefault, findResponseByStatusCode } from './InternalHelpers';
+
+const msg = (code: number) => `Unable to find a ${code} response definition`;
```
XVincentX: If not used at least 2-3 times, maybe let's not rush in create an external function.

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
...
+            },
+            matchedResponse => some(matchedResponse),
+          ),
+        );
+      },
+    ),
+  );
+}
+
+function tryToFindDefaultResponse(httpResponses: IHttpOperationResponse[], logger: Logger) {
```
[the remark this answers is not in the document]
XVincentX: This function is not required anymore — `createResponseFromDefault` is returning an `Option`, so all this code is not required. We just need to map in case we want to log something

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
...
+  logger: Logger,
+  order: number[] = matchingOrder,
+): Option<IHttpOperationResponse> {
+  return pipe(
+    head(order),
+    fold(
+      () => flatten(tryToFindDefaultResponse(httpResponses, logger)),
+      possibleMatch => {
+        return pipe(
+          findResponseByStatusCode(httpResponses, `${possibleMatch}`),
```
XVincentX:
```suggestion
          findResponseByStatusCode(httpResponses, possibleMatch.toString()),
```

packages/http/src/mocker/negotiator/responseFinder.ts
```diff
@@ -0,0 +1,52 @@
...
+          fold(
```
XVincentX: You do not need to use `fold` if you're not changing the `some` part — you should instead use a `alt` to provide an alternative option.

```ts
const t = pipe(
   trySomething,
   fold(()=> alternative, val => some(val))
)
```

-> becomes

```ts
const t = pipe(
   trySomething,
   alt(()=> alternative)
)
```

In this specific case

```ts
        return pipe(
          findResponseByStatusCode(httpResponses, `${possibleMatch}`),
          alt(() => {
              logger.trace(msg(possibleMatch));
              return matchResponse(httpResponses, logger, tail(order));
          }),
        );
```

XVincentX:
Superseded by #628

PR #568, opened 2019-08-29
feat: handle */* accept header

lag-of-death:
Fixes #492 

## Checklist

- [x] Tests have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## What kind of change does this PR introduce?

Bug fix.

## Does this PR introduce a breaking change?

No.

PR #602, opened 2019-09-11
test harness: do not swallow exceptions

XVincentX:
Instead of swallowing the error, let's cleanup using the `finally` clause

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,7 @@ Prism is using TypeScript's incremental compiler capability that sometimes does 
 
 ```sh
 rm -rf packages/**/dist
-rm -rf packages/**/.tsbuildinfo
+rm -rf packages/**/*.tsbuildinfo
 ```
 
 …then try again. If it does not work, you can generally blame [Vincenzo](https://github.com/XVincentX) and write him.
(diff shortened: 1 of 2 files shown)


lag-of-death:
yeah, I don't know. This seems to grow. Wasn't https://github.com/stoplightio/prism/pull/568 clearer? And also, while at it, can you remove these `delete`s like https://github.com/stoplightio/prism/blob/bc8742a146893f5a8337481788bbac386a68c5b2/test-harness/index.ts#L71 ?

XVincentX:
@lag-of-death I'm missing the link between this and #568 — were you referring to another Pull Request?

Moreover, that delete is required because in that branch (which is — XML payload detected) we still need to use gavel to validate the status code and the headers — but not the payload (since that's handled by a separate library). That `delete` is to not trigger Gavel's internal validation. Does that make sense/clarify?

lag-of-death:
@XVincentX, mi mistake, had #576 in mind (before that `delete` changes). Nvm then, I'm gonna approve, but will handle these `deletes` in one of my PRs

XVincentX:
Yeah in theory we can also simply provide a new object without these properties. I guess that's what you were proposing somehow, right?

lag-of-death:
@XVincentX, yeah :+1:

PR #603, opened 2019-09-11
handle head requests

### LINKED ISSUE #575: HEAD requests must not expect to find a response body schema
**Describe the bug**
When defining HEAD requests, we do not define response bodies on purpose, because HEAD never returns a body. However, in the current response selection logic, HEAD is processed the same as any other request, and when it fails to find any suitable response schema (as it shouldn't), it returns a `406` response (I have attempted to vary the `Accept` header values requests out of curiosity, but always saw `406` client error responses). **NOTE**: curl, by default, when not supplied with a specific `Accept` header, inserts `Accept: */*`, which is consistent with RFC7231:
```
A request without any Accept header field implies that 
the user agent will accept any media type in response.
```

**To Reproduce**

1. Given this OpenAPI document (this is a fragment WIP of the ACME protocol defined in RFC8555, which is the APIs used for "Lets Encrypt")
```
openapi: '3.0.1'
# ...
  /key/new-nonce:
    description: Obtain an initialization nonce for inclusion in subsequent ACME requests
    summary: Key Service Nonce Request
    head:
      summary: getNewNonce
      description: Request a new replay protection nonce
      operationId: getNewNonce
      responses:
        '200':
          headers:
            Replay-Nonce:
              description: The Replay nonce is used for each request to the key service. It may only be used once, after which it is no longer valid. Each key service response will include
              schema:
                type: string
# ... other paths...
```
2. Run this CLI command 
`curl --head http://localhost:4010/key/new-nonce`
3. See error
```
[HTTP SERVER] head /key/new-nonce ℹ  info      Request received
[NEGOTIATOR] ℹ  info      Request contains an accept header: application/jose+json, application/problem+json, */*
[VALIDATOR] ✔  success   The request passed the validation rules. Looking for the best response
[NEGOTIATOR] ⚠  warning   Unable to find a content for */*
[HTTP SERVER] head /key/v1/new-nonce ✖  error     Request terminated with error: https://stoplight.io/prism/errors#NOT_ACCEPTABLE: The server cannot produce a representation for your accept header
```
the problem bit being `[NEGOTIATOR] ⚠  warning   Unable to find a content for */*`

**Expected behavior**
A `200` response with `Replay-Nonce: <nonce value>` and a null body.

 - Library version:  3.0.3

lag-of-death:
Fixes #575 

## Checklist

Bug fix


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -153,7 +153,16 @@ const helpers = {
           );
         } else {
           logger.warn(`Unable to find a content for ${mediaTypes}`);
-          return left(ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`));
+
+          if (_httpOperation.method === 'head') {
```
XVincentX: Hmm, I think we might want to make this check as soon as possible so we can skip the whole negotiation process if possible?

In this case though, the response body will always be empty even if the user has effectively specified a content for the HEAD verb. @philsturgeon what do you think? Maybe Spectral should somehow notice this?

lag-of-death: @XVincentX, should we do the re-write here? Moving up the check would require changes in finding the `response`

philsturgeon: When in doubt, the HTTP specification is a good baseline for what how to handle most situations. RFC 7231 says this about HEAD:

```
4.3.2.  HEAD

   The HEAD method is identical to GET except that the server MUST NOT
   send a message body in the response (i.e., the response terminates at
   the end of the header section).  The server SHOULD send the same
   header fields in response to a HEAD request as it would have sent if
   the request had been a GET, except that the payload header fields
   (Section 3.3) MAY be omitted.  This method can be used for obtaining
   metadata about the selected representation without transferring the
   representation data and is often used for testing hypertext links for
   validity, accessibility, and recent modification.

   A payload within a HEAD request message has no defined semantics;
   sending a payload body on a HEAD request might cause some existing
   implementations to reject the request.

   The response to a HEAD request is cacheable; a cache MAY use it to
   satisfy subsequent HEAD requests unless otherwise indicated by the
   Cache-Control header field (Section 5.2 of [RFC7234]).  A HEAD
   response might also have an effect on previously cached responses to
   GET; see Section 4.3.5 of [RFC7234].
```

The important bit here is "The HEAD method is identical to GET except that the server MUST NOT send a message body in the response". If they have defined a response we don't need to show it. 

**Personal opinion:** Maybe 1 user ever is going to define a head with a body, then when it doesn't work they'll google around and find out they should not have done that.

**Future feature idea:** if no HEAD is defined and they run a GET then it should just use the GET header content and strip the body. Lots of web application frameworks offer this as functionality too.

test-harness/specs/head_requests.oas3.txt
```diff
@@ -0,0 +1,27 @@
...
+                type: string
+                example:
+                  abc
+====server====
+mock -p 4010
+====command====
+curl --head http://localhost:4010/todos
+====expect====
+HTTP/1.1 200 OK
+content-type: */*
```
philsturgeon: I would not accept to see a wildcard on the response, that is a shortcut for requesting Accept, no?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
lag-of-death: @philsturgeon, is this OK ?
>Future feature idea: if no HEAD is defined and they run a GET then it should just use the GET header content and strip the body. Lots of web application frameworks offer this as functionality too.

test-harness/specs/head_requests.oas3.txt
lag-of-death: @philsturgeon, do you mean that I should add -H "Accept: whatever" ?

lag-of-death:
@philsturgeon, could you take a look at the questions ^ ? thanks!

test-harness/specs/head_requests.oas3.txt
philsturgeon: Nope im talking about this specific line. Why is `content-type: */*` being returned? I do not know why */* is returned.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
philsturgeon: Is what ok?

lag-of-death: @philsturgeon, is what you written OK? You said `and they run a GET then it should just use the GET header content and strip the body`. Why should I strip anything if I sand GET? I'm coooooonfused

test-harness/specs/head_requests.oas3.txt
lag-of-death: You're right. It shouldn't have `content-type` at all

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: Noted. So when an HEAD request comes In we'll skip the whole negotiation process.

lag-of-death: @XVincentX, so I'm moving this back to in-progress and will skip the negotiation in this PR

XVincentX: Yes please. If you're in trouble or overwhelmed with other stuff please let me know, I'm now back 100% and I can share some of the work.

lag-of-death: I will if I do, thanks :+1:

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -33,24 +33,36 @@ class HttpMocker
...
-          mockConfig.mediaTypes = acceptMediaType.split(',');
-        }
-
-        return mockConfig;
-      }),
-      chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
-      chain(result => assembleResponse(result, payloadGenerator)),
-    );
+    const isHead = input.data.method === 'head';
+    const r = get(resource, ['responses', '0'], { code: 200, headers: [] });
```
XVincentX: This should only be happening if `isHead` is true, otherwise the result will never be used.

XVincentX: This won't be enough, it's simply relying on the first response; imagine my first response is a 500 instead of a 200 (or also a 204) — this is not what we want to return. We still need to go through some parts of the negotiation process. It would be a good idea to add a test to make sure it behaves correctly on this.

1. We still need input validation (query string and whatever)
2. We need to still go through the regular response checking.

I guess I was kind of wrong when I was saying we can skip most of the negotiation process, I am sorry.

I think the best way to plug the code would be here:

https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L140

We can detect here if we're dealing with an `head` request and in such case early return a right with an `empty` body skipping all the code we see. If we could write a log line as well, that'd be great.

Does that make sense?

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -33,24 +33,36 @@ class HttpMocker
...
+
+    return isHead
+      ? withLogger(logger => {
+          logger.info('Returning an empty response for a HEAD request.');
+
+          return right({
+            statusCode: +r.code,
```
XVincentX: Aha! Nice trick!

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -33,24 +33,36 @@ class HttpMocker
...
-        return mockConfig;
-      }),
-      chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
-      chain(result => assembleResponse(result, payloadGenerator)),
-    );
+    const isHead = input.data.method === 'head';
+    const r = get(resource, ['responses', '0'], { code: 200, headers: [] });
+
+    return isHead
+      ? withLogger(logger => {
```
XVincentX: I think we can easily find a way to put this block in the one that's inside and have an early return, without having to construct two `withLogger` blocks. Basically the idea would be to defer such check as much as possible. Does that make sense?

test-harness/specs/head_requests.oas3.txt
```diff
@@ -0,0 +1,26 @@
...
+                type: string
+                example:
+                  abc
+====server====
+mock -p 4010
+====command====
+curl --head http://localhost:4010/todos
+====expect====
+HTTP/1.1 200 OK
+nonce: abc
```
XVincentX: Can we also add a test to make sure that if we request an `HEAD` and this is not defined in the document, we get a 404?

Just to make sure the router stops the request entirely. I know this is covered in unit tests as well, but you never know I guess :)

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Prism is correctly returning a payload-less response to a HEAD request #603
```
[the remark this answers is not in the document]
philsturgeon: You don't need to fix em all right now that can be done when polishing things in the release branch.

```
- HEAD requests no longer fail with 406 Not Acceptable #603
```

packages/http/src/mocker/HttpMocker.ts
philsturgeon: What does this do? Is there a more readable equivalent?

lag-of-death: @philsturgeon, if you know JS, then this is readable as readable gets

XVincentX: @lag-of-death I think we can use `parseInt` as well — it _will_ effectively improve the readability, right?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
lag-of-death: > @philsturgeon, is what you written OK? You said `and they run a GET then it should just use the GET header content and strip the body`. Why should I strip anything if I sand GET? I'm coooooonfused

@philsturgeon  ?

XVincentX: Ignore this. @lag-of-death this is a future feature we won't implement right now.

lag-of-death: @XVincentX, but what does it mean? I send GET and I strip the response body? I don't get it

packages/http/src/mocker/HttpMocker.ts
lag-of-death: @XVincentX, don't know, the same for me. Is your preference to change it to `parseInt`?

XVincentX: I'm fine with both, I was just pointing out there's an alternative that (maybe) is more readable.

philsturgeon: This was a long conversation @lag-of-death, please can you be aware and respectful of peoples time. I asked if there was a more readable alternative and you could have said "Yes" and made the code more readable. If I was asking, somebody else will also appreciate it. 

As a general rule, readable is usually preferable to clever. 🙌

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: @lag-of-death No I think @philsturgeon's idea is that if you do not specify a response for an HEAD request explicitly in the OpenAPI document, Prism could intelligently just take the GET request (if defined), strip its body and return it to the client.

We can skip this for now. Does that clarify?

lag-of-death: @XVincentX, this would make sense, yes. But I have a feeling this is not what @philsturgeon said. Maybe that's what he had in mind? I just want to be sure, is all (to be on the same page and not to make wrong assumptions).

philsturgeon: Yep I mentioned it as a future idea if you want to take note of it in another issue or if its easy to do now, whatever. Y'all are free to use your own judgement and manage more of the feature planning yourself.

CHANGELOG.md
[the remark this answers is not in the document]
XVincentX: I agree, your description is much better

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -135,6 +134,15 @@ const helpers = {
     const { mediaTypes, dynamic, exampleKey } = desiredOptions;
 
     return withLogger(logger => {
+      if (_httpOperation.method === 'head') {
+        logger.warn(`Responding with an empty body to a HEAD request.`);
```
XVincentX: This should not be a warn, there's nothing to be worried about. We can use `info`

packages/http/src/mocker/negotiator/types.ts
```diff
@@ -4,7 +4,7 @@ import { IHttpOperationConfig, JSONSchema } from '../../';
 
 export interface IHttpNegotiationResult {
   code: string;
-  mediaType: string;
+  mediaType?: string;
```
XVincentX: Ah ok I understand now, if the request is `HEAD` there is no media type we're effectively returning.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -525,26 +526,52 @@ describe('NegotiatorHelpers', () => {
...
+        it('should return a payload-less response', () => {
+          const actualResponse = helpers.negotiateOptionsBySpecificResponse(
+            { ...httpOperation, method: 'head' },
+            { dynamic: false, mediaTypes: ['*/*'] },
+            { code: '200' },
+          )(logger);
+
+          expect(Either.isRight(actualResponse)).toBeTruthy();
+
+          pipe(
```
XVincentX: use the `assertRight` function.

```ts
          assertRight(
            actualResponse,
            response => {
              expect(response).not.toHaveProperty('bodyExample');
              expect(response).not.toHaveProperty('mediaType');
              expect(response).not.toHaveProperty('schema');
            },
          );
```

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents [#577](https://github.com/stoplightio/prism/pulls/577)
 - Prism is correctly returning the response when the request has `*/*` as Accept header [#578](https://github.com/stoplightio/prism/pulls/578)
 - Prism is correctly returning a single root node with the payload for XML data [#578](https://github.com/stoplightio/prism/pulls/578)
+- HEAD requests no longer fail with 406 Not Acceptable #603
```
XVincentX: This is not in the right section. It should go In the `Fixed` part of `Unreleased` section

PR #604, opened 2019-09-12
fix responding with application/json

### LINKED ISSUE #599: Accept header with multiple types default to "text/plain"
**Describe the bug**
When I request prism with an "accept" header that has multiple types (ex: `application/json, text/plain, */*`), prism will try to send back "text/plain" response.

**To Reproduce**

```
openapi: 3.0.0
info:
  title: My title
  version: 1.0.0
paths:
  /check-email:
    post:
      tags:
        - User
      summary: Check the availability of an email
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                email:
                  type: string
                  example: john.doe@domain.tld
        required: true
      responses:
        "204":
          description: OK, the email is available
          content:
            application/json:
              schema:
                nullable: true
        "400":
          description: Bad Request, the email is not available
          content:
            application/json:
              schema:
                nullable: true
```
2. Run this CLI command `http POST 127.0.0.1:4010/check-email email=d@d.com Accept:'application/json, text/plain, */*'`
3. See error 
```
{
    "detail": "Cannot serialise complex objects as text/plain",
    "status": 500,
    "title": "Cannot serialise complex objects as text/plain",
    "type": "https://stoplight.io/prism/errors#NO_COMPLEX_OBJECT_TEXT_PLAIN"
}
```

**Expected behavior**
I expected the server to respond with a `204` response.

 - Library version: 3.1.0
 - OS: Ubuntu 18.04.3 LTS (Bionic Beaver)

**Additional context**
I am using "axios" to make the call and it defaults to `application/json, text/plain, */*'` for its 'Accept' header.

lag-of-death:
Fixes #599 

## Checklist

Handling `application/json` in **Accept** header is inconsistent.
This PR implements a small-footprint solution. But **we should really consider a solution similar to the one proposed in**: https://github.com/stoplightio/prism/pull/568 

The issue:

We receive: Accept: `application/json, text/plain`
Fastify looks into the serializers and sees that there is no serializer for `application/json`, but founds `text/plain` serializer and uses this. This is wrong. This happens because our `application/json` serializer is a fallback serializer (when no serializer in `serializers` (https://github.com/stoplightio/prism/blob/master/packages/http-server/src/serializers.ts) is found)

## Why the added charset=utf8?

This is to solve the inconsistency we have. This was also addressed in #568, but didn't make it to `master`. 

An example of the inconsistency from `master`:

https://github.com/stoplightio/prism/blob/887e2d0210b2ef71d076c810898c67ec3d357d82/packages/http-server/src/__tests__/server.oas.spec.ts#L309-L319

https://github.com/stoplightio/prism/blob/887e2d0210b2ef71d076c810898c67ec3d357d82/packages/http-server/src/__tests__/server.oas.spec.ts#L323-L332

So, in the first example we have `application/json; charset=utf-8`, but in the second one only `application/json`.
And why is this happening? Because https://github.com/stoplightio/prism/pull/568#discussion_r319552501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 # Unreleased
 
+## Added
+
+## Fixed
+
+- Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. #604
+
 # 3.1.0 (2019-09-03)
 
 ## Added
@@ -22,7 +28,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
-- Prism is correctly serializing application/json responses
 
 # 3.0.4 (2019-08-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Prism is correctly serializing application/json responses
 
 # 3.0.4 (2019-08-20)
 
--- a/packages/http-server/src/serializers.ts
+++ b/packages/http-server/src/serializers.ts
@@ -6,15 +6,8 @@ const xmlSerializer = new j2xParser({});
 export default [
   {
     regex: {
-      test: (value: string) => !!typeIs.is(value, ['application/*+json']),
+      test: (value: string) => !!typeIs.is(value, ['application/*+json', 'application/json']),
       toString: () => 'application/*+json',
-    },
-    serializer: JSON.stringify,
-  },
-  {
-    regex: {
-      test: (value: string) => !!typeIs.is(value, ['application/json']),
-      toString: () => 'application/json',
     },
     serializer: JSON.stringify,
   },
--- a/packages/http-server/src/serializers.ts
+++ b/packages/http-server/src/serializers.ts
@@ -13,6 +13,13 @@ export default [
   },
   {
     regex: {
+      test: (value: string) => !!typeIs.is(value, ['application/json']),
+      toString: () => 'application/json',
+    },
+    serializer: JSON.stringify,
+  },
+  {
+    regex: {
       test: (value: string) => !!typeIs.is(value, ['application/xml', 'application/*+xml']),
       toString: () => 'application/*+xml',
     },
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -41,10 +41,6 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
   });
 
   server.addHook('onSend', (request, reply, payload, done) => {
-    if (reply.getHeader('content-type') === 'application/json') {
-      reply.header('content-type', 'application/json; charset=utf-8');
-    }
-
     if (request.headers.accept === '*/*' && typeof payload !== 'string') {
       const serializer = reply.hasHeader('content-type')
         ? serializers.find(s => s.regex.test(reply.getHeader('content-type')!))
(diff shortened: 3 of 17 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Prism is correctly serializing application/json responses
```
philsturgeon: These changelog messages are not as effective at communicating what happened.

"Prism is" we dont need this because we know its Prism
"correctly" we know we fixed something this in in the fixed section

What was fixed? What was it doing before? What is the scenario in which application/json scenarios were broken?

People read changelogs to find out if a specific situation that they might have noticed has been fixed, but this is not giving them that information. It's just kinda vaguely saying "something about json works better than before"

packages/http-server/src/server.ts
```diff
@@ -41,6 +41,10 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
   });
 
   server.addHook('onSend', (request, reply, payload, done) => {
+    if (reply.getHeader('content-type') === 'application/json') {
```
philsturgeon: What is the reasoning for this? If the api designer has defined the header as `application/json`why do we want to force the utf8 charset on there?

test-harness/specs/E2E-Accept-Content-Negotiation-AC-15.oas2.txt
```diff
@@ -0,0 +1,24 @@
...
+      produces:
+        - application/json
+        - text/plain
+      responses:
+        "204":
+          description: OK, the email is available
+          examples:
+            application/json:
+              schema:
+                nullable: true
```
philsturgeon: This test might just show that no body is output when schema is a null, maybe the more standard `type: string` would create more confidence that this test is doing what we want for the reasons we want.

packages/http-server/src/server.ts
StefanDywersant: What about support for `application/{dataType}+json` or `application/json; charset=cp1250`?

Source: https://tools.ietf.org/html/rfc6839

lag-of-death: @philsturgeon, @StefanDywersant, `charset=utf-8` was already enforced by `fastify` with its default `application/json` serializer. We sometimes returned `application/json` without `charset=utf-8` because... the response was wrongly serialized by some other serializer (like instead of a serializer for JSON, the data was serialized with a serializer for XML). `application/{whatever}+json` is here https://github.com/stoplightio/prism/blob/master/packages/http-server/src/serializers.ts#L9
haven't found other charset than utf-8 logic in the code

packages/http-server/src/serializers.ts
```diff
@@ -11,6 +11,13 @@ export default [
     },
     serializer: JSON.stringify,
   },
+  {
+    regex: {
```
XVincentX: You do not need to create a new section, you can simply modify the one at line 9 and add `application/json` in the array

packages/http-server/src/server.ts
XVincentX: > What about support for application/{dataType}+json

This scenario is already supported in the serialiser. I would not bother about the `charset` — it is actually something that is inferred by the browser most of the times. I am quite sure there was also a proposal to remove it completely from the content type header.

test-harness/specs/E2E-Accept-Content-Negotiation-AC-15.oas2.txt
XVincentX: I agree with @philsturgeon here and more importantly `204` responses do not have content and so Fastify will *delete it* no matter if you're trying to send it to the client, so let's change this to a regular 200.

test-harness/specs/E2E-Accept-Content-Negotiation-AC-15.oas3.txt
```diff
@@ -0,0 +1,21 @@
+====test====
+Sending multiple values in Accept header selects the first one matched.
+====spec====
+openapi: 3.0.2
+paths:
+  /check-email:
+    get:
+      responses:
+        "204":
```
XVincentX: Ditto, do not use 204 here.
```suggestion
        "200":
```

test-harness/specs/E2E-Accept-Content-Negotiation-AC-4.oas2.txt
```diff
@@ -22,7 +22,7 @@ mock -p 4010
 curl -i -X GET http://localhost:4010/todos -H "accept: application/xml, application/json"
 ====expect====
 HTTP/1.1 200 OK
-content-type: application/json
+content-type: application/json; charset=utf-8
```
XVincentX: It's ok to return just `application/json`

CHANGELOG.md
XVincentX: The correct changeling here would be a line explaining that Prism is now giving higher priority to `application/json` Accept content type, and it's not a "fallback" format anymore.

XVincentX: Moreover this is not in the right section, it should go in the `Unreleased` part

PR #605, opened 2019-09-12
generate dynamic properties

### LINKED ISSUE #570: Prism mock dynamic response generation missing required fields on oneOf/allOf schemas
**Describe the bug**
I used `prism mock api.yaml -d` to spin up a mock server for my openAPI spec, the dynamically generated responses always miss some required fields.

**To Reproduce**
1. Sample API spec:
```
openapi: 3.0.0
info:
  title: Sample API
  version: v1
paths:
  /services/resources/{resource_id}:
    get:
      summary: Get a resource detail.
      description: Returns a single resource.
      operationId: getV1Resource
      responses:
        200:
          description: get a resource
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Resource'
        404:
          description: not found
components:
  schemas:
    Resource:
      type: object
      discriminator:
        propertyName: resourceType
      oneOf:
        - $ref: '#/components/schemas/A'
        - $ref: '#/components/schemas/B'
        - $ref: '#/components/schemas/C'
        - $ref: '#/components/schemas/D'
    A:
      required:
        - name
        - resourceType
      type: object
      properties:
        resourceType:
          type: string
        name:
          type: string
        color:
          type: string
        length:
          type: integer
    B:
      required:
        - name
        - resourceType
      type: object
      properties:
        resourceType:
          type: string
        name:
          type: string
        description:
          type: string
        age:
          type: integer
    C:
      required:
        - data
        - resourceType
      type: object
      properties:
        resourceType:
          type: string
        data:
          type: string
        password:
          type: string
        hint:
          type: string
    D:
      required:
        - value
        - resourceType
      type: object
      properties:
        resourceType:
          type: string
        value:
          type: string
        note:
          type: string
```
2. Run this CLI command `prism mock api.yaml -d`
3. Use postman to get sample response

**Expected behavior**
The expected response should have field 'resourceType' for all, and should have 'name' field for A and B, but the actual responses never have those fields 

Some actual responses;

{
    "color": "elit irure voluptate sunt",
    "length": 80304934
}

{
    "description": "deserunt ut in",
    "age": -65249860
}

{
    "data": "exercitation",
    "password": "nisi",
    "hint": "magna enim ullamco Ut consequat"
}

 - Library version: 3.0.4
 - OS: MacOS 10.13.6

**Additional context**
if I use 'static' generation without '-d' in the command, I can see all the field but it will always return schema A.

e.g.
{
    "resourceType": "string",
    "name": "string",
    "color": "string",
    "length": 0
}

lag-of-death:
Fixes #570

## Checklist

Bug fix.

`json-schema-faker` doesn't generate `required` properties when `alwaysFakeOptionals` is on. So we now have this hacky solution. What happens is `required` attribute is removed from the passed-in schema. So, all of the properties are optional during faking. But, with `alwaysFakeOptionals: true`, all of the properties are faked. This solution is meant to be temporary until it's resolved in `json-schema-faker`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is correctly handling the `csv` collection format argument property in OAS2 documents #577
 - Prism is correctly returning the response when the request has `*/*` as Accept header #578
 - Prism is correctly returning a single root node with the payload for XML data #578
+- Prism is generating dynamic properties for both required and optional ones
 
 # 3.0.4 (2019-08-20)
 
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -22,6 +22,7 @@
     "ajv-oai": "^1.0.0",
     "axios": "^0.18.0",
     "caseless": "^0.12.0",
+    "deep-cleaner": "^1.2.1",
     "faker": "^4.1.0",
     "fp-ts": "^2.0.5",
     "json-schema-faker": "json-schema-faker/json-schema-faker",
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -7,6 +7,8 @@ import * as jsf from 'json-schema-faker';
 // @ts-ignore
 import * as sampler from 'openapi-sampler';
 
+const dc = require('deep-cleaner');
+
 jsf.extend('faker', () => faker);
 
 jsf.option({
@@ -21,7 +23,7 @@ jsf.option({
 });
 
 export function generate(source: JSONSchema): unknown {
-  return jsf.generate(cloneDeep(source));
+  return jsf.generate(dc(cloneDeep(source), 'required'));
 }
 
 export function generateStatic(source: JSONSchema): unknown {
--- a/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
+++ b/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
@@ -77,5 +77,263 @@ describe('JSONSchema generator', () => {
 
       return expect(generate(schema)).toBeTruthy();
     });
+
+    describe('multiple properties', () => {
+      describe('no required properties specified', () => {
+        it('generates all of the properties', () => {
+          const schema: JSONSchema = {
+            type: 'object',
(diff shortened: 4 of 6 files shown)


philsturgeon:
> json-schema-faker doesn't generate required properties when alwaysFakeOptionals is on.

is this true? This sounds backwards. Why would "always fake optionals" mean "do not fake required"? Is this a bug in JSF?

lag-of-death:
> > json-schema-faker doesn't generate required properties when alwaysFakeOptionals is on.
> 
> is this true? This sounds backwards. Why would "always fake optionals" mean "do not fake required"? Is this a bug in JSF?

@philsturgeon, sadly, but yes and yes

mallachari:
Just wondering. There's [@stoplightio/json-schema-faker](https://github.com/stoplightio/json-schema-faker) fork used by some projects. Maybe we could use it and fix that bug there instead of the hack? Only if it's a quick fix of course, otherwise there's no point in that as it should be fixed in the original repo. What do you think?

philsturgeon:
That fork is no longer used by any projects and is scheduled for deletion. We should submit fixes to projects via PRs instead of maintaining forks, it’s an extra hassle we don’t need :)

If they’ve got issues let’s send em fixed!

mallachari:
> We should submit fixes to projects via PRs instead of maintaining forks, it’s an extra hassle we don’t need :)

That's much better approach indeed. I was just thinking of something that might be faster, but you're right. I didn't know that this fork is no longer used.

> If they’ve got issues let’s send em fixed!

Agree with that!

XVincentX:
Ok apparently this is done on purpose by the library, so there's something going on we're missing.

image

XVincentX:
Note 2: It works correctly if the `oneOf` has only one element:

image

XVincentX:
Effectively when there's more than a schema in a `oneOf` the derived target schema is missing the required properties:

image

XVincentX:
@philsturgeon @lag-of-death 

After couple of hours of debug I've identified the offending commit in Json Schema Faker; it has been put on purpose to handle another case; basically the problem is that two schemas are sharing the same required properties and this is confusing the software.

I'm closing this since this is not something we should fix in Prism directly or, if we really need to, it must be the last shot.

For more informations:

https://github.com/json-schema-faker/json-schema-faker/commit/16286c217336a84e5c7128a981f6ec1374371297#r35110880

PR #606, opened 2019-09-13
allow empty responses

### LINKED ISSUE #584: Negotiator incorrectly returns a 406 for 204's with no defined response
**Describe the bug**
When Prism is mocking an OA3 specification containing a 204 response, it incorrectly returns an error since it can not find a valid response layout and specification for the same. 

As per [RFC 7231](https://tools.ietf.org/html/rfc7231), [Section 6.3.5](https://tools.ietf.org/html/rfc7231#section-6.3.5):

> The 204 (No Content) status code indicates that the server has successfully fulfilled the request and that there is no additional content to send in the response payload body.

The bug can be traced back to `@stoplight/prism-http/src/mocker/negotiator/NegotiatorHelper.ts` on [Line 154](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L154). 

**To Reproduce**

1. Given this OpenAPI document:

```yaml
openapi: 3.0.0
info:
  title: Test
  version: '1.0'
servers:
paths:
  /auth/logout:
    get:
      summary: Deletes the current session
      tags: []
      responses:
        '204':
          description: No Content
        '404':
          description: No session with the provided session ID was found.
          content:
            application/json:
              schema:
                type: object
                properties:
                  errors:
                    type: array
                    items:
                      type: object
                      properties:
                        code:
                          type: string
                        statusCode:
                          type: string
                        description:
                          type: string
                      required:
                        - code
                        - statusCode
                        - description
                required:
                  - errors
              examples:
                '404':
                  errors:
                    - code: SESSION_NFOUND
                      statusCode: '404'
                      message: No session with the provided ID was found.
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                sessionId:
                  type: string
                  format: uuid
              required:
                - sessionId
```

2. Run this CLI command: `prism mock ./openapi.yaml`
3. Make a `curl` GET request: `curl -XGET http://127.0.0.1:4010/auth/logout`
4. See error:
```json
{
    "type": "https://stoplight.io/prism/errors#NOT_ACCEPTABLE",
    "title": "The server cannot produce a representation for your accept header",
    "status": 406,
    "detail": "Unable to find content for */*"
}
```

**Expected behavior**
Prism should just return 204 with no body or content unless explicitly added to the response definition.

**Screenshots**
Screen Shot 2019-09-05 at 12 17 24 PM

 - Prism version: v3.1.0
 - OS: macOS 10.14.6

**Possible Fix**
The simplest fix is to add an explicit check for status code 204 in the negotiator and let the application continue from there instead of falling back on the negotiatingByMediaType which inevitably gives the standardized `ProblemJson` response.

I have a working PR for this and would love to submit it. 😄

### LINKED ISSUE #592: Empty responses are not supported for 401's due to failed security scheme
The [openAPI](https://swagger.io/docs/specification/describing-responses/) spec says that I can specify a response without a body by not providing `content:` or `schema:`. However, if one of these is not present, the errors generated by the new securitySchema check will not be used, and prism will use it's default response.

**To Reproduce**

1.  Have a response defined like this attached to an endpoint as a 401

```
InvalidAuthorizationTokenResponse:
      description: "The provided authorization token is invalid or expired."
```

```
responses:
        200:
        401:
```

and a security scheme like:
```
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

2. Hit the endpoint without an authorization token
3. The default error message is sent since the 401 has no schema or example:

```
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

**Expected behavior**
I'd like to have my custom, bodyless 401 returned.

lag-of-death:
Fixes #584
Fixes #592

## Checklist

Bug fix.

No empty responses were allowed.


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -142,12 +170,7 @@ const helpers = {
         return pipe(
           httpContent,
           Option.fold(
-            () => {
-              logger.warn(`Unable to find a content for ${mediaTypes}`);
-              return Either.left<Error, IHttpNegotiationResult>(
-                ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`),
-              );
-            },
+            () => whenNoContent(logger, response, headers || [], mediaTypes),
```
StefanDywersant: You should also apply the support for empty content here: https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L316

It will let you "hit two birds with one stone". This [PR](https://github.com/stoplightio/prism/pull/609) introduces empty response for invalid requests, however you solution looks more generic.

If you feel that this is not in scope I'll create new PR based on yours.

lag-of-death: thanks @StefanDywersant for pointing this out, will take care of this in this PR

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -22,6 +23,30 @@ import {
 } from './InternalHelpers';
 import { IHttpNegotiationResult, NegotiatePartialOptions, NegotiationOptions } from './types';
 
+export const warnMsg = (contentTypes: string[]) => `Unable to find content for ${contentTypes}`;
+
+const whenNoContent = (
+  logger: Logger,
+  response: IHttpOperationResponse,
+  headers: IHttpHeaderParam[],
+  mediaTypes: string[],
```
XVincentX: I would really specify the return type explicitly — it helps the readability but more importantly the type checker will help us in case we're going to refactor this function, effectively disabling the inference.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -22,6 +23,30 @@ import {
 } from './InternalHelpers';
 import { IHttpNegotiationResult, NegotiatePartialOptions, NegotiationOptions } from './types';
 
+export const warnMsg = (contentTypes: string[]) => `Unable to find content for ${contentTypes}`;
```
XVincentX: In general if we're not using this at least 2-3 times, let's refrain from creating another function.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -22,6 +23,30 @@ import {
...
+  mediaTypes: string[],
+) => {
+  const contents = (response.contents && response.contents.filter(c => !c.mediaType.includes('*/*'))) || [];
+  const isContentEmpty = contents.length === 0;
+  const wasAcceptHeaderLeftUnspecified = mediaTypes.filter(mt => !mt.includes('*/*')).length === 0;
+
+  if (isContentEmpty && wasAcceptHeaderLeftUnspecified) {
+    logger.warn(`${warnMsg(mediaTypes)}. Sending an empty response.`);
+
+    return Option.some({
```
XVincentX: I am assuming this is an `IHttpResponse` object?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -143,9 +168,18 @@ const helpers = {
           httpContent,
           Option.fold(
             () => {
-              logger.warn(`Unable to find a content for ${mediaTypes}`);
-              return Either.left<Error, IHttpNegotiationResult>(
-                ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`),
+              return pipe(
+                whenNoContent(logger, response, headers || [], mediaTypes),
```
XVincentX: The readability of this is not clear. `whenNoContent` — what happens?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -143,9 +168,18 @@ const helpers = {
...
+                Option.fold(
+                  () => {
+                    logger.warn(warnMsg(mediaTypes));
+
+                    return Either.left<Error, IHttpNegotiationResult>(
+                      ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`),
+                    );
+                  },
+                  r => Either.right(r),
```
XVincentX: You do not need to use `fold` if all you need is convert this to an `Either`

`Either.fromOption` is the function you're looking for.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -313,8 +347,25 @@ const helpers = {
...
+                    Option.fold(
+                      () => {
+                        logger.trace(
+                          `Unable to find a content with a schema defined for the response ${response.code}`,
+                        );
+                        return Either.left(
+                          new Error(`Neither schema nor example defined for ${response.code} response.`),
+                        );
+                      },
+                      r => Either.right(r),
```
XVincentX: ditto, `Either.fromOption` is the function you should be using here.

packages/http/src/mocker/HttpMocker.ts
```diff
@@ -119,12 +119,13 @@ function assembleResponse(
         const mockedBody = computeBody(negotiationResult, payloadGenerator);
         const mockedHeaders = computeMockedHeaders(negotiationResult.headers || [], payloadGenerator);
 
+        const headers = negotiationResult.mediaType
```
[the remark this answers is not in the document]
lag-of-death: This is to not have `Content-Type` when a response has no payload.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
lag-of-death: 1st use: https://github.com/stoplightio/prism/pull/606/files#diff-47034315930357c592d473ec0fe19afcR179
  
2nd use:  https://github.com/stoplightio/prism/pull/606/files#diff-47034315930357c592d473ec0fe19afcR184

lag-of-death:
@XVincentX, I believe the feedback is addressed now, ready for another round

XVincentX:
@lag-of-death Can you rebase when you get a chance?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -153,9 +183,21 @@ const helpers = {
           httpContent,
           Option.fold(
             () => {
-              logger.warn(`Unable to find a content for ${mediaTypes}`);
-              return Either.left<Error, IHttpNegotiationResult>(
-                ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`),
+              return pipe(
+                optionallyGetPayloadlessResponse(response, headers || [], mediaTypes),
```
XVincentX: We should come up with a better function name, this one is kind of awkward.

lag-of-death: @XVincentX, I'm all ears :)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -23,6 +25,34 @@ import {
...
+  return pipe(
+    fromNullable(response.contents),
+    Option.map(contents => contents.filter(c => !c.mediaType.includes('*/*'))),
+    Option.chain(filteredContents => {
+      const wasAcceptHeaderLeftUnspecified = mediaTypes.filter(mt => !mt.includes('*/*')).length === 0;
+
+      return Option.fromPredicate((c: { length: number }) => {
+        const isContentEmpty = c.length === 0;
+
+        return isContentEmpty && wasAcceptHeaderLeftUnspecified;
```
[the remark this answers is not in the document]
lag-of-death: Sure :)
so our `c` is `contents.filter(c => !c.mediaType.includes('*/*')` and our `wasAcceptHeaderLeftUnspecified` is `mediaTypes.filter(mt => !mt.includes('*/*')).length === 0`, so we know that we are excluding `*/*` from the equation here.
So `isContentEmpty && wasAcceptHeaderLeftUnspecified` is saying:
`if there is no specific content type defined in the operation` AND `Accept header sent from the user was different than */* or left empty`, then later 
```
return {
        code: response.code,
        headers,
      };
```
which basically means `return a payloadless response`
Hope it clarifies ;)

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -138,8 +138,7 @@ describe.each([['petstore.no-auth.oas2.yaml', 'petstore.no-auth.oas3.yaml']])('s
       url: '/pets/findByTags',
     });
 
-    expect(response.statusCode).toBe(422);
-    checkErrorPayloadShape(response.payload);
+    expect(response.statusCode).toBe(400);
```
lag-of-death: because now empty responses are allowed

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -146,9 +175,7 @@ describe('NegotiatorHelpers', () => {
             logger,
           );
 
-          assertLeft(negotiationResult, e =>
-            expect(e.message).toBe('Neither schema nor example defined for 422 response.'),
-          );
+          assertRight(negotiationResult, e => expect(e).toStrictEqual({ code: '422', headers: [] }));
```
lag-of-death: because it now can be an empty response

packages/http-server/src/__tests__/server.oas.spec.ts
XVincentX: Oh whoa, this is really cool!

packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -19,7 +19,9 @@ export function findExampleByKey(httpContent: IHttpContent, exampleKey: string)
 }
 
 export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpOperationResponse, 'contents'> {
-  return !!v.contents;
+  const x = v.contents || [];
+
+  return !!x.filter(y => !y.mediaType.includes('*/*')).length;
```
XVincentX: We're going to remove this hack when http-spec will be fixed, right?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -1,12 +1,14 @@
 import { ProblemJsonError } from '@stoplight/prism-core';
 import { IHttpOperation, IHttpOperationResponse, IMediaTypeContent } from '@stoplight/types';
+import { IHttpHeaderParam } from '@stoplight/types';
 import * as Either from 'fp-ts/lib/Either';
 import { NonEmptyArray } from 'fp-ts/lib/NonEmptyArray';
+import { fromNullable } from 'fp-ts/lib/Option';
```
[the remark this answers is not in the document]
XVincentX: Ping

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -170,6 +222,27 @@ const helpers = {
...
+                //       : {};
+                //
+                //   return Object.assign(
+                //     {
+                //       headers: headers || [],
+                //       ...rest,
+                //     },
+                //     yo,
+                //   );
+                // }),
```
XVincentX: Can we remove this?

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -527,13 +554,27 @@ describe('NegotiatorHelpers', () => {
         });
       });
 
-      describe('httpContent does not exist ', () => {
+      describe('httpContent does not exist', () => {
```
XVincentX: Maybe we can specify that "the response exists, but there is no httpcontent"

test-harness/specs/E2E-Accept-Content-Negotiation-AC-7.oas2.txt
```diff
@@ -7,6 +7,8 @@ swagger: "2.0"
 paths:
   /todos:
     get:
+      produces:
+        - text/plain
```
XVincentX: Yes!

packages/http/src/mocker/negotiator/InternalHelpers.ts
lag-of-death: Yes :)

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -153,9 +183,21 @@ const helpers = {
...
+                optionallyGetPayloadlessResponse(response, headers || [], mediaTypes),
+                Option.map(payloadlessResponse => {
+                  logger.warn(`${warnMsg(mediaTypes)}. Sending an empty response.`);
+
+                  return payloadlessResponse;
+                }),
+                Either.fromOption(() => {
+                  logger.warn(warnMsg(mediaTypes));
+
+                  return ProblemJsonError.fromTemplate(
```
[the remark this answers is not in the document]
lag-of-death: hm.. weren't we doing this:
```typescript
return Either.left<Error, IHttpNegotiationResult>(
                ProblemJsonError.fromTemplate(NOT_ACCEPTABLE, `Unable to find content for ${mediaTypes}`),
```
wanted to stay consistent

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -23,6 +24,42 @@ import {
...
+      contents.filter(c => {
+        return !c.schema && !(c.examples || []).length;
+      }),
+    ),
+    Option.chain(responses => {
+      return Option.fromPredicate((noContentResponses: { length: number }) => {
+        const noAcceptAndNoContentResponses =
+          !noContentResponses.length && !mediaTypes.filter((mt: string) => !mt.includes('*/*')).length;
+
+        return noAcceptAndNoContentResponses || ['400', '422'].includes(response.code);
```
XVincentX: Why are you checking for 400 and 422 responses manually? Don't they follow the same flow?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -23,6 +24,42 @@ import {
...
+    ),
+    Option.map(contents =>
+      contents.filter(c => {
+        return !c.schema && !(c.examples || []).length;
+      }),
+    ),
+    Option.chain(responses => {
+      return Option.fromPredicate((noContentResponses: { length: number }) => {
+        const noAcceptAndNoContentResponses =
+          !noContentResponses.length && !mediaTypes.filter((mt: string) => !mt.includes('*/*')).length;
```
XVincentX: I really think we should put a comment on this condition, I still struggle to understand it correctly.

Also, with includes `Accept: application/json,*/*` will match. Is this intentional? I guess so (if the client accepts */* then an empty response is viable, right?)

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -84,6 +95,24 @@ describe('NegotiatorHelpers', () => {
       });
 
       describe('and has no static contents', () => {
+        it('returns an empty payload response', () => {
```
XVincentX: Let's put a word or two about the fact we're looking for a response for an invalid Request

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -35,14 +35,12 @@ const findEmptyResponse = (
...
-    Option.chain(responses => {
-      return Option.fromPredicate((noContentResponses: { length: number }) => {
-        const noAcceptAndNoContentResponses =
-          !noContentResponses.length && !mediaTypes.filter((mt: string) => !mt.includes('*/*')).length;
-
-        return noAcceptAndNoContentResponses || ['400', '422'].includes(response.code);
-      })(responses);
-    }),
+    Option.chain(responses =>
+      Option.fromPredicate(
```
XVincentX: `pipe(responses, Option.fromPredicate())`

lag-of-death:
hi @XVincentX, I think I've addressed the feedback. The logic for `findEmptyResponse` is now greatly simplified. 
The previous version was kind of redundant because some of the conditional checks were already performed up the chain - somewhere in `findEmptyResponse`'s caller or higher.
There are also a couple of more tests so that hopefully no edge case is missed.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -23,6 +24,28 @@ import {
...
 
+const outputNoContentFoundMessage = (contentTypes: string[]) => `Unable to find content for ${contentTypes}`;
+
+const findEmptyResponse = (
+  response: IHttpOperationResponse,
+  headers: IHttpHeaderParam[],
+  mediaTypes: string[],
+): Option.Option<IHttpNegotiationResult> => {
+  return pipe(
+    Option.fromPredicate((contentTypes: string[]) => {
```
XVincentX: `pipe(mediaTypes, Option.fromPredicate(…), Option.map`

PR #621, opened 2019-09-16
fix: use default/examples in json-schema-faker

### LINKED ISSUE #586: Support combination of static example and dynamic values
**User story.**
As an API provider, I can do use static data in combination with dynamic data, so that I can provide more useful examples.

I have a monetary amount object type that has two fields: amount and currency code. I'd like to have a dynamic amount and always `USD` as currency code.

Perhaps have another field like "x-faker" called "x-static" and have that take precedence.

StefanDywersant:
Closes #586 

## Checklist

Bug fix

Current: neither default nor examples are used for dynamic content generation using json-schema-faker. 

New: default/examples are used if available

No

### DIFF
--- a/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
+++ b/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
@@ -79,7 +79,7 @@ describe('JSONSchema generator', () => {
     });
 
     describe('when default value is set', () => {
-      it('uses it instead of generating one', () => {
+      it('uses it instead of generating value', () => {
         const schema: JSONSchema = {
           type: 'string',
           default: 'a-string-which-is-hard-to-be-randomly-generated-although-who-knows?',
@@ -90,7 +90,7 @@ describe('JSONSchema generator', () => {
     });
 
     describe('when examples are set', () => {
-      it('uses it instead of generating one', () => {
+      it('uses it instead of generating value', () => {
         const schema: JSONSchema = {
           type: 'string',
           examples: ['a-string-which-is-hard-to-be-randomly-generated-although-who-knows?'],
--- a/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
+++ b/packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
@@ -89,7 +89,7 @@ describe('JSONSchema generator', () => {
       });
     });
 
-    describe('when example value is set', () => {
+    describe('when examples are set', () => {
       it('uses it instead of generating one', () => {
         const schema: JSONSchema = {
           type: 'string',
@@ -97,6 +97,22 @@ describe('JSONSchema generator', () => {
         };
 
         expect(generate(schema)).toEqual('a-string-which-is-hard-to-be-randomly-generated-although-who-knows?');
+      });
+    });
+
+    describe('when default value and examples are set', () => {
+      it('uses either default value or examples', () => {
+        const schema: JSONSchema = {
+          type: 'string',
+          examples: ['an-examples-string-which-is-hard-to-be-randomly-generated-although-who-knows?'],
+          default: 'a-default-string-which-is-hard-to-be-randomly-generated-although-who-knows?',
+        };
+
+        expect(generate(schema)).toEqual(
+          expect.stringMatching(
+            /^a(n-examples|-default)-string-which-is-hard-to-be-randomly-generated-although-who-knows\?$/,
+          ),
+        );
       });
     });
   });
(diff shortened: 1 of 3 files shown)


StefanDywersant:
> looks good :) can we also add one test where we check what happens when we have both `examples` and `default`?

I'm afraid that it explodes if I do that :(

```

                             \         .  ./
                           \      .:";'.:.."   /
                               (M^^.^~~:.'").
                         -   (/  .    . . \ \)  -
  O                         ((| :. ~ ^  :. .|))
 |\\                     -   (\- |  \ /  |  /)  -
 |  T                         -\  \     /  /-
/ \[_]..........................\  \   /  /
```

lag-of-death:
@StefanDywersant, if it's `json-schema-faker` that explodes, then maybe before invoking `generate` let's remove `default` and put it inside `examples`? WDYT?

StefanDywersant:
> @StefanDywersant, if it's `json-schema-faker` that explodes, then maybe before invoking `generate` let's remove `default` and put it inside `examples`? WDYT?

I added the test for both default value and examples.

In such case jsf takes randomly one of them, so it exploded ;)

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -77,5 +77,43 @@ describe('JSONSchema generator', () => {
 
       return expect(generate(schema)).toBeTruthy();
     });
+
+    describe('when default value is set', () => {
+      it('uses it instead of generating one', () => {
```
XVincentX:
```suggestion
      it('uses it instead of generating a value', () => {
```

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -77,5 +77,43 @@ describe('JSONSchema generator', () => {
...
+          type: 'string',
+          default: 'a-string-which-is-hard-to-be-randomly-generated-although-who-knows?',
+        };
+
+        expect(generate(schema)).toEqual('a-string-which-is-hard-to-be-randomly-generated-although-who-knows?');
+      });
+    });
+
+    describe('when examples are set', () => {
+      it('uses it instead of generating one', () => {
```
XVincentX:
```suggestion
      it('uses it instead of generating a value', () => {
```

packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -77,5 +77,43 @@ describe('JSONSchema generator', () => {
...
+      it('uses either default value or examples', () => {
+        const schema: JSONSchema = {
+          type: 'string',
+          examples: ['an-examples-string-which-is-hard-to-be-randomly-generated-although-who-knows?'],
+          default: 'a-default-string-which-is-hard-to-be-randomly-generated-although-who-knows?',
+        };
+
+        expect(generate(schema)).toEqual(
+          expect.stringMatching(
+            /^a(n-examples|-default)-string-which-is-hard-to-be-randomly-generated-although-who-knows\?$/,
```
XVincentX: Aha! nice!

StefanDywersant:
> Ok the changes look good.
> 
> 2 things:
> 
> 1. Can we add a line on the CHANGELOG file explaining the change?
> 2. I want to make sure @philsturgeon is ok with the change. I recall we had a discussion when we were writing this piece of functionality and we specifically decided that `-d` would be fully dynamic (as in — ignore defaults and examples). @philsturgeon are you ok with this?

1. changelog line added 👍 
2. waiting with merge for @philsturgeon :)

philsturgeon:
Hey! So basically the idea is "static" is repeatable and the more information it has (enum, examples, default) the more useful that predictable mock response will be. Dynamic is always going to give you different randomly generated data, and so we actually do not _want_ to use the examples or default. 

Does this make sense?

chrisjallen:
Sorry to be late to the party on this, @karol-maciaszek is right, we need a means for having both. So by removing the -D flag on start up, is it possible to mark in yaml which content can be dynamic vs statically generated? Seems like a deal breaker to me that in some examples we need to pick from  a predefined set of examples, and in others we can rely on x-faker properties to do the heavy lifting of populating examples for us dynamically.... I realise this is an old post, is this question still valid?

chrisjallen:
Solutions: 
### so if mixing is a problem, import a separate yaml file with ability to call out a dynamic mode much like the cli:
$ref: ./someotherfile.yaml?prefer=static

### inline properties using same language as cli
in this case the x-prefer allows choice between dynamic or static (so example) route
````
  x-prefer: dynamic
  properties:
    name:
      type: string
      x-faker: lorum
   - examples:
      ...blah
````

PR #628, opened 2019-09-18
fix: return correct error code based on the validation type

### LINKED ISSUE #591: When request doesn't contain proper security scheme, return the user-defined 401 first

XVincentX:
1. Remove 403 leftovers
2. Generalize the `findResponse` so that somebody can provide an array of what we're looking for
3. Look fo the right response according to the validation type (security comes first, validation comes later)

Closes #591


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -248,22 +249,28 @@ const helpers = {
...
-          logger.trace('Unable to find a 400 response definition.');
-          return findResponseByStatusCode(httpResponses, '401');
-        }),
-        Option.alt(() => findResponseByStatusCode(httpResponses, '403')),
+        others.reduce(
+          (previous, current, index) =>
+            pipe(
+              previous,
+              Option.alt(() => {
+                logger.trace(`Unable to find a ${statusCodes[index === 0 ? first : 1 + index]} response definition`);
```
lag-of-death: @XVincentX,
if `statusCodes` equals ['400', '422', '404'], then
for `index === 0`: statusCodes['400'],
for `index === 1`: statusCodes['2']
for `index === 2`: statusCodes['3']

or am I reading this wrong?

XVincentX: That's effectively wrong, I've fixed it.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. [#604](https://github.com/stoplightio/prism/pulls/604)
 - Prism is now taking in consideration the `required` properties for combined schemas (`oneOf, allOf`). This is coming through an update to the Json Schema Faker Library [#623](https://github.com/stoplightio/prism/pulls/623)
 - Prism will never have enough information to return a `403` status code; all these occurences have been now replaced with a `401` status code which is more appropriate [#625](https://github.com/stoplightio/prism/pulls/625)
+- Prism is now negotiating the error response dinamically based on the validation result (security or schema validation) instead of always returning a static order of responses [#628](https://github.com/stoplightio/prism/pulls/628)
```
lag-of-death:
```suggestion
- Prism is now negotiating the error response dynamically based on the validation result (security or schema validation) instead of always returning a static order of responses [#628](https://github.com/stoplightio/prism/pulls/628)
```

test-harness/specs/security-scheme-validation-AC-27.oas2.txt
```diff
@@ -0,0 +1,46 @@
...
+    schema:
+      type: string
+      example:
+        abc
+  InvalidPayload:
+    description: InvalidPayload
+    schema:
+      type: string
+      example:
+        abc
```
philsturgeon: might be worth having different example values for things like this just to make sure the body isnt mismatching somehow.

XVincentX: I did it for oas3, didn't do it for oas2. Fixed.

Issue #473, opened 2019-07-17
Proxy Server


philsturgeon:
**User story.**

As a client developer integrating with an API which has moved from the mock stage to a prototype or staging instance, I can run Prism in proxy mode, send all that API's traffic through this proxy, which will throw big loud errors if there is an issue with my request, or the response I'm getting back.

As a QA tester managing an e2e suite testing flows between a bunch of APIs, I can funnel all traffic through the Prism proxy, to make sure all the requests and responses coming through the proxy line up with what their API descriptions are advertising, in order to catch contract mismatches quicker, and without relying on those API teams to self report issues.

**Is your feature request related to a problem?**

This is a feature that Prism v1 and v2 had in various forms, and we sadly had to take it out of v3 due to time constraints, and slightly unclear specifications. The two use cases above should help shape thew return of this feature in v3.2 or later. 

**Describe the solution you'd like**

When #429 is done, we should have the means to listen for contract mismatches. Using this mechanism we can have the proxy server listen to requests and using the Prism HTTP Client JS interface, it can make those requests to the real server.

The proxy server should have two modes:

```js
prism proxy {upstream} openapi.yaml --log
prism proxy {upstream} openapi.yaml --error
```

Not sure which should be default right now, but the two modes would work like this:

1. Log would just report mismatches to logs stdout/stderr
2. Error would respond with Prism JSON errors explaining the mismatch

the goal of logs would be to have it running in staging or maybe even production (agh this will need lots of load testing), and error would be more useful for local development and testing environments.

**Additional context**

Blocked by #429 

/cc @ww-daniel-mora for input.

philsturgeon:
> a client makes a request to Prism and Prism forwards that request to an upstream

correct.

philsturgeon:
Btw the reason we are not interested in `curl http://some_upstream.com --proxy http://localhost:4010` is because http://localhost:4010 is the proxy yes, and the proxy is started with an upstream, so we don't need to place that extra burden on the client. The client can just request http://localhost:4010 and it's already going to the upstream.

This is still a proxy 👍

lag-of-death:
@philsturgeon, I understand. I was trying to clarify what type of proxy Prism is to be, which you did, so thank you for that. And also, this is the only way I believe to be able to peek into request/responses. With Prism creating a TCP tunnel (for https over http) between a client and an upstream, it would be impossible to do so.

philsturgeon:
[...]

PR #629, opened 2019-09-19
refactor: kill forwarder

XVincentX:
The forwarder concept is something that has never been used and needs to be redesigned. Per @philsturgeon request we're going to remove the whole concept and redesign it from scratch as a real proxy server when we'll hit the good time.

I haven't removed some of the related tests so I can re-enable them once the proxy feature will be back.


XVincentX:
BOOM

lag-of-death:
@XVincentX, I don't think we are to implement a real proxy server - https://github.com/stoplightio/prism/issues/473#issuecomment-533165658 The forwarder might have been useful for that

philsturgeon:
@lag-of-death I'm not sure what a "real" proxy means to you, but when a customer makes a request to prism it will make a request to an upstream server. That's a proxy :)

Some of this code might have been useful, but the name was confusing, and the code was confusing. It was all smushed in with mocking and the options and implementation had to go so we can plan a good solution from scratch. Designing a new thing when you have an old thing around always taints the new design to match the old design, and we don't want that. We want to build this code from scratch, to have no confusion with mocking. :)

lag-of-death:
thanks @philsturgeon, I responded in https://github.com/stoplightio/prism/issues/473#issuecomment-533194802 I'm thankful for clearing up the confusion :+1:

PR #630, opened 2019-09-20
feat: encoding > allowReserved support

### LINKED ISSUE #497: Support 'allowReserved' property when validating form-data body
**User story.**
As a Prism user, I can OAS3 body parameter encoding to allow reserved characters, so that I can handle more real life corner cases.

**Additional context**

Note that this test should already exist in the source code. At the time of writing this issue up it exited on a feature branch under `test-harness/specs/validate-body-params/form-data-allow-reserved-fail.oas3.txt.unsupported`

====test====
Send reserved characters in body and expect to fail validation.
====spec====
openapi: '3.0.1'
paths:
  /path:
    post:
      responses:
        200:
          content:
            text/plain:
              example: ok
      requestBody:
        content:
            schema:
              type: object
              properties:
                reserved:
                  type: string
            encoding:
              reserved:
                contentType: text/plain
                allowReserved: false
                style: form
====server====
mock -p 4010
====command====
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "reserved=:/?#[]@!$&'()*+,;"
====expect====
Connection: keep-alive

{ "to-be": "done" }

StefanDywersant:
Closes #497 

WIP: needs code style refinement

## Checklist

1) support for encoding.allowReserved flag
2) `encoding.style` is only applied for `application/application/x-www-form-urlencoded`
3) `encoding.style` is applied to all uri parts (before it was only applied to first one)
4) i'm using `qs` package to decode UrlEncoded payload. It was used internally by fastify's `form-body` plugin

No


packages/http/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@stoplight/prism-core": "^3.1.1",
+    "@types/qs": "^6.5.3",
```
XVincentX: We usually put the devDependencies (such as types) in the main package.json file of Prism.

Also, I miss `qs` package installed here, which we're going to need.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,8 +1,12 @@
 import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { IMediaTypeContent } from '@stoplight/types';
 import { get } from 'lodash';
+import { parse as deepParseUrlEncoded } from 'qs';
+import * as typeIs from 'type-is';
```
XVincentX: This needs to be installed in the http package as well.

packages/http/src/validator/validators/body.ts
```diff
@@ -19,19 +23,33 @@ export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent>
       return [];
     }
 
-    const encodingStyle = get(content, 'encodings[0].style');
+    if (mediaType && typeIs.is(mediaType, 'application/x-www-form-urlencoded')) {
```
XVincentX: I _think_ we can avoid the first check?
```suggestion
    if (typeIs.is(mediaType, 'application/x-www-form-urlencoded')) {
```

packages/http/src/validator/validators/body.ts
```diff
@@ -19,19 +23,33 @@ export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent>
...
-      if (deserializer && deserializer.supports(encodingStyle) && schema.properties) {
-        const propertySchema = Object.keys(schema.properties)[0];
-        const deserializedObject = deserializer.deserialize(propertySchema, target, Object.values(
-          schema.properties,
-        )[0] as any);
+      const encodings = get(content, 'encodings', []);
+      for (const encoding of encodings) {
+        const allowReserved = get(encoding, 'allowReserved', false);
+        const property = encoding.property;
+        const value = shallowParsedTarget[property];
```
XVincentX: What does `shallow` specifically mean here?

test-harness/specs/validate-body-params/form-data-allow-reserved-fail.oas3.txt
```diff
@@ -0,0 +1,35 @@
...
+                  type: string
+            encoding:
+              reserved:
+                contentType: text/plain
+                allowReserved: false
+                style: form
+====server====
+mock -p 4010
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "reserved=:/?#[]@!$&'()*+,;"
```
XVincentX: Maybe we can send only some of these characters?

test-harness/specs/validate-body-params/form-data-invalid-request.oas2.txt
```diff
@@ -27,7 +27,7 @@ paths:
 ====server====
 mock -p 4010
 ====command====
-curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "status=ooopsie!"
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "status=ooopsie%21"
```
XVincentX: Just to make sure I understand this correctly — since `!` is a reserved character and `allowReserved` is false by default, you have effectively encoded it so that it passes the validation and goes through the following validation step, right?

packages/http/src/validator/validators/body.ts
```diff
@@ -1,8 +1,12 @@
 import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { IMediaTypeContent } from '@stoplight/types';
 import { get } from 'lodash';
+import { parse as deepParseUrlEncoded } from 'qs';
+import * as typeIs from 'type-is';
 import { body } from '../deserializers';
 
+import { DiagnosticSeverity, Dictionary } from '@stoplight/types/dist';
```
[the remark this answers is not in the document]
StefanDywersant: Fixed 👍

packages/http/package.json
StefanDywersant: Fixed 👍

packages/http/src/validator/validators/body.ts
StefanDywersant: I had trouble naming that.

Basically it should mean that URL is split by `&` and the key-val pairs are extracted. But percent-encoded entities hadn't been decoded yet.

We need it in that form because:
1) we need key-val pairs in order to match key names with relevant encoding definitions
2) we need before-percent-decoding values in order to look for forbidden characters

test-harness/specs/validate-body-params/form-data-invalid-request.oas2.txt
StefanDywersant: Exactly. And this is the place where I doubt in OpenApi 3 spec. I mean having `allowReserved=false` by default sounds reasonable. However, I think it will end with lots of people complaining that prism has a bug.

StefanDywersant:
> I've dropped some comments, but the direction is overall ok!
> 
> Can you add a changelog line?

Changelog line added 👍

test-harness/specs/parameters-ac7.oas2.txt
```diff
@@ -26,7 +26,7 @@ paths:
 ====server====
 mock -p 4010
 ====command====
-curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "arr=a,b,c"
+curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data "arr=a%2Cb%2Cc"
```
XVincentX: We can use `--data-urlencode` to let CURL encode the data and keep the test readable

```suggestion
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded" --data-urlencode "arr=a,b,c"
```

test-harness/specs/validate-body-params/form-data-allow-reserved-fail.oas3.txt
XVincentX: I guess it's fine the way it is, ignore this :)

test-harness/specs/validate-body-params/form-data-invalid-request.oas2.txt
XVincentX: It depends, usually the clients handle the encoding automatically and even curl has an option
`--data-urlencode`.

//cc @philsturgeon for awareness but I do not see this as a problem

packages/http/src/validator/validators/body.ts
philsturgeon: Path and queryParams?

philsturgeon:
Looks great feature wise!

test-harness/specs/validate-body-params/form-data-invalid-request.oas2.txt
StefanDywersant: Updated to use `--data-urlencode` 👍

test-harness/specs/parameters-ac7.oas2.txt
StefanDywersant: Done, good catch btw. :)

packages/http/src/validator/validators/body.ts
StefanDywersant: Quite unsure what do you mean by path and queryParams. We are dealing here with body payload.

PR #633, opened 2019-09-23
[WIP] feat: add proxy

### LINKED ISSUE #473: Proxy Server

lag-of-death:
Closes #473 

## Checklist

_You can also link to an open issue here for all the info._

_What changes might users need to make in their application due to this PR?_

### DIFF
--- a/packages/cli/src/commands/proxy.ts
+++ b/packages/cli/src/commands/proxy.ts
@@ -1,0 +1,31 @@
+import { partial } from 'lodash';
+import { CommandModule } from 'yargs';
+import { documentPositional, handler, middleware, onFail } from './shared/helpers';
+import options from './shared/options';
+
+const proxyCommand: CommandModule = {
+  describe: 'Start a proxy server with the given document',
+  command: 'proxy <upstream> <document>',
+  builder: yargs =>
+    yargs
+      .positional('upstream', {
+        description: 'An address for the destination server.',
+        type: 'string',
+      })
+      .positional(...documentPositional)
+      .middleware(middleware)
+      .fail(partial(onFail, yargs))
+      .options({
+        ...options,
+        log: {
+          choices: ['log', 'error'],
+          description: 'The way mismatches are reported. Choose error for Prism error format.',
+          demandOption: false,
+          default: 'log',
+          string: true,
+        },
+      }),
+  handler: args => handler({ upstream: args.upstream, log: args.log }, args),
+};
+
+export default proxyCommand;
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -5,6 +5,7 @@ import * as fastify from 'fastify';
 import * as fastifyCors from 'fastify-cors';
 import * as formbodyParser from 'fastify-formbody';
 import { IncomingMessage, ServerResponse } from 'http';
+import * as httpProxy from 'http-proxy';
 import { defaults } from 'lodash';
 import * as typeIs from 'type-is';
 import { getHttpConfigFromRequest } from './getHttpConfigFromRequest';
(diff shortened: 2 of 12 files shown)


lag-of-death:
@XVincentX, @philsturgeon, I've created this PR so that you can give feedback early :)

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,41 @@
+import * as signale from 'signale';
+import { CommandModule } from 'yargs';
+import { createMultiProcessPrism, CreatePrismOptions, createSingleProcessPrism } from '../util/createServer';
+import getHttpOperations from '../util/getHttpOperations';
+import options from './options';
+
+const proxyCommand: CommandModule = {
+  describe: 'Start a proxy server with the given spec file',
```
philsturgeon: Let's not talk about spec files, are are trying to stick to a unwritten styleguide but we call them "description documents". So "document" is fine.

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,41 @@
...
+        type: 'string',
+      })
+      .middleware(async argv => (argv.operations = await getHttpOperations(argv.spec!)))
+      .fail((msg, err) => {
+        if (msg) yargs.showHelp();
+        else signale.fatal(err.message);
+
+        process.exit(1);
+      })
+      .options(options),
```
philsturgeon: Sharing all options between mock and proxy is a little confusing as we do not want or need --dynamic. Can we have some shared options and some specific to the command?

pytlesk4:
@philsturgeon is proxy only going to proxy requests? It would be cool if we could proxy/mock requests for a particular spec. It was something we did in the first version of Prism, you could turn mocking on for particular operations. Thought was you could start out mocking your API, and slowly over time turn mocking off for operations and start recieving real data from your API. It was possible in Prism V2, but not as easy as v1. Here is how it was done in classic: https://help.stoplight.io/prism/getting-started/swagger

I don't know if that is out of scope for this PR, or if it is a different mode. But just wanted to get it on your radar.

philsturgeon:
Let's leave this PR comments for comments about the PR. We can chat on #473.

packages/cli/src/commands/proxy.ts
XVincentX: Yes we can

packages/http-server/src/server.ts
```diff
@@ -4,21 +4,44 @@ import { IHttpOperation } from '@stoplight/types';
 import * as fastify from 'fastify';
 import * as fastifyCors from 'fastify-cors';
 import * as formbodyParser from 'fastify-formbody';
+const proxy = require('fastify-http-proxy');
```
XVincentX: Use import instead of require

lag-of-death:
alternative approach with `axios`: https://github.com/stoplightio/prism/pull/651

XVincentX:
I'm closing this for now since it does not appear to be a viable alternative.

PR #641, opened 2019-09-23
chore: upgrade http spec and remove hack

XVincentX:


examples/petstore.oas2.yaml
```diff
@@ -1,700 +1,19 @@
 swagger: '2.0'
-info:
```
lag-of-death: @XVincentX, why are we removing this?

test-harness/specs/validate-body-params/no-consumes.defined.oas2.txt
```diff
@@ -1,37 +0,0 @@
-====test====
-When I send a request to an operation that does not have a consumes defined
```
lag-of-death: @XVincentX, does the change in http-spec impact this test?

examples/petstore.oas2.yaml
XVincentX: Argh, I must have done something wrong here. I'll restore it.

test-harness/specs/validate-body-params/no-consumes.defined.oas2.txt
XVincentX: Yeah this is something we need to discuss with @philsturgeon 

basically a spec file that has neither a global or a local consumes is wrong — and this test was testing a behaviour which should not happen.

If you do not define what your API can consume, we can't even try to parse and validate your body.

philsturgeon: This seems like reasonable thinking, but if we want to be friendly we can say “they might have meant JSON” perhaps?

XVincentX: Given this change will affect other parts of Stoplight I've put a comment in the relevant PR, let's continue there @philsturgeon https://github.com/stoplightio/http-spec/pull/25#issuecomment-535035897

PR #645, opened 2019-09-24
fix: killing subprocess only if prism is running in multiprocess mode

### LINKED ISSUE #644: Cannot read property 'kill' of undefined
**Describe the bug**
When prism cannot start due to already used port it fails with additional message: `✖  fatal     Cannot read property 'kill' of undefined`

**To Reproduce**

1. start prism on one console (so it binds to 4010 port)
2. start prism on second console

**Expected behavior**
Single message with error description:
`[CLI] ✖  fatal     listen EADDRINUSE: address already in use 127.0.0.1:4010`

**Actual behavior**
It fails with additional message:
`✖  fatal     Cannot read property 'kill' of undefined`

**Screenshots**
Screen Shot 2019-09-24 at 11 16 29

 - Library version: 3.1.1
 - OS: macOS
 - Browser: curl

StefanDywersant:
Closes #644 

## Checklist

Bug fix

Cli is trying to kill subprocess even if it is not running in multi-process mode. It results in no nice-looking line `✖ fatal Cannot read property 'kill' of undefined`.

This fix moves logging of errors to the place where the context (single/multiprocess) is known. `.kill` will be invoked only in the latter context.

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 # Unreleased
 
-## Added
-
 ## Fixed
 
 - Killing sub-process only if Prism is running in multi-process mode [#645](https://github.com/stoplightio/prism/pull/645)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to Semantic Versioning.
 
+# Unreleased
+
+## Added
+
+## Fixed
+
+- Killing sub-process only if Prism is running in multi-process mode [#645](https://github.com/stoplightio/prism/pull/645)
+
 # 3.1.1 (2019-09-23)
 
 ## Fixed
--- a/packages/cli/src/util/createServer.ts
+++ b/packages/cli/src/util/createServer.ts
@@ -39,7 +39,7 @@ export async function createSingleProcessPrism(options: CreatePrismOptions) {
   pipeOutputToSignale(logStream);
 
   try {
-    return createPrismServerWithLogger(options, logInstance);
+    return await createPrismServerWithLogger(options, logInstance);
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to Semantic Versioning.
 
+# Unreleased
+
+## Added
```
lag-of-death: hm.. we should only have `##Fixed` (?)

lag-of-death:
hm.. just curious if we could do sth like:
```
====server====
mock -p 4010
====command====
mock -p 4010
```
in an e2e test? maybe the change could be tested like this? WDYT @StefanDywersant ?

CHANGELOG.md
XVincentX: It depends on what the next release is going to be.

packages/cli/src/util/createServer.ts
```diff
@@ -22,23 +22,32 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
     }
   } else {
     const logInstance = createLogger('CLI');
-    return createPrismServerWithLogger(options, logInstance);
+    try {
+      return await createPrismServerWithLogger(options, logInstance);
```
XVincentX: In a return statement, await is redundant.

```suggestion
      return createPrismServerWithLogger(options, logInstance);
```

packages/cli/src/util/createServer.ts
```diff
@@ -22,23 +22,32 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
...
+    } catch (e) {
+      logInstance.fatal(e.message);
+      cluster.worker.kill();
+    }
   }
 }
 
-export function createSingleProcessPrism(options: CreatePrismOptions) {
+export async function createSingleProcessPrism(options: CreatePrismOptions) {
```
XVincentX: No async if there is no await 
```suggestion
export function createSingleProcessPrism(options: CreatePrismOptions) {
```

PR #648, opened 2019-09-25
fix: validating output status code

### LINKED ISSUE #647: Output validator is not checking the returned status code
The Prism output validator is not checking out that the returned status code is in line with what specified in the contract/open api file;

We need this for the contract validation story; the returned status code should be within the defined responses for the current operation.

Question for @philsturgeon — if there's a default response, does it mean that *any* status code is ok? Maybe we mark is as a warning?

StefanDywersant:
Closes #647 

## Checklist

Bugfix

Response's status code is not validated against responses found in the spec.

No (we do not have proxy functionality yet, right?)


packages/http/src/validator/utils/spec.ts
```diff
@@ -5,8 +5,22 @@ export function findOperationResponse(
...
-    .filter(spec => new RegExp(`^${spec.code.replace(/X/g, '\\d')}$`).test(String(statusCode)))
-    .sort((s1, s2) => s1.code.split('X').length - s2.code.split('X').length);
+    .filter(
+      spec =>
+        new RegExp(`^${spec.code.replace(/X/g, '\\d')}$`).test(String(statusCode)) ||
+        spec.code.toLowerCase() === 'default',
+    )
+    .sort((s1, s2) => {
+      if (s1.code.toLowerCase() === 'default') {
+        return 1;
```
XVincentX: Question: shouldn't this be -1? Ideally we want default to be the last choice, not the first one.

packages/http/src/validator/utils/spec.ts
```diff
@@ -5,8 +5,22 @@ export function findOperationResponse(
...
+      }
+
+      if (s2.code.toLowerCase() === 'default') {
+        return -1;
```
XVincentX: Same question — shouldn't this be 1 since we want default to be a fallback choice?

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -145,6 +145,8 @@ describe('HttpValidator', () => {
   describe('validateOutput()', () => {
     describe('output is set', () => {
       it('validates the body and headers', () => {
+        (findResponseSpecModule.findOperationResponse as jest.Mock).mockReturnValue({ statusCode: 200 });
```
XVincentX: Note to self: where is this being spyed/mocked?

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -163,5 +165,31 @@ describe('HttpValidator', () => {
...
+              path: '/',
+              id: '1',
+              request: {},
+              responses: [{ code: '200' }],
+            },
+            element: { statusCode: 200 },
+          }),
+        ).toEqual([
+          {
+            message: 'Unable to match returned status code with those defined in spec',
```
XVincentX: Why is this failing? Aren't we providing a 200 response which is indeed a listed response?

packages/http/src/validator/utils/__tests__/spec.spec.ts
```diff
@@ -1,16 +1,60 @@
...
+    });
+  });
+
+  describe('when default response is specified with mixed case', () => {
+    it('returns default response', () => {
+      expect(
+        findOperationResponse(
+          [
+            { code: '2XX', contents: [], headers: [] },
+            { code: 'deFAULT', contents: [], headers: [] },
```
XVincentX: This is interesting; I am not sure this should work. @philsturgeon shall the default response be case insensitive?

packages/http/src/validator/utils/__tests__/spec.spec.ts
```diff
@@ -1,16 +1,60 @@
...
+        findOperationResponse(
+          [
+            { code: '2XX', contents: [], headers: [] },
+            { code: '20X', contents: [], headers: [] },
+            { code: 'default', contents: [], headers: [] },
+            { code: '1XX', contents: [], headers: [] },
+          ],
+          200,
+        ),
+      ).toEqual({ code: '20X', contents: [], headers: [] });
```
XVincentX: I'm assuming this has been specified because it's the most specific, correct?

packages/http/src/validator/utils/spec.ts
StefanDywersant: `1` on s1 moves it down.

Screen Shot 2019-09-25 at 14 23 57

packages/http/src/validator/utils/spec.ts
StefanDywersant: Same here

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## Fixed
 
 - Killing sub-process only if Prism is running in multi-process mode [#645](https://github.com/stoplightio/prism/pull/645)
+- Validating output status code against available response specs [#648](https://github.com/stoplightio/prism/pull/648)
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/http/src/validator/__tests__/index.spec.ts
StefanDywersant: Hah, we are mocking fetching of response here:
```
        (findResponseSpecModule.findOperationResponse as jest.Mock).mockReturnValue(undefined);
```

So thanks to beauty of unit tests it does not matter what's inside the spec. And sth needs to be there because of typing.

However, I agree that it may be misleading. I changed slightly the spec.

packages/http/src/validator/utils/__tests__/spec.spec.ts
StefanDywersant: I blindly assumed that someone can use write "default" using any case.

packages/http/src/validator/utils/__tests__/spec.spec.ts
StefanDywersant: Yes

packages/http/src/validator/__tests__/index.spec.ts
StefanDywersant: On the beginning of the file. This is a pure unit test with the whole environment mocked.

packages/http/src/validator/utils/__tests__/spec.spec.ts
XVincentX: The spec says it's either a number or `default`, named responses are not allowed. 

@philsturgeon will do the call here, but I think we should not test/support this scenario.

StefanDywersant: It is not "named response". It is "default" string being case insensitive vs sensitive.

Just in case some folk will make:
```
...
responses:
  200: ...
  3XX: ...
  DEFAULT:
...
```

XVincentX: Yeah I get it; my point is that if the response is not a number and it's not `default` then it's a named response (as in a regular string).

I still think we should not support these scenarios at all as they're invalid OAS documents and spectral will likely refuse them anyway.

In any case, @philsturgeon (he's bombarded with mentions now :trollface:) will answer the question for us.

packages/http/src/validator/utils/spec.ts
```diff
@@ -5,8 +5,22 @@ export function findOperationResponse(
   statusCode: number,
 ): IHttpOperationResponse | undefined {
   const sortedSpecs = responseSpecs
-    .filter(spec => new RegExp(`^${spec.code.replace(/X/g, '\\d')}$`).test(String(statusCode)))
```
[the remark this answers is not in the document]
StefanDywersant: Just added some Option witchcraft but I have a feeling that it is not enough. Could you take a look and guide me on what I can do more?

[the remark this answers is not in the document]
XVincentX: I think you're on the right track; I'll drop some comments on how to improve this but you're not that far away from what I was looking for.

packages/http/src/validator/index.ts
```diff
@@ -2,6 +2,7 @@ import { IPrismDiagnostic, ValidatorFn } from '@stoplight/prism-core';
 import { DiagnosticSeverity, IHttpOperation } from '@stoplight/types';
 import * as caseless from 'caseless';
 
+import { toUndefined } from 'fp-ts/lib/Option';
```
XVincentX: Ok here's something to keep in mind; in general if you're going to a Monad (`Option, Either, Task, TaskEither`) you should "never" go back to the "bad world". `toUndefined` is a method effectively bringing back in the uncertainty and it should be avoided.

The how depends on the context and the code. Follow this review, and we'll figure out this specific one.

packages/http/src/validator/index.ts
```diff
@@ -37,7 +38,14 @@ const validateInput: ValidatorFn<IHttpOperation, IHttpRequest> = ({ resource, el
 const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource, element }) => {
   const results: IPrismDiagnostic[] = [];
   const mediaType = caseless(element.headers || {}).get('content-type');
-  const responseSpec = resource.responses && findOperationResponse(resource.responses, element.statusCode);
+  const responseSpec = resource.responses && toUndefined(findOperationResponse(resource.responses, element.statusCode));
+
+  if (!responseSpec) {
```
XVincentX: …in this specific case, we have two paths: one when the responseSpec is not defined (we add an element in the array)

packages/http/src/validator/index.ts
```diff
@@ -37,7 +38,14 @@ const validateInput: ValidatorFn<IHttpOperation, IHttpRequest> = ({ resource, el
...
+    results.push({
+      message: 'Unable to match returned status code with those defined in spec',
+      severity: DiagnosticSeverity.Error,
+    });
+  }
 
   return results
```
XVincentX: …and one when the responseSpec is defined; you can see indeed from the checks (`response&&responseSpec.contents`) that this piece of code is not executed if the response is not defined.

For this, we can use `fold`, which will accept two functions: one if Option has a content, one if the content does *not*. 

So we can rewrite this into

```ts
  return pipe(findOperationResponse(resource.responses, element.statusCode),
    fold(() => {
      results.push({
        message: 'Unable to match returned status code with those defined in spec',
        severity: DiagnosticSeverity.Error,
      });
      return results;

    }, responseSpec => {
      const mediaType = caseless(element.headers || {}).get('content-type');

      return results
        .concat(bodyValidator.validate(element.body, (responseSpec && responseSpec.contents) || [], mediaType))
        .concat(headersValidator.validate(element.headers || {}, (responseSpec && responseSpec.headers) || []));
    }))
```

packages/http/src/validator/utils/__tests__/spec.spec.ts
```diff
@@ -1,16 +1,69 @@
...
-          { code: '20X', contents: [], headers: [] },
-          { code: '1XX', contents: [], headers: [] },
-        ],
-        200,
-      ),
-    ).toEqual({ code: '20X', contents: [], headers: [] });
+  describe('when response for given code exists', () => {
+    it('returns found response', () => {
+      expect(
+        toUndefined(
```
XVincentX: You can use the `assertSome` utility function that will verify for you that the `Option` has a content inside and execute the statements you require.

We can rewrite this test to be

```ts
    it('returns found response', () => {
      assertSome(
        findOperationResponse(
          [
            { code: '2XX', contents: [], headers: [] },
            { code: '20X', contents: [], headers: [] },
            { code: 'default', contents: [], headers: [] },
            { code: '1XX', contents: [], headers: [] },
          ],
          200,
        ), value => expect(value).toEqual({ code: '20X', contents: [], headers: [] })
      )
    });
```

In case you're wondering how annoying it is, I agree with you. However I'm in touch with the jest maintainers and they'll release the ability to introspect wrapped values soon making test this stuff a breeze: https://github.com/facebook/jest/issues/8540

packages/http/src/validator/utils/spec.ts
```diff
@@ -1,12 +1,27 @@
...
 
-  return sortedSpecs[0];
+      if (s2.code.toLowerCase() === 'default') {
+        return -1;
+      }
+
+      return s1.code.split('X').length - s2.code.split('X').length;
+    });
+
+  return fromNullable(sortedSpecs[0]);
```
XVincentX: This part is correct! However fp-ts provides some utilities that will help you avoid using `fromNullable` and other stuff.

In particular you could do:

```ts
  return pipe(sortedSpecs, head);
```

but your solution works perfectly anyway.

packages/http/src/validator/index.ts
StefanDywersant: Two things here:
1) original code did not returned immediately when there was no responseSpec
1.1) now I think it should return, so the example provided by you is fine
2) theoretically if I want to preserve original flow, how should I model the situation where depending on existence of responseSpec we are pushing a diagnostic to array or not?

XVincentX: @StefanDywersant 

In case you want to execute an action *only* when the `Option` is populated, you can either

1. Use a fold, with an empty handler for the "none" part.

`pipe(option, fold(()=>{}, value=> array.push(value))`

2. You can also use a map, which will only go through the "happy" part.

`pipe(option, map(value=>array.push(value))`

packages/http/src/validator/index.ts
philsturgeon: Can we avoid the word spec? It’s a description doc

packages/http/src/validator/utils/__tests__/spec.spec.ts
philsturgeon: It’s difficult to maintain a difficult balance between “only supporting what the spec supports” and “supporting what people do” but yeah the options here are “it’s an integer as a string because YAML” or its “default“ but we can ignore everything else. Let’s not waste our time on conventions like 3xx == 300 or anything else.

XVincentX: Noted. @StefanDywersant can you make sure we only look for `default` and not for any other of its combinations?

packages/http/src/validator/utils/spec.ts
StefanDywersant: Changed, it looks cooler with that pipe/head approach :)

packages/http/src/validator/utils/__tests__/spec.spec.ts
StefanDywersant: Done  👍

packages/http/src/validator/index.ts
StefanDywersant: Pushing to array creates a side-effect of invoking function, so I guess is not the FP-way.

packages/http/src/validator/index.ts
StefanDywersant: Changed to `responseDescDoc`.

@philsturgeon Why is `description doc` better than `spec`?

XVincentX: I would say consistency; if we're using description Document in Stoplight generally, it makes sense to stick to the same convention everywhere; the spec — I think — refers to the specification (oas2, oas3)

StefanDywersant: I the poll is still open my vote goes on `specification`. The world calls it `specification` and I don't think it is good to invent new names. Especially longer ones ;)

packages/http/src/validator/index.ts
XVincentX: Generally speaking, you're right. The idea would be that map would return a new array.

However, it really depends; there are a lot of functional libraries that implement stuff in a non functional way, but as long it is isolated, it's fine anyway.

An example would be

```ts
function doSomething(a: number): Array<number> {
   const array: Array<number> = [];
   array.push(a);
   return a;
}
```

You can see this function is creating and pushing stuff in the array which is effectively non functional; however if you look at the code at the whole it's still ok as the creation and the destruction of the "non functional object" is confined in that function.

The code we're looking at fortunately for us is in this case. 

Does that make sense, clarify?

StefanDywersant: The functional world is built on the foundation of an imperative one ;)
I just hoped that there are some helper functions provided by FP to solve it.

Thanks for the clarification!

XVincentX: We can still refactor it — you're more than welcomed to refactor it if you want; I try to do this opportunistically when I see opportunities!

StefanDywersant: Sure, just need to learn as much as I can about that FP. My goal is to leave every touched code as functional :)

packages/http/src/validator/utils/__tests__/spec.spec.ts
StefanDywersant: Done 👍

PR #649, opened 2019-09-25
fix: proper form data decoding

StefanDywersant:
## Checklist

Bugfix

Currently, body encoded with form-data is improperly decoded. We are using `qs` package which is not capable of decoding all styles, especially those described in encoding > style.

The new behavior is that we support all encoding styles through our deserializers.

No.


packages/http/src/validator/validators/__tests__/body.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { HttpParamStyles } from '@stoplight/types/dist';
```
lag-of-death:
```suggestion
import { HttpParamStyles } from '@stoplight/types';
```

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,138 @@
...
-    if (!schema) {
+    try {
+      return pipe(
+        schema ? Either.right(target) : Either.left([]),
+        Either.chain(partial(maybeValidateFormBody, schema!, content, mediaType)),
+        Either.chain(partial(validateBody, schema!)),
+        Either.fold(partial(applyPrefix, this.prefix), () => []),
+      );
+    } catch (e) {
+      require('fs').appendFileSync('myk.txt', JSON.stringify(e.stack));
```
lag-of-death: what does this do?

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,138 @@
...
-        if (!allowReserved && typeof value === 'string' && value.match(/[\/?#\[\]@!$&'()*+,;=]/)) {
-          return [
-            {
-              path: [prefix, property],
-              message: 'Reserved characters used in request body',
-              severity: DiagnosticSeverity.Error,
-            },
-          ];
-        }
+function validateBody(schema: JSONSchema, target: any) {
```
lag-of-death: can `target` be something more concrete?

packages/http/src/validator/validators/body.ts
StefanDywersant: Shush! That is nothing ;)

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,138 @@
...
-    }
+function validateAgainstReservedCharacters(encodedUriParams: Dictionary<string, string>, encodings: IHttpEncoding[]) {
+  return pipe(
+    encodings,
+    Array.reduce<IHttpEncoding, IPrismDiagnostic[]>([], (diagnostics, encoding) => {
+      const allowReserved = get(encoding, 'allowReserved', false);
+      const property = encoding.property;
+      const value = encodedUriParams[property];
+
+      if (!allowReserved && typeof value === 'string' && value.match(/[\/?#\[\]@!$&'()*+,;=]/)) {
```
lag-of-death:
```typescript
if (!allowReserved && typeof value === 'string' && value.match(/[\/?#\[\]@!$&'()*+,;=]/)) {
  return diagnostics.concat({
          path: [property],
          message: 'Reserved characters used in request body',
          severity: DiagnosticSeverity.Error,
        });
} else {
  return diagnostics;
}
```
`.push` does not play well in FP

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,138 @@
...
+  encodings: IHttpEncoding[],
+  decodedUriParams: Dictionary<string, string>,
+) {
+  if (!schema.properties) {
+    return decodedUriParams;
   }
+
+  return pipe(
+    Object.keys(schema.properties),
+    Array.reduce({}, (deserialized, property) => {
```
lag-of-death: can this reduce be mutation-free?

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,138 @@
...
     const [key, ...rest] = pair.split('=');
     result[key] = rest.join('=');
     return result;
   }, {});
 }
+
+function decodeUriEntities(target: Dictionary<string, string>) {
+  return Object.entries(target).reduce((result, [k, v]) => {
+    result[decodeURIComponent(k)] = decodeURIComponent(v);
+    return result;
```
lag-of-death:
```
return { [[decodeURIComponent(k)]: decodeURIComponent(v), ...result }
```

packages/http/src/validator/validators/__tests__/body.spec.ts
StefanDywersant: Done 👍

packages/http/src/validator/validators/body.ts
StefanDywersant: That discussion may take lot of time, but let me explain my motivation.

Basically I consider this as a conflict between:
mutation-free vs. speed
or
perfectionism vs. rationality

Like in this example, the poor js engine will need to create a brand new object, assign all properties (we have no control over how many of those are there!), put another property and repeat those ops property-count times. 

I choose speed/rationality in small reducers over mutation-free/perfectionism. I think the reducer is small, well defined and mutating the value won't make any trouble.

packages/http/src/validator/validators/body.ts
StefanDywersant: Same as here: https://github.com/stoplightio/prism/pull/649#discussion_r329123342

packages/http/src/validator/validators/__tests__/body.spec.ts
lag-of-death:
```suggestion
import { HttpParamStyles } from '@stoplight/types';
```

packages/http/src/validator/validators/body.ts
lag-of-death: okey-dokey, but let's add least do this https://github.com/stoplightio/prism/pull/649#discussion_r329107225 ;)

packages/http/src/validator/validators/body.ts
lag-of-death: get ya

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
-    }
+function validateBody(schema: JSONSchema, target: any) {
+  const diagnostics = validateAgainstSchema(target, schema);
+  return diagnostics.length ? Either.left(diagnostics) : Either.right(target);
+}
+
+function maybeValidateFormBody(
+  schema: JSONSchema,
+  content: IMediaTypeContent,
+  mediaType: string | undefined,
```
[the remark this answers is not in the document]
XVincentX: …although it's not mandatory, we can make it happen in another step.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+  target: any,
+) {
```
XVincentX: In general I really suggest to annotate the return type explicitly. This makes the debugger way easier sometimes.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+function deserializeFormBody(
+  schema: JSONSchema,
+  encodings: IHttpEncoding[],
+  decodedUriParams: Dictionary<string, string>,
+) {
+  if (!schema.properties) {
+    return decodedUriParams;
   }
+
+  return pipe(
```
XVincentX: Is `pipe` really required here?

packages/http/src/validator/validators/body.ts
XVincentX: In general I am ok with such tradeoffs; messes will have to happen somewhere anyway but as long you can keep the stuff contained in one place where is not harming anybody but still a functional interface, I am ok with that.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+          deserialized[property] = deserializer.deserialize(property, decodedUriParams, propertySchema as JSONSchema);
+        }
+      }
+
+      return deserialized;
+    }),
+  );
+}
+
+function getContent(specs: IMediaTypeContent[], mediaType?: string) {
```
XVincentX: If so, I think we need to use a better name, such as `getContentByMediaTypeOrFirst` or even better, do not do `getOrElse` and return the `Option` as it is; let the caller decide the fallback mechanism

XVincentX: I also think that `mediaType` should not be nullable; either you find a way to provide it or, you know, find another way.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
-      return specs[0];
-    }
+  return Either.right(target);
+}
+
+function applyPrefix(prefix: string, diagnostics: IPrismDiagnostic[]): IPrismDiagnostic[] {
+  return diagnostics.map(d => ({ ...d, path: [prefix, ...(d.path || [])] }));
+}
+
+function validateAgainstReservedCharacters(encodedUriParams: Dictionary<string, string>, encodings: IHttpEncoding[]) {
```
XVincentX: I think I wrote this somewhere else, but for monads I suggest to explicitly put the return type, it's going to help TypeScript a lot when doing type checking. 
```suggestion
function validateAgainstReservedCharacters(encodedUriParams: Dictionary<string, string>, encodings: IHttpEncoding[]): Either.Either<IPrismDiagnostic[], Dictionary<string, string>> {
```

XVincentX:
Now that I think about it, would it make sense to extract this logic into a different/external library? I am quite sure this could be used somewhere else in Stoplight @philsturgeon @marbemac

packages/http/src/validator/validators/__tests__/body.spec.ts
StefanDywersant: Done

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
-          return [
-            {
-              path: [prefix, property],
-              message: 'Reserved characters used in request body',
-              severity: DiagnosticSeverity.Error,
-            },
-          ];
-        }
+    return pipe(
+      schema ? Either.right(target) : Either.left([]),
```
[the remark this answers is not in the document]
StefanDywersant: Actually the equivalent of
```
schema ? Either.right(target) : Either.left([])
```
is
```
      Either.fromNullable([])(schema),
      Either.map(() => target),
```

I did choose the following approach:
1) left: `IPrismValidation[]` travels here
2) if 1) occurs, no more right processing is needed (early return)
3) right: `target` travels in that box
4) target is mapped by maybeValidateFormBody (deserialized according to encoding)

Basically it was hard to transform code with such amount of conditionals and exit point to FP.
The thing I'm most concerned about is whether `left` is the right place for validation errors. These are expected outcomes from validations. I wonder whether the same rules apply to `left` as to exceptions. Exceptions should be always unexpected and never used for "control". On the other hand if the above is more relaxed in fp-ts's Either, then we are good.

Do you have an opinion on that?

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+      Either.chain(partial(maybeValidateFormBody, schema!, content, mediaType)),
```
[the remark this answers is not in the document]
StefanDywersant: `right` is already occupied by `target`. Now I start to think that not only `target` should travel through `right`. Maybe it should be a touple of { schema, target } (both required).
What do you guys think?

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+      Either.chain(partial(validateBody, schema!)),
```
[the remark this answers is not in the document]
StefanDywersant: The actual method has four parameters:
```
function maybeValidateFormBody(
  schema: JSONSchema,
  content: IMediaTypeContent,
  mediaType: string | undefined,
  target: any,
)
```

Only three of them binded. The last one, `target` is injected by `chain`. I used partial just to omit this=undefined. It just looks nicer, nothing more.

StefanDywersant:
> Now that I think about it, would it make sense to extract this logic into a different/external library? I am quite sure this could be used somewhere else in Stoplight @philsturgeon @marbemac

I think about two library candidates: request-validation, request-deserialize.
Former checks req agains spec, latter deals with all those crazy serialization styles.

At least request-deserialization is a unique code. At the time of writing (1yr ago), there was no similar tool.

packages/http/src/validator/validators/body.ts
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,131 @@
...
+  if (mediaType && typeIs.is(mediaType, 'application/x-www-form-urlencoded')) {
+    const encodings = get(content, 'encodings', []);
+    const encodedUriParams = splitUriParams(target);
 
-    return validateAgainstSchema(target, schema).map(error =>
-      Object.assign({}, error, { path: [prefix, ...(error.path || [])] }),
+    return pipe(
+      validateAgainstReservedCharacters(encodedUriParams, encodings),
+      Either.map(decodeUriEntities),
+      Either.map(deserializeFormBody.bind(undefined, schema, encodings)),
```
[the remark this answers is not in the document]
StefanDywersant: I've put lodash's `partial` here, it's compliant with the code above. Unless we resign from using `partial` at all.

packages/http/src/validator/validators/body.ts
StefanDywersant: Done 👍

packages/http/src/validator/validators/body.ts
StefanDywersant: Done 👍

packages/http/src/validator/validators/body.ts
StefanDywersant: I could achieve the same using old good `Object.keys().reduce()`. I'm unsure which to pick TBH. Personally I'm always fan of producing short code. On the other hand we are trying to be as FP as possible, so that's why I've chosen pipe. Any vote will tip the balance on any approach.

packages/http/src/validator/validators/body.ts
StefanDywersant: I changed function name as you suggested - it's way better now.

The method will always return sth, so it is not worth returning Option here. `mediaType` is nullable at entry point level (`validate` method). And it is ok since we can deal with such a situation (by choosing first available).

packages/http/src/validator/validators/body.ts
StefanDywersant: At this point, it can be `string|object`. I guess. I'm quite sure that at some point we should decide how `body` is transported around prism infrastructure. Unfortunately, it means some refactor which is out of this issue's scope.

packages/http/src/validator/validators/body.ts
StefanDywersant: Yeah, but this is simple `Array.reduce`. IMO is exactly the same case as in those two other places (mutable vs immutable). I vote on leaving it as is.

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,141 @@
...
+}
+
+function maybeValidateFormBody(
+  schema: JSONSchema,
+  content: IMediaTypeContent,
+  mediaType: Option.Option<string>,
+  target: any,
+): Either.Either<IPrismDiagnostic[], any> {
+  if (
+    Option.getEq(fromEquals((a: string, b: string) => typeIs.is(a, b) as boolean)).equals(
```
lag-of-death:
```suggestion
    Option.getEq(fromEquals((a: string, b: string) => !!typeIs.is(a, b))).equals(
```

packages/http/src/validator/validators/body.ts
lag-of-death: If the choice about `reduce()` vs `pipe()`, then none of these is more FP than the other

packages/http/src/validator/validators/body.ts
lag-of-death: if `we are trying to be as FP as possible` (from https://github.com/stoplightio/prism/pull/649/files#r329567219) is to be satisfied, then we go with `.concat`, if not, then `.push` it is

philsturgeon:
> Now that I think about it, would it make sense to extract this logic into a different/external library? I am quite sure this could be used somewhere else in Stoplight @philsturgeon @marbemac

I think we could hold off for a while on doing this. We have talked a bit recently about how the validation logic in general needs to be refactored, about how middlewares could be useful, and about how it could become its own Lerna package. 

I think all of these things could be done together in the future while we refactor things, and maybe it could become its own package outside of the repo. Either way we can kick that can down the road whilst we sort out Prism more IMO.

packages/http/src/validator/validators/__tests__/body.spec.ts
```diff
@@ -45,5 +46,57 @@ describe('HttpBodyValidator', () => {
...
+                  },
+                  required: ['key'],
+                },
+              },
+            ],
+            'application/x-www-form-urlencoded',
+          ),
+        ).toEqual([]);
+      });
+    });
```
XVincentX: Maybe we could 
1. add a failing test
2. Break down this super test into a set of sub tests? (nested, array…)

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,141 @@
...
-              path: [prefix, property],
-              message: 'Reserved characters used in request body',
-              severity: DiagnosticSeverity.Error,
-            },
-          ];
-        }
+    return pipe(
+      Either.fromNullable([])(schema),
+      Either.map(() => target),
+      Either.chain(partial(maybeValidateFormBody, schema!, content, Option.fromNullable(mediaType))),
```
XVincentX: The reason why we can't really use `partial` here is that we need to use the `schema` that's returned from the `Either.fromNullable` computation, not the one coming from the function parameters (which can be null and you're effectively using the `!` to make it appear ok)

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,141 @@
...
-            {
-              path: [prefix, property],
-              message: 'Reserved characters used in request body',
-              severity: DiagnosticSeverity.Error,
-            },
-          ];
-        }
+    return pipe(
+      Either.fromNullable([])(schema),
+      Either.map(() => target),
```
XVincentX: `target` is already available from the function parameter and does not need any change or "fp-isation", so we can just use that directly.

packages/http/src/validator/validators/body.ts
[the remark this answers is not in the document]
XVincentX: As I mentioned in the previous comment, the problem is that we need to use the `schema` parameter coming from the chain, not from the function's parameters, where the `!` is required.

packages/http/src/validator/validators/body.ts
```diff
@@ -66,8 +66,8 @@ function findContentByMediaTypeOrFirst(specs: IMediaTypeContent[], mediaType: st
 function validateBodyIfNotFormEncoded(mediaType: string, schema: JSONSchema, target: unknown) {
   return pipe(
     mediaType,
-    Option.fromPredicate(mt => !!!typeIs.is(mt, ['application/x-www-form-urlencoded'])),
-    Option.chain(() => Option.some(validateBody(schema, target))),
+    Option.fromPredicate(mt => !typeIs.is(mt, ['application/x-www-form-urlencoded'])),
+    Option.map(() => validateBody(schema, target)),
```
XVincentX: Oh very nice catch, I didn't realise it!

packages/http/src/validator/validators/body.ts
```diff
@@ -1,82 +1,147 @@
...
+          Option.map(diagnostics => applyPrefix(this.prefix, diagnostics)),
+        ),
+      ),
+      Option.getOrElse<IPrismDiagnostic[]>(() => []),
+    );
+  }
+}
+
+function validateBody(schema: JSONSchema, target: any): IPrismDiagnostic[] {
+  return validateAgainstSchema(target, schema);
```
XVincentX: We can probably remove this redundant thing.

packages/http/src/validator/validators/__tests__/body.spec.ts
StefanDywersant: 1. added failing test
2. I simplified the test to only ensure that deserializer works properly

StefanDywersant:
> Thanks a lot for this, I hope it was not that painful to make it happen. I'm merging when tests will pass.

I learned lots while working on this + thank you for guiding while doing so!

PR #650, opened 2019-09-26
feat: contract testing: add missing output checks

### LINKED ISSUE #429: Contract Validation
**User Story**

As a user of the Prism HTTP Client (#511) I have been building an entire end-to-end testing system  which uses Prism as a HTTP client, and I would love to enable contract testing against a provided API description document. 

**Details**

Contract violations for Prism would be a case of getting a HTTP response, and checking several things:

_Requests_

1. Are provided parameters (headers, query, path, etc) valid? Error
1. Is the response mime type defined in the document? Error
1. Is the body valid to the contents of `schema`

_Responses_

1. Is response status code specified in the document? This could be a warning if 300-599, but an error if 200-299.
1. Is the response mime type defined in the document? This could be an error
1. Is the body valid to the contents of `schema`

All of these things could trigger a `DiagnosticSeverity`, just like request validation, because this is exactly the same concept as request validation. Some of the logic from validation could be refactored into a new `contract-violations` folder, and we just make some "request validation" code which turns contract violations into HTTP problems.

**Why?**

Various aspects of our system care about knowing if the client matches against a real OpenAPI description document: 

- Prism HTTP Server
- Prism CLI
- Request Maker (Violation/Warning Component)
- Future Governance / Compliance Reports

All of this functionality will originate in Prism as a HTTP client. Contract violations can bubbled up to logs in Prism CLI, and request maker can store Violations in a "Store" for the Violations component to read from.

**Acceptance Criteria**
- Unit testing of new functionality at 85%
- Prism HTTP Client reports violations for real HTTP requests and responses when description doc is provided
- Prism HTTP Client reports violations for mock server requests
- Prism CLI logs warnings for any contract violation (and tested in test harness)
- Prism HTTP mock server continues to convert violations to HTTP Problems
- ~Integration testing showing warnings are being logged~

**Additional Context**

Blocked by #511

XVincentX:
This PR is implementing two checks missing in the output validation that are required according to #429 

1. If the media type of the response does not match the media type in the document, we add an error
2. If the returned status code is *not* the one defined in the document and it's between 200 and 299, then it's an error, otherwise a warning.

I've added tests and reorganised these since I think we were a bit too aggressive with the mocking (and not cleaning them up)

Closes #429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Support for encoding > allowReserved flag when validating application/x-www-form-urlencoded body [#630](https://github.com/stoplightio/prism/pull/630)
 - Validating output status code against available response specs [#648](https://github.com/stoplightio/prism/pull/648)
+- Support for Contract Testing [#650](https://github.com/stoplightio/prism/pull/650)
 
 ## Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Fixed
 
-- Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. [#604](https://github.com/stoplightio/prism/pulls/604)
-- Prism is now taking in consideration the `required` properties for combined schemas (`oneOf, allOf`). This is coming through an update to the Json Schema Faker Library [#623](https://github.com/stoplightio/prism/pulls/623)
-- Prism will never have enough information to return a `403` status code; all these occurences have been now replaced with a `401` status code which is more appropriate [#625](https://github.com/stoplightio/prism/pulls/625)
-- Prism is now negotiating the error response dynamically based on the validation result (security or schema validation) instead of always returning a static order of responses [#628](https://github.com/stoplightio/prism/pulls/628)
+- Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. [#604](https://github.com/stoplightio/prism/pull/604)
+- Prism is now taking in consideration the `required` properties for combined schemas (`oneOf, allOf`). This is coming through an update to the Json Schema Faker Library [#623](https://github.com/stoplightio/prism/pull/623)
+- Prism will never have enough information to return a `403` status code; all these occurences have been now replaced with a `401` status code which is more appropriate [#625](https://github.com/stoplightio/prism/pull/625)
+- Prism is now negotiating the error response dynamically based on the validation result (security or schema validation) instead of always returning a static order of responses [#628](https://github.com/stoplightio/prism/pull/628)
 - Prism is now selecting proper serializer when Accept header contains content type which is missing in spec. This is a result of simplifying serializer selection approach. [#620](https://github.com/stoplightio/prism/pull/620)
 - HEAD requests no longer fail with 406 Not Acceptable [#603](https://github.com/stoplightio/prism/pull/603)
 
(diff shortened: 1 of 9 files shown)


XVincentX:
The CLI does not have any knowledge of what's going on in the server in general, it just receives messages and forwards them to the standard output. I'm investigating some solutions, I'll get back soon on this.

philsturgeon:
Gotcha. Yeah the suggestions I've been making about moving the HTTP Server logic to a HTTP Middleware, validation logic to another Middleware, etc would mean that the server is just spinning up a server, registering the mocking and validation middlewares, and then adding an extra logging middleware. I think trying to move towards middleware-all-the-things will make a lot of this a lot cleaner, but I understand it cannot be done overnight.

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
+# Prism Client
+
+Prism inclues an Axios style HTTP Client that you can use to seamlessy perform requests to both a real server and a mocked document.
```
philsturgeon:
```suggestion
Prism includes an full featured HTTP Client that you can use to seamlessly perform requests to both a real server and a mocked document. The client is modeled after Axios so it may feel familiar.
```

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+
+```ts
+const client = await createClientFromResource('examples/petstore.oas2.yaml', {
+  mock: true,
+  validateRequest: true,
+  validateResponse: true,
+});
+```
+
+### Create from OAS document in string format
```
philsturgeon:
```suggestion
### Create from OpenAPI string
```

Remember that we don't really talk about OAS for SEO reasons, and human readability. OpenAPI is a known brand, so whilst you might use oas for internal variables we generally want to say OpenAPI on anything user facing.

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+
+```ts
+const oasFile = `
```
philsturgeon:
```suggestion
const descriptionDoc = `
```

docs/guides/client.md
XVincentX: I thought it was a good idea to say we inspired to Axios?

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+`;
+
+const client = await createClientFromString(oasFile, {
+  mock: true,
+  validateRequest: true,
+  validateResponse: true,
+});
+```
+
+### Create from manual Http Operations (you might want _NOT_ use this)
```
philsturgeon: Keep titles "Short, Sweet, And Using Title Case"

docs/guides/client.md
XVincentX: Ignore me, your suggestion includes that.

packages/http/README.md
```diff
@@ -10,6 +10,10 @@ In essence it's an HTTP client (similar to axios - in fact, we use axios to make
 
 The goal of this document is to provide you with some basic code examples to get you started and to cover some of the advanced scenarios.
 
+## Important
+
+If you're a regular user and not a PRO, you might to want to use the [User facing client](../../docs/guides/client.md) which
```
philsturgeon: what is the difference between regular and pro? also message is chopped

XVincentX: The "old", or be precise, the internal client, offers a single method and the idea is that this will have breaking changes over time, that's why we want to suggest *NOT* to use it.

Request Maker should migrated from that.

philsturgeon: The internal client is just that, internal. It's not documented and maybe going away or changing drastically over time. The HTTP Client is what people should be using, so let's just talk about the HTTP Client?

Users are either:

- Using Prism CLI
- Using Prism HTTP Client
- Using Prism HTTP Server? 

We didn't document how to use the HTTP server yet. We've only ever talked about CLI so far as I know.

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+openapi: 3.0.2
+paths:
+  /hello:
+    get:
+      responses:
+        200:
+          description: hello
+`;
+
+const client = await createClientFromString(oasFile, {
```
XVincentX:
```suggestion
const client = await createClientFromString(descriptionDoc, {
```

docs/guides/client.md
XVincentX:
```suggestion
### Create From Manual Http Operations
```

philsturgeon:
Great job! I can’t wait to play around with a beta of this with proxy command. It’s gonna be awesome!

packages/http/README.md
XVincentX: So you're fundamentally suggesting to remove this page, right?

packages/http-server/src/server.ts
```diff
@@ -108,6 +108,13 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
             reply.headers(output.headers);
           }
 
+          response.validations.output.forEach(
+            validation =>
+              validation.severity === DiagnosticSeverity.Error
+                ? request.log.error(validation.message)
+                : request.log.warn(validation.message),
```
StefanDywersant: You should add handling for other severities (Informational, Hint). I checked out of curiosity whether we use them currently and we don't, but this will create minor bug if we start.

CHANGELOG.md
```diff
@@ -19,42 +20,42 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Fixed
 
-- Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. [#604](https://github.com/stoplightio/prism/pulls/604)
-- Prism is now taking in consideration the `required` properties for combined schemas (`oneOf, allOf`). This is coming through an update to the Json Schema Faker Library [#623](https://github.com/stoplightio/prism/pulls/623)
-- Prism will never have enough information to return a `403` status code; all these occurences have been now replaced with a `401` status code which is more appropriate [#625](https://github.com/stoplightio/prism/pulls/625)
-- Prism is now negotiating the error response dynamically based on the validation result (security or schema validation) instead of always returning a static order of responses [#628](https://github.com/stoplightio/prism/pulls/628)
+- Prism is now giving precedence to `application/json` instead of using it as a "fallback" serializer, fixing some conditions where it wouldn't get triggered correctly. [#604](https://github.com/stoplightio/prism/pull/604)
```
lag-of-death: strange, these doesn't seem to belong to this PR (?)

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
+# Prism Client
+
+Prism includes an full featured HTTP Client that you can use to seamlessly perform requests to both a real server and a mocked document. The client is modeled after Axios so it may feel familiar.
```
lag-of-death: `a full
```suggestion
Prism includes a fully-featured HTTP Client that you can use to seamlessly perform requests to both a real server and a mocked document. The client is modeled after Axios so it may feel familiar.
```

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Create from filename or http resource
```
lag-of-death:
```suggestion
### Create from a filename or http resource
```

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -71,83 +70,89 @@ describe('HttpValidator', () => {
...
+                path: '/',
+                id: '1',
+                request: {},
+                responses: [{ code: '200' }],
+              },
+              resourceExtension,
+            ),
+            element: { method: 'get', url: { path: '/' } },
+          }),
+        ).toHaveLength(length);
```
lag-of-death: hm.. is this enough to say that `headers validation is enabled`? Yeah, I know it was already like this, but maybe we could save the snapshot or do something like `toBe`?

packages/http/src/validator/index.ts
```diff
@@ -41,19 +42,39 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+        const mismatchingMediaTypeError = pipe(
+          Option.fromNullable(operationResponse.contents),
+          Option.map(contents =>
+            pipe(
+              contents,
+              findFirst(c => c.mediaType === mediaType),
+              Option.map<IMediaTypeContent, IPrismDiagnostic[]>(() => []),
+              Option.getOrElse<IPrismDiagnostic[]>(() => [
+                {
+                  message: `The received media type ${mediaType} does not match the one specified in the document`,
```
StefanDywersant: `mediaType` might be undefined which leads to "The received media type **undefined** does not match the one specified in the document". I think undefined mediaType deserves better handling which will improve user experience.

packages/http/src/validator/index.ts
```diff
@@ -41,19 +42,39 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
-        return [
-          {
-            message: 'Unable to match returned status code with those defined in spec',
-            severity: DiagnosticSeverity.Error,
-          },
-        ];
+    Option.fold<IHttpOperationResponse, IPrismDiagnostic[]>(
+      () => [
+        {
+          message: 'Unable to match returned status code with those defined in spec',
```
lag-of-death:
```suggestion
          message: 'Unable to match the returned status code with those defined in spec',
```

packages/http/src/validator/index.ts
```diff
@@ -41,19 +42,39 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+          severity:
+            element.statusCode >= 200 && element.statusCode <= 299
```
lag-of-death:
```suggestion
            _.inRange(element.statusCode, 200, 300)
```

XVincentX: I like it!

packages/http/src/validator/index.ts
lag-of-death: that's sadly true https://github.com/stoplightio/prism/pull/650/files#diff-f26acda03be74b077a817c80a629f514R37

XVincentX: Whoops, very good catch. I'll check it out and see what we can do here.

XVincentX: I've removed the variable printing; I guess we will understand if the error message is good enough or not when we'll have the validation in place with the proxy.

packages/http/src/validator/__tests__/index.spec.ts
XVincentX: Good point, effectively this test is a little bit flaky; however the validation is one of the pieces we'll refactor soon, so I am not going to bother so much at this point.

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+
+Once you've got a client instance:
+
+1. You can perform the request using the generic method:
+
+```ts
+client.request('https://google.it', { method: 'get' }).then(response => console.log(response));
+```
+
+The response object has all the informations you need, including the used configuration object.
```
lag-of-death:
```suggestion
The response object has all the information you need, including the used configuration object.
```

docs/guides/client.md
```diff
@@ -0,0 +1,83 @@
...
+
+2. You can override the configuration object on the request level if you prefer
+
+```ts
+client
+  .request('https://google.it', { method: 'get' }, { validateResponse: false })
+  .then(response => console.log(response));
+```
+
+This disables the validation response _only for the current request_
```
lag-of-death:
```suggestion
This disables response validation _only for the current request_
```

packages/http-server/src/server.ts
```diff
@@ -108,6 +108,12 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
             reply.headers(output.headers);
           }
 
+          response.validations.output.forEach(validation => {
+            if (validation.severity === DiagnosticSeverity.Error) request.log.error(validation.message);
```
lag-of-death: let's make `{}` great again ;)

packages/http/README.md
lag-of-death: @XVincentX, can we finish the sentence? It suddenly ends with `which`

packages/http-server/src/server.ts
```diff
@@ -108,6 +108,16 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
             reply.headers(output.headers);
           }
 
+          response.validations.output.forEach(validation => {
+            if (validation.severity === DiagnosticSeverity.Error) {
```
lag-of-death: so beautiful :+1:

StefanDywersant: `switch` would be beautifuler :D

packages/http/src/validator/index.ts
```diff
@@ -41,19 +43,36 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+        const mismatchingMediaTypeError = pipe(
+          Option.fromNullable(operationResponse.contents),
+          Option.map(contents =>
+            pipe(
+              contents,
+              findFirst(c => c.mediaType === mediaType),
+              Option.map<IMediaTypeContent, IPrismDiagnostic[]>(() => []),
+              Option.getOrElse<IPrismDiagnostic[]>(() => [
+                {
+                  message: `The received media type does not match the one specified in the document`,
```
StefanDywersant: @XVincentX you are misbehaving ;)

PR #664, opened 2019-10-01
refactor: simplify validation

XVincentX:
Guess what, I was on a plane again with no internet :trollface: 

Seriously though, this PR introduces some simplifications on the validation and factory side that are going to be needed to break down a little bit more the things that Prism does so that we do not have to use a single giant mega method `prism.process` anymore, but rather compose the way Prism does the things.

In this PR:

* Code climate is gone, I'm just tired of it; it has 0 value and and only costs (spam, emails, fake failures).
* The `route` method is not taking the configuration object anymore, since it was never using it
* The `mock` method is not taking the whole configuration object (with details it shouldn't look into, but only mock specific config (dynamic, static, and other stuff)
* The validation object for the `input` property is no more nested, so we can now do

`input.validations` instead of `input.validations.input` (see https://github.com/stoplightio/prism/pull/664/files#diff-31e61963df18017e4fb3c71929c2e37bL58) which was effectively a little bit weird.


packages/http-server/src/server.ts
```diff
@@ -49,13 +49,86 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+        baseUrl: query.__server,
+      },
+      headers,
+      body,
+    };
+
+    request.log.info({ input }, 'Request received');
+    try {
+      const operationSpecificConfig = getHttpConfigFromRequest(input);
+      const mockConfig = Object.assign({}, opts.config.mock, operationSpecificConfig);
```
StefanDywersant: How about:
```suggestion
      const mockConfig = { ...opts.config.mock, ...operationSpecificConfig };
```
or
```suggestion
      const mockConfig = _.defaults(opts.config.mock, operationSpecificConfig);
```
?

`Object.assign` is soo ECMA 2015 :/

packages/core/src/factory.ts
```diff
@@ -18,31 +18,9 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
-            return TaskEither.right<Error, Resource | undefined>(undefined);
-          },
-          value => TaskEither.right(value),
-        ),
+        TaskEither.fromEither(components.route({ resources, input })),
         TaskEither.chain(resource => {
-          // validate input
-          if (config.validateRequest && resource && components.validateInput) {
+          if (config.validateRequest && resource) {
             inputValidations.push(
```
lag-of-death: if we're doing this refactor, let's not force the monad to go outside its `default minimal context` - https://drboolean.gitbooks.io/mostly-adequate-guide-old/content/ch9.html. So if we need `inputValidations`, let's pass it through the chain or use `Reader` monad. And also, most importantly, let's not `.push`, especially inside of a monad

packages/core/src/factory.ts
```diff
@@ -100,6 +76,21 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
             },
           };
         }),
+        TaskEither.mapLeft(error => {
+          if (components.validateInput) {
+            const { message, name, status } = error as ProblemJsonError;
+            // let's just stack it on the inputValidations
+            // when someone simply wants to hit an URL, don't block them
+            inputValidations.push({
```
lag-of-death: let' maybe pass `inputValidations` as `output` and `resource` are

packages/http-server/src/server.ts
```diff
@@ -49,13 +49,86 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+        path: (url || '/').split('?')[0],
+        query,
+        baseUrl: query.__server,
+      },
+      headers,
+      body,
+    };
+
+    request.log.info({ input }, 'Request received');
+    try {
```
lag-of-death: if we get rid of `throw new Error('Unable to find any decent response for the current request.');` and `throw new Error(`Cannot find serializer for ${contentType}`);` then, possibly, this `try/catch` could be turned into an Either

packages/http-server/src/server.ts
XVincentX: Ahaha, you can definitely see I'm 30 :trollface:

packages/core/src/factory.ts
XVincentX: Hmm I fail to understand what you mean here, can you elaborate?

packages/core/src/factory.ts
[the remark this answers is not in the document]
XVincentX: It is indeed the aim; we're struggling so much with these that it's time to make them functional so they can be composed and stuff, but this is going to require a bigger effort that I couldn't find in a 2.5 hours flight :)

packages/http-server/src/server.ts
XVincentX: That's a good point, however it's out of the scope of this PR to use fp-ts in the http-server — because we do not really know yet what's going to happen to this package in general. We'll get up a plan the day we'll have clear ideas.

packages/core/src/factory.ts
lag-of-death: `inputValidations` should be composed inside the chain, and passed down through the chain (ie. like `output`). It shouldn't be a global that's being mutated - `const inputValidations: IPrismDiagnostic[] = [];`. LMK if this explains :)

packages/core/src/factory.ts
XVincentX: I do not think it's going to happen in this PR, unfortunately.

packages/core/src/types.ts
```diff
@@ -18,24 +18,22 @@ export interface IPrismConfig {
...
-  validateOutput?: ValidatorFn<Resource, Output>;
+  route: (opts: { resources: Resource[]; input: Input }) => Either<Error, Resource>;
+  validateInput: ValidatorFn<Resource, Input>;
+  validateOutput: ValidatorFn<Resource, Output>;
   mock: (
     opts: {
       resource: Resource;
       input: IPrismInput<Input>;
-      config?: Config;
+      config: Config['mock'];
```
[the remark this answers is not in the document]
XVincentX: Correct, but on the other hand we figured that every optional parameter we take would be another `Option<T>` to handle and so, if not necessary, we prefer to pass all the parameters explicitly.

Plus, this makes much easier to chain them with a `Reader` in case it's going to be needed on day.

packages/core/src/factory.ts
XVincentX: Although it's what I'd like to do, this is really complicated at this stage and I'm going to have to pass on.

The problem is that in order to accumulate all these errors the validators should be kind of refactored and the returned types unified, and we're really not at this stage right now.

I've tried this in the past and the result is basically TypeScript screaming type errors everywhere, and their resolution is not easy at all :(

packages/core/src/factory.ts
lag-of-death: Ok, sure :+1:

packages/core/src/factory.ts
lag-of-death: I'm not that hardcore myself, but if TypeScript becomes a problem, you can always try PureScript ;)

PR #667, opened 2019-10-01
Form data encoding improvements

XVincentX:


packages/http/src/validator/validators/body.ts
```diff
@@ -139,3 +53,95 @@ function decodeUriEntities(target: Dictionary<string, string>) {
...
+}
+
+export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent> {
+  constructor(private prefix: string) {}
+
+  public validate(target: any, specs: IMediaTypeContent[], mediaType?: string): IPrismDiagnostic[] {
+    const mediaTypeWithContentAndSchema = pipe(
+      Option.fromNullable(mediaType),
+      Option.chain(mt => findContentByMediaTypeOrFirst(specs, mt)),
+      Option.alt(() => Option.some({ content: specs[0] || {}, mediaType: 'random' })),
```
StefanDywersant: I think we should keep `mediaType` optional here. It is optional on the input level and it's better to keep it like that.

PR #668, opened 2019-10-02
Restore/refactor the forwarder

XVincentX:
This PR puts the forwarder back with the change that if requested path is not declared in the spec document (aka: there's no http operation) the request won't be executed.

It also removes some features we never used and are not even exposed in the new client (CancelToken and responseType)

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -80,27 +80,23 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
 
       const { output } = response;
 
-      if (output) {
-        reply.code(output.statusCode);
+      reply.code(output.statusCode);
 
-        if (output.headers) {
-          reply.headers(output.headers);
-        }
-
-        response.validations.output.forEach(validation => {
-          if (validation.severity === DiagnosticSeverity.Error) {
-            request.log.error(`${validation.path} — ${validation.message}`);
-          } else if (validation.severity === DiagnosticSeverity.Warning) {
-            request.log.warn(`${validation.path} — ${validation.message}`);
-          } else {
-            request.log.info(`${validation.path} — ${validation.message}`);
-          }
-        });
-
-        reply.serializer((payload: unknown) => serialize(payload, reply.getHeader('content-type'))).send(output.body);
-      } else {
-        throw new Error('Unable to find any decent response for the current request.');
+      if (output.headers) {
+        reply.headers(output.headers);
       }
+
+      response.validations.output.forEach(validation => {
+        if (validation.severity === DiagnosticSeverity.Error) {
+          request.log.error(`${validation.path} — ${validation.message}`);
+        } else if (validation.severity === DiagnosticSeverity.Warning) {
+          request.log.warn(`${validation.path} — ${validation.message}`);
+        } else {
+          request.log.info(`${validation.path} — ${validation.message}`);
+        }
+      });
+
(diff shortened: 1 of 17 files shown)


packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -159,44 +157,10 @@ describe('Http Client .request', () => {
...
-                  url: {
-                    path: '/x-pet',
-                  },
-                },
-                resources,
-                config,
-              ),
-            ).rejects.toThrowError(ProblemJsonError.fromTemplate(NO_BASE_URL_ERROR));
-          });
-        });
```
XVincentX: This are not relevant anymore.

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -271,9 +235,6 @@ describe('Http Client .request', () => {
       });
     });
 
-    // TODO will be fixed by https://stoplightio.atlassian.net/browse/SO-260
-    test.todo('GET /pet without an optional body parameter');
-
```
XVincentX: I have no clue what this is about.

packages/http/src/forwarder/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  const baseUrl =
+    resource.servers && resource.servers.length > 0 ? resolveServerUrl(resource.servers[0]) : input.url.baseUrl;
+
+  if (!baseUrl) {
+    return TaskEither.left(ProblemJsonError.fromTemplate(NO_BASE_URL_ERROR));
+  }
+
+  return TaskEither.tryCatch<Error, IHttpResponse>(async () => {
+    const response = await axios.request<unknown>({
+      method: input.method as any,
```
XVincentX: I know, there's nothing I can do here.

packages/http-server/src/server.ts
```diff
@@ -71,7 +71,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
     request.log.info({ input }, 'Request received');
     try {
       const operationSpecificConfig = getHttpConfigFromRequest(input);
-      const mockConfig = { ...opts.config.mock, ...operationSpecificConfig };
+      const mockConfig = opts.config.mock === false ? false : { ...opts.config.mock, ...operationSpecificConfig };
```
lag-of-death: do we have a way to `not to mock` from the CLI? Seems to me like it's only in the tests that we can force the forwarder to be used. Opts passed to `createHttpServer` stays the same in this PR (are not changed). Should we enable the forwarder from CLI or is this PR only a part of changes for the proxying?

XVincentX: That's correct. This PR right now will only enable this from the Client JS API; once we'll figure out how the proxy should work we'll then enable the relative command.

packages/http/src/forwarder/index.ts
```diff
@@ -41,18 +42,12 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
...
-    return server.url;
-  }
-
-  return server.url.replace(/{(.*?)}/g, (_match, variableName) => {
-    const variable = server.variables![variableName];
-    if (!variable) {
-      throw new Error(`Variable '${variableName}' is not defined, cannot parse input.`);
-    }
+  const variables = server.variables;
+  if (!variables) return server.url;
```
lag-of-death: @XVincentX, you win, no `{}` ;)

PR #669, opened 2019-10-02
Proxy command

### LINKED ISSUE #473: Proxy Server

XVincentX:
This PR implements the basic proxy feature for Prism using axios and the forwarder concept.

To check out how it works, please see the update documentation page (part of this PR)

In order to test this I've added a secondary container running an instance of httpbin in CircleCI so that we can test the proxy capabilities while staying in a local network and have the tests still relatively fast and reliables.

Moreover, I had to modify the harness framework so that the injected file must be written with `${file}` so that we do not incur in problems with positional arguments. Given I was there, I've moved some tests in directories because the `spec` directory was just huge and impossible to manage.

There are also some `eslint` leftovers that I had to fix otherwise it would just yell at me how bad I am at writing software.

Closes #473


packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,46 @@
...
+import { CommandModule } from 'yargs';
+import { createMultiProcessPrism, CreateProxyServerOptions, createSingleProcessPrism } from '../util/createServer';
+import sharedOptions from './sharedOptions';
+
+const proxyCommand: CommandModule = {
+  describe: 'Start a proxy server with the given spec file',
+  command: 'proxy <spec> <upstream>',
+  builder: yargs =>
+    yargs
+      .positional('spec', {
```
philsturgeon: Can we talk about "document" as spec could mean "the local file" or could mean "OpenAPI v2, v3, JSON Schema" etc.

XVincentX: I'm sorry I keep doing this mistake.

docs/guides/cli.md
```diff
@@ -35,10 +35,11 @@ Connection: keep-alive
 
 Responses will be mocked using realistic data that conforms to the type in the description.
 
-## Determine Response Status
+### Dynamic configuration through CLI
```
philsturgeon:
```suggestion
### Runtime Configuration in HTTP
```

docs/guides/cli.md
```diff
@@ -52,6 +53,71 @@ Connection: keep-alive
 
 The body, headers, etc. for this response will be taken from the API description document.
 
+```bash
```
StefanDywersant: I believe it should be a couple of lines below.

docs/guides/cli.md
```diff
@@ -52,6 +53,71 @@ Connection: keep-alive
...
+This commands creates an HTTP server that will proxy all the requests to the specified upstream server. Prism will analyze the request coming in and the response coming back from the upstream server and report the discrepancies with what's declared in the provided OpenAPI document.
+
+```
+prism proxy https://raw.githack.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore.yaml https://petstore.swagger.io/v2
+
+[CLI] …  awaiting  Starting Prism…
+[HTTP SERVER] ℹ  info      Server listening at http://127.0.0.1:4010
+[CLI] ●  note      GET        http://127.0.0.1:4010/pets
+[CLI] ●  note      POST       http://127.0.0.1:4010/pets
+[CLI] ●  note      GET        http://127.0.0.1:4010/pets/{petId}
```
StefanDywersant: :(

packages/cli/src/commands/sharedOptions.ts
```diff
@@ -0,0 +1,41 @@
+import { Dictionary } from '@stoplight/types';
+import { Options } from 'yargs';
+
+const sharedOptions: Dictionary<Options> = {
```
StefanDywersant: Good idea!

packages/cli/src/util/createServer.ts
```diff
@@ -26,41 +27,48 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
   } else {
     const logInstance = createLogger('CLI');
     try {
-      return await createPrismServerWithLogger(options, logInstance);
+      return createPrismServerWithLogger(options, logInstance).catch(() => cluster.worker.kill());
```
StefanDywersant: I'd stay with the previous solution with `await`. Having two ways of catching errors is hard to read.

docs/guides/cli.md
XVincentX: You are right :)

docs/guides/cli.md
XVincentX: What's up?

docs/guides/cli.md
```diff
@@ -52,6 +53,63 @@ Connection: keep-alive
 
 The body, headers, etc. for this response will be taken from the API description document.
 
+#### Request a specific example from a resource
+
+You can request a specific example from the named of your document by using the `__example` query string paramenter.
```
lag-of-death:
```suggestion
You can request a specific example from your document by using the `__example` query string paramenter.
```

docs/guides/cli.md
```diff
@@ -52,6 +53,63 @@ Connection: keep-alive
...
+curl -v -s http://localhost:4010/pets/10 > /dev/null
+
+< HTTP/1.1 404 Not Found
+< sl_validation: [{"severity":"Error","message":"The received media type does not match the one specified in the document"},{"location":["body"],"severity":"Error","code":"type","message":"should be object"}]
+```
+
+You can see there's a `sl_validation` header which is a JSON Payload with all the violations found in the response.
+
+```bash
+prism proxy https://raw.githack.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore.yaml https://petstore.swagger.io/v2 --log httpResponse
```
lag-of-death: hm.. maybe `httpBody` or `httpPayload` insted of `httpResponse`? There is `httpHeader` above. And having `httpHeader` vs `httpResponse` makes it a bit confusing

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      cors: true,
+      log: 'stdout',
+      multiprocess: true,
+      host: '0.0.0.0',
+      port: 4010,
+      ...additionalCallOptions,
+    });
+  });
+
+  test(`starts ${command} server with httpResponse log option `, async () => {
```
lag-of-death:
```suggestion
  test(`starts ${command} server with httpBody log option `, async () => {
```

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,46 @@
...
+      .coerce('upstream', value => {
+        try {
+          return new URL(value);
+        } catch (e) {
+          throw new Error(`Invalid upstream URL provided: ${value}`);
+        }
+      })
+      .middleware(async argv => (argv.operations = await getHttpOperationsFromResource(argv.document!)))
+      .options(sharedOptions),
+  handler: parsedArgs => {
```
lag-of-death: `handler` could also be shared

packages/cli/src/util/createServer.ts
```diff
@@ -23,45 +24,54 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
...
   if (options.operations.length === 0) {
     throw new Error('No operations found in the current file.');
   }
 
+  const config: IHttpProxyConfig | IHttpConfig = isProxyServerOptions(options)
+    ? {
+        mock: false,
+        validateRequest: true,
+        validateResponse: true,
+        checkSecurity: true,
```
lag-of-death: these 3 are identical to the counterparts down below, could have sth like:
```
const shared = { 
  validateRequest: true,
  validateResponse: true,
  checkSecurity: true,
}
```

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -15,27 +15,15 @@ function improveSchema(schema: JSONSchema) {
...
-    }
+  } else if (newSchema.type === 'string' && !newSchema.format && !newSchema.enum) {
+    newSchema['x-faker'] = 'lorem.word';
+  } else if (newSchema.type === 'object' && newSchema.properties) {
+    newSchema.properties = Object.entries(newSchema.properties).reduce((r, [k, v]) => {
+      r[k] = improveSchema(v);
+      return r;
+    }, {});
+  } else if (newSchema.type === 'array' && typeof newSchema.items === 'object') {
+    newSchema.items = improveSchema(newSchema.items);
```
lag-of-death: It sure does look complicated. Can we add some tests to it? I've checked the coverage for this and it might need some additional tests. Coverage is lacking for the "red lines":
coverage

test-harness/specs/violations/print-as-http-response.txt
```diff
@@ -0,0 +1,33 @@
+====test====
+Prism generates dynamic values for schema when no static examples are provided.
+
+Given there are examples defined for an operation
+And this operation can produce application/json response
+And this operation response defines a JSON Schema (schema field)
+And I make a request to this operationAnd the request has Accept header of application/json
+And I add __dynamic=true to the query path
+Then I should get a response with dynamically generated JSON as payload
```
lag-of-death: nah, this does not seem to be right ;)

lag-of-death:
@XVincentX, I've left some suggestions/comments. Initially, I thought that we could maybe reuse some of the e2e tests from https://github.com/stoplightio/prism/pull/651/files#diff-705dfbdb1cdce8ac455b13c0f720b8bf but as I understand, there is no validating that `upstream` is listed in a spec, right? Great job :+1:

packages/cli/src/util/createServer.ts
XVincentX: Good point. Let's share these. 101ae0c

packages/cli/src/commands/__tests__/commands.spec.ts
XVincentX: Not too sure I understand this. Are you proposing to change the option name?

packages/cli/src/util/createServer.ts
XVincentX: Yeah this is one of these refactors that didn't really work :trollface: — will put it back.

packages/http/src/mocker/generator/HttpParamGenerator.ts
XVincentX: Yeah I agree with you that this has been a little bit under-tested. I've added some more tests for this. 923bee5

XVincentX:
@lag-of-death Good point on the `upstream` thing.

Basically yes, we would need to spin up a server somehow to make the validation thing affordable; which is actually kind of doable.

CircleCI can run multiple containers and so in theory we could have an instance of https://httpbin.org running as a separate container and use that as a proxy.

The downside is that in such case — Harness will be only and forever on CircleCI, with no possibility of running them elsewhere (such as Azure Pipelines or something). I'll try to do some tests.

packages/cli/src/commands/__tests__/commands.spec.ts
lag-of-death: @XVincentX, yes. Because a response is comprised of headers and a payload. For now we have `httpHeader` and `httpResponse`. Think we should change `httpResponse` to either `httpBody` or `httpPayload`. If we do this we name the entities correctly. We can think of it as `response` being an aggregate and `header(s)` and `body`, entities it aggregates. WDYT?

XVincentX: Sounds reasonable. @philsturgeon What do you think?

packages/cli/src/commands/proxy.ts
lag-of-death: it really could ;) so that we have the same logic for `proxy` and `mock`. But maybe we don't want to make this `handler` shared? I'm thinking, do we want `proxy` to also be restarted (hot-reloaded) when the doc is being changed? If so, we might just make `handler` a common functionality. And this is assuming that the hot-reloading will indeed take place here, of which I am still unsure

docs/guides/cli.md
lag-of-death: hm.. @XVincentX, I don't think we want to leave `from the named of` as is, I'm confused by this

docs/guides/cli.md
```diff
@@ -52,6 +53,63 @@ Connection: keep-alive
...
+
+You can override the `dynamic` keyword (which decides whether the generated example is static or dynamic) through the `__dynamic` query string parameter.
+
+```bash
+curl -v http://127.0.0.1:4010/pets/123?__dynamic=false
+```
+
+## Proxy
+
+This commands creates an HTTP server that will proxy all the requests to the specified upstream server. Prism will analyze the request coming in and the response coming back from the upstream server and report the discrepancies with what's declared in the provided OpenAPI document.
```
lag-of-death:
```suggestion
This command creates an HTTP server that will proxy all the requests to the specified upstream server. Prism will analyze the request coming in and the response coming back from the upstream server and report the discrepancies with what's declared in the provided OpenAPI document.
```

.dockerignore
```diff
@@ -1,8 +1,9 @@
+**/.circleci
+**/.github
+**/docs
+**/examples
 **/node_modules
 **/__tests__
-**/examples
 **/test-harness
 **/.tsbuildinfo
 **/dist
```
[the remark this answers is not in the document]
StefanDywersant: Cool, catastrophy avoided.

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
```diff
@@ -50,4 +50,50 @@ describe('HttpParamGenerator', () => {
       });
     });
   });
+
+  describe('improveSchema', () => {
+    describe.each(['number', 'integer'])('when feed with a %s', type => {
+      // @ts-ignore
+      const improvedSchema = improveSchema({ type });
+
+      it('shuold have a minimum and a maximum', () => {
```
StefanDywersant:
```suggestion
      it('should have a minimum and a maximum', () => {
```

StefanDywersant:
@XVincentX what are the pros of choosing axios as a backbone for doing requests?

node-fetch has few advantages:
1) window.fetch compliant interface which means using some kind of standard
2) 1)
3) effectively one dependency less (ref-resolves depend on node-fetch already)
4) if there will be super-fast lib in the future it is more likely that it will follow fetch interface

XVincentX:
@StefanDywersant Yes, now that even the resolver is using node fetch I think it makes sense to remove axios from our codebase.

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,46 @@
+import { getHttpOperationsFromResource } from '@stoplight/prism-http';
+import { pick } from 'lodash';
+import { CommandModule } from 'yargs';
+import { createMultiProcessPrism, CreateProxyServerOptions, createSingleProcessPrism } from '../util/createServer';
+import sharedOptions from './sharedOptions';
+
+const proxyCommand: CommandModule = {
+  describe: 'Start a proxy server with the given document file',
+  command: 'proxy <document> <upstream>',
```
StefanDywersant: Having two positional args may be confusing:
1) folks won't remember which one should be first
2) their order does not result from some logical explanation: e.g. `cp <from> <to>` <- this has logical explanation

Moreover, most of the time we can implicate upstream host. It should be in OAS document.

To sum this up, I'd recommend optional: `-u / --upstream` which when not provided results in host form OAS "api description".

StefanDywersant: + a failure when neither -u nor oas/host is provided

packages/cli/src/commands/proxy.ts
StefanDywersant: For now, it doesn't look like sth worth sharing. Effectively `pick(...)` varies depending on the command. L42 is one and only common thing. But, according to what you say about hot-reload it may be valuable to share it when there will be more code.

packages/cli/src/commands/proxy.ts
lag-of-death: @StefanDywersant, I think that the conclusion was that `upstream` is not gonna be validated against what's in spec. cc @XVincentX

StefanDywersant: Oh, thank you for that lovely gif :)

However, I believe our cli tools should more or less look like iphone interfaces rather than not iphone. that being said, even if you just woke up or u're drunk you should be able to spin proxy instance without looking to the manual.

lag-of-death: I do get the point :) I'm fine with the current version though

.circleci/config.yml
```diff
@@ -120,9 +120,6 @@ workflows:
             tags:
               only: /^v.*/
           requires:
-            - build
-            - harness
-            - harness-macos
```
lag-of-death: are we sure just requiring `publish` will also wait for these steps? I mean, I know these steps ARE required by `publish` step, but is it safe to assume so?

packages/cli/src/commands/mock.ts
```diff
@@ -1,71 +1,41 @@
...
         type: 'string',
       })
-      .middleware(async argv => argv.operations = await getHttpOperationsFromResource(argv.spec!))
-      .fail((msg, err) => {
-        if (msg) yargs.showHelp();
-        else signale.fatal(err.message);
-
-        process.exit(1);
-      })
+      .middleware(async argv => (argv.operations = await getHttpOperationsFromResource(argv.document!)))
```
StefanDywersant: That looks like a dirty hack that can even fail in certain circumstances.

How about removing that middleware and resolve operations in handler below?

StefanDywersant: I did not find any information that middleware accepts a promise.

lag-of-death: Looks OK to me, well, in the handler you wouldn't need the mutation I guess. But, why do you think it would fail? The only situation for this to fail would be not having `document` positional argument. But since it is positional it will be checked by `yargs`

lag-of-death: @StefanDywersant, behold https://github.com/yargs/yargs/pull/1257

StefanDywersant: @lag-of-death that explains everything, thanks!

.circleci/config.yml
```diff
@@ -131,8 +128,4 @@ workflows:
...
-            - publish
```
lag-of-death: ditto

CHANGELOG.md
```diff
@@ -12,11 +12,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Validating output status code against available response specs [#648](https://github.com/stoplightio/prism/pull/648)
 - Support for Contract Testing [#650](https://github.com/stoplightio/prism/pull/650)
 - The CLI will now propose operation paths with meaningful examples [#671](https://github.com/stoplightio/prism/pull/671)
+- Prism now includes a new `proxy` command to act as a validation server [#669](https://github.com/stoplightio/prism/pull/669)
```
lag-of-death: hm.. is `validation server` something that most of the folks will understand? Or is it only a term that we have an idea what this is about?

.circleci/config.yml
XVincentX: Yes, according to the CircleCI documentation.

.circleci/config.yml
XVincentX: ditto :)

CHANGELOG.md
XVincentX: I'll articulate better.

docs/guides/cli.md
XVincentX: Yes, we should probably replace that. I'll see what I can do.

packages/cli/src/commands/mock.ts
XVincentX: In any case I want to get rid of this middleware and leverage chokidar at some point to see the first read as a "special" reload so we do not have to do this anymore.

packages/cli/src/util/createServer.ts
```diff
@@ -95,10 +101,25 @@ function pipeOutputToSignale(stream: Readable) {
...
-  host?: string;
+  host: string;
   port: number;
   operations: IHttpOperation[];
+  multiprocess: boolean;
+  log: 'stdout' | 'httpResponse' | 'httpHeader';
 };
+
+export interface CreateProxyServerOptions extends CreateBaseServerOptions {
+  dynamic: false;
```
StefanDywersant: `dynamic` makes sense only in mock command, right? If so, let move it there.

packages/http-server/src/server.ts
```diff
@@ -80,23 +71,44 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
 
       const { output } = response;
 
-      reply.code(output.statusCode);
-
-      if (output.headers) {
-        reply.headers(output.headers);
+      if (response.validations.output.length > 0) {
+        const violations = response.validations.output.map(detail => ({
```
StefanDywersant: Is `violations` supported by any type? How about creating one + a builder which transforms `IDiagnostic` into that type.

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+import { createMultiProcessPrism, createSingleProcessPrism } from '../../util/createServer';
+import mockCommand from '../mock';
+import proxyCommand from '../proxy';
+
+const parser = yargs.command(mockCommand).command(proxyCommand);
+
+jest.mock('../../util/createServer', () => ({
+  createMultiProcessPrism: jest.fn(),
+  createSingleProcessPrism: jest.fn(),
+}));
```
StefanDywersant: `jest.mock('../../util/createServer');` overrides exported functions with `jest.fn()` for you. You don't need to supply implementation of module, the code inside `createServer` won't run.

packages/cli/src/commands/proxy.ts
XVincentX: I'm ok with both. @philsturgeon what do you think?

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
```diff
@@ -50,4 +50,50 @@ describe('HttpParamGenerator', () => {
       });
     });
   });
+
+  describe('improveSchema', () => {
```
StefanDywersant:
```suggestion
  describe('improveSchema()', () => {
```

docs/guides/cli.md
XVincentX: I'm resolving this comment since we're having the same discussion below.

packages/cli/src/commands/__tests__/commands.spec.ts
XVincentX: Oh cool, didn't know that. 8f2685e

packages/cli/src/commands/proxy.ts
XVincentX: This will be shared and refactored — but I need #689 merged before working on this.

packages/http/src/forwarder/index.ts
```diff
@@ -1,36 +1,25 @@
 import { IPrismComponents } from '@stoplight/prism-core';
```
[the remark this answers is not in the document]
XVincentX: This is effectively a forwarding thing; the proxy is done by the http server that will take your request and send it back. If you're using the client library — you're indeed forwarding something.

packages/http-server/src/server.ts
XVincentX: Hmm can you elaborate more? I am not sure I understand what you mean

packages/http/src/forwarder/index.ts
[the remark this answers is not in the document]
StefanDywersant: Fair enough, resolving.

packages/http-server/src/server.ts
StefanDywersant: Yes, `violations` is currently a dynamic object. I'd be nice to have a type for it: `IHeaderDiagnostic` or sth similar.

Additionally, there could be a builder:
`const violations = response.validations.output.map(fromPrismDiagnostic)` (more or less)

XVincentX: Ah ok ok I understand it now. Will see what I can do about it, sure!

StefanDywersant:
@XVincentX @lag-of-death since we are doing proxy we should fill in `X-Forwarded-*` headers.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

StefanDywersant:
> @XVincentX @lag-of-death since we are doing proxy we should fill in `X-Forwarded-*` headers.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

Even better, there is a standardized version of that header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

packages/cli/src/util/createServer.ts
XVincentX: 28ebfc2

test-harness/index.ts
```diff
@@ -45,8 +45,7 @@ describe('harness', () => {
 
     test(testText, done => {
       const [command, ...args] = parsed.command.split(' ').map(t => t.trim());
-      const serverArgs = [...parsed.server.split(' ').map(t => t.trim()), tmpFileHandle.name];
-
+      const serverArgs = [...parsed.server.split(' ').map(t => t.trim())].map(t => t === '${file}' ? tmpFileHandle.name : t);
```
StefanDywersant:
```suggestion
      const serverArgs = parsed.server.split(' ').map(t => t.trim().replace('${file}, tmpFileHandle.name));
```

packages/http-server/src/serialize.ts
```diff
@@ -42,6 +42,8 @@ export const serialize = (payload: unknown, contentType?: string) => {
   const serializer = contentType ? serializers.find(s => s.regex.test(contentType)) : undefined;
 
   if (!serializer) {
+    if (typeof payload === 'string') return payload;
```
StefanDywersant: Missing test

packages/http/src/types.ts
```diff
@@ -17,6 +17,11 @@ export interface IHttpConfig extends IPrismConfig {
   mock: false | IHttpOperationConfig;
 }
 
+export interface IHttpProxyConfig extends IHttpConfig {
```
StefanDywersant: I think there is sth not right with the config types.

Below looks more reasonable:
`IHttpProxyConfig extends IHttpBaseConfig { upstream: .. }`
`IHttpMockConfig extends IHttpBaseConfig { mock: .. }`
`IHttpConfig { common stuff }`

WDYT?

packages/http/src/index.ts
```diff
@@ -7,16 +7,18 @@ import route from './router';
...
 export * from './mocker/serializer/style';
 export { generate as generateHttpParam } from './mocker/generator/HttpParamGenerator';
 
-import { IHttpConfig, IHttpRequest, IHttpResponse, PickRequired, PrismHttpComponents } from './types';
+import { IHttpConfig, IHttpRequest, IHttpResponse, PickRequired, PrismHttpComponents, IHttpProxyConfig } from './types';
 
 export const createInstance = (
-  defaultConfig: IHttpConfig,
-  components?: PickRequired<Partial<PrismHttpComponents>, 'logger'>,
+  defaultConfig: IHttpConfig | IHttpProxyConfig,
```
StefanDywersant: Please take a look at the comment on `http/src/types.ts`.

packages/http/src/forwarder/index.ts
```diff
@@ -1,53 +1,49 @@
...
+  TaskEither.tryCatch<Error, IHttpResponse>(() => {
+    const partialUrl = parse(baseUrl);
+    return fetch(
+      format({
+        ...partialUrl,
+        pathname: posix.join(partialUrl.pathname || '', input.url.path),
+        query: input.url.query,
+      }),
+      {
+        headers: defaults(omit(input.headers, ['host', 'accept']), {
```
StefanDywersant:
```suggestion
        headers: defaults(omit(input.headers, ['host']), {
```

No need to omit since you are overriding it anyway.

packages/http/src/forwarder/index.ts
```diff
@@ -1,53 +1,49 @@
...
+          accept: 'application/json, text/plain, */*',
+          'user-agent': `Prism/${prismVersion}`,
+        }),
+      }
+    )
+      .then(response =>
```
StefanDywersant: How about using async/await here?

packages/http/src/forwarder/index.ts
```diff
@@ -1,53 +1,49 @@
...
+        Promise.all([
+          response,
```
StefanDywersant: ..then no magic like this would be needed

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -158,51 +164,32 @@ describe('Http Client .request', () => {
...
-      });
-
       describe('Prism user-agent header', () => {
         describe('when the defaults are used', () => {
           it('should use Prism/<<version>> for the header', async () => {
-            const userAgent = await checkUserAgent(config, prism, resources);
+            const userAgent = await checkUserAgent(config, prism, resources, {}, 'https://stoplight.io');
 
-            expect(userAgent).toBe(`Prism/${prismVersion}`);
+            expect(userAgent).toContain(`Prism/${prismVersion}`);
```
StefanDywersant: Why `.toContain` instead of `.toBe`?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -158,51 +164,32 @@ describe('Http Client .request', () => {
...
+              prism,
+              resources,
+              {
+                'user-agent': 'Other_Agent/1.0.0',
+              },
+              'https://stoplight.io'
+            );
 
-            expect(userAgent).toBe('Other_Agent/1.0.0');
+            expect(userAgent).toContain('Other_Agent/1.0.0');
```
StefanDywersant: ditto

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,114 @@
+import * as utils from '@stoplight/prism-http';
```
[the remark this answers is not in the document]
XVincentX: I do not know why I called it in this way. I'll rename.

I cannot use `jest.mock` because I need to mock the returned values in some functions, see line 10,11

package.json
```diff
@@ -57,12 +56,10 @@
...
     "ts-node": "^8.3.0",
     "tsconfig-paths": "^3.8.0",
-    "typescript": "^3.6.4"
-  },
-  "prettier": {
-    "printWidth": 100,
-    "singleQuote": true,
-    "trailingComma": "es5"
+    "typescript": "^3.6.4",
+    "@types/node-fetch": "^2.5.2",
```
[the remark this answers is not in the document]
XVincentX: Done!

packages/http/src/__tests__/http-prism-instance.spec.ts
XVincentX: I have no idea why I changed this. I am so old.

packages/http/src/__tests__/http-prism-instance.spec.ts
XVincentX: I have no idea why I changed this. I am so old.

packages/http/src/forwarder/index.ts
XVincentX: Sure. c9962bf

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+By default, the output violations will be reported on the standard output, but you can also specify to embed these in the HTTP Response:
+
+```bash
+prism proxy https://raw.githack.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore.yaml https://petstore.swagger.io/v2 --log httpHeader
+… …
+
+curl -v -s http://localhost:4010/pets/10 > /dev/null
+
+< HTTP/1.1 404 Not Found
+< sl_validation: [{"severity":"Error","message":"The received media type does not match the one specified in the document"},{"location":["body"],"severity":"Error","code":"type","message":"should be object"}]
```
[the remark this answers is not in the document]
XVincentX: We had a discussion about this with @philsturgeon and he pointed out a very good article on why we should not use `X-` headers anymore: https://www.mnot.net/blog/2009/02/18/x-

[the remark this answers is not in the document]
StefanDywersant: Sounds fair. Why not SL-Validation then? - still sounds more along with existing headers than _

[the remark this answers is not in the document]
XVincentX: Fair enough.
99fc0f3

StefanDywersant:
> Damn 222 files!

I'll charge him 2 pints for reviewing that.

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+[HTTP SERVER] ℹ  info      Server listening at http://127.0.0.1:4010
+[CLI] ●  note      GET        http://127.0.0.1:4010/pets
+[CLI] ●  note      POST       http://127.0.0.1:4010/pets
+[CLI] ●  note      GET        http://127.0.0.1:4010/pets/10
+```
+
+By default, the output violations will be reported on the standard output, but you can also specify to embed these in the HTTP Response:
+
+```bash
+prism proxy https://raw.githack.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore.yaml https://petstore.swagger.io/v2 --log httpHeader
```
philsturgeon: Did we stick with '--log httpHeader' we talked about some alternatives before, maybe the docs didn't get updated?

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+… …
+
+curl -v -s http://localhost:4010/pets/10 > /dev/null
+
+< HTTP/1.1 404 Not Found
+< sl-validation: [{"severity":"Error","message":"The received media type does not match the one specified in the document"},{"location":["body"],"severity":"Error","code":"type","message":"should be object"}]
```
philsturgeon: I'm looking into the `warning` header as a more standard alternative to this. https://tools.ietf.org/html/rfc7234#section-5.5

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+```
+
+You can see there's a `sl-validation` header which is a JSON Payload with all the violations found in the response.
```
philsturgeon:
```suggestion
You can see there's a `sl-validation` header which is a JSON object with all the violations found in the response.
```

philsturgeon:
This is really impressive but we've got some rough corners to round off. 

1. Our error format does not explain if the problem was in the request or the response. 

```
[{"location":["body"],"severity":"Error","code":"required","message":"should have required property 'nondefined'"}]
```

2. We are hinting at more information in some of our errors, but not providing it: 

```
[{"location":["body","status"],"severity":"Error","code":"enum","message":"should be equal to one of the allowed values"}]
```

I got this by changing the Pet definition to this:

```yaml
      status:
        type: string
        enum:
          - availablefoo
          - pending
          - sold
```

running `./cli-binaries/prism-cli proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2 --log httpHeader` and hitting `http://127.0.0.1:4010/pet/1`.

@lag-of-death the reload is really good, great job!

philsturgeon:
Where is "code" coming from? This seems like a new concept and im not sure I understand it.

docs/guides/cli.md
XVincentX: We're still using `--log httpHeader`; I do not remember we agreed on something else but I might not recall correctly

docs/guides/cli.md
XVincentX: I had a quick look into and it seems to be mostly cache related?

XVincentX:
@philsturgeon About the error messages, currently we're just forwarding what AJV is returning us. 

I know that spectral is using https://github.com/atlassian/better-ajv-errors at the moment, I can try to have a look and see if it improves the situation a bit.

docs/guides/cli.md
philsturgeon: I don't think we agreed on an exact new set of options but we discussed some alternatives. I just listed a few ideas elsewhere in this PR we can try.

docs/guides/cli.md
philsturgeon: 2nd paragraph says: 

>    Warnings can be used for other purposes, both cache-related and otherwise.  The use of a warning, rather than an error status code, distinguishes these responses from true failures.

philsturgeon:
> @philsturgeon About the error messages, currently we're just forwarding what AJV is returning us.
> 
> I know that spectral is using https://github.com/atlassian/better-ajv-errors at the moment, I can try to have a look and see if it improves the situation a bit.
> 
> > Where is "code" coming from? This seems like a new concept and im not sure I understand it.
> 
> Same thing, it's mostly coming from ajv. I'll look into and see what I can do here.

I think we'll need to make our own Violation type which puts things how we like. AJV might change over time so just punting that out to the header could cause a lot of trouble. 

We want to have most of the same information that the RFC 7807 JSON has, so that people can actually use this information to track down mismatches. If I saw these headers I would still have a lot of questions about what the actual problem was.

- request or response
- what was the incorrect value
- any way of suggesting a more correct value (is there an example, default, format, or enum values) 

If we use the Warning header then things might need to be more text based. In the past I've used code like this:

```

function buildHumanErrors(errors) {
  return errors.map(function(error) {
    if (error.params.missingProperty) {
      const property = pointer.get(userSchema, '/properties/' + error.params.missingProperty);
      return property.title + ' is a required field';
    }
    const property = pointer.get(userSchema, '/properties' + error.dataPath);
    if (error.keyword == 'format' && property.example) {
      return property.title + ' is in an invalid format, e.g: ' + property.example;
    }
    return property.title + ' ' + error.message;
  });
}
```

More context here: https://apisyouwonthate.com/blog/the-many-amazing-uses-of-json-schema-client-side-validation

packages/cli/src/util/createServer.ts
```diff
@@ -23,57 +23,64 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
...
-  }
+  return createPrismServerWithLogger(options, logInstance);
 }
 
-async function createPrismServerWithLogger(options: CreatePrismOptions, logInstance: Logger) {
-  if (options.operations.length === 0) {
+async function createPrismServerWithLogger(options: CreateBaseServerOptions, logInstance: Logger) {
+  const operations = await getHttpOperationsFromResource(options.document);
+
+  if (operations.length === 0) {
```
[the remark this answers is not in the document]
XVincentX: JavaScript objects are always compared by reference:

image

packages/cli/src/util/createServer.ts
```diff
@@ -97,10 +104,25 @@ function pipeOutputToSignale(stream: Readable) {
...
+type CreateBaseServerOptions = {
   dynamic: boolean;
   cors: boolean;
-  host?: string;
+  host: string;
   port: number;
-  operations: IHttpOperation[];
+  document: string;
+  multiprocess: boolean;
+  log: 'stdout' | 'httpResponse' | 'httpHeader';
```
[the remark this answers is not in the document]
XVincentX: That's a very good point. I'll change the option type.

XVincentX:
@philsturgeon If it's ok with you I would keep the "validation" format out of this PR.

This one is already gigantic and the validation format is something we're suffering in the input validation as well, so I'd say it makes sense to keep it out from here.

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
XVincentX: @philsturgeon

Now that I am reading the proposal, with these options I would be able to understand where the error is going. The only reason I got it is because you wrote it in the parentheses. On this side, having `stdout vs httpHeader vs httpBody/httpResponse` easier to get (I guess).

We can definitely add such information in the CLI documentation, but the parameter does not look that intuitive.

Thoughts?

docs/guides/cli.md
XVincentX: I'm sorry I totally missed that part. Seems to fit very well with what we're looking for. Do you want me to make the change?

packages/http-server/src/serialize.ts
XVincentX: There's an harness test verifying that in such case the payload is just let go.

packages/cli/src/util/createServer.ts
```diff
@@ -97,10 +104,25 @@ function pipeOutputToSignale(stream: Readable) {
   });
 }
 
-export type CreatePrismOptions = {
+function isProxyServerOptions(options: CreateBaseServerOptions): options is CreateProxyServerOptions {
+  return !options.dynamic && 'upstream' in options;
```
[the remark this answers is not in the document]
philsturgeon: Ok we can keep it

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
philsturgeon: Ok, but some main points:

- no camel case in the CLI
- logging should be default 
- httpResponse is vague we are trying to make it fail for us, maybe —validation enables that, or —error 

Maybe logging and header could always happen and the only switch is —error?

@marbemac need second opinion

docs/guides/cli.md
philsturgeon: Yeah let’s do it. It might be tricky to fit the information into text but if people want json they can enable the —error mode or whatever.

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
marbemac: I'm catching up - by "log" do we mean where to put contract validation errors?

docs/guides/cli.md
XVincentX: @philsturgeon So if I understand correctly in the `warning` header we're just going to put the error message and no other informations.

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
XVincentX: @marbemac That is correct!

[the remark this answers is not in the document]
marbemac: Got it - so yeah for me `log` has a pretty clear existing meaning that !== where contract test results are put. Ideas for a different name? `--report-contract-violations=stdout | header | response`?

marbemac:
> If it's ok with you I would keep the "validation" format out of this PR.

If we want to start safe - could map over the results and just output a really simple object with a basic message property. Can always add to that later without breaking things.

packages/cli/src/util/createServer.ts
XVincentX: This's been done.

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      host: '0.0.0.0',
+      port: 4010,
+      ...additionalCallOptions,
+    });
+  });
+
+  test(`starts ${command} server with httpResponse log option `, () => {
+    parser.parse(`${command} /path/to -m -h 0.0.0.0 --log httpResponse ${upstream}`);
+
+    expect(createSingleProcessPrism).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
XVincentX: This's been done.

packages/http/src/types.ts
XVincentX: Isn't is in this way already?

```ts
export interface IHttpConfig extends IPrismConfig {
  mock: false | IHttpOperationConfig;
}

export interface IHttpProxyConfig extends IHttpConfig {
  mock: false;
  upstream: URL;
}

export type IMockHttpConfig = IHttpConfig & { mock: IHttpOperationConfig };

```

XVincentX: 5ecf360ca253ce09a6668e2c107e02ef5862133f

packages/http/src/index.ts
XVincentX: 5ecf360ca253ce09a6668e2c107e02ef5862133f

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
XVincentX: I am really struggling to find a good name :) — on the other hand I want to make sure we do not spend ages on this…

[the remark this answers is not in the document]
marbemac: `--report-contract-violations`, `--reporter`, `--report-violations`? 

`log` just has so much baggage, and we'll probably want to reserve that flag for tradditional logs if/when we have that.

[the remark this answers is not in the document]
XVincentX: I'll go with `--report-violations`

packages/cli/src/util/createServer.ts
```diff
@@ -23,6 +23,8 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
     if (worker.process.stdout) {
       pipeOutputToSignale(worker.process.stdout);
     }
+
```
philsturgeon: If it’s an async function doesn’t it always return a promise? You could return true or just return nothing instead of this?

packages/cli/src/util/createServer.ts
```diff
@@ -97,10 +108,25 @@ function pipeOutputToSignale(stream: Readable) {
...
+type CreateBaseServerOptions = {
   dynamic: boolean;
   cors: boolean;
-  host?: string;
+  host: string;
   port: number;
-  operations: IHttpOperation[];
+  document: string;
+  multiprocess: boolean;
+  reportViolations: 'stdout' | 'httpBody' | 'httpHeader';
```
philsturgeon: I increasingly thing we just need a single flag not a series of options. 

Would it be technically possible to have it always log to stdout (useful for prod), always report to headers (useful for everyone but not the most verbose format, so they can check the logs, and make an option that allows the body to show up?

Again we don’t want any camelCase in our values we don’t canelCase anywhere in our CLI (if we do it’s a bug we missed).

—fail or —errors or —panic or something, anything, I think might be easier.

philsturgeon: Also "report violations" is terminology I would use for sending actual failures to the stoplight API, which we might want to do independently of HTTP validation of "logging" (stdout).

packages/cli/src/util/createServer.ts
XVincentX: True that, I'm just used to the old school. Fixed. f508a02

packages/cli/src/util/createServer.ts
XVincentX: > If stdout is default then we don't need to offer it as a mode which can be turned on right? Should it not always be on?

Being default does not mean it's always on. If you do `httpHeader` (or `http-header`, whatever we'll come up with) — then stdout won't happen.

---

> Would it be technically possible and reasonable to 1) always log to stdout (useful for prod), 2) always report to headers (useful for everyone but not the most verbose format, so they can check the logs, and 3) make an option that allows the body to show up?

1) yes
2) yes
3) yes

Do you want me to do that? If so, what would be the option name? `--errors-for-violation`? Something else?

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import proxyCommand from '../proxy';
+
+const parser = yargs.command(mockCommand).command(proxyCommand);
+
+jest.mock('../../util/createServer', () => ({
+  createMultiProcessPrism: jest.fn().mockResolvedValue([]),
+  createSingleProcessPrism: jest.fn().mockResolvedValue([]),
+}));
+
+jest.spyOn(prismHttp, 'getHttpOperationsFromResource').mockResolvedValue([]);
```
StefanDywersant: I'd transform this line into jest.mock() or the line above into jest.spyOn. Just to make it more perfect :)

packages/cli/src/commands/mock.ts
```diff
@@ -64,17 +28,14 @@ const mockCommand: CommandModule = {
...
-      multiprocess: boolean;
-      spec: string;
-    };
+      document,
+      reportViolations,
+    } = (parsedArgs as unknown) as CreateMockServerOptions;
 
     const createPrism = multiprocess ? createMultiProcessPrism : createSingleProcessPrism;
-    const options = { cors, dynamic, port, host, operations };
+    const options = { cors, dynamic, port, host, document, multiprocess, reportViolations };
```
StefanDywersant: The `multiprocess` option has done its job and is no longer needed. However, ignore the comment if it adds too much spaghetti into the codebase.

packages/cli/src/commands/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      port: 4010,
+      ...additionalCallOptions,
+    });
+  });
+
+  test(`starts ${command} server on custom port`, () => {
+    parser.parse(`${command} /path/to -p 666 ${upstream}`);
+
+    expect(createMultiProcessPrism).not.toHaveBeenCalled();
+    expect(createSingleProcessPrism).toHaveBeenLastCalledWith({
```
StefanDywersant: I'm thinking of changing those expectations into something more like:
`expect(createSingleProcessPrism).toHaveBeenLastCalledWith(expect.objectContaining({ port: 666 }))`. Less code, almost the same confidence.

packages/cli/src/util/createServer.ts
```diff
@@ -97,10 +108,25 @@ function pipeOutputToSignale(stream: Readable) {
   });
 }
 
-export type CreatePrismOptions = {
+function isProxyServerOptions(options: CreateBaseServerOptions): options is CreateProxyServerOptions {
+  return !options.dynamic && 'upstream' in options;
```
StefanDywersant:
```suggestion
  return 'upstream' in options;
```

`options.dynamic = false` would be also truthy here. I think to check for upstream existence is enough.

packages/http/src/getHttpOperations.ts
```diff
@@ -19,23 +19,20 @@ const httpAndFileResolver = new Resolver({
 });
 
 export async function getHttpOperationsFromResource(file: string): Promise<IHttpOperation[]> {
-  const fileContent = /^https?:\/\//i.exec(file)
-    ? (await axios.get(file, { transformResponse: res => res })).data
-    : fs.readFileSync(file, { encoding: 'utf8' });
+  const isRemote = /^https?:\/\//i.test(file);
+  const fileContent = isRemote ? await fetch(file).then(d => d.text()) : fs.readFileSync(file, { encoding: 'utf8' });
```
StefanDywersant: It's quite hard to get on which part `await` has an effect. On the other hand, it looks pretty. Please extract it to separate func or ignore the comment if you also think it's pretty ;)

Sidenote: Pre-C programmers would not believe how many complicated things are possible using a single line of code.

CONTRIBUTING.md
```diff
@@ -119,44 +119,6 @@ rm -rf packages/**/dist
 rm -rf packages/**/*.tsbuildinfo
```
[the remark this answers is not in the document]
philsturgeon: Sure

packages/cli/src/util/createServer.ts
philsturgeon: > Being default does not mean it's always on. If you do httpHeader (or http-header, whatever we'll come up with) — then stdout won't happen.

yeah i was suggesting we should make it be default and always on.

<hr>

Yes. --errors is fine, it will be explained in the manual and we can get user feedback on what the name should be if people are unsure. `--errors-for-violation` is long and also assumes people know what violations are, which they dont have a reason to yet.

packages/cli/package.json
```diff
@@ -7,17 +7,15 @@
   },
   "bugs": "https://github.com/stoplightio/prism/issues",
   "dependencies": {
-    "@stoplight/http-spec": "^2.2.3",
+    "@stoplight/http-spec": "2.2.3",
```
philsturgeon: Are we pegging to a specific version intentionally? we normally add `^` but this removes one.

XVincentX: Fixed. I do not know how this happened.

packages/cli/src/util/createServer.ts
XVincentX: > yeah i was suggesting we should make it be default and always on.

stdout? stdout and http headers?

packages/cli/src/commands/__tests__/commands.spec.ts
XVincentX: I cannot really do that — jest.mock won't make me mock the resolved value in a single statement, which I need.

packages/cli/src/commands/mock.ts
XVincentX: Yeah — basically I need to go through and modify different statement to be `Omit<CreateBaseServerOptions, 'multiprocess'>` — but it's not really worth it at this point.

I acknowledge that I'm losing the hand with the options in the CLI — I hope to refactor this very soon on the airplane to Los Angeles, 12 hours flight.

packages/cli/src/commands/__tests__/commands.spec.ts
XVincentX: Very good point. Addressed in 88da6af

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,46 @@
...
+      })
+      .positional('upstream', {
+        description: 'URL to a target server.',
+        type: 'string',
+      })
+      .coerce('upstream', value => {
+        try {
+          return new URL(value);
+        } catch (e) {
+          throw new Error(`Invalid upstream URL provided: ${value}`);
```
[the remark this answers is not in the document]
XVincentX: I've tried this in the past already but there's no really value in the error message.

Here's its content:

`Invalid URL: asidjasiod`

test-harness/specs/proxy/proxy.validation.httpResponse.txt
```diff
@@ -0,0 +1,34 @@
+====test====
+Receiving a response we cannot serialize but it's a string gets passed through
```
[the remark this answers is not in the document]
XVincentX: Fixed!

CONTRIBUTING.md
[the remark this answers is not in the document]
XVincentX: I'd wait TypeScript 3.7 which should be fixing this problem and be out very soon. If not, we'll add an automatic command.

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+curl -v -s http://localhost:4010/pets/10 > /dev/null
+
+< HTTP/1.1 404 Not Found
+< warning: body should be object
+```
+
+You can see there's a `warning` header which is a JSON object with all the violations found in the response.
+
+```bash
+prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2 --report-violations httpResponse
```
philsturgeon:
```suggestion
prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2 --errors
```

packages/cli/src/commands/proxy.ts
```diff
@@ -0,0 +1,46 @@
...
+    const p: CreateProxyServerOptions = pick(
+      (parsedArgs as unknown) as CreateProxyServerOptions,
+      'dynamic',
+      'cors',
+      'host',
+      'port',
+      'document',
+      'multiprocess',
+      'upstream',
+      'reportViolations'
```
philsturgeon: `errors`?

packages/cli/src/commands/sharedOptions.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+
+  multiprocess: {
+    alias: 'm',
+    description: 'Forks the http server from the CLI for faster log processing.',
+    boolean: true,
+    default: process.env.NODE_ENV === 'production',
+  },
+
+  'report-violations': {
```
philsturgeon:
```suggestion
  'errors': {
```

packages/cli/src/util/createServer.ts
philsturgeon: Logging is always happening in proxy just like in mocking, and warning is always happening too.

No, switches should be enabling a thing, so adding `--errors` would be turning it on.

XVincentX: So to recap:

1. Always log stuff on the stdout ✅ 
2. Always send the `warning` header in the response ✅ 
3. `--errors` flag will decide whether the response should be returned back or "transformed into a 422 with the error detail"

docs/guides/cli.md
```diff
@@ -61,6 +62,65 @@ Connection: keep-alive
...
+```bash
+prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2 --errors
+… …
+
+curl -v http://localhost:4010/pets/10
+
+< HTTP/1.1 422 Unprocessable Entity
+```
+
+The response body contains the found output violations.
```
philsturgeon: Let’s show them how it looks in case they’re a visual learner and need to see what the JSON looks like in order for this to click.

packages/cli/src/util/runner.ts
```diff
@@ -1,38 +1,40 @@
...
-export function runPrismAndSetupWatcher(createPrism: CreatePrism, options: CreatePrismOptions, spec: string) {
+export function runPrismAndSetupWatcher(createPrism: CreatePrism, options: CreateMockServerOptions) {
   return createPrism(options).then(possiblyServer => {
     if (possiblyServer) {
       let server: IPrismHttpServer = possiblyServer;
 
-      const watcher = chokidar.watch(spec, {
-        persistent: false,
+      const watcher = chokidar.watch(options.document, {
+        persistent: os.platform() === 'darwin',
```
philsturgeon: Can we get a comment here explaining what or why. Is this a temporary thing, is there a GitHub issue explaining a bug that caused this workaround, or documentation somewhere about why we are doing this? The reasoning could be forgotten over time.

XVincentX: It's basically an OS-specific flag that we need to explicitly set, unfortunately. See https://github.com/paulmillr/chokidar#persistence

I'll add a comment.

docs/guides/cli.md
XVincentX: Done

XVincentX:
@philsturgeon @StefanDywersant @lag-of-death 

Please review these changes: https://github.com/stoplightio/prism/compare/92c6b21675cbaf814d074e710de9dd01bb25357c...35d3a1eefb8ba72034675cac5ac5d80078253c00

* Back to `sl-violations` header, since the `warning` is deprecated and probably removed soon
* Back to JSON serialised object in the `sl-violations` header; we'll find out a different format in a following story
* Modified CLI docs — added a violation example
* Renamed flag to boolean `--errors`, defaulted to false; violations will *always* be logged and put as a header. 
* Added chokidar OS-specific comment
* Both response and request violations are sent to the output, and the path is prefixed accordingly with `request/response`
* Modified harness tests and added a new one for warnings

docs/guides/cli.md
```diff
@@ -61,6 +62,66 @@ Connection: keep-alive
...
+The output violations will be reported on the standard output and as a response header
+
+```bash
+prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2
+… …
+
+curl -v -s http://localhost:4010/pets/10 > /dev/null
+
+< HTTP/1.1 404 Not Found
+< sl-violations: body should be object
```
lag-of-death: shouldn't it be a json?

XVincentX: 211da2f

docs/guides/cli.md
```diff
@@ -61,6 +62,67 @@ Connection: keep-alive
...
+In case you want to return an error HTTP response in case there's an `Error` in the request or the response, you can use the `--errors` flag.
+
+```bash
+prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2 --errors
+… …
+
+curl -v -X POST http://localhost:4010/pet/
+
+< HTTP/1.1 422 Unprocessable Entity
+{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request/response is not valid and the --errors flag is set, so Prism is generating this error for you.","validation":[{"location":["request"],"severity":"Error","code":401,"message":"Invalid security scheme used"}]}
```
XVincentX: @philsturgeon I'm starting to think that 422 is not a good status code, since this error can either be because of an invalid request but also because of an invalid response. Any idea?

docs/guides/cli.md
```diff
@@ -94,7 +94,7 @@ prism proxy examples/petstore.oas2.yaml https://petstore.swagger.io/v2
 [CLI] ●  note      GET        http://127.0.0.1:4010/pets/10
 ```
 
-The output violations will be reported on the standard output and as a response header
+The output violations will be reported on the standard output and as a response header (`sl-violations`)
```
lag-of-death: isn't it or?

packages/cli/src/commands/sharedOptions.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  multiprocess: {
+    alias: 'm',
+    description: 'Forks the http server from the CLI for faster log processing.',
+    boolean: true,
+    default: process.env.NODE_ENV === 'production',
+  },
+
+  errors: {
+    description: 'Specifies whether response violations will be sent as an error response',
```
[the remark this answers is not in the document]
XVincentX: @lag-of-death 

1. All the violations will *always* be put in the stdout (log on the console)
2. All the violations will *always* be put in the `sl-violations` header
3. That flag says "if among all the violations there's one that's an error, then return an error response instead of the original one"

Does this clarify?

docs/guides/cli.md
XVincentX: ?

packages/cli/src/commands/sharedOptions.ts
lag-of-death:
```suggestion
    description: 'if among all the violations there's one that's an error, then return an error response instead of the original one',
```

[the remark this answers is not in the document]
lag-of-death: yes:) https://github.com/stoplightio/prism/pull/669#discussion_r338538343

packages/cli/src/util/createServer.ts
```diff
@@ -34,7 +36,7 @@ export async function createMultiProcessPrism(options: CreatePrismOptions) {
   }
 }
 
-export async function createSingleProcessPrism(options: CreatePrismOptions) {
+async function createSingleProcessPrism(options: CreateBaseServerOptions) {
```
lag-of-death:
```suggestion
async function createBaseServer(options: CreateBaseServerOptions) {
```

packages/core/src/factory.ts
```diff
@@ -1,65 +1,68 @@
...
 import { getOrElse, map } from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/lib/pipeable';
 import * as TaskEither from 'fp-ts/lib/TaskEither';
 import { defaults } from 'lodash';
-import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, ProblemJsonError } from './types';
+import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, IPrismProxyConfig } from './types';
 import { validateSecurity } from './utils/security';
 
+function isProxyConfig(p: IPrismConfig): p is IPrismProxyConfig {
+  return !p.mock;
```
lag-of-death: hm.. maybe worth adding?
```suggestion
  return !p.mock && p.proxy;
```

packages/core/src/factory.ts
```diff
@@ -1,65 +1,68 @@
...
-              }),
+              })
             );
           }
 
-          const inputValidationResult = config.checkSecurity
-            ? inputValidations.concat(
-                pipe(
+          config.checkSecurity
+            ? inputValidations.push(
```
lag-of-death: so the mutation was needed so that https://github.com/stoplightio/prism/pull/669/files#diff-47c5dc2d65fd624c869f5f08d0cfb56aL73 have it or there was some other reason?

packages/http-server/src/server.ts
```diff
@@ -139,3 +149,15 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
   };
   return prismServer;
 };
+
+const createErrorObjectWithPrefix = (locationPrefix: string) => (detail: IPrismDiagnostic) => {
+  const location = detail.path || [];
+  location.unshift(locationPrefix);
```
lag-of-death: just a suggestion: `[locationPrefix].concat(location)`

packages/core/src/factory.ts
XVincentX: proxy is a property that does not exist

packages/http-server/src/server.ts
```diff
@@ -80,23 +71,42 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
 
       const { output } = response;
 
-      reply.code(output.statusCode);
-
-      if (output.headers) {
-        reply.headers(output.headers);
+      const inputOutputValidationErrors = response.validations.output
```
lag-of-death: is it `input` and `output` or only `output`?

XVincentX: It's both. You can see they're concatenated.

packages/core/src/factory.ts
XVincentX: Yeah; do not worry about this so much because the validation refactor #731 will remove all this bad stuff.

packages/core/src/factory.ts
lag-of-death: @XVincentX `upstream` then?

XVincentX: We can add it effectively, something like `return !p.mock && 'upstream' in p;` but @StefanDywersant made me remove it :trollface:

Let's keep it this way for now.

packages/core/src/types.ts
```diff
@@ -3,7 +3,7 @@ import { Either } from 'fp-ts/lib/Either';
 import { ReaderEither } from 'fp-ts/lib/ReaderEither';
 import { TaskEither } from 'fp-ts/lib/TaskEither';
 import { Logger } from 'pino';
-export type IPrismDiagnostic = Omit<IDiagnostic, 'range'>;
+export type IPrismDiagnostic = Omit<IDiagnostic, 'range' | 'path'> & { path?: string[] };
```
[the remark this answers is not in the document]
XVincentX: This is the original definition. I'm just redefining it and narrowing to be only a `string[]` and not `string or number[]` — which is really weird.

I think I did a mistake back in the time to reuse the `IDiagnostic` definition; apparently it does not make that much sense for us.

We'll revisit this soon.

docs/guides/cli.md
philsturgeon: Ahh yeah, so if we have a violation on a response it can be a 500, as the client is not at fault.

docs/guides/cli.md
```diff
@@ -61,6 +62,67 @@ Connection: keep-alive
...
+… …
+
+curl -v -s http://localhost:4010/pet/10 > /dev/null
+
+< sl-violations: [{"location":["request"],"severity":"Error","code":401,"message":"Invalid security scheme used"}]
+```
+
+You can see there's a `sl-violations` header which is a JSON object with all the violations found in the response.
+
+In case you want to return an error HTTP response in case there's an `Error` in the request or the response, you can use the `--errors` flag.
```
philsturgeon:
```suggestion
The header is a handy way to see contract mismatches or incorrect usage in a way that doesn't block the client, so you can monitor all/some production traffic this way record the problems. If you want Prism to make violations considerably more clear, run the proxy command with the `--errors` flag. This will turn any request or response violation into a [RFC 7807 HTTP Problem Details Error](https://tools.ietf.org/html/rfc7807) just like validation errors on the mock server.
```

docs/guides/cli.md
philsturgeon: I think my comment got lost somewhere but agree, response violations = 500

test-harness/specs/proxy/proxy.validation.httpResponse.txt
```diff
@@ -0,0 +1,34 @@
...
+                type: number
+                default: 0
+====server====
+proxy -p 4010 ${document} http://httpbin --errors
+====command====
+curl -i http://localhost:4010/json
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+
+{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request/response is not valid and the --errors flag is set, so Prism is generating this error for you.","validation":[{"location":["response","body"],"severity":"Error","code":"required","message":"should have required property 'title'"},{"location":["response","body"],"severity":"Error","code":"required","message":"should have required property 'description'"}]}
```
philsturgeon: Have we got a test somewhere for more complex cases like a value not in an enum? We have had trouble with things like that in the past and I'd like to make sure its well covered.

test-harness/specs/violations/print-as-http-header.txt
```diff
@@ -0,0 +1,27 @@
...
+                  value: 123
+              schema:
+                  type: object
+====server====
+mock -p 4010 ${document}
+====command====
+curl -i http://localhost:4010/path
+====expect====
+HTTP/1.1 200 OK
+sl-violations: [{"location":["response","body"],"severity":"Error","code":"type","message":"should be object"}]
```
philsturgeon: I think you explained but I forgot what "code: type" is

docs/guides/cli.md
XVincentX: 500 it is.

test-harness/specs/violations/print-as-http-header.txt
XVincentX: This is just the ajv validation object's `keyword` forwarded here. Not so great — but the idea is to fix it back when we'll find our own format.

test-harness/specs/proxy/proxy.validation.httpResponse.txt
XVincentX: Added

PR #671, opened 2019-10-03
feat: generating path examples

### LINKED ISSUE #505: Use path parameter example/default when displaying list of operations
**User story.**
As a CLI and Studio users, I want the URLs with path parameters to be instantly usable if possible so that I can click them, or copy them into a HTTP client.

Currently we show URLs like this:

In studio the current implementation is repeating the path with param templates showing up, then the full URL with the same stuff.

We should use example values, default, the first item from an enum, or if x-faker is there we can use a value from that. The goal here is to make the URL useful, without somebody having to copy and paste it, only to tweak bits and probably put an example value in there themselves.

This logic should go into the Prism JS API itself, so studio and Prism CLI do not both need to figure out how to do it.

StefanDywersant:
Closes #505 

## Checklist

Feature

Prism CLI lists available operation paths to the console. This feature adds example parameter values to those paths. This both path and query parameters.

No


packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
...
+import * as Either from 'fp-ts/lib/Either';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { JSONSchema } from '../../../types';
+
+function serializeAndImplode(separator: string, name: string, value: Array<string | number | boolean>) {
+  return (
+    encodeURIComponent(name) +
+    '=' +
+    value
+      .map(String)
```
XVincentX: This is clever, I didn't know this trick.

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
+import { HttpParamStyles } from '@stoplight/types';
+import * as Either from 'fp-ts/lib/Either';
```
XVincentX: I do not really know if the serialiser should be in the http package; this is not going to be used here so I think maybe it makes sense to leave it in the CLI?

packages/http/src/mocker/serializer/style/deepObject.ts
```diff
@@ -0,0 +1,19 @@
...
+    return Object.keys(value)
+      .map(key => serialize(name, [...path, isPositiveInteger(key) ? '' : key], value[key]))
+      .join('&');
+  } else {
+    return `${name}${path.map(key => `[${key}]`).join('')}=${value}`;
+  }
+}
+
+function isPositiveInteger(str: string) {
+  return /^\+?\d+$/.test(str) && parseInt(str) >= 0;
```
XVincentX: Do we need the regex check here?

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -0,0 +1,189 @@
+import { assertLeft, assertRight } from '@stoplight/prism-http/src/__tests__/utils';
+import { HttpParamStyles } from '@stoplight/types/dist';
```
XVincentX:
```suggestion
import { HttpParamStyles } from '@stoplight/types';
```

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    it('fails when not supported style supplied', () => {
+      assertLeft(
+        createExamplePath({
+          id: '123',
+          path: '/path/{p}',
+          method: 'get',
+          request: {
+            path: [
+              { name: 'p', style: HttpParamStyles.PipeDelimited as any, examples: [{ key: 'foo', value: 'test' }] },
```
XVincentX: The good thing of TypeScript is that it will avoid us to make some troubles and so we can avoid both the runtime checks and testing for such conditions. I'd remove this test completely and the runtime checks in the function as well.

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    describe('schema is present', () => {
+      it('generates example from schema', () => {
+        assertSome(
+          generate({
+            name: 'a',
+            style: HttpParamStyles.Form,
+            schema: { type: 'string' },
+          }),
+          v => expect(v).toEqual(expect.any(String)),
```
XVincentX: Can we use the `format` schema parameter so we can narrow down the generated string so we can check out the schema is being used for real?

packages/http/src/mocker/serializer/style/delimited.ts
StefanDywersant: I got two args for leaving it here:
- it fits by domain here
- I feel that at some point we will create a lib with those serializers / deserializers, so it will be easier to extract (a folk doing it might not even know that there is something in cli package)

I'd still vote on leaving it here.

packages/http/src/mocker/serializer/style/deepObject.ts
StefanDywersant: Actually it looks like the shortest way of checking for a positive integer. Other ways include casting, checking for (-)Infinity, NaNs, dirty stuff basically.

StefanDywersant: Note: we are looking for a positive integer inside a string.

packages/cli/src/util/__tests__/paths.spec.ts
StefanDywersant: Crap :]

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+  IHttpQueryParam,
+  INodeExample,
+  INodeExternalExample,
+} from '@stoplight/types';
+import * as Either from 'fp-ts/lib/Either';
+import * as Option from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get } from 'lodash';
+// @ts-ignore
+const { parse } = require('uri-template');
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/createServer.ts
```diff
@@ -3,11 +3,14 @@ import { createServer as createHttpServer } from '@stoplight/prism-http-server';
 import { IHttpOperation } from '@stoplight/types';
 import chalk from 'chalk';
 import * as cluster from 'cluster';
+import * as Either from 'fp-ts/lib/Either';
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/__tests__/__snapshots__/paths.spec.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
...
+    Either.fold(
+      e => {
+        throw e;
+      },
+      serialized => serialized,
+    ),
+  );
+}
+
+export const serializeWithCommaDelimitedStyle = serializeWithDelimitedStyle.bind(undefined, ',');
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+      default:
+        values[spec.name] = value;
+    }
+    return values;
+  }, {});
+}
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    Either.tryCatch(() => {
```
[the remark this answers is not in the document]
StefanDywersant: You are completely right. With a fresh look today I believe I found concise way for turning all remaining functions to return Monads.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+      const specs = get(operation, 'request.path', []);
+      return { template: createPathUriTemplate(operation.path, specs), values: generateParamValues(specs) };
+    }, Either.toError),
+    Either.chain(({ template, values }) => {
+      const specs = get(operation, 'request.query', []);
+      try {
+        return Either.right({
```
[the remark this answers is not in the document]
StefanDywersant: Refactored 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (restParams) {
+    path += `${formExplodedParams || formImplodedParams ? '&' : '?'}${restParams}`;
+  }
+
+  return path;
+}
+
+function createParamUriTemplate(name: string, style: HttpParamStyles, explode: boolean) {
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+  const starOrVoid = explode ? '*' : '';
+  switch (style) {
+    case HttpParamStyles.Simple:
+      return `{${name}${starOrVoid}}`;
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    Either.tryCatch(() => {
+      const specs = get(operation, 'request.path', []);
+      return { template: createPathUriTemplate(operation.path, specs), values: generateParamValues(specs) };
+    }, Either.toError),
+    Either.chain(({ template, values }) => {
+      const specs = get(operation, 'request.query', []);
```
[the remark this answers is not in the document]
StefanDywersant: I'd leave specs as lodash getter with a default value. There is no logic that needs to be invoked when `.query` is empty besides returning a default value.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,139 @@
...
+  const starOrVoid = explode ? '*' : '';
+  switch (style) {
+    case HttpParamStyles.Simple:
+      return `{${name}${starOrVoid}}`;
+    case HttpParamStyles.Label:
+      return `{.${name}${starOrVoid}}`;
+    case HttpParamStyles.Matrix:
+      return `{;${name}${starOrVoid}}`;
+    default:
+      throw new Error(`Unsupported parameter style: ${style}`);
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
...
+  return pipe(
+    Either.right(value),
+    Either.chain(
+      v =>
+        Array.isArray(v)
+          ? Either.right(v)
+          : Either.left(new Error('Space/pipe/comma.. delimited style is only applicable to array parameter')),
+    ),
+    Either.map(() => (explode ? serializeAndExplode(name, value) : serializeAndImplode(separator, name, value))),
+    Either.fold(
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
...
+export function serializeWithDelimitedStyle(
+  separator: string,
+  name: string,
+  value: Array<string | number | boolean>,
+  explode?: boolean,
+) {
+  return pipe(
+    Either.right(value),
+    Either.chain(
+      v =>
```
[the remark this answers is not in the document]
StefanDywersant: Fixed 👍

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,50 @@
...
+        Array.isArray(v)
```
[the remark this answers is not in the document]
StefanDywersant: I was a bit defensive here. Basically value in some cases is directly taken from spec's (api description's :> ) examples. That as we know not necessarily is an array. Our types also are vanished after compilation.

Now the question: where we should defend against bad input data? On one hand, it looks scary if we verify arg even if it has defined type. On the other hand, if we don't it is pretty easy to generate non-understandable errors. WDYT?

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,31 @@
...
+import { generate as generateDynamicExample } from './JSONSchema';
+
+function isExternalExample(example: INodeExample | INodeExternalExample): example is INodeExternalExample {
+  return !!(example as INodeExternalExample).externalValue;
+}
+
+function generateStaticExample(examples: Array<INodeExample | INodeExternalExample>) {
+  return pipe(
+    Option.fromNullable(examples),
+    Option.chain(exs => Option.fromNullable(exs[Math.floor(Math.random() * exs.length)])),
```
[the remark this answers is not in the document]
StefanDywersant: Yes, `.examples` can be either not-defined (L+14) or an empty array (L+15).

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,31 @@
...
+    Option.map(example => (isExternalExample(example) ? example.externalValue : example.value)),
```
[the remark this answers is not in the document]
StefanDywersant: Good point, I wasn't aware of what IHttpExternalExample actually is.

I changed this part so it returns `.value` or `Option.none`. The latter, as you say will never happen. Will see :P

packages/http/src/mocker/serializer/style/delimited.ts
XVincentX: Yes effectively I think you made the good choice. Good call 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
+import { generate } from '@stoplight/prism-http/src/mocker/generator/HttpParamGenerator';
+import { serializeWithDeepObjectStyle } from '@stoplight/prism-http/src/mocker/serializer/style/deepObject';
+import {
```
XVincentX: This is likely the reason why the harness tests are failing (well at least we know the test harness is working)

Basically, while in _test_ files it's ok to import from the source files directly, this will not work in production. The reason is that after the TypeScript compilation this becomes `dist` — and such directory does not exist anymore.

If you need stuff from the http-package, just export it explicitly from the `index.ts` file and grab it from there

@StefanDywersant

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+  return pipe(
+    generate(spec),
+    Either.fromOption(() => new Error(`Cannot generate value for: ${spec.name}`)),
+    Either.chain(value => {
+      switch (spec.style) {
+        case HttpParamStyles.DeepObject:
+          return serializeWithDeepObjectStyle(spec.name, value);
+
+        case HttpParamStyles.PipeDelimited:
+          return serializeWithPipeDelimitedStyle(spec.name, value as Array<string | number | boolean>, spec.explode);
```
XVincentX: I think _this_ is the place where we should do the `Array.isArray` and return an error in case it's not working, not in the function; so we can

1. Eliminate the cast
2. Make the internal function care about its job and not having to validate the arguments. Does it make sense @StefanDywersant?

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+      return pipe(
+        createPathUriTemplate(operation.path, specs),
+        Either.map(template => ({ template, values })),
+      );
+    }),
+  );
+}
+
+function generateTemplateAndValuesForQueryParams(template: string, operation: IHttpOperation) {
+  const specs = get(operation, 'request.query', []);
```
XVincentX: This is up to you, but I usually prefer make this function accept a `QueryParameter[]` directly instead of getting the whole `IHttpOperation`, get the value from the property and then use it.

Up to you though, it's not blocking anything.

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,64 @@
+import { IHttpOperation, IHttpParam, IHttpPathParam } from '@stoplight/types';
+import { INodeExample, INodeExternalExample } from '@stoplight/types';
+import * as Array from 'fp-ts/lib/Array';
+import * as Option from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { JSONSchema } from '../../types';
+import { generate as generateDynamicExample } from './JSONSchema';
+
+function improveSchema(schema: JSONSchema) {
+  const newSchema = { ...schema };
```
XVincentX: Maybe let's use deepClone by lodash?

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (newSchema.type === 'integer') {
```
XVincentX: This part of code is valid for integers as well

```suggestion
  if (newSchema.type === 'integer' || newSchema.type === 'number') {
```

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,64 @@
...
+      newSchema.minimum = 1;
+    }
+
+    if (!newSchema.maximum) {
+      newSchema.maximum = 1000;
+    }
+  }
+
+  if (newSchema.type === 'string') {
+    // @todo: use faker to generate single-worded string
```
XVincentX: Do it! :trollface:

packages/http/src/mocker/generator/HttpParamGenerator.ts
StefanDywersant: Actually it is cloned anyway by this recurrence so cloning it again would be showing no respect for CPU cycles.

packages/http/src/mocker/generator/HttpParamGenerator.ts
StefanDywersant: Done 👍

packages/http/src/mocker/generator/HttpParamGenerator.ts
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+// @ts-ignore
+import { parse } from 'uri-template';
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    generateTemplateAndValuesForPathParams(operation),
+    Either.chain(({ template: pTemplate, values: pValues }) => {
+      return pipe(
+        generateTemplateAndValuesForQueryParams(pTemplate, operation),
+        Either.map(({ template: qTemplate, values: qValues }) => {
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+import { get } from 'lodash';
+// @ts-ignore
+import { parse } from 'uri-template';
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    generateTemplateAndValuesForPathParams(operation),
+    Either.chain(({ template: pTemplate, values: pValues }) => {
+      return pipe(
+        generateTemplateAndValuesForQueryParams(pTemplate, operation),
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+        Either.map(({ template: qTemplate, values: qValues }) => {
+          return { template: qTemplate, values: { ...pValues, ...qValues } };
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+import * as Either from 'fp-ts/lib/Either';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get } from 'lodash';
+// @ts-ignore
+import { parse } from 'uri-template';
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    generateTemplateAndValuesForPathParams(operation),
+    Either.chain(({ template: pTemplate, values: pValues }) => {
```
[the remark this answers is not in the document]
StefanDywersant: 👍

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Support for encoding > allowReserved flag when validating application/x-www-form-urlencoded body [#630](https://github.com/stoplightio/prism/pull/630)
 - Validating output status code against available response specs [#648](https://github.com/stoplightio/prism/pull/648)
 - Support for Contract Testing [#650](https://github.com/stoplightio/prism/pull/650)
+- Example parameters in operations list [#671](https://github.com/stoplightio/prism/pull/671)
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,27 @@
+import { IHttpOperation, IHttpParam, IHttpPathParam } from '@stoplight/types';
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,27 @@
...
+import { INodeExample, INodeExternalExample } from '@stoplight/types';
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as Array from 'fp-ts/lib/Array';
```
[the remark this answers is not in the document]
StefanDywersant: Somehow my PhpStorm always notices unused imports and now it stopped. Php sucks.

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as Option from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { generate as generateDynamicExample } from './JSONSchema';
+
+function generateStaticExample(examples: Array<INodeExample | INodeExternalExample>) {
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+    .filter(spec => spec.required !== false)
+    .filter(spec => spec.explode)
+    .map(spec => spec.name)
+    .join(',');
+  const formImplodedParams = formSpecs
+    .filter(spec => spec.required !== false)
+    .filter(spec => !spec.explode)
+    .map(spec => spec.name)
+    .join(',');
+  const restParams = specs
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,168 @@
...
+
+function createQueryUriTemplate(path: string, specs: IHttpQueryParam[]) {
+  // defaults for query: style=Form exploded=false
+  const formSpecs = specs.filter(spec => (spec.style || HttpParamStyles.Form) === HttpParamStyles.Form);
+  const formExplodedParams = formSpecs
+    .filter(spec => spec.required !== false)
+    .filter(spec => spec.explode)
+    .map(spec => spec.name)
+    .join(',');
+  const formImplodedParams = formSpecs
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/serializer/style/deepObject.ts
```diff
@@ -0,0 +1,20 @@
+import { HttpParamStyles } from '@stoplight/types';
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/serializer/style/deepObject.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as Either from 'fp-ts/lib/Either';
+
+export function serializeWithDeepObjectStyle(name: string, value: any) {
+  return Either.right(serialize(name, [], value));
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,40 @@
...
+): Either.Either<Error, string> {
+  return pipe(
+    Array.isArray(value)
+      ? Either.right(value)
+      : Either.left(new Error('Space/pipe/comma delimited style is only applicable to array parameter')),
+    Either.map(v => (explode ? serializeAndExplode(name, v) : serializeAndImplode(separator, name, v))),
+  );
+}
+
+export const serializeWithCommaDelimitedStyle = partial(serializeWithDelimitedStyle, ',');
```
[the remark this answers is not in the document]
StefanDywersant: CommaDelimitedStyle is one of our supported styles. I added that function in order to increase PR size ;)

packages/cli/src/util/paths.ts
StefanDywersant: Yeah, you completely right. Checking for it inside the serializer was a bad decision. Fixed.

packages/cli/src/util/paths.ts
StefanDywersant: It is like that just for `createExamplePath` (entry point) to look more balanced. I'd vote on leaving it like that instead of moving that dirty lodash's get to our entry point.

packages/cli/src/util/paths.ts
StefanDywersant: @XVincentX Http's `index.ts` looks pretty concise. Only `createInstance` is actually exported. That is another vote on moving serializers to CLI which makes that solution win the election. WDYT?

packages/cli/src/util/paths.ts
StefanDywersant: Forget ^. I exported it from http package.

packages/cli/src/util/paths.ts
StefanDywersant: Forget ^. I exported it from http package.

packages/http/src/mocker/generator/HttpParamGenerator.ts
XVincentX: I was suggesting to use clone instead of the ... but your solution is just ok as it is!

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { assertNone, assertSome } from '@stoplight/prism-core/src/utils/__tests__/utils';
+import { HttpParamStyles } from '@stoplight/types/dist';
```
XVincentX:
```suggestion
import { HttpParamStyles } from '@stoplight/types';
```

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
XVincentX: Yes, I'd really say let's use the format keyword and try to generate an email or something.

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,32 @@
+import { partial } from 'lodash';
+
+function serializeAndImplode(separator: string, name: string, value: Array<string | number | boolean>) {
```
XVincentX: We actually do not care of the array type.
```suggestion
function serializeAndImplode(separator: string, name: string, value: Array<unknown>) {
```

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,32 @@
...
+    encodeURIComponent(name) +
+    '=' +
+    value
+      .map(String)
+      .map(encodeURIComponent)
+      .join(separator)
+  );
+}
+
+function serializeAndExplode(name: string, value: Array<string | number | boolean>) {
```
XVincentX:
```suggestion
function serializeAndExplode(name: string, value: Array<unknown>) {
```

packages/http/src/mocker/serializer/style/delimited.ts
```diff
@@ -0,0 +1,32 @@
...
+  return value
+    .map(String)
+    .map(v => `${encodeURIComponent(name)}=${encodeURIComponent(v)}`)
+    .join('&');
+}
+
+export function serializeWithDelimitedStyle(
+  separator: string,
+  name: string,
+  value: Array<string | number | boolean>,
```
XVincentX:
```suggestion
  value: Array<unknown>,
```

packages/http/src/mocker/serializer/style/delimited.ts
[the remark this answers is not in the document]
XVincentX: Yeah it's not a big deal, I was just wondering if this was a leftover.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,179 @@
...
+          return Either.right(serializeWithDeepObjectStyle(spec.name, value));
+
+        case HttpParamStyles.PipeDelimited:
+          if (Array.isArray(value)) {
+            return Either.right(
+              serializeWithPipeDelimitedStyle(spec.name, value as Array<string | number | boolean>, spec.explode),
+            );
+          } else {
+            return Either.left(new Error('Pipe delimited style is only applicable to array parameter'));
+          }
```
XVincentX: For these case you can use the utility function `Either.fromPredicate` which will do all the `if` and else for you and makes the code way more readable:

```ts
return pipe(
  value,
  Either.fromPredicate(
    Array.isArray,
    () => new Error('Pipe delimited style is only applicable to array parameter'),
  ),
  Either.map(v => serializeWithPipeDelimitedStyle(spec.name, v, spec.explode)),
);

```

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        case HttpParamStyles.SpaceDelimited:
+          if (Array.isArray(value)) {
+            return Either.right(
+              serializeWithSpaceDelimitedStyle(spec.name, value as Array<string | number | boolean>, spec.explode),
+            );
+          } else {
+            return Either.left(new Error('Space delimited style is only applicable to array parameter'));
+          }
```
XVincentX: ditto, you can change this with

```ts
return pipe(
  value,
  Either.fromPredicate(
    Array.isArray,
    () => new Error('Pipe delimited style is only applicable to array parameter'),
  ),
  Either.map(v => serializeWithPipeDelimitedStyle(spec.name, v, spec.explode)),
);
```

You can also see that this code is very similar to the one handling the other case; if we could DRY it and reuse that's be even better.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,179 @@
+import { generateHttpParam } from '@stoplight/prism-http';
+import { serializeWithDeepObjectStyle } from '@stoplight/prism-http';
+import { serializeWithPipeDelimitedStyle, serializeWithSpaceDelimitedStyle } from '@stoplight/prism-http';
```
XVincentX: You can group all these imports together.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,179 @@
...
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get } from 'lodash';
+// @ts-ignore
+import { parse } from 'uri-template';
+
+export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
+  return pipe(
+    generateTemplateAndValuesForPathParams(operation),
+    Either.chain(({ template: pathTemplate, values: pathValues }) => {
+      return pipe(
```
XVincentX: Nitpick (you can ignore this) if you're just returning pipe, use a single statement, it improves the readability a lot.

packages/cli/src/util/paths.ts
```diff
@@ -0,0 +1,179 @@
...
+    Either.fromOption(() => new Error(`Cannot generate value for: ${spec.name}`)),
+    Either.chain(value => {
+      switch (spec.style) {
+        case HttpParamStyles.DeepObject:
+          return Either.right(serializeWithDeepObjectStyle(spec.name, value));
+
+        case HttpParamStyles.PipeDelimited:
+          if (Array.isArray(value)) {
+            return Either.right(
+              serializeWithPipeDelimitedStyle(spec.name, value as Array<string | number | boolean>, spec.explode),
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/mocker/generator/__tests__/HttpParamGenerator.spec.ts
StefanDywersant: 👍

packages/http/src/mocker/serializer/style/delimited.ts
StefanDywersant: Why more permissive `unknown`?

XVincentX: Actually is way _less_ permissive — unknown is basically a "we have no idea, so we won't make any assumption" and will prevent anybody inside the other functions to try to do some stuff (such as concatenate or add them)

packages/cli/src/util/paths.ts
StefanDywersant: 👍

packages/cli/src/util/paths.ts
StefanDywersant: There are two differences between them. Error message and called function. I'd say it would be overengineering to make a common function for that and definitely would decrease readability.

XVincentX: I'm ok keeping it in this way.

packages/cli/src/util/paths.ts
StefanDywersant: Learned `Either.fromPredicate` 👍 Now its only 999 tiny helper functions to learn left ;)

packages/http/src/mocker/serializer/style/delimited.ts
StefanDywersant: Yeah, but it will allow objects or arrays which we don't want.

XVincentX: Oh gosh, you're right. You can ignore this, thanks for spotting it.

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!newSchema.minimum) {
+      newSchema.minimum = 1;
+    }
+
+    if (!newSchema.maximum) {
+      newSchema.maximum = 1000;
+    }
+  }
+
+  if (newSchema.type === 'string') {
```
XVincentX:
```suggestion
  else if (newSchema.type === 'string') {
```

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!newSchema.format && !newSchema.enum) {
+      newSchema['x-faker'] = 'random.word';
+    }
+  }
+
+  if (newSchema.type === 'object') {
```
XVincentX:
```suggestion
  else if (newSchema.type === 'object') {
```

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (newSchema.properties) {
+      newSchema.properties = Object.entries(newSchema.properties).reduce((r, [k, v]) => {
+        r[k] = improveSchema(v);
+        return r;
+      }, {});
+    }
+  }
+
+  if (newSchema.type === 'array') {
```
XVincentX:
```suggestion
  else if (newSchema.type === 'array') {
```

PR #674, opened 2019-10-07
Check all markdown links on PRs

XVincentX:
Add markdown links checker so we do not give broken links to people and we know if any of these die along the way.


philsturgeon:
Is this one confirmed failing builds for bad links? I addedan action for studio and it does not work. https://github.com/stoplightio/studio/

XVincentX:
It does. Thanks to this I indeed fixed a lot of docs dead links. @philsturgeon

PR #685, opened 2019-10-07
[POC] refactor: validators

lag-of-death:
Just a POC for validators refactoring.


lag-of-death:
@XVincentX, I've put body serializer out of body validator. That body serializer is now higher - in TaskEither chain. Input validator returns Either with a semigroup as its left value. ATM, only body serializer/validator is used (for POC purposes). I've attached comments to the code, hope I didn't obfuscate things. LMK, if this needs explaining. Happy to hear your thoughts :+1:

lag-of-death:
@XVincentX, any feedback appreciated greatly! :)

packages/http/src/validator/index.ts
```diff
@@ -1,62 +1,174 @@
...
+    sequenceT(getValidation(getSemigroup<IPrismDiagnostic>()))(
+      reqBodyValidation,
+      bodyValidation,
+      headersValidation,
+      queryValidation,
+    ),
+    map(() => []),
+  );
+
+  console.log('violations', violations);
```
StefanDywersant: Remove

packages/core/src/types.ts
```diff
@@ -1,8 +1,12 @@
 import { IDiagnostic } from '@stoplight/types';
+import { IHttpOperationRequest, IHttpOperationResponse, IMediaTypeContent } from "@stoplight/types";
 import { Either } from 'fp-ts/lib/Either';
 import { ReaderEither } from 'fp-ts/lib/ReaderEither';
 import { TaskEither } from 'fp-ts/lib/TaskEither';
+import { JSONSchema4, JSONSchema6, JSONSchema7 } from "json-schema";
```
XVincentX: We have our own type that already has all three schemas in one thing.

packages/core/src/types.ts
```diff
@@ -16,12 +20,31 @@ export interface IPrismConfig {
   validateResponse: boolean;
 }
 
-export type ValidatorFn<Resource, T> = (opts: { resource: Resource; element: T }) => IPrismDiagnostic[];
+export type ValidatorFn<Resource, T> = (opts: { resource?: Resource; element: T }) => IPrismDiagnostic[];
```
XVincentX: If we're validating, both the resource to validate against and the element must be present. What'd be the use case for `resource` to be undefined?

XVincentX:
Thanks for the hard work. I'll take over on this.

PR #689, opened 2019-10-08
feat: auto reload a Prism instance when the doc changes

### LINKED ISSUE #673: Reload Mock Server on Change
**User story.**
As a developer, I can use my most up-to-date document for mocking without restarting the server, so that I can get quick feedback on my changes and work through changes iteratively.

When working with developers new to Prism or API mocking in general, I've found they expect the server to automatically use their changes. However, they have to restart the server. Loading new changes automatically is something they expect.

It also comes up later when someone makes a change and forgets to restart.

I'd like to see the mock server reloaded automatically. I'd like for the terminal to show it's been reloaded. We can use something like `nodemon` to handle this, but that's not what people are expecting.

lag-of-death:
## Checklist

_feature_

HTTP server that Prism uses is now shut down and restarted when the doc it was loaded with is changed. 

Closes #673


lag-of-death:
@XVincentX, this implementation:
- was confined to the CLI
- used chokidar
- allows for reacting only on **changes** in the document a Prism instance was loaded with
- was tested against both the binary and against `run cli` "mode" (and for both, single and multi-process Prism was tested)

I haven't added a changelog line nor did I update the docs. Happy to hear your thoughts

docs/guides/cli.md
```diff
@@ -35,6 +35,12 @@ Connection: keep-alive
 
 Responses will be mocked using realistic data that conforms to the type in the description.
 
+### Auto-reloading
+
+Prism watches for changes made to a document it was loaded with.
+When they happen, Prism restarts its HTTP server to reflect changes to operations.
```
XVincentX: Can we also explicit what happens if the target (modified) document is incorrect?

packages/cli/src/index.ts
```diff
@@ -3,7 +3,8 @@
 import * as yargs from 'yargs';
 import mockCommand from './commands/mock';
 
-const _v = yargs
+// tslint:disable-next-line:no-unused-expression
```
XVincentX: I'd rather leave it, if you do not mind.

packages/cli/src/util/createServer.ts
```diff
@@ -48,7 +50,39 @@ export async function createSingleProcessPrism(options: CreatePrismOptions) {
...
+  const watcher = chokidar.watch(spec);
+
+  let server = await createFastifyServerWithLogger(options, logInstance);
+
+  watcher.on('change', () => {
+    logInstance.start('Restarting Prism…');
+
+    server.fastify
+      .close()
+      .then(() => getHttpOperationsFromResource(spec))
```
XVincentX: I think we should try to load the operations first, so that in case we cannot correctly parse them we can just stop processing this and be like "Yeah something is wrong so we're going to keep the server running with the original file.

packages/cli/src/util/createServer.ts
```diff
@@ -48,7 +50,39 @@ export async function createSingleProcessPrism(options: CreatePrismOptions) {
...
+      .then(operations => {
+        if (operations.length === 0) {
+          logInstance.note('No operations found in the current file. Loading the document from the initial load.');
+
+          return createFastifyServerWithLogger(options, logInstance);
+        } else {
+          return createFastifyServerWithLogger({ ...options, operations }, logInstance);
+        }
+      })
+      .then(newServer => {
```
XVincentX: One line! 
```suggestion
      .then(newServer => server = newServer)
```

packages/cli/src/commands/mock.ts
```diff
@@ -57,15 +57,24 @@ const mockCommand: CommandModule = {
...
+      operations,
+      spec,
+    } = (parsedArgs as unknown) as CreatePrismOptions & {
       multiprocess: boolean;
+      spec: string;
     };
 
     if (multiprocess) {
-      return createMultiProcessPrism({ cors, dynamic, port, host, operations });
+      return createMultiProcessPrism({ cors, dynamic, port, host, operations, spec });
```
XVincentX: Ok I understand why you're sending the spec down here; I think we could probably move the chokidar part in this part of the file so we can simply call this function again?

packages/cli/src/util/runner.ts
```diff
@@ -0,0 +1,38 @@
+import { getHttpOperationsFromResource } from '@stoplight/prism-http';
+import { IPrismHttpServer } from '@stoplight/prism-http-server/src/types';
+import * as chokidar from 'chokidar';
+import { CreatePrismOptions } from './createServer';
+
+export function runPrism(createPrism: Function, options: CreatePrismOptions, spec: string) {
+  return createPrism(options).then((server: IPrismHttpServer) => {
+    if (server) {
+      const watcher = chokidar.watch(spec, { awaitWriteFinish: { stabilityThreshold: 500, pollInterval: 100 } });
```
lag-of-death: `awaitWriteFinish` was needed because sometimes `fs.readFileSync(file, { encoding: 'utf8' })` (https://github.com/stoplightio/prism/blob/dbe60ed1a1b7cd62a6778dbe70d18913c6f8e14d/packages/http/src/getHttpOperations.ts#L24) would give back an empty string, so, most probably the file was being accessed in the middle of it being saved

packages/cli/src/util/runner.ts
```diff
@@ -0,0 +1,38 @@
+import { getHttpOperationsFromResource } from '@stoplight/prism-http';
+import { IPrismHttpServer } from '@stoplight/prism-http-server/src/types';
+import * as chokidar from 'chokidar';
+import { CreatePrismOptions } from './createServer';
+
+export function runPrism(createPrism: Function, options: CreatePrismOptions, spec: string) {
+  return createPrism(options).then((server: IPrismHttpServer) => {
+    if (server) {
```
lag-of-death: this check is needed because `server` is only returned by the fork https://github.com/stoplightio/prism/blob/5c7a6958275f9b838593821654a9348834172223/packages/cli/src/util/createServer.ts#L29

docs/guides/cli.md
XVincentX: FYI — In markdown you need to use two new lines to make an actual new line :trollface: 

This is how the docs look like at the moment: 

image

packages/cli/src/util/runner.ts
XVincentX: We also need to set the `ignoreInitial` to true, otherwise we get a reload event just when we loaded the file for the first time.

```suggestion
      const watcher = chokidar.watch(spec, { ignoreInitial:true, awaitWriteFinish: { stabilityThreshold: 500, pollInterval: 100 } });
```

Or what we could do (and it would be probably more correct) — we keep the flag to false and we remove the first "automatic" loading we're doing in yargs and we treat the first loading as a reload event as well. What do you think @lag-of-death ?

packages/cli/src/util/runner.ts
```diff
@@ -0,0 +1,38 @@
+import { getHttpOperationsFromResource } from '@stoplight/prism-http';
+import { IPrismHttpServer } from '@stoplight/prism-http-server/src/types';
+import * as chokidar from 'chokidar';
+import { CreatePrismOptions } from './createServer';
+
+export function runPrism(createPrism: Function, options: CreatePrismOptions, spec: string) {
```
XVincentX: This function has some inconsistencies. Sometimes you use `async/await` and sometimes you go with the `Promises`. I do not mind which one you prefer, but pick one and keep using it for the whole flow.

packages/cli/src/util/runner.ts
```diff
@@ -0,0 +1,38 @@
...
+  return createPrism(options).then((server: IPrismHttpServer) => {
+    if (server) {
+      const watcher = chokidar.watch(spec, { awaitWriteFinish: { stabilityThreshold: 500, pollInterval: 100 } });
+
+      watcher.on('change', () => {
+        server.fastify.log.info('Restarting Prism...');
```
XVincentX: Do not use the logger in the fastify server — this one shouldn't be touched at all. Just import `signale` in this file and log the things directly.

packages/cli/src/commands/mock.ts
```diff
@@ -57,15 +58,23 @@ const mockCommand: CommandModule = {
...
     };
 
-    if (multiprocess) {
-      return createMultiProcessPrism({ cors, dynamic, port, host, operations });
-    }
+    const createPrism = multiprocess ? createMultiProcessPrism : createSingleProcessPrism;
+    const options = { cors, dynamic, port, host, operations };
 
-    return createSingleProcessPrism({ cors, dynamic, port, host, operations });
+    return runPrism(createPrism, options, spec);
```
XVincentX:
```suggestion
    return runPrismAndSetupWatcher(createPrism, options, spec);
```

or anything that gives the idea of what's going on.

packages/cli/src/util/runner.ts
lag-of-death: @XVincentX, hm.. I think that https://github.com/stoplightio/prism/blob/38f1acfa8f2907a2d41b6d47ad84a48b02054025/packages/cli/src/util/runner.ts#L11 is the answer. `change` is the only event I listen to. I think that's all we need here. In case someone deletes the file, then we don't do a thing and keep the Prism running

docs/guides/cli.md
lag-of-death: It was intended ;> Just added line breaks so that if you read it in a text/code editor, you have them :)

packages/cli/src/util/runner.ts
```diff
@@ -12,25 +12,28 @@ export function runPrismAndSetupWatcher(createPrism: Function, options: CreatePr
...
+          .then(operations => {
             if (operations.length === 0) {
               server.fastify.log.info(
                 'No operations found in the current file, continuing with the previously loaded spec.',
               );
             } else {
-              await server.fastify.close();
+              return server.fastify.close()
+                .then(() => {
+                  server.fastify.log.info('Loading the updated operations...');
```
XVincentX: Let's try to limit the indentation and leverage the chaining

```ts
export function runPrismAndSetupWatcher(createPrism: Function, options: CreatePrismOptions, spec: string) {
  return createPrism(options).then((server: IPrismHttpServer) => {
    if (server) {
      const watcher = chokidar.watch(spec, { awaitWriteFinish: { stabilityThreshold: 500, pollInterval: 100 } });

      watcher.on('change', () => {
        server.fastify.log.info('Restarting Prism...');

        getHttpOperationsFromResource(spec)
          .then(operations => {
            if (operations.length === 0) {
              server.fastify.log.info(
                'No operations found in the current file, continuing with the previously loaded spec.',
              );
            } else {
              return server.fastify.close().then(() => operations);
            }
          }).then(operations => {
            server.fastify.log.info('Loading the updated operations...');

            return createPrism({ ...options, operations });
          })
          .then(newServer => server = newServer)
          .catch(() => {
            server.fastify.log.info('Something went terribly wrong, trying to start Prism with the original document.');

            return server.fastify
              .close()
              .then(() => createPrism(options))
              .catch(() => process.exit(1))
          });
      });
    }
  });
}
```

docs/guides/cli.md
XVincentX: Can we break this in at least two small paragraphs?

packages/cli/src/util/runner.ts
lag-of-death: hm.. @XVincentX, I think this won't work because of:
```
if (operations.length === 0) {
  server.fastify.log.info(
    'No operations found in the current file, continuing with the previously loaded spec.',
  );
}
```
so then, in `}).then(operations => {`, `operations` is undefined

docs/guides/cli.md
lag-of-death: Broke it into two paragraphs. The first one is about the fact Prism now has hot-reloading and the second one is about when Prism is not gonna auto-reload.

packages/cli/src/util/runner.ts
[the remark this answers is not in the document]
lag-of-death: with this solution, I was getting `Object is possibly 'undefined'` in line `server.fastify.log.info('Restarting Prism...');`, I added `CreatePrism` type though

packages/cli/src/util/runner.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type CreatePrism = (options: CreatePrismOptions) => Promise<IPrismHttpServer | undefined>;
+
+export function runPrismAndSetupWatcher(createPrism: CreatePrism, options: CreatePrismOptions, spec: string) {
+  return createPrism(options).then(server => {
+    if (server) {
+      const watcher = chokidar.watch(spec, { awaitWriteFinish: { stabilityThreshold: 500, pollInterval: 100 } });
+
+      watcher.on('change', () => {
+        server!.fastify.log.info('Restarting Prism...');
```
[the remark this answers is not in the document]
XVincentX: This is because the reference could change by the time this even triggers. You can avoid this by using another variable and write

`const s = server;` and use this one

PR #694, opened 2019-10-09
chore: using ref-resolvers from separate package

StefanDywersant:
## Checklist

_You can also link to an open issue here for all the info._

_What changes might users need to make in their application due to this PR?_


packages/http/src/getHttpOperations.ts
```diff
@@ -1,12 +1,25 @@
...
+
+const httpAndFileResolver = new Resolver({
+  resolvers: {
+    https: { resolve: resolveHttp },
+    http: { resolve: resolveHttp },
+    file: { resolve: resolveFile },
+  },
+  parseResolveResult: async opts => {
+    opts.result = parse(opts.result);
+    return opts;
```
lag-of-death:
```suggestion
    return { ...opts, result: parse(opts.result) };
```

lag-of-death:
@StefanDywersant, you can grab that chance and get rid of the mutation! ;)

packages/http/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@stoplight/prism-core": "^3.1.1",
+    "@stoplight/ref-resolvers": "^1.0.2",
```
XVincentX: Hmm, the json-ref-resolver should probably be next to the ref-resolvers package; if not, something is wrong. Can you check it out?

yarn.lock
```diff
@@ -3980,9 +3997,9 @@ ignore@^5.1.1:
   integrity sha512-vdqWBp7MyzdmHkkRWV5nY+PfGRbYbahfuvsBCh277tq+w9zyNi7h5CYJCK0kmzti9kU+O/cB7sE8HvKv6aXAKQ==
 
 immer@^3.2.0:
-  version "3.2.0"
-  resolved "https://registry.yarnpkg.com/immer/-/immer-3.2.0.tgz#53686471e9dd2b070e0fb5500c6fdecd3a99375f"
-  integrity sha512-+a2R8z9eELHst6aht++nzVzJ8LJ+Hsg49qttfg9Kc/vmoxEdPXw5/rV6+4DYWGgnq+B36KbLr4OTaGtS9mDjtg==
+  version "3.3.0"
```
XVincentX: Where is this update coming from?

packages/http/src/getHttpOperations.ts
StefanDywersant: Scout approach, fixed! :)

yarn.lock
```diff
@@ -1179,7 +1179,17 @@
...
+  version "3.1.2"
+  resolved "https://registry.yarnpkg.com/@stoplight/json/-/json-3.1.2.tgz#ae89d1974f0b6f8bc5778f792aeed4664c95ec62"
+  integrity sha512-7zt0zCN94m90X1sL7FrD7jXhPtkIup07MyG/00sELVrCvzb2wgNaiE+x2N7WSbk3Z0wS08e8B6kow81GQY+Isw==
+  dependencies:
+    "@stoplight/types" "^11.0.0"
+    jsonc-parser "~2.1.1"
+    lodash "^4.17.15"
+    safe-stable-stringify "^1.1"
+
+"@stoplight/json@^3.1.1":
```
XVincentX: These can probably be the same version

yarn.lock
```diff
@@ -1189,6 +1199,13 @@
     lodash "^4.17.15"
     safe-stable-stringify "^1.1"
```
XVincentX: You can get rid of this.

packages/http/package.json
StefanDywersant: Good spot. I moved `json-ref-resolver` to http package where it is actually used.

StefanDywersant: Oh, the same applies to yaml package.

XVincentX: Fix them all 🚎

yarn.lock
```diff
@@ -1179,7 +1179,17 @@
     urijs "~1.19.1"
     vscode-uri "^2.0.3"
 
-"@stoplight/json@^3.1.0", "@stoplight/json@^3.1.1":
+"@stoplight/json@^3.1.0":
```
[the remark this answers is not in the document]
StefanDywersant: 👍

yarn.lock
StefanDywersant: 👍

PR #695, opened 2019-10-10
Switch to ESLint

XVincentX:
Remove TSLint (deprecated) for ESLint.

I've enabled ESLint rules as well and I've fixed some of the warnings we were receiving. Also, some recommended ESLint rules for TypeScript have been disabled, but the aim it so enable them back with the passing of the time opportunistically.


packages/http/src/resolvers/http-and-file.ts
```diff
@@ -22,8 +22,8 @@ export const httpAndFileResolver = new Resolver({
     },
   },
 
-  parseResolveResult: async opts => {
+  parseResolveResult: opts => {
```
philsturgeon: would we not want async and await instead of the Promise.resolve() ?

XVincentX: Normally yes; the thing is this is a false async function, so there's never going to be an await; this is the reason ESLint is complaining.

I'm ok removing the rule if bothers the team though, I'm not opinionated on this one.

packages/http/src/forwarder/index.ts
```diff
@@ -12,7 +12,7 @@ const { version: prismVersion } = require('../../package.json');
 const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig>['forward'] = (
   resource: IHttpOperation,
   input: IHttpRequest,
-  timeout: number = 0,
```
lag-of-death: hm.. ? why remove this?

packages/http/src/router/__tests__/utils.ts
```diff
@@ -9,11 +9,11 @@ export function pickOneHttpMethod(): HttpMethod {
   return chance.pickone(httpMethods);
 }
 
-export function pickSetOfHttpMethods(count: number = 2): HttpMethod[] {
+export function pickSetOfHttpMethods(count = 2): HttpMethod[] {
```
lag-of-death: hm.. so this is in the spirit of `prefer inference over explicitness` kinda thing?

lag-of-death:
@XVincentX, looks ok, you might consider https://eslint.org/docs/rules/curly though ;)

packages/http/src/forwarder/index.ts
XVincentX: This is an automatic fix of ESLint, I guess for some reason decided that inferred arguments are better. Will see if I can revert this.

packages/http/src/router/__tests__/utils.ts
XVincentX: Apparently so — I'm kind of ok here, it's a number and it's pretty clear it's a number.

PR #701, opened 2019-10-14
fix: type import rewrite

XVincentX:
This is kind of a nasty TS to JS bug I found while trying to use Prism for its hosted version

Fundamentally given that it is a monorepo when generating the types for the current project the references points to the source code:

image

It goes without saying that such path does not exist in the published package, since that is instead found in `@stoplight/prism-core` package.

If you're using JavaScript, you won't even notice this bug. However, if you're using TypeScript and the `http` package, you'll notice some functions will return `any` instead of the correct interface (because TypeScript can't find the types, it will assume `any`).

This PR will use `ttypescript` project (a declarative wrapper for the TS compiler API) and adding a dropbox developed plugin that will rewrite the generated files path correctly

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,9 @@
     "tmp": "^0.1.0",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.3.0",
+    "ts-transform-import-path-rewrite": "^0.2.1",
     "tsconfig-paths": "^3.8.0",
+    "ttypescript": "^1.5.7",
     "typescript": "^3.6.4"
   },
   "prettier": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -56,7 +56,9 @@
     "tmp": "^0.1.0",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.3.0",
+    "ts-transform-import-path-rewrite": "^0.2.1",
     "tsconfig-paths": "^3.8.0",
+    "ttypescript": "^1.5.7",
```
StefanDywersant: I hope we'll get rid of it when typescript starts supporting plugins via config.

XVincentX: This is not happening anytime soon unfortunately :(

PR #702, opened 2019-10-14
Path params validation

### LINKED ISSUE #690: Path params are not validated
**Describe the bug**

Path-level path parameters are not validated with their schema.

**To Reproduce**

```yaml
openapi: 3.0.0
servers:
paths:
  '/path/{param}':
    parameters:
      - type: string
        name: param
        in: path
        required: true
        schema:
          type: string
          pattern: '^[A-Z]$'
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties: {}
```

or (get-level)

```yaml
openapi: 3.0.0
servers:
paths:
  '/path/{param}':
    get:
      parameters:
        - type: string
          name: param
          in: path
          required: true
          schema:
            type: string
            pattern: '^[A-Z]$'
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties: {}
```
```

```
curl 127.0.0.1:4010/path/123
```

**Expected behavior**

Get a validation error in response.

**Actual**

I get empty JSON object and OK status

 - Library version: v3.1.1

### PR DESCRIPTION
**Pull request by StefanDywersant**
Closes #690 

## Checklist

Feature

Path parameters are now validated against schema.

It does if someone relies on path params is not validated.

### DIFF
--- a/packages/cli/src/util/__tests__/paths.spec.ts
+++ b/packages/cli/src/util/__tests__/paths.spec.ts
@@ -75,7 +75,7 @@ describe('createExamplePath()', () => {
           },
           responses: [{ code: '200' }],
         }),
-        r => expect(r).toEqual('/path?p[a][aa]=1&p[a][ab]=2'),
+        r => expect(r).toEqual('/path?p%5Ba%5D%5Baa%5D=1&p%5Ba%5D%5Bab%5D=2'),
       );
     });
 
@@ -194,7 +194,7 @@ describe('createExamplePath()', () => {
         }),
         r =>
           expect(r).toEqual(
-            '/path/test1/.test1,test2/;p3=test1,test2?q1=test1&q2=test1%20test2&q3=test1%7Ctest2&q4=test1&q4=test2&q5[a][]=test1&q5[a][]=test2&q5[b][ba]=1&q5[b][bb]=2',
+            '/path/test1/.test1,test2/;p3=test1,test2?q1=test1&q2=test1%20test2&q3=test1%7Ctest2&q4=test1&q4=test2&q5%5Ba%5D%5B%5D=test1&q5%5Ba%5D%5B%5D=test2&q5%5Bb%5D%5Bba%5D=1&q5%5Bb%5D%5Bbb%5D=2',
           ),
       );
     });
--- a/packages/cli/src/util/paths.ts
+++ b/packages/cli/src/util/paths.ts
@@ -16,7 +16,7 @@ import * as Either from 'fp-ts/lib/Either';
 import { pipe } from 'fp-ts/lib/pipeable';
 import { get } from 'lodash';
 // @ts-ignore
-import { parse } from 'uri-template';
+import { URI } from 'uri-template-lite';
 
 export function createExamplePath(operation: IHttpOperation): Either.Either<Error, string> {
   return pipe(
@@ -29,7 +29,7 @@ export function createExamplePath(operation: IHttpOperation): Either.Either<Erro
         }),
       );
     }),
-    Either.map(({ template, values }) => parse(template).expand(values)),
+    Either.map(({ template, values }) => new URI.Template(template).expand(values)),
   );
 }
 
--- a/packages/cli/src/util/paths.ts
+++ b/packages/cli/src/util/paths.ts
@@ -29,7 +29,7 @@ export function createExamplePath(operation: IHttpOperation): Either.Either<Erro
         }),
       );
     }),
-    Either.map(({ template, values }) => new URI.Template(template).expand(values)),
+    Either.map(({ template, values }) => URI.expand(template, values)),
   );
 }
 
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -32,7 +32,7 @@
     "pino": "^5.13.2",
     "tslib": "^1.10.0",
     "type-is": "^1.6.18",
-    "uri-templates": "^0.2.0",
+    "uri-template-lite": "^19.4.0",
     "urijs": "^1.19.1"
   },
   "publishConfig": {
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -32,6 +32,7 @@
     "pino": "^5.13.2",
     "tslib": "^1.10.0",
     "type-is": "^1.6.18",
+    "uri-templates": "^0.2.0",
     "urijs": "^1.19.1"
   },
   "publishConfig": {
--- a/packages/http/src/__tests__/fixtures/index.ts
+++ b/packages/http/src/__tests__/fixtures/index.ts
@@ -274,7 +274,7 @@ export const httpInputs: IHttpRequest[] = [
   {
     method: 'post',
     url: {
-      path: '/',
+      path: '/todos',
       query: {
         overwrite: 'yes',
       },
--- a/packages/http/src/__tests__/fixtures/index.ts
+++ b/packages/http/src/__tests__/fixtures/index.ts
@@ -274,7 +274,7 @@ export const httpInputs: IHttpRequest[] = [
   {
     method: 'post',
     url: {
-      path: '/todos',
+      path: '/',
       query: {
         overwrite: 'yes',
       },
--- a/packages/http/src/validator/__tests__/index.spec.ts
+++ b/packages/http/src/validator/__tests__/index.spec.ts
@@ -151,7 +151,7 @@ describe('HttpValidator', () => {
     describe('path validation in enabled', () => {
       describe('request is set', () => {
         describe('request.path is set', () => {
-          it('validates path', () => {
+          it('calls the path validator', () => {
             validateInput({
               resource: {
                 method: 'get',
--- a/packages/http/src/validator/__tests__/index.spec.ts
+++ b/packages/http/src/validator/__tests__/index.spec.ts
@@ -176,27 +176,6 @@ describe('HttpValidator', () => {
           });
         });
       });
-
-      describe('request.url.path does not match the path from resource', () => {
-        it('throws error', () => {
-          expect(
-            () => validateInput({
-              resource: {
-                method: 'get',
-                path: '/a/{a}',
-                id: '1',
-                request: {
-                  path: [
-                    { name: 'a', style: HttpParamStyles.Simple },
-                  ]
-                },
-                responses: [{ code: '200' }],
-              },
-              element: { method: 'get', url: { path: '/b/1' } },
-            })
-          ).toThrowError('Received path is not a match for path from api description');
-        });
-      });
     });
   });
 
--- a/packages/http/src/validator/__tests__/index.spec.ts
+++ b/packages/http/src/validator/__tests__/index.spec.ts
@@ -1,5 +1,5 @@
 import { IPrismDiagnostic } from '@stoplight/prism-core';
-import { DiagnosticSeverity, IHttpOperation } from '@stoplight/types';
+import { DiagnosticSeverity, HttpParamStyles, IHttpOperation } from '@stoplight/types';
 import { IHttpRequest } from '../../types';
 import {
   bodyValidator,
@@ -9,7 +9,6 @@ import {
   validateInput,
   validateOutput
 } from '../index';
-import { HttpParamStyles } from '@stoplight/types/dist';
 
 const mockError: IPrismDiagnostic = {
   message: 'mocked C is required',
--- a/packages/http/src/validator/__tests__/index.spec.ts
+++ b/packages/http/src/validator/__tests__/index.spec.ts
@@ -1,7 +1,15 @@
 import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { DiagnosticSeverity, IHttpOperation } from '@stoplight/types';
 import { IHttpRequest } from '../../types';
-import { bodyValidator, headersValidator, queryValidator, validateInput, validateOutput } from '../index';
+import {
+  bodyValidator,
+  headersValidator,
+  pathValidator,
+  queryValidator,
+  validateInput,
+  validateOutput
+} from '../index';
+import { HttpParamStyles } from '@stoplight/types/dist';
 
 const mockError: IPrismDiagnostic = {
   message: 'mocked C is required',
@@ -16,6 +24,7 @@ describe('HttpValidator', () => {
       jest.spyOn(bodyValidator, 'validate').mockReturnValue([mockError]);
       jest.spyOn(headersValidator, 'validate').mockReturnValue([mockError]);
       jest.spyOn(queryValidator, 'validate').mockReturnValue([mockError]);
+      jest.spyOn(pathValidator, 'validate').mockReturnValue([mockError]);
     });
 
     afterAll(() => jest.restoreAllMocks());
@@ -47,7 +56,7 @@ describe('HttpValidator', () => {
               {
                 request: { body: { contents: [] }, path: [], query: [], headers: [], cookie: [] },
               },
-              2,
+              3,
             ),
           );
         });
@@ -63,7 +72,7 @@ describe('HttpValidator', () => {
                 request: { body: { contents: [], required: true } },
                 responses: [{ code: '200' }],
               },
-              3,
+              4,
             ),
           );
         });
@@ -90,7 +99,7 @@ describe('HttpValidator', () => {
       };
 
       describe('request is not set', () => {
-        it('validates headers', validate(undefined, 2));
+        it('validates headers', validate(undefined, 3));
       });
     });
 
@@ -98,7 +107,7 @@ describe('HttpValidator', () => {
       const validate = (
         resourceExtension?: Partial<IHttpOperation>,
         inputExtension?: Partial<IHttpRequest>,
-        length = 2,
+        length = 3,
       ) => () => {
         expect(
           validateInput({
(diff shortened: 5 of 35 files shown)

### CODE REVIEW

File: test-harness/specs/parameters-ac9.oas3.txt [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+Then I should get a 422 response back
+And the payload of the response should be in application/problem+json content-type
+====spec====
+openapi: 3.0.0
+paths:
+  '/path/{param}':
+    parameters:
+      - name: param
+        in: path
+        required: true
```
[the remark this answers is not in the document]
philsturgeon: If a path param is not provided in this path `/foo/{bar}/baz` that would be `/foo//baz`? That would just be a 404 I guess?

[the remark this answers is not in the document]
XVincentX: Good point, we can either take this as a validation error (missing required parameter) or return 404 as a missing route error. Hard to say for me which is the best.

packages/http/src/validator/deserializers/index.ts
```diff
@@ -18,4 +20,10 @@ export const query = new HttpParamDeserializerRegistry([
   new DeepObjectStyleDeserializer(),
 ]);
 
+export const path = new HttpParamDeserializerRegistry([
+  new SimpleStyleDeserializer(),
+  new LabelStyleDeserializer(),
+  new MatrixStyleDeserializer(),
+]);
```
XVincentX: One thing to keep in mind is that we can in the future employ a library doing all this for us (https://github.com/litejs/uri-template-lite#readme) — this apparently passes all the URI Templates specification tests and has both a "match", and "expand" function we could use.

@StefanDywersant Would it be possible to do a quick (no more than 1H) test on how hard would it be to use this library instead of writing our deserialiser?

packages/http/src/validator/deserializers/style/__tests__/label.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import { LabelStyleDeserializer } from '../label';
+import * as createObjectFromKeyValListModule from '../utils';
+
+describe('LabelStyleDeserializer', () => {
+  const labelStyleDeserializer = new LabelStyleDeserializer();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
```
XVincentX: Is this needed? We're not mocking anything in this test file.

packages/http/src/__tests__/fixtures/index.ts
```diff
@@ -274,7 +274,7 @@ export const httpInputs: IHttpRequest[] = [
   {
     method: 'post',
     url: {
-      path: '/',
+      path: '/todos',
```
XVincentX: Why this change?

packages/http/src/validator/validators/__tests__/__snapshots__/query.spec.ts.snap
```diff
@@ -1,28 +1,28 @@
...
-    "code": "required",
-    "message": "should have required property 'aparam'",
+    "code": "deprecated",
+    "message": "Query param param is deprecated",
     "path": Array [
       "query",
+      "param",
     ],
-    "severity": 0,
+    "severity": 1,
```
XVincentX: Aha ok it took me ages to understand what was going here — all it's happening is simply reversing the order :trollface:

StefanDywersant: Haha, me too ;) I should have comment it somehow to save reviewers time.

packages/http/src/validator/index.ts
```diff
@@ -35,7 +37,8 @@ const validateInput: ValidatorFn<IHttpOperation, IHttpRequest> = ({ resource, el
 
   return results
     .concat(headersValidator.validate(element.headers || {}, (request && request.headers) || []))
-    .concat(queryValidator.validate(element.url.query || {}, (request && request.query) || []));
+    .concat(queryValidator.validate(element.url.query || {}, (request && request.query) || []))
+    .concat(pathValidator.validate(getPathParams(element.url.path, resource.path), (request && request.path) || []));
```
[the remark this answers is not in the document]
StefanDywersant: It is exactly as you stated: `path` is a string. There is `getPathParams` function if you take a closer look.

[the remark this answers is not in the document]
StefanDywersant: However, I'd vote on extracting path params at an earlier stage. They are needed here but will be needed also for WebHooks. How do you feel about that?

packages/http/src/validator/index.ts
```diff
@@ -77,4 +80,14 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
   );
 };
 
+function getPathParams(path: string, template: string) {
+  const matches = new RegExp('^' + template.replace(/{(.+?)}/g, (_, name) => `(?<${name}>.*)`)).exec(path);
```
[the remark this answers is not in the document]
StefanDywersant: Believe me or not but I did some lookup over the internet. I apparently failed.

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -122,23 +131,73 @@ describe('HttpValidator', () => {
...
 
       describe('input.url.query is not set', () => {
         it("validates query assuming it's empty", validate(undefined, { url: { path: '/' } }));
       });
     });
+
+    describe('path validation in enabled', () => {
+      describe('request is set', () => {
+        describe('request.path is set', () => {
+          it('validates path', () => {
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -1,7 +1,15 @@
...
-import { bodyValidator, headersValidator, queryValidator, validateInput, validateOutput } from '../index';
+import {
+  bodyValidator,
+  headersValidator,
+  pathValidator,
+  queryValidator,
+  validateInput,
+  validateOutput
+} from '../index';
+import { HttpParamStyles } from '@stoplight/types/dist';
```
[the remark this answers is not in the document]
StefanDywersant: Done 👍

packages/http/src/validator/deserializers/style/__tests__/label.spec.ts
StefanDywersant: Definitely not, I also removed it from the rest of deserializer tests. Good catch & bit smaller codebase.

packages/http/src/__tests__/fixtures/index.ts
StefanDywersant: Not sure, most probably some leftover. I removed it.

packages/http/src/validator/index.ts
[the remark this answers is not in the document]
XVincentX: Yeah so the problem right now is that we're kind of using two Uri Template libraries

`urijs` (which is being removed in #669)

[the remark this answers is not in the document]
XVincentX: Yeah so the thing is we are already an Uri Template library `urijs` — that is being removed in #669;

Then we have another one `uri-templates` that is used for the example generation; we need to verify which library among the following ones can do the following things:

1. Parsing and expanding (regular case)
2. Desubstitution (grab parameters from a real url and its template)

Here are the libraries we have:

* https://medialize.github.io/URI.js/uri-template.html
* https://github.com/geraintluff/uri-templates
* https://github.com/litejs/uri-template-lite

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
StefanDywersant: Prism returns 422 for `/foo//baz`. I just added a test case covering that.

1) 422 is a correct error for such case:
a) the path is known (thus not 404) however, the `bar` param is missing.
b) imagine that `bar` is defined as a number & request path = `/foo/2118/baz` - this also results in 422
c) having that said, invalid format and missing param are similar cases and should both lead to 422

2. On the other hand we have a case where param is last part of path: `/foo/{bar}`.
The input `/foo` results in 404 (test case: https://github.com/stoplightio/prism/blob/master/test-harness/specs/parameters-ac1.oas3.txt).

Even though 2) I think 1) is still correct behavior.

packages/http/src/validator/index.ts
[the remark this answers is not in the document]
StefanDywersant: The full var name is `resource.request.path`. These are out request path parameters specifications (confused with that "api desc" stuff, how should I name it here?), not actual request input values.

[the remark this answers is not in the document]
StefanDywersant: Aha, and the source of confusion might be that we have:
`operation.url.path: string` - the actual path
`operation.request.path: stuff[]` - path param specs

packages/http/src/validator/deserializers/index.ts
StefanDywersant: There are two things making it a little harder to implement:
1) OpenApi's param serialization is a superset of what is inside URI Template RFC (deepobject, form, ..)
2) in order to perform extraction we'd need to generate URI Template - again, OpenApi does not help us here because OAS's `path` is not a URI Template.

packages/http/src/validator/index.ts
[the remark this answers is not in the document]
StefanDywersant: I'd initially vote in the `-lite` lib, however it:
```
(karol@MacBook-Pro-Karol prism)$ yarn add -W --dev @types/uri-template-lite
yarn add v1.19.1
[1/4] 🔍  Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/@types%2furi-template-lite: Not found".
```

packages/http/src/validator/deserializers/index.ts
philsturgeon: yeah all the csv supporting, split, matrix, pipes, etc is all OpenAPI-specific. This is not a complication of OpenAPI, more of a recognitiion of all the things people do in the wild which do not perfectly rely on standards. If everything was perfectly built using defined standards, our tooling would be easier!

philsturgeon:
Is there somewhere in this PR, or elsewhere in the existing codebase, where we deal with the awkard logic for "each type of parameter style has a different defaults for types? This stuff: 

image

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
philsturgeon: It's probably just a personal opinion but I would expect `/foo//bar` to be a 404. Path parameters are a bit of a invented concept which doesnt really exist in HTTP, so... considering it an "invalid parameter" is mostly just down to us naming everything "validation" when really it is "looking for various types of failures are responding accordingly". 

What happens if you go to `https://github.com/stoplightio//pull/702`? 

For other failures, I think we could probably get some clarity here by checking common JavaScript routing libraries. Any routing system which allows a `/foo/:integer/bar`? If you send a string in as that param, does that 404 or 422? 

I think we consider too many things "validation", and that is confusing us when it comes to figuring out what prism should do.

StefanDywersant:
> Is there somewhere in this PR, or elsewhere in the existing codebase, where we deal with the awkard logic for "each type of parameter style has a different defaults for types? This stuff:
> 
> image

Path (in this PR): https://github.com/stoplightio/prism/pull/702/files#diff-5c7aab71e5a0049cfb2006afe47c8a35R11
Query: https://github.com/stoplightio/prism/pull/702/files#diff-5c7aab71e5a0049cfb2006afe47c8a35R11
Headers: https://github.com/stoplightio/prism/blob/master/packages/http/src/validator/validators/headers.ts#L12
Body: might be missing - need to double check

packages/http/src/validator/index.ts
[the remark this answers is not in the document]
XVincentX: Yeah. We can leave as it is though, no actions point for this.

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
XVincentX: @StefanDywersant How hard would it be to treat this as a 404?

packages/http/src/validator/index.ts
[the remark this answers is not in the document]
XVincentX: Don't be discouraged by that; we can easily provide the types for that and contribute back to the community. If the library quality is effectively better, let's go for it.

packages/http/src/validator/deserializers/index.ts
XVincentX: What did we do to deserve this? :trollface: — resolving :)

packages/http/package.json
```diff
@@ -32,6 +32,7 @@
     "pino": "^5.13.2",
     "tslib": "^1.10.0",
     "type-is": "^1.6.18",
+    "uri-templates": "^0.2.0",
```
XVincentX: Can we use the same library everywhere? URIJS is going away with proxy command, but the example generation is using another library (see the CLI). It'd be cool to replace it and use the same here as well.

StefanDywersant: My bad that I did not unified those libraries.

I looked on [uri-templates](https://github.com/geraintluff/uri-templates#readme) and [uri-template-lite](https://github.com/litejs/uri-template-lite#readme). Both seem to support full spec. The latter is 3x smaller and the code looks less dense. My vote goes on `-lite` variant.

Now, cli and http uses `uri-template-lite`.

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -75,7 +75,7 @@ describe('createExamplePath()', () => {
           },
           responses: [{ code: '200' }],
         }),
-        r => expect(r).toEqual('/path?p[a][aa]=1&p[a][ab]=2'),
+        r => expect(r).toEqual('/path?p%5Ba%5D%5Baa%5D=1&p%5Ba%5D%5Bab%5D=2'),
```
StefanDywersant: This an similar lines need comment: the library is performing decodeURIComponent on passed arguments. That actually fixes a "bug". I used quotes because both versions put into the browser will work. The current one is more correct.

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
StefanDywersant: The logic for checking `deprecated` and `required` is one place for all parameter types (query, body, header, and path). It would require some gymnastics to treat path-param-required differently. If I add that in this PR it will be soon redone during validators refactor. I'd do that inside validator-refactor PR to avoid doubling the work.

Additionally, I still have some doubts about whether it is a good approach:

`https://github.com/stoplightio//pull/702` - returns 404 because there is no such repo called ''. That's because it would not pass validation during repo creation.

Let's consider another example. Github repo-name-validation disallows '_'.

`https://github.com/stoplightio/_gigi_dagostino/pull/702` - shall it return 404 or 422?

I'm sure that it will respond with 404, again, the repo can't exist because it would not pass repo-name-validation. I'd say that 404 is a data-driven error and Prism has no database.


Summary:
`Prism pretends that it has some database and guesses whether a resource exists`
vs
`Prism focuses on request validation against the API description`

or

`Developers are lazy and don't validate input params if they don't need to, so Prism is`
vs
`Developers are lazy and don't validate input params if they don't need to, so Prism isn't`

packages/cli/src/util/paths.ts
```diff
@@ -29,7 +29,7 @@ export function createExamplePath(operation: IHttpOperation): Either.Either<Erro
         }),
       );
     }),
-    Either.map(({ template, values }) => parse(template).expand(values)),
+    Either.map(({ template, values }) => new URI.Template(template).expand(values)),
```
XVincentX: We can avoid to create an instance and collect it right away.

```suggestion
    Either.map(({ template, values }) => URI.expand(template, values)),
```

packages/cli/src/util/__tests__/paths.spec.ts
XVincentX: Hmm, I am afraid I do not understand what's going on here — can you elaborare a little bit more?

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
XVincentX: Ok I _think_ I understand what you're saying.

The basic idea would be that since Prism has the knowledge of your API it can proactively say "hey I know what you're trying to do — and you're missing a parameter so here's a nice 422 saying that this is not ok" — while a regular web server won't effectively know what's going on and has no choice than answering 404.

Interesting; on one side, we can provide a helpful error message to the user; on the other, we want to try to behave as much as possible as a real web server.

> It would require some gymnastics to treat path-param-required differently

Yeah I understand; moreover this kind of error should be thrown by the router component and not the deserialiser/validator. It's kind of a big effort right now and I think it's be better to wait until the validation stuff has been refactored.

I vote to keep (at least temporarily) the 422 status code, and reason about this in the future

packages/cli/src/util/paths.ts
StefanDywersant: Done 👍

packages/cli/src/util/__tests__/paths.spec.ts
StefanDywersant: Sure. Previous library (`uri-templates`) did not care about whether input params are already uri encoded or not. That's why chars like '[' or ']' were visible in final urls. The new lib - `uri-template-lite` does encodeUriComponent on each params - that's the source of difference.

Both forms are valid though if you put it to the browser (browsers automatically encode some special characters) but I believe it is better that we provide already encoded form. So no relaying on browsers, our example uris will be ready to use.

test-harness/specs/parameters-ac9.oas3.txt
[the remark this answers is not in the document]
StefanDywersant: Agree! The moment of the decision whether prism is more-validation or more-mocking is coming :)

packages/http/package.json
XVincentX: Saturday task: create typings for such library :trollface:

test-harness/specs/parameters-ac10.oas3.txt
```diff
@@ -0,0 +1,34 @@
...
+                properties: {}
+====server====
+mock -p 4010 ${document}
+====command====
+curl -i http://127.0.0.1:4010/path//foo
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+content-type: application/problem+json
+
+{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["path"],"severity":"Error","code":"required","message":"should have required property 'param'"}]}
```
lag-of-death:
```suggestion
{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["path"],"severity":"Error","code":"required","message":"should have required property 'param'"}]}
```

Not sure if this is a change for this PR, but maybe? Surely saying `request body` here ain't right

test-harness/specs/parameters-ac9.oas3.txt
```diff
@@ -0,0 +1,36 @@
...
+                properties: {}
+====server====
+mock -p 4010 ${document}
+====command====
+curl -i http://127.0.0.1:4010/path/fp
+====expect====
+HTTP/1.1 422 Unprocessable Entity
+content-type: application/problem+json
+
+{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request body payload","status":422,"detail":"Your request body is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["path","param"],"severity":"Error","code":"pattern","message":"should match pattern \"^[A-Z]+$\""}]}
```
lag-of-death: ditto

test-harness/specs/parameters-ac10.oas3.txt
StefanDywersant: Looks like not right, I'll check where is that coming from. Good spot!

PR #713, opened 2019-10-15
fix: logger component is always required

XVincentX:
It turns out that although we're claiming that the `components` object is not required, it is indeed required because the logger instance is always required. 

This PR fixes this so that when using the http raw client package to have the hosted Prism version I do not incur in the problem of having something required not passed because of TypeScript not complaining.

### DIFF
--- a/packages/http-server/README.md
+++ b/packages/http-server/README.md
@@ -9,7 +9,7 @@ import { createServer } from '@stoplight/prism-http-server';
 
 const server = createServer({
   path: './api.oas2.json',
-  {}
+  {logger: createLoggerInstance() }
 });
 
 server.listen(3000).then(() => {
--- a/packages/http-server/README.md
+++ b/packages/http-server/README.md
@@ -7,6 +7,7 @@ Usage:
 ```js
 import { createServer } from '@stoplight/prism-http-server';
 
+const operations = await getHttpOperationsFromResource('./api.oas2.json');
 const server = createServer({
   operations,
   { logger: createLoggerInstance() }
--- a/packages/http-server/README.md
+++ b/packages/http-server/README.md
@@ -8,8 +8,8 @@ Usage:
 import { createServer } from '@stoplight/prism-http-server';
 
 const server = createServer({
-  path: './api.oas2.json',
-  {logger: createLoggerInstance() }
+  operations,
+  { logger: createLoggerInstance() }
 });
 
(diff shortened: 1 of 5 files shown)


packages/http-server/README.md
```diff
@@ -9,7 +9,7 @@ import { createServer } from '@stoplight/prism-http-server';
 
 const server = createServer({
   path: './api.oas2.json',
```
StefanDywersant: This part is outdated. `createServer` signature is `const createServer = (operations: IHttpOperation[], opts: IPrismHttpServerOpts): IPrismHttpServer`.

packages/http-server/src/server.ts
```diff
@@ -14,7 +14,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
   const { components, config } = opts;
 
   const server = fastify({
-    logger: (components && components.logger) || createLogger('HTTP SERVER'),
+    logger: components.logger,
```
StefanDywersant: I think we should lean towards having more default parameters. Ideally, a user should be able to spin up the server with just: `createServer([])`

XVincentX: The final user will always be able. They use the public client which has everything defaulted so they'll be able to use it with no parameters at all.

On the other hand, this is the internal client that I am using to create the hosted Prism version and this one requires this stuff.

Does that clarify?

packages/http-server/README.md
```diff
@@ -8,8 +8,8 @@ Usage:
 import { createServer } from '@stoplight/prism-http-server';
 
 const server = createServer({
-  path: './api.oas2.json',
-  {}
+  operations,
```
StefanDywersant:
```suggestion
  await getHttpOperationsFromResource('./api.oas2.json'),
```

Additionally, all should be wrapped into async func.

XVincentX: I've changed; ultimately this is an example and I think people will understand what we're talking about (async await) — I hope at least.

StefanDywersant: Some will, some won't. I believe it would be a couple less wtf's addressed to us if we do docs working out of a box.
Leaving to choice to you!

packages/http-server/src/server.ts
StefanDywersant: Yes it does!

PR #716, opened 2019-10-18
Callbacks

### LINKED ISSUE #331: OpenAPI v3 Callbacks
As you can see in the OAS3 docs: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#callback-object-example callbacks are now part of the specification.

It would be nice if prims could load the callback objects from the document and trigger the corresponding callback requests.

Here you have some ideas:

Given the [callback-example](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/callback-example.yaml) document from the specification, if we run:

`prism mock callback-example.yaml --callback_delay=30000`

The CLI output could be something like:
```
…  awaiting  Starting Prism…
★  star      Dynamic example generation enabled.
✔  success   Prism is listening on http://127.0.0.1:4010
●  note          POST       http://127.0.0.1:4010/streams
●  onData      POST       {$request.query.callbackUrl}/data  delay 30s.
```
This means that after the 30s. of delay a request with sample data based on the `requestBody` defined inside the `callbacks/onData` will be triggered to the `{$request.query.callbackUrl}/data` URL.

In case no delay is provided the callback request will be trigger immediately.

The CLI could also log when the request is made to the callback URL:
```
…  awaiting  Starting Prism…
★  star      Dynamic example generation enabled.
✔  success   Prism is listening on http://127.0.0.1:4010
●  note          POST       http://127.0.0.1:4010/streams
```
30s. after a request `POST http://127.0.0.1:4010/streams?callbackURL=http://callback.com`
```
●  callback request          POST       http://callback.com/data
```

The OAS3 example adds the callbackURL as a query param but I guess it can be also send as a body param.

StefanDywersant:
Closes #331 

## Checklist

_feature_

This PR introduces Callback support in Prism (issue #331).

Screen Shot 2019-11-06 at 15 55 28

No

## Dependencies (needed to be merged first)
1. https://github.com/stoplightio/types/pull/54
2. https://github.com/stoplightio/http-spec/pull/31


packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,58 @@
+import { IHttpRequest, IHttpResponse } from '../types';
+import * as Option from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get as _get } from 'lodash';
+import { pointerToPath } from '@stoplight/json';
+
```
XVincentX:
```suggestion
import { head, lookup } from 'fp-ts/lib/Array';

```

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,58 @@
...
+export function resolveRuntimeExpression(expr: string, request: IHttpRequest, response: IHttpResponse): Option.Option<unknown> {
+  const parts = expr.split(/[.#]/);
+  return pipe(
+    pipe(
+      parts[0],
```
XVincentX:
```suggestion
      head(parts),
```

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,58 @@
...
+      Option.fromPredicate(part => part === '$method'),
```
XVincentX:
```suggestion
      Option.chain(part => pipe(part, Option.fromPredicate(part => part === '$method'))),
```

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,58 @@
...
+      parts[0],
+      Option.fromPredicate(part => part === '$statusCode'),
+      Option.map(() => String(response.statusCode)),
+    )),
+    Option.alt(() => pipe(
+      parts[0],
+      Option.fromPredicate(part => part === '$request'),
+      Option.chain(() => pipe(
+        pipe(
+          parts[1],
```
XVincentX:
```suggestion
          lookup(1, parts),
```

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,58 @@
...
+          Option.fromPredicate(part => part === 'header'),
+          Option.chain(() => Option.fromNullable(request.headers && request.headers[parts[2]])),
+        ),
+        Option.alt(() => pipe(
+          parts[1],
+          Option.fromPredicate(part => part === 'query'),
+          Option.chain(() => Option.fromNullable(request.url.query && request.url.query[parts[2]])),
```
XVincentX: Maybe we could organise this as a series of functions so that the final pipe results a little bit easier to read?

```ts
  return pipe(
    readMethod(),
    Option.alt(() => readStatus()),
    Option.alt(() => readRequest()),
    Option.alt(() => readResponse()),
);
```

and then the single functions will effectively do the stuff.

StefanDywersant: Done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: 👍

XVincentX:
> However, I ended feeling that it is too much to carry the whole engine just for resolving such simple syntax.

Yeah that's totally legit. I'm jumping on this right now!

packages/http/src/util/__tests__/runtimeExpression.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+describe('resolveRuntimeExpression', () => {
+  it('resolves $method', () => {
+    assertSome(
+      resolveRuntimeExpression('$method', { method: 'get', url: { path: '' } }, { statusCode: 200 }),
+       value => expect(value).toEqual('get'),
+    );
+  });
+
+  it.skip('resolves $url', () => {
```
XVincentX: Aha! You thought I wouldn't see this? :trollface: 

You bad boy

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
+import { IHttpRequest, IHttpResponse } from '../types';
+import * as Option from 'fp-ts/lib/Option';
+import { lookup } from 'fp-ts/lib/Array';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get as _get } from 'lodash';
```
XVincentX: Why the renaming?

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+import { pointerToPath } from '@stoplight/json';
+
+export function resolveRuntimeExpressions(input: string, request: IHttpRequest, response: IHttpResponse) {
+  return input.replace(/{(.+?)}/g, (_, expr) =>
```
XVincentX: Can we put a comment on what is this RegEx going to look for? Every time I pass on this line I'm going to regex101.com to read what it is going to do :D

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+      Option.fromNullable(reqRes.body),
+      Option.chain(body =>
+        pipe(
+          lookup(2, parts),
+          Option.chain(part => Option.tryCatch(() => pointerToPath('#' + part))),
+          Option.chain(path => Option.fromNullable(_get(body, path)))
+        )
+      )
+    );
+  }
```
XVincentX: Let's put all these functions outside of `resolveRuntimeExpression` so when I am reviewing I can jump straight to the real function body and avoid reading all this boilerplate stuff.

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+      isPart(0, '$statusCode'),
+      Option.map(() => String(response.statusCode))
+    );
+  }
+
+  function tryRequestHeader() {
+    return pipe(
+      isPart(1, 'header'),
+      Option.chain(() => lookup(2, parts)),
+      Option.chain(part => Option.fromNullable(request.headers && request.headers[part]))
```
XVincentX:
```suggestion
      Option.chain(part => pipe(Option.fromNullable(request.headers), Option.mapNullable(request => request.headers[part])))
```

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+
+  function tryRequestQuery() {
+    return pipe(
+      isPart(1, 'query'),
+      Option.chain(() => lookup(2, parts)),
+      Option.chain(part => Option.fromNullable(request.url.query && request.url.query[part]))
```
XVincentX: Same as above, maybe it's better to use `Option.fromNullable` and `Option.mapNullable`

packages/http/src/util/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  function tryResponseBody() {
+    return pipe(
+      isPart(1, 'body'),
+      Option.chain(() => readBody(response))
+    );
+  }
+
+  function readBody(reqRes: IHttpRequest | IHttpResponse) {
```
XVincentX: If we're using a property in common, we can probably do either `IHttpRequest | IHttpResponse` or even better narrow the type to the body itself

```suggestion
  function readBody(body: unknown) {
```

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,82 @@
...
+  return withLogger(logger => {
+    const { url, ...requestInit } = assembleRequest({ resource: callback, request, response });
+    const logViolation = violationLogger(logger);
+
+    logger.info({ name: 'CALLBACK' }, `${callback.callbackName}: Making request to ${url}...`);
+
+    return pipe(
+      TaskEither.tryCatch(() => fetch(url, requestInit), Either.toError),
+      TaskEither.chain(parseResponse),
+      TaskEither.map(element => validateOutput({ resource: callback, element })),
```
XVincentX: Note to you: when #731 will be merged you might need to change this — or maybe no. Keep it In mind though.

packages/http/src/mocker/generator/HttpParamGenerator.ts
```diff
@@ -51,7 +52,7 @@ function pickStaticExample(
   );
 }
 
-export function generate(param: IHttpParam): Option.Option<unknown> {
+export function generate(param: { schema?: JSONSchema4 | JSONSchema6 | JSONSchema7; examples?: Array<INodeExample | INodeExternalExample> }): Option.Option<unknown> {
```
XVincentX: What is this change about?

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +51,52 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
+  response: IHttpResponse;
+}) {
+  return withLogger(logger =>
+    pipe(
+      Option.fromNullable(resource.callbacks),
+      Option.alt(() => Option.some([] as IHttpCallbackOperation[])),
+      Option.map(callbacks =>
+        pipe(
+          callbacks,
+          map(callback => runCallback({ callback, request, response })(logger)())
```
XVincentX: Maybe let's use the full name so we do not get confused with what's going on.
```suggestion
          Array.map(callback => runCallback({ callback, request, response })(logger)())
```

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +51,52 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
+  response,
+}: {
+  resource: IHttpOperation;
+  request: IHttpRequest;
+  response: IHttpResponse;
+}) {
+  return withLogger(logger =>
+    pipe(
+      Option.fromNullable(resource.callbacks),
+      Option.alt(() => Option.some([] as IHttpCallbackOperation[])),
```
XVincentX:
```suggestion
      Option.alt(() => Option.some<IHttpCallbackOperation[]>([])),
```

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +51,52 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
-    chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
-    chain(result => assembleResponse(result, payloadGenerator))
+    Reader.chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
+    Reader.chain(result => assembleResponse(result, payloadGenerator)),
+    Reader.chain(response =>
+      withLogger(logger =>
+        pipe(
+          response,
+          Either.map(r => runCallbacks({ resource, request: input.data, response: r })(logger)),
+          Either.chain(() => response)
```
XVincentX: Ok this is becoming difficult to review because of types error that are happening because of missing packages (updated types and http spec) — but this is likely wrong. 

I think in this case you need to use `ReaderEither` and one of its internal mapping methods to avoid doing all this.

### Welcome to monad composition :)

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,82 @@
...
+import { head } from 'fp-ts/lib/Array';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { generate as generateHttpParam } from '../generator/HttpParamGenerator';
+import { validateOutput } from '../../validator';
+import { parseResponse } from '../../util/response';
+import withLogger from '../../withLogger';
+import { violationLogger } from '../../util/logger';
+import { Logger } from 'pino';
+
+export function runCallback({ callback, request, response }: { callback: IHttpCallbackOperation, request: IHttpRequest, response: IHttpResponse }): Reader.Reader<Logger, Task.Task<void>> {
```
XVincentX: Same again, I cannot review this 100% because TypeScript is throwing me errors everywhere but I think this is the wrong abstraction. According to the code, we should use `ReaderTaskEither` and avoid the final `fold` call

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,82 @@
...
+        );
+      }),
+      TaskEither.fold(
+        error => {
+          logger.error({ name: 'CALLBACK' }, `${callback.callbackName}: Request failed: ${error.message}`);
+          return async () => undefined;
+        },
+        () => async () => undefined
+      ),
+    );
```
XVincentX: We need to avoid this and rely on `ReaderTaskEither` to return it. The logging sentence can be done using a `mapLeft`

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,82 @@
...
+    method: resource.method,
+  };
+}
+
+function assembleBody(request?: IHttpOperationRequest): Option.Option<string> {
+  return pipe(
+    Option.fromNullable(request && request.body && request.body.contents),
+    Option.chain(head),
+    Option.chain(generateHttpParam),
+    Option.map(JSON.stringify),
```
XVincentX: This is tricky. The problem is that `JSON.stringify` can effectively throw an exception, specifically when fed with circular objects. We need to decide if we want to pretend this is never going to happen or not.

I suggest we try to handle this accordingly — and use + `Either.fromOption` + `Either.stringifyJSON` function to return an Either. (Also `Option.fromEither(Either.stringifyJSON(value))`)

packages/http/src/mocker/generator/HttpParamGenerator.ts
StefanDywersant: I hit a problem that `IHttpParam` (https://github.com/stoplightio/types/blob/master/src/http-spec.ts#L72) and `IHttpContent` (https://github.com/stoplightio/types/blob/master/src/http-spec.ts#L123) are separate interfaces. So I changed `generate()` signature to define only the fields it needs to operate (`examples` and `schema`). Side effect is that we can put `IHttpParam` and `IHttpContent` because the field names match :)

packages/http/src/mocker/index.ts
StefanDywersant: 👍

packages/http/src/mocker/index.ts
StefanDywersant: Oh, I'm quite against doing `import * as Array from 'fp-ts...`. `Array` is an existing object in JS and mixing those names may cause a big bada boom.

Maybe `FPArray`. Or just stay with `map`? WDYT?

XVincentX: Ah ok, that is a good point. Let's leave it as it is, if it causes confusion we will revisit this.

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +51,52 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
     }),
-    chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
-    chain(result => assembleResponse(result, payloadGenerator))
+    Reader.chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
+    Reader.chain(result => assembleResponse(result, payloadGenerator)),
+    Reader.chain(response =>
+      withLogger(logger =>
+        pipe(
+          response,
+          Either.map(r => runCallbacks({ resource, request: input.data, response: r })(logger)),
```
XVincentX: Hmm, I am almost sure that running the callback is an independent operation from `negotiateResponse` and `assembleResponse ` — meaning we can throw the callbacks as soon as we have the response in place.

I do not know how much hard would be to do this (I am assuming we'd need to use a `sequenceT/S` or something; while this is not urgent, we might start thinking about this.

---

Also, given this will return a `ReaderTaskEither` — map won't be enough here anymore.

XVincentX:
```suggestion
          Either.map(response => runCallbacks({ resource, request: input.data, response })(logger)),
```

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +51,52 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
+  response,
+}: {
+  resource: IHttpOperation;
+  request: IHttpRequest;
+  response: IHttpResponse;
+}) {
+  return withLogger(logger =>
+    pipe(
+      Option.fromNullable(resource.callbacks),
+      Option.alt<IHttpCallbackOperation[]>(() => Option.some([])),
```
XVincentX: I would move this check outside of this function and not call it at all in case the user is not able to provide the callback

packages/http/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    it('logs error', () => {
+      violationLogger(logger)({ severity: DiagnosticSeverity.Error, message: 'Test' });
+      expect(logger.error).toHaveBeenCalledWith({ name: 'VALIDATOR' }, 'Violation: Test');
+      expect(logger.warn).not.toHaveBeenCalled();
+      expect(logger.info).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('violation is an warning', () => {
+    it('logs error', () => {
```
XVincentX:
```suggestion
    it('logs warning', () => {
```

packages/http/src/utils/__tests__/logger.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      violationLogger(logger)({ severity: DiagnosticSeverity.Warning, message: 'Test' });
+      expect(logger.error).not.toHaveBeenCalled();
+      expect(logger.warn).toHaveBeenCalledWith({ name: 'VALIDATOR' }, 'Violation: Test');
+      expect(logger.info).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('violation is an info', () => {
+    it('logs error', () => {
```
XVincentX:
```suggestion
    it('logs info', () => {
```

packages/http/src/utils/__tests__/parseResponse.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+        const response = {
+          headers: new Map([['content-type', 'application/json']]),
+          json: jest.fn().mockRejectedValue(new Error('Big Bada Boom')),
+          text: jest.fn(),
+        } as any;
+
+        assertLeft(
+          await parseResponseBody(response)(),
+          error => expect(error.message).toEqual('Big Bada Boom'),
+        );
```
XVincentX: This is probably my fault for the bad naming but I think we should create separate helpers for `TaskEither` so we do not have to execute the promise right now and have this weird `await` everywhere and `any` as returned body

packages/http/src/utils/__tests__/parseResponse.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  describe('body is not json', () => {
+    describe('body is readable', () => {
+      it('returns body text', async () => {
+        const response = {
+          headers: new Map([['content-type', 'text/html']]),
+          json: jest.fn(),
+          text: jest.fn().mockResolvedValue('<html>Test</html>'),
+        } as any;
```
XVincentX: Instead of using the `as any` cast I would modify the `parseResponseBody` response to use `Pick`:

`response: Pick<Response, 'json' | 'text' | 'headers'>`

It might sound weird but I believe it's going to be way better than having a bunch of casts everywhere.

packages/http/src/utils/parseResponse.ts
```diff
@@ -0,0 +1,33 @@
+import { Response } from 'node-fetch';
+import { is as typeIs } from 'type-is';
+import * as Either from 'fp-ts/lib/Either';
+import * as TaskEither from 'fp-ts/lib/TaskEither';
+import { mapValues } from 'lodash';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { Dictionary } from '@stoplight/types';
+import { IHttpResponse } from '../types';
+
+export function parseResponseBody(response: Response): TaskEither.TaskEither<Error, any> {
```
XVincentX:
```suggestion
export function parseResponseBody(response: Response): TaskEither.TaskEither<Error, unknown> {
```

packages/http/src/utils/runtimeExpression.ts
```diff
@@ -0,0 +1,128 @@
+import { IHttpRequest, IHttpResponse } from '../types';
```
XVincentX: Note to self: I still have not reviewed this.

examples/callbacks/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Run!
+
+Subscribe to callback
+
+```bash
+curl -v -H'Content-type: application/json' -d'{ "url": "http://localhost:4011/notify", "token": "ssecurre" }' http://127.0.0.1:4010/subscribe
+```
+
+Now, the console for `4010` service should show:
+```$xslt
```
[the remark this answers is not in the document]
StefanDywersant: `$xslt` is a way how `PHPStorm` expresses its shitty support for editing MD documents. Fixed.

examples/callbacks/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+prism mock -p 4010 service.oas3.yaml
+``` 
+
+Start service exposing `/notify` operation used for receiving callback requests (referred as `4011`)
+
+```bash
+prism mock -p 4011 client.oas3.yaml
+```
```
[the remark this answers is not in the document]
StefanDywersant: 👍

examples/callbacks/README.md
```diff
@@ -0,0 +1,48 @@
+# Mocking Callbacks with Prism
+
+This example shows how Prism mocks callbacks.
```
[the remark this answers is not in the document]
StefanDywersant: I reworked callback examples and docs today. The example is more real-life (subscription to notification about a certain invoice from payment service). The docs are in guides/ and it is linked to mocking-responses guide.

examples/callbacks/README.md
```diff
@@ -0,0 +1,48 @@
+# Mocking Callbacks with Prism
```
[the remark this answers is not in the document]
StefanDywersant: 👍

packages/http/src/utils/__tests__/logger.spec.ts
StefanDywersant: Done 👍

packages/http/src/utils/__tests__/logger.spec.ts
StefanDywersant: Done 👍

packages/http/src/forwarder/index.ts
```diff
@@ -1,47 +1,40 @@
...
-  }, toError);
+        {
+          headers: defaults(omit(input.headers, ['host', 'accept']), {
+            accept: 'application/json, text/plain, */*',
+            'user-agent': `Prism/${prismVersion}`,
+          }),
+        }
+      );
+    }, toError),
+    TaskEither.chain(parseResponse)
```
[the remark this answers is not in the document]
StefanDywersant: I separated `parseResponse` because it is also used in the `forwarder`.

packages/http/src/util/response.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as TaskEither from 'fp-ts/lib/TaskEither';
+import { mapValues } from 'lodash';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { Dictionary } from '@stoplight/types';
+import { IHttpResponse } from '../types';
+
+export function parseResponseBody(response: Response): TaskEither.TaskEither<Error, any> {
+  return TaskEither.tryCatch(
+    () =>
+      typeIs(response.headers.get('content-type') || '', ['application/json', 'application/*+json'])
```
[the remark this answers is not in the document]
StefanDywersant: Just checked, `@types/type-is@1.6.3` disagrees with that.

packages/http/src/util/runtimeExpression.ts
StefanDywersant: `get` is too common. However, I'm good with any approach. We could agree on `get`, `_get` or `_.get`.

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Will you be ok with moving the actual algorithm to the top of the function? I would vote on leaving those small functions defined inside because of the nice way they are binding the enclosing function arguments. If I move them outside I will need to do it explicitly and it does not look so clever anymore :(

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Way better than &&, done 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Done 👍

packages/http/src/mocker/index.ts
StefanDywersant: > Hmm, I am almost sure that running the callback is an independent operation from `negotiateResponse` and `assembleResponse ` — meaning we can throw the callbacks as soon as we have the response in place.

Callbacks require the full response object of enclosing operation. See `runtimeExpression` syntax. They can access response object to retrieve certain data.

The real-life example for that is:
1) `POST /invoices` - create invoice, at request stage invoice id does not yet exists
2) response returns the invoice data together with the id
3)  payment notification callback: `{$request.body#/url}?invoiceId={$response.body#/id}` - client wants invoice id in query param because it is in java and it would require 3 classes and 6 interfaces to decode body

packages/http/src/mocker/index.ts
StefanDywersant: I don't recall why it is there, maybe some leftover. Thanks & fixed.

packages/http/src/mocker/callback/callbacks.ts
StefanDywersant: Adjusted to recent validators refactor 👍

packages/http/src/util/runtimeExpression.ts
StefanDywersant: Good catch, fixed 👍

packages/http/src/util/response.ts
```diff
@@ -0,0 +1,33 @@
+import { Response } from 'node-fetch';
+import { is as typeIs } from 'type-is';
+import * as Either from 'fp-ts/lib/Either';
+import * as TaskEither from 'fp-ts/lib/TaskEither';
+import { mapValues } from 'lodash';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { Dictionary } from '@stoplight/types';
+import { IHttpResponse } from '../types';
+
+export function parseResponseBody(response: Response): TaskEither.TaskEither<Error, any> {
```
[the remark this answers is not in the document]
StefanDywersant: Fixed. We need to add pre-commit hook which will look for `any` in the patch and write "You probably could do better in line #".

packages/http/src/utils/parseResponse.ts
StefanDywersant: Done 👍

packages/http/src/mocker/callback/callbacks.ts
StefanDywersant: Improved 👍

packages/core/src/__tests__/utils.ts
```diff
@@ -9,34 +10,42 @@ export function assertNone<A>(e: Option.Option<A>) {
...
   );
 }
+
+export async function assertResolvesRight<L, A>(e: TaskEither.TaskEither<L, A>, onRight: (a: A) => void = () => {}) {
+  assertRight(await e(), onRight);
+}
+
+export async function assertResolvesLeft<L, A>(e: TaskEither.TaskEither<L, A>, onLeft: (a: L) => void = () => {}) {
+  assertLeft(await e(), onLeft);
+}
```
XVincentX: Nice! We're set!

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,97 @@
...
+            pipe(violations, map(logViolation));
+            return violations;
+          }),
+          Either.alt(() => Either.right(element)),
+        );
+      }),
+      TaskEither.mapLeft(error => {
+        logger.error({ name: 'CALLBACK' }, `${callback.callbackName}: Request failed: ${error.message}`);
+        return error;
+      }),
```
XVincentX: This flow here is a little bit weird; I think the reason why the code's been wrote in this way is that `chain` carries on the same error type — and so we can't move from `Error` to `NonEmptyArray<IPrismDiagnostic[]>`.

However, since we're only logging the error and not "handling" it for real, we can use a little trick to make this flow a little bit better. I've prepared a PR so you can take a look into and merge if if you agree:

https://github.com/stoplightio/prism/pull/772

In any case I'm going to look a little bit into the theory and see how you handle such situation properly, and then we can refactor in case that makes sense.

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,97 @@
...
+    url: resolveRuntimeExpressions(resource.path, request, response),
+    headers: Option.toUndefined(assembleHeaders(resource.request, bodyAndMediaType && bodyAndMediaType.mediaType)),
+    body: bodyAndMediaType && bodyAndMediaType.body,
+        method: resource.method,
+};
+}
+
+function assembleBody(request?: IHttpOperationRequest): Option.Option<{ body: string, mediaType: string }> {
+  return pipe(
+    Option.fromNullable(request),
```
XVincentX: This check is required by both `assembleBody` and `assembleHeaders` — it means that most likely we can move this control up to the caller so they can both use it.

packages/http/src/mocker/generator/HttpParamGenerator.ts
XVincentX: Then I guess you can just do

```suggestion
export function generate(param: IHttpParam | IHttpContent): Option.Option<unknown> {
```

So that only the fields in common can be used inside this function as long you do not use a type guard.

packages/http/src/utils/runtimeExpression.ts
XVincentX: Note to self 2: I still have not reviewed this and I should be just ashamed of myself.

docs/guides/callbacks.md
```diff
@@ -0,0 +1,58 @@
+# Mocking Callbacks with Prism
+
+## What are callbacks?
```
XVincentX: New lines between all the paragraphs — or if you're lazy you can just run prettier on it:

`npx prettier --write ./docs/guides/callbacks.md`

I did that for you in #772

packages/http/src/util/response.ts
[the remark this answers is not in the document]
XVincentX: We could enable TypeScript to not allow any *at all*, but I am afraid we're not there yet. We can get there though!

packages/http/src/util/response.ts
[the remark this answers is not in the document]
XVincentX: Puah — it's just not true though.

packages/http/src/util/runtimeExpression.ts
XVincentX: We've bee using `get` everywhere here in Prism. It's not a big deal though, just mentioning for consistency.

packages/http/src/util/runtimeExpression.ts
XVincentX: That's ok — let's put the main function up so that we can easily look at the real code and jump to the helper functions in case it's needed.

packages/http/src/mocker/callback/callbacks.ts
XVincentX: I checked out again this flow — I suggest to use the following that is a little bit more verbose/readable

```ts
function assembleBody(request?: IHttpOperationRequest): Option.Option<{ body: string; mediaType: string }> {
  return pipe(
    Option.fromNullable(request),
    Option.mapNullable(request => request.body),
    Option.mapNullable(body => body.contents),
    Option.chain(head),
    Option.chain(param =>
      pipe(
        param,
        generateHttpParam,
        Option.map(body => ({ body, mediaType: param.mediaType }))
      )
    ),
    Option.chain(({ body, mediaType }) =>
      pipe(
        Either.stringifyJSON(body, () => undefined),
        Either.map(body => ({ body, mediaType })),
        Option.fromEither
      )
    )
  );
}
```

I've pushed this as part of #772

examples/callbacks/README.md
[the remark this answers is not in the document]
XVincentX: Come to the dark side:

[SublimeText](https://www.sublimetext.com)

docs/guides/callbacks.md
StefanDywersant: Resolved via merging #772

docs/guides/callbacks.md
```diff
@@ -0,0 +1,58 @@
...
+
+```
+[HTTP SERVER] post /notify ℹ  info      Request received
+    [NEGOTIATOR] ℹ  info      Request contains an accept header: */*
+    [VALIDATOR] ✔  success   The request passed the validation rules. Looking for the best response
+    [NEGOTIATOR] ✔  success   Found a compatible content for */*
+    [NEGOTIATOR] ✔  success   Responding with the requested status code 200
+```
+
+After subscribing via `/subscribe`, Prism successfully invoked `/notify` callback with mocked payload.
```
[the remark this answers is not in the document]
StefanDywersant: Limitations section added 👍

packages/core/src/__tests__/utils.ts
StefanDywersant: 👍

packages/http/src/mocker/callback/callbacks.ts
StefanDywersant: I tried that way but it looked worse than how it looks now. I vote for keeping the way it is now.

packages/http/src/mocker/generator/HttpParamGenerator.ts
StefanDywersant: Good point, updated 👍

packages/http/src/mocker/index.ts
```diff
@@ -41,20 +44,51 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
...
-    chain(result => assembleResponse(result, payloadGenerator))
+    Reader.chain(mockConfig => negotiateResponse(mockConfig, input, resource)),
+    Reader.chain(result => assembleResponse(result, payloadGenerator)),
+    Reader.chain(response =>
+      withLogger(logger =>
+        pipe(
+          response,
+          Either.map(response => runCallbacks({ resource, request: input.data, response })(logger)),
+          Either.chain(() => response)
+        )
```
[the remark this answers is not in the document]
StefanDywersant: I thought about the placement of callbacks on the beginning, my thoughts on that:
1. Moving to `core` is not a good idea, they are specific to `http`. They may make no sense for other types of specifications
2. Callbacks are part of mocking. They are not applicable to other existing commands (meaning proxy for now)

packages/http/src/mocker/callback/callbacks.ts
StefanDywersant: #772 applied

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,97 @@
...
+import { generate as generateHttpParam } from '../generator/HttpParamGenerator';
+import { validateOutput } from '../../validator';
+import { parseResponse } from '../../utils/parseResponse';
+import withLogger from '../../withLogger';
+import { violationLogger } from '../../utils/logger';
+import { Logger } from 'pino';
+
+export function runCallback({ callback, request, response }: { callback: IHttpCallbackOperation, request: IHttpRequest, response: IHttpResponse }): Reader.Reader<Logger, TaskEither.TaskEither<Error, unknown>> {
+  return withLogger(logger => {
+    const { url, ...requestInit } = assembleRequest({ resource: callback, request, response });
```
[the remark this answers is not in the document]
StefanDywersant: Applied from #772

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -0,0 +1,97 @@
...
+import { head } from 'fp-ts/lib/Array';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { generate as generateHttpParam } from '../generator/HttpParamGenerator';
+import { validateOutput } from '../../validator';
+import { parseResponse } from '../../utils/parseResponse';
+import withLogger from '../../withLogger';
+import { violationLogger } from '../../utils/logger';
+import { Logger } from 'pino';
+
+export function runCallback({ callback, request, response }: { callback: IHttpCallbackOperation, request: IHttpRequest, response: IHttpResponse }): Reader.Reader<Logger, TaskEither.TaskEither<Error, unknown>> {
```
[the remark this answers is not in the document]
StefanDywersant: Applied from #772 👍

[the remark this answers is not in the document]
StefanDywersant: The names are better and better. Like german names :]

docs/guides/callbacks.md
```diff
@@ -0,0 +1,64 @@
+# Mocking Callbacks with Prism
+
+## What are callbacks?
+
+Callback in OpenApi 3 defines an outgoing, asynchronous request that your service will make to some other service. Typical real-life example is a code repository. You can subscribe to certain events on a repo (like commit or tag) and your api will start receiving notifications for those events. Another example is one-time notifications. You can subscribe to a `invoice paid` event and a callback will be invoked when such payment is processed.
+
+##### Sources:
+
+- [Callback Docs](https://swagger.io/docs/specification/callbacks/)
+- [Callback Object Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject)
```
philsturgeon:
```suggestion
- [Callback Object Specification](http://spec.openapis.org/oas/v3.0.2#callbackObject)
```

docs/guides/mock-responses.md
```diff
@@ -153,3 +153,7 @@ When we call `curl http://127.0.0.1:4010/pets/123?__dynamic=true`, the operation
 The more descriptive your description is, the better job Prism can do at creating a mock response.
 
 _**Tip:** If your team needs help creating better quality API description documents, take a look at [Spectral](https://stoplight.io/spectral/). You could enforce the use of `example` properties, or similar._
+
+### Mocking OpenApi 3 callbacks
```
philsturgeon:
```suggestion
### Mocking OpenAPI Callbacks
```

It's always "OpenAPI", and they will be in v4 and v5 too so we dont need to specify v3 for things. It'll just get wrong over time. :)

packages/http/src/mocker/index.ts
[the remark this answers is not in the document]
philsturgeon: Keep in mind that maybe callbacks (an outbound event triggered by an incoming event) will be the first step towards AsyncAPI support, so maybe some of it could be core. If that is just nonsense at this point dont worry we can refactor later, but keep it in mind.

philsturgeon:
This is so cool, amazing work @StefanDywersant. 

One thing could do with a bit of polish: I was using https://webhook.site/ which is a cool request bin type thing which captures webhooks for testing. Apparently their API is not responding quite how Prism would like: 

Screen Shot 2019-11-06 at 15 55 28

I guess it's because the callback has rather specific requirements for the callback:

```yaml
      callbacks:
        actions:
          '{$request.body#/url}?token={$request.body#/token}':
            post:
              summary: 'Receive notification about certain invoice'
              requestBody:
                required: true
                content:
                  application/json:
                    schema:
                      $ref: '#/components/schemas/notification'
              responses:
                200:
                  description: 'Notification successfully processed'
                  content:
                    text/plain:
                      examples:
                        ok:
                          value: 'ok'
```

Here is what it responded with:

```
HTTP/1.1 200 OK
Cache-Control: no-cache, private
Content-Encoding: gzip
Content-Type: text/plain; charset=UTF-8
Date: Wed, 06 Nov 2019 15:02:21 GMT
Server: nginx/1.10.3
Transfer-Encoding: chunked
Vary: Accept-Encoding
```

Anyhow, the error message it gives is not as descriptive as it could be. 

>  [VALIDATOR] ✖  error     Violation: The received media type does not match the one specified in the document

Maaaybe that should actually be displayed as [CALLBACK]? Or if not, a super clear message:

```
Violation: The description document expects text/plan or application/json but instead it got application/nonsense
```

Beyond the error message, I suppose if somebody wanted to avoid this problem they would simply not specify a `content` right?

So instead of:

```yaml
              responses:
                200:
                  description: 'Notification successfully processed'
                  content:
                    text/plain:
                      examples:
                        ok:
                          value: 'ok'
```

it would be:

```yaml
              responses:
                200:
                  description: 'Notification successfully processed'
```

docs/guides/callbacks.md
StefanDywersant: Done 👍

docs/guides/mock-responses.md
StefanDywersant: That's right, fixed 👍

packages/http/src/mocker/index.ts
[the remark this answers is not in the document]
StefanDywersant: Callbacks have well-defined, single-point-of-contact. We can move them around when new requirements appear.

StefanDywersant:
> Beyond the error message, I suppose if somebody wanted to avoid this problem they would simply not specify a `content` right?

If you do not specify the content prism will also report an error that it couldn't match. Basically callbacks validation acts exactly the same as normal request validation.

That points to a possible problem which I also noticed doing more or less the same as you are doing with `webhook.site`: when `contents` are not defined, shouldn't prism accept whatever mediaType of response?

@philsturgeon @XVincentX

XVincentX:
@StefanDywersant You're correct, there might be a little bug in the `mismatchMediaType` function — which checks for something even though there's nothing to look for.

Let's keep this separate though. I'll investigate into this — in meantime I would just change the error message from a generic thing to something more like "I received *THIS* and instead this is what's available in the document"

philsturgeon:
@StefanDywersant ahh well this just shows that my next job of "play with proxy server and contract testing to make sure it works as expected" should have been done sooner. My bad. 

If the API does not specify content, then anything which comes in is a bonus, maybe a warning or an info, but certainly not an error. It's the sort of thing we'd report to folks via Studio somehow (long way off) saying "Hey we found some content, looks like this is JSON, here is the schema. Wanna save that to your description?" etc.

How do I make a callback which says "I dont care what they respond with im just looking for a 200? A lot of callbacks work this way, they just want a "Yep!" and nothing more.

StefanDywersant:
> How do I make a callback which says "I dont care what they respond with im just looking for a 200? A lot of callbacks work this way, they just want a "Yep!" and nothing more.

@XVincentX just submitted an issue for that https://github.com/stoplightio/prism/issues/774

packages/http/src/mocker/index.ts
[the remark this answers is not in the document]
XVincentX: Yeah for now we can leave it as it is since it's working properly anyway and not critical. However I also think this will be probably moved at a certain point out of this and put in the factory file. I'll just add a comment in the code.

philsturgeon:
@StefanDywersant sorry yeah, I didnt see his comment on this crap ferry wifi. Our comments passed each other like ships in the night. If you do the thing he suggested I've got a thumb ready and waiting!

StefanDywersant:
> @StefanDywersant sorry yeah, I didnt see his comment on this crap ferry wifi. Our comments passed each other like ships in the night. If you do the thing he suggested I've got a thumb ready and waiting!

Done in a22024aa2a0e381c0f8ef5e25e6466450a39b8e2

PR #720, opened 2019-10-19
feat: file support for harness

XVincentX:
This has been back ported from #669 

By merging this in master the diff of the Proxy PR shall become significantly smaller and help people get that 🐉 over the finish line.

### DIFF
--- a/test-harness/README.md
+++ b/test-harness/README.md
@@ -36,7 +36,7 @@ Test text, can be multi line.
 ====spec====
 openapi 2/3 document
 ====server====
-command line arguments to run Prism with. You can use the ${file} template variable
+command line arguments to run Prism with. You have to use the ${file} template variable.
 ====command====
 curl command to get the response
 ====expect====
--- a/test-harness/README.md
+++ b/test-harness/README.md
@@ -36,7 +36,7 @@ Test text, can be multi line.
 ====spec====
 openapi 2/3 document
 ====server====
-command line arguments to run Prism with. You have to use the ${file} template variable.
+command line arguments to run Prism with. You have to use the ${document} template variable.
 ====command====
 curl command to get the response
 ====expect====
@@ -60,7 +60,7 @@ paths:
             'application/json':
               example: hello
 ====server====
-mock -p 4010 ${file}
+mock -p 4010 ${document}
 ====command====
 curl -i -X GET http://localhost:4010/todos -H "accept: application/json"
 ====expect====
--- a/test-harness/README.md
+++ b/test-harness/README.md
@@ -36,7 +36,7 @@ Test text, can be multi line.
 ====spec====
 openapi 2/3 document
 ====server====
-command line arguments to run Prism with
+command line arguments to run Prism with. You can use the ${file} template variable
 ====command====
 curl command to get the response
 ====expect====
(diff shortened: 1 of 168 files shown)


test-harness/README.md
```diff
@@ -36,7 +36,7 @@ Test text, can be multi line.
 ====spec====
 openapi 2/3 document
 ====server====
-command line arguments to run Prism with
+command line arguments to run Prism with. You can use the ${file} template variable
```
lag-of-death: @XVincentX, you `can` or `have to`? Also, don't we always mean `document` or `spec` by `file`? Maybe it's worth renaming? WDYT?

test-harness/index.ts
```diff
@@ -45,8 +45,7 @@ describe('harness', () => {
 
     test(testText, done => {
       const [command, ...args] = parsed.command.split(' ').map(t => t.trim());
-      const serverArgs = [...parsed.server.split(' ').map(t => t.trim()), tmpFileHandle.name];
-
+      const serverArgs = parsed.server.split(' ').map(t => t.trim().replace('${file}', tmpFileHandle.name));
```
StefanDywersant:
```suggestion
      const serverArgs = parsed.server.split(' ').map(t => t.replace('${file}', tmpFileHandle.name));
```

Trim has no effect here as you are splitting by ` `

test-harness/README.md
XVincentX: You're correct, it's compulsory.

Phil has always been against using `spec` — so definitely not that term. Let's leave it as it is right now (otherwise merging this in the Proxy PR is going to be a mess) and we can change the name if we find something appropriate.

lag-of-death: The process times out for me, when I remove ${file} from the tests

test-harness/index.ts
StefanDywersant:
```suggestion
      const serverArgs = parsed.server.split(/ +/).map(t => t.replace('${file}', tmpFileHandle.name));
```
^ this could be "more-what-author-wanted"

test-harness/README.md
XVincentX: Docs rewrote.

test-harness/index.ts
XVincentX: What does that regex do?

test-harness/index.ts
```diff
@@ -41,12 +41,11 @@ describe('harness', () => {
...
     afterAll(() => tmpFileHandle.removeCallback(undefined, undefined, undefined, undefined));
 
-    const testText = `${file}${os.EOL}${parsed.test}`;
+    const testText = `${document}${os.EOL}${parsed.test}`;
 
     test(testText, done => {
       const [command, ...args] = parsed.command.split(' ').map(t => t.trim());
-      const serverArgs = [...parsed.server.split(' ').map(t => t.trim()), tmpFileHandle.name];
-
+      const serverArgs = parsed.server.split(' ').map(t => t.replace('${document}', tmpFileHandle.name));
```
StefanDywersant:
```suggestion
      const serverArgs = parsed.server.split(/ +/).map(t => t.replace('${document}', tmpFileHandle.name));
```

I'm repeating the comment as I think it has been lost. The above suggestion should be overall better.

XVincentX:
Nice to see comments that otherwise wouldn't have popped up from the gigantic PR by the way :)

test-harness/index.ts
XVincentX: The trim is still required because of newlines.

StefanDywersant:
@XVincentX changelog line is missing

XVincentX:
It was on purpose, since we do not want to document internal changes that do not affect the users (library updates and shit). However this might affect users writing tests. Adding a line.

PR #731, opened 2019-10-22
refactor: validation returns eithers

XVincentX:
This PR is a start to a refactor of the validators approach. 

As I mentioned somewhere, the idea is to have 3 PRs that will do the job

Here's what I did:

1. The validator generic signature now does not return a `IPrismDiagnostic[]`, but rather an `Either<NonEmptyArray<IPrismDiagnostic>, T>` — which means either a validated `T` or an array of validation errors. https://github.com/stoplightio/prism/pull/731/files#diff-2991521f751aa9cf68976ceef512e19a
2. I have created a sequence for an Either with the applicative method modified to concat the errors, as the article we were reading about was suggesting: https://github.com/stoplightio/prism/pull/731/files#diff-d6f4104cbf0cdb51855c413bc0079241R43-R44
3. All the validators have been modified to return the Either — https://github.com/stoplightio/prism/pull/731/files#diff-ec5f393de032d67a4f08a12ecde5456cR54, and then all sequenced together to get the final request/response validation https://github.com/stoplightio/prism/pull/731/files#diff-52fafd78c752aca126e4fd8c4ce567a6R98-R103
4. All the improvements then finally converge in the `factory.ts` file where you can see that finally validation and mocking are two different `TaskEither` steps melted together with chain calls.


packages/core/src/factory.ts
```diff
@@ -18,62 +22,71 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
+                validations =>
+                  Either.right<Error, { resource: Resource; validations: IPrismDiagnostic[] }>({
+                    resource,
+                    validations,
+                  }),
+                () =>
+                  Either.right<Error, { resource: Resource; validations: IPrismDiagnostic[] }>({
+                    resource,
+                    validations: [],
+                  })
```
lag-of-death: note to self: so `Either.right` because we also want to gather output validation results (?)

packages/core/src/factory.ts
```diff
@@ -18,62 +22,71 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
+        ),
+        TaskEither.chain(({ resource, validations }) => {
+          const produceOutput = isProxyConfig(config)
             ? components.forward(input, config.upstream.href)
             : TaskEither.fromEither(
                 components.mock({
                   resource,
                   input: {
-                    validations: inputValidations,
+                    validations,
```
lag-of-death: note to self: both request and security validation results

packages/core/src/factory.ts
```diff
@@ -18,62 +22,71 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
+            Option.map<NonEmptyArray<IPrismDiagnostic>, IPrismDiagnostic[]>(t => t),
+            Option.getOrElse<IPrismDiagnostic[]>(() => [])
+          );
 
           return {
             input,
             output,
             validations: {
-              input: inputValidations,
+              input: validations,
```
lag-of-death: maybe be frank about `validations` from the beginning and call it `inputValidations`? so here, we would end up with:
```
input: inputValidations,
output: outputValidations,
```

packages/core/src/utils/security/index.ts
```diff
@@ -91,12 +92,20 @@ function getAuthResults(securitySchemes: SecurityScheme[][], someInput: unknown,
...
 
   if (!securitySchemes.length) {
-    return none;
+    return Either.right(someInput);
   } else {
-    return gatherValidationResults(securitySchemes, someInput, resource);
+    return pipe(
+      gatherValidationResults(securitySchemes, someInput, resource),
+      Either.fromOption(() => someInput),
+      Either.swap,
```
lag-of-death: note to self: ok, because "different Either.swap that are just weird"

packages/core/src/utils/security/index.ts
```diff
@@ -91,12 +92,20 @@ function getAuthResults(securitySchemes: SecurityScheme[][], someInput: unknown,
...
+      Either.mapLeft<IPrismDiagnostic, NonEmptyArray<IPrismDiagnostic>>(e => [e])
```
lag-of-death: we could make it point-free with `(Array.of)` instead of `(e => [e])`

packages/http/src/validator/index.ts
```diff
@@ -24,65 +27,82 @@ export const queryValidator = new HttpQueryValidator(queryDeserializerRegistry,
...
+        )
+      )
+    ),
+    Either.fromOption(() => {}),
+    Either.swap
+  );
+
+  return pipe(
+    sequenceValidation(
+      validateBody,
```
lag-of-death: would it be feasible for the composition of `validateBody` to be put outside of `validateInput`?

packages/http/src/validator/index.ts
```diff
@@ -24,65 +27,82 @@ export const queryValidator = new HttpQueryValidator(queryDeserializerRegistry,
...
 };
 
+const findResponseByStatus = (responses: NonEmptyArray<IHttpOperationResponse>, statusCode: number) =>
+  pipe(
+    findOperationResponse(responses, statusCode),
+    Either.fromOption<IPrismDiagnostic>(() => ({
+      message: 'Unable to match the returned status code with those defined in spec',
+      severity: inRange(statusCode, 200, 300) ? DiagnosticSeverity.Error : DiagnosticSeverity.Warning,
+    })),
+    Either.mapLeft<IPrismDiagnostic, NonEmptyArray<IPrismDiagnostic>>(error => [error])
```
lag-of-death:
```suggestion
    Either.mapLeft<IPrismDiagnostic, NonEmptyArray<IPrismDiagnostic>>(Array.of)
```

packages/http/src/validator/index.ts
```diff
@@ -24,65 +27,82 @@ export const queryValidator = new HttpQueryValidator(queryDeserializerRegistry,
...
+
+const mismatchMediaType = (response: IHttpOperationResponse, mediaType: string) =>
+  pipe(
+    Option.fromNullable(response.contents),
+    Option.chain(findFirst(c => c.mediaType === mediaType)),
+    Either.fromOption<IPrismDiagnostic>(() => ({
+      message: `The received media type does not match the one specified in the document`,
+      severity: DiagnosticSeverity.Error,
+    })),
+    Either.mapLeft<IPrismDiagnostic, NonEmptyArray<IPrismDiagnostic>>(e => [e])
```
lag-of-death:
```suggestion
    Either.mapLeft<IPrismDiagnostic, NonEmptyArray<IPrismDiagnostic>>(Array.of)
```

packages/http/src/validator/validators/__tests__/headers.spec.ts
```diff
@@ -2,6 +2,7 @@ import { HttpParamStyles } from '@stoplight/types';
 import { query as registry } from '../../deserializers';
 import { HttpHeadersValidator } from '../headers';
 import * as validateAgainstSchemaModule from '../utils';
+import { assertRight } from '@stoplight/prism-core/src/utils/__tests__/utils';
 
 describe('HttpHeadersValidator', () => {
   const httpHeadersValidator = new HttpHeadersValidator(registry, 'header');
```
lag-of-death: so, for this PR validator classes like this - `HttpHeadersValidator` - stay classes and operate on `this` inside theirs `validate` behavior?

lag-of-death:
@XVincentX, shouldn't the refactor address `validators` vs `serializers` in the first place? Is this something that will be addressed in the next PRs?

packages/core/src/factory.ts
XVincentX: Fundamentally no matter if the validation fails or succeeds — we still want to call the mock method with empty validation array or not — and the only way to make the chain continue is to make sure the Either returned is a right.

The successive step would be to break down the `mock` method into the `mockSuccess` and `mockFailure` (or something similar) so that we can properly use the fp-ts flows.

I still do not have a clear plan about it though, but this is the direction we should be taking.

packages/core/src/factory.ts
XVincentX: Let me know once you clarify the question so I can look into.

packages/core/src/factory.ts
XVincentX: Good point. 277d9b7

packages/http/src/validator/validators/__tests__/headers.spec.ts
XVincentX: Correct. The removing of the class would be a following PR, I do not want to bloat this too much or we'll end up like the Proxy PR that was huge and took ages to review and merge.

XVincentX:
> shouldn't the refactor address validators vs serializers in the first place? Is this something that will be addressed in the next PRs?

This will be a follow up PR. This is just the beginning of breaking it down so that we can easily at least compose the steps (route, validate input, mock, validate output, send response) that is required for Hosted Prism. We'll iterate and move on as we go.

packages/core/src/factory.ts
lag-of-death: it was more of a `note to self`, no question here, I'm good :+1:

packages/http/src/validator/index.ts
XVincentX: We can definitely do that but why?

lag-of-death: Composing `validateBody` and running the sequence seems like two responsibilities to me. If we can spot a distinct piece of code, that we could move out and hence lower the complexity of `validateInput`, then why not go for it?

XVincentX: Aaaaaaah now I get what you mean. It totally makes sense! Will do

XVincentX: Done!

azure-pipelines.yml
```diff
@@ -27,13 +27,11 @@ jobs:
 
       - script: npx pkg --output ./cli-binaries/prism-cli ./packages/cli/
         displayName: Create binary
-
-      - bash: yarn test.harness
-        displayName: Run harness
-        condition: not(always())
+        condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
```
lag-of-death: is this change needed because of the refactor?

azure-pipelines.yml
```diff
@@ -27,13 +27,11 @@ jobs:
...
 
       - publish: ./cli-binaries/
         artifact: prism-cli-win.exe
+        condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
```
lag-of-death: ditto

packages/core/src/factory.ts
```diff
@@ -14,60 +18,66 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
   defaultConfig: Config,
   components: IPrismComponents<Resource, Input, Output, Config>
 ): IPrism<Resource, Input, Output, Config> {
+  type ResourceAndValidation = {
+    resource: Resource;
+    inputValidations: IPrismDiagnostic[];
+  };
+
+  const inputValidation = (
```
lag-of-death: something more verb-ish, like:

```suggestion
  const getInputValidation = (
```

packages/core/src/factory.ts
```diff
@@ -14,60 +18,66 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
   defaultConfig: Config,
   components: IPrismComponents<Resource, Input, Output, Config>
 ): IPrism<Resource, Input, Output, Config> {
+  type ResourceAndValidation = {
```
lag-of-death: just to be sure, this was intentional for the type to be put inside `factory` ?

packages/core/src/utils/security/index.ts
lag-of-death: @XVincentX, I must stay that's interesting. So how does `e => [e]` guarantee that? I'm just asking out of curiosity :+1:

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -71,80 +73,44 @@ describe('HttpValidator', () => {
...
-              resourceExtension,
-            ),
-            element: { method: 'get', url: { path: '/' } },
-          }),
-        ).toHaveLength(length);
-      };
-
       describe('request is not set', () => {
-        it('validates headers', validate(undefined, 3));
+        it('does not validate headers', validate(undefined, undefined, 0));
```
lag-of-death: there's a couple of these, does it mean that there's a user-facing change in Prism behavior or is it just a part of the refactor?

azure-pipelines.yml
XVincentX: ditto

packages/http/src/validator/validators/utils.ts
```diff
@@ -35,3 +39,6 @@ export const validateAgainstSchema = (value: any, schema: JSONSchema, prefix?: s
     throw new Error(`AJV validation error: "${error}"`);
   }
 };
+
+export const sequenceValidation = sequenceT(getValidation(getSemigroup<IPrismDiagnostic>()));
```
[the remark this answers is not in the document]
XVincentX: In reality there are at least 3 new concepts here:

`sequenceT` and `sequenceS` — and then the `getValidation` that modifies the `ap` method of the `applicative` that the `Either` is implementing; from there, you then instruct the applicative how to concatenate instead of stopping by using a `semigroup`.

Do you want more? :trollface: 


> That line opened a whole new chapter of my FP learning.

We're just getting started.

packages/http/src/mocker/__tests__/functional.spec.ts
```diff
@@ -2,7 +2,7 @@ import * as Ajv from 'ajv';
 
 import { createLogger } from '@stoplight/prism-core';
 import { httpOperations, httpRequests } from '../../__tests__/fixtures';
-import { assertLeft, assertRight } from '../../__tests__/utils';
+import { assertRight, assertLeft } from '@stoplight/prism-core/src/utils/__tests__/utils';
```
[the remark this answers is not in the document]
XVincentX: It was not.

packages/core/src/factory.ts
XVincentX: It is, because the `Resource` generic type is defined there. Putting it outside would not help. This is another place we'll need to break, but not in this PR.

packages/core/src/types.ts
```diff
@@ -16,13 +17,16 @@ export interface IPrismConfig {
   validateResponse: boolean;
 }
 
+export type ValidatorFn<Resource, T> = (opts: {
+  resource: Resource;
+  element: T;
+}) => Either<NonEmptyArray<IPrismDiagnostic>, T>;
```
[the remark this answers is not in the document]
XVincentX: Absolutely. I've felt the same pain multiple times and it's time to separate the deserialisation and the validation process. We will sit down and have a conversation about this. This PR is just the beginning.

packages/http/src/validator/validators/params.ts
```diff
@@ -40,14 +44,20 @@ export class HttpParamsValidator<Target> implements IHttpValidator<Target, IHttp
...
+      })
     );
 
-    return validateAgainstSchema(parameterValues, schema, prefix).concat(deprecatedWarnings);
+    const errors = validateAgainstSchema(parameterValues, schema, prefix).concat(deprecatedWarnings);
+
+    return pipe(
+      fromArray(errors),
+      Either.fromOption(() => target),
+      Either.swap
```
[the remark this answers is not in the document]
XVincentX: Unfortunately, not at the moment. It's not *that* bad, but it can be better. We will iterate on this.

packages/http/src/validator/__tests__/index.spec.ts
XVincentX: Good question. So previously our validators were ran even if there was no need to run them. You can see it here: https://github.com/stoplightio/prism/pull/731/files#diff-52fafd78c752aca126e4fd8c4ce567a6L45-L47

You can see that the functions were called anyway even if there was nothing to validate. In the following lines instead you can see there's a check to make sure it's the case to call them or otherwise call directly `Either.right`

In this tests, the validators were mocked to return an error in any case, so no matter what was happening, all the validators were always called and returning an error. Body + Query + Headers = 3 validators — that explains the number.

PR #743, opened 2019-10-25
feat: colorize values of params

### LINKED ISSUE #696: Colorized parameters in operation examples
**User story.**
As a Prism user, I can parameters in example operations to be colorized, so that I can distinguish them from fixed URL parts.

It is not easy to distinguish variable params from fixed parts of URL without looking into spec. Especially for string-based params.

Generated parameter values should be colorized.

**Additional context**
Which part is param and which is fixed? You got 1 second to answer.

Screen Shot 2019-10-11 at 14 29 36

lag-of-death:
Closes #696

## Checklist

_feature_

Currently, values for params are in the same color the whole path is.
Now, the values are in bold and colored.

No.

colored_params


packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  POST_PARAM_VALUE_TAG,
+  attachTagsToParamsValues,
+  transformPathParamsValues
+} from '../colorizer';
+
+describe('colorizer', () => {
+
+  describe('transformPathParamsValues()', () => {
+    it('colorizes tagged values of query params', () => {
+      const path = `/no_auth/pets/findByStatus?status=${PRE_PARAM_VALUE_TAG},sold,pending,${POST_PARAM_VALUE_TAG}`;
```
philsturgeon: This is a good test but I think it is based on a bug? This status is an enum and should only ever contain one thing, but we're passing in all the things. Functionally it doesnt matter because it is just an arbitrary string, but it makes it look like this sort of thing is intentional.

XVincentX:
Looks very promising! I'll take a look at the code tomorrow.

packages/cli/src/util/__tests__/colorizer.spec.ts
lag-of-death: @philsturgeon, it might be a bug. But maybe it was intended @StefanDywersant? cc @XVincentX

StefanDywersant: @philsturgeon I agree that it does not look correct, but it is perfectly correct. See the schema of `status` parameter (https://github.com/stoplightio/prism/blob/master/examples/petstore.oas3.json#L112). It is array of enums.

lag-of-death: @StefanDywersant, @philsturgeon pointed out that "This status is an enum and should only ever contain one thing, but we're passing in all the things.", so it is an enum, but it still should only produce one value for the generated path to be valid (?)

packages/cli/src/util/createServer.ts
```diff
@@ -78,13 +79,14 @@ async function createPrismServerWithLogger(options: CreateBaseServerOptions, log
 
   operations.forEach(resource => {
     const path = pipe(
-      createExamplePath(resource),
+      createExamplePath(resource, attachTagsToParamsValues),
```
StefanDywersant: Why not push only one function to `createPathExample` which will cover the whole colorizing process? WDYT?

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Dictionary } from '@stoplight/types';
+
+type ParamValue = number | string | string[] | unknown;
+export type ValuesTransformer = (values: Dictionary<ParamValue>) => Dictionary<string | string[]>;
+
+export const PRE_PARAM_VALUE_TAG = 'PRE_PARAM_VALUE_TAG';
+export const POST_PARAM_VALUE_TAG = 'POST_PARAM_VALUE_TAG';
+
+export function transformPathParamsValues(path: string, transform: (aString: string) => string): string {
+  const taggedParamsValues = new RegExp(`(${PRE_PARAM_VALUE_TAG},?)(.*?)(,?${POST_PARAM_VALUE_TAG})`, 'gm');
```
StefanDywersant: I'd be cautious with omitting those `,`. What if a value is exactly `,,,,,`?

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,24 @@
...
+  return path.replace(taggedParamsValues, transform('$2'));
+}
+
+export const attachTagsToParamsValues: ValuesTransformer = values => {
+  return mapValues(values, attachPrePostTags);
+};
+
+const attachPrePostTags = (paramValue: ParamValue) => {
+  return isArray(paramValue)
+    ? [PRE_PARAM_VALUE_TAG].concat(uniq(paramValue)).concat(POST_PARAM_VALUE_TAG)
```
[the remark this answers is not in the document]
lag-of-death: Good point! :+1:

packages/cli/src/util/createServer.ts
```diff
@@ -78,13 +79,14 @@ async function createPrismServerWithLogger(options: CreateBaseServerOptions, log
...
       Either.getOrElse(() => resource.path)
     );
 
-    logInstance.note(`${resource.method.toUpperCase().padEnd(10)} ${address}${path}`);
+    logInstance.note(
+      `${resource.method.toUpperCase().padEnd(10)} ${address} ${transformPathParamsValues(path, chalk.bold.cyan)}`
```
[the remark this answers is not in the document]
lag-of-death: yeah, but I can revert it :+1:

packages/cli/src/util/createServer.ts
lag-of-death: @StefanDywersant, do you mean `createExamplePath` and `transformPathParamsValues`? If yes, then I don't how we could do this with only one function. It would be possible if https://github.com/stoplightio/prism/blob/d13c4742c4b21d1f5da0c8f013f2a3e091ec353d/packages/cli/src/util/paths.ts#L36 was not serializing the path. Then, the colorization could take place inside `attachTagsToParamsValues` (under a different name of course ;>). But maybe you had something else in mind? Or maybe you have some solution in mind?

packages/cli/src/util/colorizer.ts
lag-of-death: I'll add some tests for this use case, thanks!

lag-of-death:
@StefanDywersant, I've addressed the feedback. Colorizing is now optional as `transformValues` now defaults to `_.identity`. Also, `_.uniq` was removed completely, I must have been a left-over. Do you think this version is OK, or you'd rather see some more changes? Also, what about https://github.com/stoplightio/prism/pull/743#discussion_r339491634 ?

packages/cli/src/util/__tests__/colorizer.spec.ts
philsturgeon: Ha! Ok it's correct! Just looks super odd :D Thanks for clearing that up.

packages/cli/src/util/createServer.ts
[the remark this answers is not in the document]
XVincentX: Please do — if you put the extra space the terminals won't be able to let you use the whole URL

image

You can see the selection stops at `http://localhost:4010`

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -12,7 +13,7 @@ describe('createExamplePath()', () => {
           method: 'get',
           request: { path: [{ name: 'p', style: HttpParamStyles.Simple, examples: [{ key: 'foo', value: 'test' }] }] },
           responses: [{ code: '200' }],
-        }),
+        }, identity),
```
XVincentX: The parameter is optional and defaulted to identity, you can remove this.

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -25,7 +26,7 @@ describe('createExamplePath()', () => {
           method: 'get',
           request: { path: [{ name: 'p', style: HttpParamStyles.Label, examples: [{ key: 'foo', value: 'test' }] }] },
           responses: [{ code: '200' }],
-        }),
+        }, identity),
```
XVincentX: The parameter is optional and defaulted to identity, you can remove this.

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -38,7 +39,7 @@ describe('createExamplePath()', () => {
           method: 'get',
           request: { path: [{ name: 'p', style: HttpParamStyles.Matrix, examples: [{ key: 'foo', value: 'test' }] }] },
           responses: [{ code: '200' }],
-        }),
+        }, identity),
```
XVincentX: ditto

packages/cli/src/util/__tests__/paths.spec.ts
```diff
@@ -53,7 +54,7 @@ describe('createExamplePath()', () => {
           method: 'get',
           request: { query: [{ name: 'p', style: HttpParamStyles.Form, examples: [{ key: 'foo', value: 'test' }] }] },
           responses: [{ code: '200' }],
-        }),
+        }, identity),
```
XVincentX: yeah you get the point

packages/cli/src/util/createServer.ts
lag-of-death: @XVincentX, I don't think this is weird. With this solution, `createExamplePath`, is open for extension because the dependency is injected and hence it is loosely coupled. But if you really rather see `createExamplePath` strongly dependent on `attachTagsToParamsValues`, then LMK.

packages/cli/src/util/paths.ts
```diff
@@ -29,7 +33,7 @@ export function createExamplePath(operation: IHttpOperation): Either.Either<Erro
         })
       )
     ),
-    Either.map(({ template, values }) => URI.expand(template, values))
+    Either.map(({ template, values }) => URI.expand(template, transformValues(values)))
```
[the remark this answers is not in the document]
lag-of-death: `= identity` was introduced after @StefanDywersant's comment

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  return path.replace(taggedParamsValues, transform('$2'));
+}
+
+export const attachTagsToParamsValues: ValuesTransformer = values => {
+  return mapValues(values, attachPrePostTags);
+};
+
+const attachPrePostTags = (paramValue: ParamValue) => {
+  return [PRE_PARAM_VALUE_TAG].concat(isArray(paramValue) ? paramValue : `${paramValue}`).concat(POST_PARAM_VALUE_TAG);
```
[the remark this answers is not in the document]
lag-of-death: @XVincentX, I'm not sure I follow, basically `.join` on a string or number won't end nice. The list (which here I'm adding the tags to) is, as it was before, provided to `URI.expand` for multiple-value parameters: https://github.com/stoplightio/prism/blob/ba472b9288a287603fed252a482d08ee12f09dc5/packages/cli/src/util/paths.ts#L36

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,22 @@
+import { mapValues, isArray } from 'lodash';
+import { Dictionary } from '@stoplight/types';
+
+type ParamValue = number | string | string[] | unknown;
+export type ValuesTransformer = (values: Dictionary<ParamValue>) => Dictionary<string | string[]>;
+
+export const PRE_PARAM_VALUE_TAG = 'PRE_PARAM_VALUE_TAG';
+export const POST_PARAM_VALUE_TAG = 'POST_PARAM_VALUE_TAG';
```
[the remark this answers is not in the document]
lag-of-death: @XVincentX, it's now `~pre~` and `~post~` since special chars were escaped by `URL.expand` - https://github.com/stoplightio/prism/blob/ba472b9288a287603fed252a482d08ee12f09dc5/packages/cli/src/util/paths.ts#L36

packages/cli/src/util/paths.ts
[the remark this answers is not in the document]
XVincentX: Ah ok, nice!

packages/cli/src/util/colorizer.ts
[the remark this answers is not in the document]
XVincentX: Ok I understand what you're trying to do now. So maybe use `String(paramValue)` so that the conversion is explicit.

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,23 @@
...
+export const PRE_PARAM_VALUE_TAG = '~pre~';
+export const POST_PARAM_VALUE_TAG = '~post~';
+
+export const transformPathParamsValues = ((
+  taggedParamsValues: RegExp,
+  path: string,
+  transform: (aString: string) => string
+): string => {
+  return path.replace(taggedParamsValues, transform('$2'));
+}).bind({}, new RegExp(`(${PRE_PARAM_VALUE_TAG},)(.*?)(,${POST_PARAM_VALUE_TAG})`, 'gm'));
```
XVincentX: Hmm maybe instead of binding you can just create the RegEx ahead of the time as a constant and use it in there — this should improve the readability.

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        it('tags numeric values', () => {
+          const val = 170;
+          const valuesOfParams = {
+            petId: val
+          };
+
+          expect(attachTagsToParamsValues(valuesOfParams)).toStrictEqual({
+            petId: [
+              PRE_PARAM_VALUE_TAG,
+              `${val}`,
```
XVincentX:
```suggestion
              val,
```

packages/cli/src/util/colorizer.ts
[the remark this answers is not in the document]
lag-of-death: @XVincentX, isn't `${paramValue}` explicit? I don't see how `String(paramValue)` serves the purpose better (?)

packages/cli/src/util/__tests__/colorizer.spec.ts
lag-of-death: nah, it should stay as is - as a string

packages/cli/src/util/colorizer.ts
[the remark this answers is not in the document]
XVincentX: I thought it was a little bit cleaner but I do not really mind.

packages/cli/src/util/createServer.ts
XVincentX: I understand your point — but on the other hand at the moment I really do not see any of us extending this function and at this point it's just an additional complexity when reading the code. We can put it back in case the needs arises with the pass of the time.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Path parameters are now validated against schema [#702](https://github.com/stoplightio/prism/pull/702)
 - The Test Harness framework now requires the `${document}` parameter explicitly [#720](https://github.com/stoplightio/prism/pull/720)
 - Prism now includes a new `proxy` command that will validate the request coming in, send the request to an upstream server and then validate the response coming back [#669](https://github.com/stoplightio/prism/pull/669)
+- Prism has values for path/query params bolded and in color [#743](https://github.com/stoplightio/prism/pull/743)
```
XVincentX: Maybe it would be worth to update the `CLI.md` docs as well?

packages/cli/src/util/__tests__/colorizer.spec.ts
lag-of-death: yeah :+1:

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const attachTagsToParamsValues: ValuesTransformer = values => {
+  return mapValues(values, attachPrePostTags);
+};
+
+const attachPrePostTags = (paramValue: unknown) => {
+  return [PRE_PARAM_VALUE_TAG].concat(isArray(paramValue) ? paramValue : `${paramValue}`).concat(POST_PARAM_VALUE_TAG);
```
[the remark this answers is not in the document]
lag-of-death: That's tough. I think a solution could be this:
1. forking `uri-template-lite`
2. changing the signature of `.expand` from `function expand(template, data)` to `function expand(template, data, cb)` and invoke `cb` on a generated value for each of the passed params
3. so it would be sth like `.expand(template, data, chalk.bold.red)`
this way we have an encoded path with the expanded parameters, which values are colored (since the coloring info itself was not escaped) 
WDYT @StefanDywersant, @XVincentX ?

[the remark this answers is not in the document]
philsturgeon: If there is any "nice to have" functionality being considered here, please keep in mind that getting PRs to other forks merged and released can take a long time, so go with a lesser solution until that's done. In general it is better to have "ok" now and "awesome" later, than to try and maintain yet another fork.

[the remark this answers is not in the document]
XVincentX: @lag-of-death Yeah I was thinking about the same solution as well but I think we should not go in such direction

1. This is a job for the URI Template library in general
2. This is something that should be happening in Prism

In theory, instead of forking the library, we could try to look for another one that instead of simply matching and expanding would give us access to the matching data that we could use to reconstruct the URL somehow.

I think already in such state this improvement will provide a lot of value to the users, so I suggest to create a low-priority followup issue that we can work on with the passing of the time.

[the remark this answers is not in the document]
StefanDywersant: @lag-of-death now that I think it could be solved by wrapping in ~pre~/~post~ each value of array.

More or less sth like that:
```suggestion
  return isArray(paramValue) ? paramValue.map(v => `${PRE_PARAM_VALUE_TAG}${v}${POST_PARAM_VALUE_TAG}`) : `${PRE_PARAM_VALUE_TAG}${paramValue}${POST_PARAM_VALUE_TAG}`;
```

This way we should colorize only the actual values, not the stuff around.

Another option worth checking is to wrap whole uri templates into colorization tags. See here: https://github.com/stoplightio/prism/blob/feature/colorize-params-values/packages/cli/src/util/paths.ts#L180. If you do `~pre~{?${formExplodedParams}*}~post~` then it should also work.

The bright side is: it is possible that expand will not touch special characters, so you could possibly put colorizing tags immediately (worth checking!).
The dark side is: it won't colorize values only, for some complex examples (matrix & explode) it will colorize bit more (**;id=1;id=2;id=3** instead of ;id=**1**;id=**2**;id=**3**).

I'd check above.

----

@XVincentX @philsturgeon  I think it is worth solving because leaving it as is will introduce bug with the wrong path being generated. However, those two solutions described above should be fairly easy to implement without forking external libs.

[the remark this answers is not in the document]
lag-of-death: @StefanDywersant, I went for the first proposed solution, thanks! :+1:

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        errors: false,
+        multiprocess: false,
+        host: 'localhost',
+        port: 9999,
+        cors: false,
+        document: './examples/test.yaml',
+        dynamic: false
+      }, logInstance)
+    })
+  }
```
XVincentX: I would remove this test.

1. It has an `if` statement that is usually not ok for a test; in case the colours are not supported on the testing platform, this whole thing is dropped
2. It's doing weird things (ts-ignore, doing assertions in `mockImplementation` functions) which makes it hard to read and understand.

In general I feel the value of testing this compared to the cost of such test is too low to justify them.

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+import { createLogger } from '@stoplight/prism-core';
+import * as httpServer from '@stoplight/prism-http-server';
+
+const logStream = new PassThrough();
+const logInstance = createLogger('CLI', undefined, logStream);
+
+describe('colorizer', () => {
+
+  if (chalk.supportsColor) {
+    it('colorizes values for parameters', async () => {
```
XVincentX: You can remove the async…

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      jest.spyOn(logInstance, 'note').mockImplementation((aString) => {
+        // @ts-ignore
+        const coloredParam = aString.match(/(endpoint\/)(.*)/)[2];
+
+        expect(coloredParam).toContain("\u001b[1m\u001b[36m");
+        expect(coloredParam).toContain("\u001b[39m\u001b[22m");
+      });
+
+      await createPrismServerWithLogger({
```
XVincentX: …and return instead of awaiting.

```suggestion
      return createPrismServerWithLogger({
```

examples/test.yaml
```diff
@@ -0,0 +1,12 @@
...
+  /endpoint/{val}:
+    get:
+      parameters:
+        - in: path
+          name: val
+          schema:
+            type: string
+      responses:
+        200:
+          description: abc
```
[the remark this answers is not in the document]
lag-of-death: Not anymore because of https://github.com/stoplightio/prism/pull/743#discussion_r340919426 
`examples/test.yaml` was only needed for that test

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,23 @@
+import { mapValues, isArray } from 'lodash';
```
StefanDywersant:
```suggestion
import { isArray, mapValues } from 'lodash';
```

Our linter R.I.P.

packages/cli/src/util/createServer.ts
```diff
@@ -50,7 +51,7 @@ async function createSingleProcessPrism(options: CreateBaseServerOptions) {
   }
 }
 
-async function createPrismServerWithLogger(options: CreateBaseServerOptions, logInstance: Logger) {
+export async function createPrismServerWithLogger(options: CreateBaseServerOptions, logInstance: Logger) {
```
StefanDywersant: Unneeded export.

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Dictionary } from '@stoplight/types';
+
+export type ValuesTransformer = (values: Dictionary<unknown>) => Dictionary<string | string[]>;
+
+export const PRE_PARAM_VALUE_TAG = '~pre~';
+export const POST_PARAM_VALUE_TAG = '~post~';
+
+const taggedParamsValues = new RegExp(`(${PRE_PARAM_VALUE_TAG})(.*?)(${POST_PARAM_VALUE_TAG})`, 'gm');
```
StefanDywersant:
```suggestion
const taggedParamsValues = new RegExp(`${PRE_PARAM_VALUE_TAG}(.*?)${POST_PARAM_VALUE_TAG}`, 'gm');
```

The capture groups around tags are not needed.

packages/cli/src/util/colorizer.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const transformPathParamsValues = (path: string, transform: (aString: string) => string): string => {
+  return path.replace(taggedParamsValues, transform('$2'));
```
StefanDywersant: After removing unneeded capture capture groups you can:

```suggestion
  return path.replace(taggedParamsValues, transform('$1'));
```

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    it('colorizes tagged values of path params', () => {
+      const paramVal = 651;
+      const path = `/no_auth/pets/${PRE_PARAM_VALUE_TAG}${paramVal}${POST_PARAM_VALUE_TAG}`;
+
+      expect(transformPathParamsValues(path, chalk.bold.blue)).toBe(`/no_auth/pets/${chalk.bold.blue(`${paramVal}`)}`);
+    });
+
+    it('colorizes values that are equal to multiple or single , characters', () => {
```
StefanDywersant: We don't need that test anymore since you have changed the way of handling param arrays.

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+          const valuesOfParams = {
+            petId: val
+          };
+
+          expect(attachTagsToParamsValues(valuesOfParams)).toStrictEqual({
+            petId: `${PRE_PARAM_VALUE_TAG}${val}${POST_PARAM_VALUE_TAG}`
+          });
+        });
+
+        it('tags values that are equal to multiple or single , characters', () => {
```
StefanDywersant: ditto we don't need anymore

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      describe('tagging single values', () => {
+        it('tags string values', () => {
+          const val = 'dignissimos';
+          const valuesOfParams = {
+            name: val
+          };
+
+          expect(attachTagsToParamsValues(valuesOfParams)).toStrictEqual({
+            name: `${PRE_PARAM_VALUE_TAG}${val}${POST_PARAM_VALUE_TAG}`,
```
StefanDywersant:
```suggestion
            name: `${PRE_PARAM_VALUE_TAG}dignissimos${POST_PARAM_VALUE_TAG}`,
```

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        it('tags numeric values', () => {
+          const val = 170;
+          const valuesOfParams = {
+            petId: val
+          };
+
+          expect(attachTagsToParamsValues(valuesOfParams)).toStrictEqual({
+            petId: `${PRE_PARAM_VALUE_TAG}${val}${POST_PARAM_VALUE_TAG}`
```
StefanDywersant:
```suggestion
            petId: `${PRE_PARAM_VALUE_TAG}170${POST_PARAM_VALUE_TAG}`
```

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    it('colorizes tagged values of query params', () => {
+      const paramVal = 'sold,pending';
+      const path = `/no_auth/pets/findByStatus?status=${PRE_PARAM_VALUE_TAG}${paramVal}${POST_PARAM_VALUE_TAG}`;
+
+      expect(transformPathParamsValues(path, chalk.bold.blue)).toBe(`/no_auth/pets/findByStatus?status=${chalk.bold.blue(paramVal)}`);
+    });
+
+    it('colorizes tagged values of path params', () => {
+      const paramVal = 651;
+      const path = `/no_auth/pets/${PRE_PARAM_VALUE_TAG}${paramVal}${POST_PARAM_VALUE_TAG}`;
```
StefanDywersant:
```suggestion
      const path = `/no_auth/pets/${PRE_PARAM_VALUE_TAG}651${POST_PARAM_VALUE_TAG}`;
```

packages/cli/src/util/__tests__/colorizer.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+} from '../colorizer';
+
+describe('colorizer', () => {
+
+  describe('transformPathParamsValues()', () => {
+    it('colorizes tagged values of query params', () => {
+      const paramVal = 'sold,pending';
+      const path = `/no_auth/pets/findByStatus?status=${PRE_PARAM_VALUE_TAG}${paramVal}${POST_PARAM_VALUE_TAG}`;
+
+      expect(transformPathParamsValues(path, chalk.bold.blue)).toBe(`/no_auth/pets/findByStatus?status=${chalk.bold.blue(paramVal)}`);
```
StefanDywersant:
```suggestion
      expect(transformPathParamsValues(path, chalk.bold.blue)).toBe(`/no_auth/pets/findByStatus?status=${chalk.bold.blue('sold,pending')}`);
```

PR #776, opened 2019-11-06
fix(build): Use ttsc for compiling

BinaryMuse:
#701 introduced a change to fix imports like `"../../core/src"` showing up in `.d.ts` files. However, if you install the latest version (v3.1.1) of `stoplight/prism-cli`, you'll see that the relative imports still exist in the built package, as no version has been published after this change.

Additionally, the `build` step in `package.json` still reads: `tsc --build ./packages/tsconfig.build.json`; this PR updates it to use `ttsc` instead, so that type files are created correctly when building locally. The CircleCI config has also been updated to use this build script instead of the one-off `npx` command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ jobs:
           command: yarn --frozen-lockfile
       - run:
           name: Build all code to JavaScript
-          command: npx ttsc --build ./packages/tsconfig.build.json
+          command: yarn build
       - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
       - run:
           name: Publish
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "lint": "eslint ./packages/**/src/**/*.ts",
     "lint.fix": "yarn lint --fix",
-    "build": "tsc --build ./packages/tsconfig.build.json",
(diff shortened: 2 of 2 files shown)


XVincentX:
Hey @BinaryMuse 

> However, if you install the latest version (v3.1.1) of stoplight/prism-cli, you'll see that the relative imports still exist in the built package, as no version has been published after this change.

This is correct! We haven't been able to publish (yet) 3.2.0 as we're doing some final e2e tests. However, we have a tagged beta that you can install in case you want to try it out right now.

Thank for the PR, I missed this step!

rmosolgo:
Hey, yes, I'm trying to build a _dynamic_ mock server that, given a sha, can fetch the OpenAPI for that sha, then serve mock requests 😅  I know it's a bit off the beaten track for this project, but since you all have done so much great work in this area already, it'd be silly for me to reinvent the wheel without trying this out first. 

For the moment, it works fine: it's just a very small OpenAPI spec, so fetching, parsing and mocking in a single request cycle is fine. But we'll see how it grows 😆 

This was my first go at TypeScript, so I called for help when I encountered the error mentioned above. Thanks for your work on this! I'll keep an eye out for the new release.

XVincentX:
Ah nice, understood. So basically what's making you use Prism programmatically is the data source. Neat stuff!

PR #777, opened 2019-11-06
fix: trigger media type validation only when defined in the response

### LINKED ISSUE #774: Do not validate the output media type if the document does not specify any content
@StefanDywersant ahh well this just shows that my next job of "play with proxy server and contract testing to make sure it works as expected" should have been done sooner. My bad. 

If the API does not specify content, then anything which comes in is a bonus, maybe a warning or an info, but certainly not an error. It's the sort of thing we'd report to folks via Studio somehow (long way off) saying "Hey we found some content, looks like this is JSON, here is the schema. Wanna save that to your description?" etc.

How do I make a callback which says "I dont care what they respond with im just looking for a 200? A lot of callbacks work this way, they just want a "Yep!" and nothing more.

_Originally posted by @philsturgeon in https://github.com/stoplightio/prism/pull/716#issuecomment-550370944_

XVincentX:
Closes #774

* Calls the media type return validator only if the selected response has contents with a media type
* Uses now `typeIs.is` function library so that we can match stuff with charset headers as well instead of a simple equality check

### DIFF
--- a/packages/http/src/validator/index.ts
+++ b/packages/http/src/validator/index.ts
@@ -1,9 +1,16 @@
 import { IPrismDiagnostic, ValidatorFn } from '@stoplight/prism-core';
-import { DiagnosticSeverity, IHttpOperation, IHttpOperationResponse, IHttpOperationRequest } from '@stoplight/types';
+import {
+  DiagnosticSeverity,
+  IHttpOperation,
+  IHttpOperationResponse,
+  IHttpOperationRequest,
+  IMediaTypeContent,
+} from '@stoplight/types';
 import * as caseless from 'caseless';
-import { findFirst } from 'fp-ts/lib/Array';
+import { findFirst, isNonEmpty } from 'fp-ts/lib/Array';
 import * as Option from 'fp-ts/lib/Option';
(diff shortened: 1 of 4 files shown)


packages/http/src/validator/index.ts
```diff
@@ -84,14 +91,13 @@ const findResponseByStatus = (responses: NonEmptyArray<IHttpOperationResponse>,
...
-    Option.fromNullable(response.contents),
-    Option.chain(findFirst(c => c.mediaType === mediaType)),
+    contents,
+    findFirst(c => !!typeIs.is(mediaType, [c.mediaType])),
     Either.fromOption<IPrismDiagnostic>(() => ({
-      message: `The received media type "${
-        mediaType ? mediaType : ''
-      }" does not match the one specified in the document`,
+      message: `The received media type "${mediaType ||
+        ''}" does not match the one specified in the current response: ${contents.map(c => c.mediaType).join(',')}`,
```
StefanDywersant:
```suggestion
        ''}" does not match the one${contents.length ? 's' : ''} specified in the current response: ${contents.map(c => c.mediaType).join(', ')}`,
```

PR #782, opened 2019-11-08
Make the internal http client return the raw TaskEither and not the Promise

XVincentX:
This PR modifies the internal client we're using in the http server and in some other places to no throw exceptions anymore.

Since this is an _internal_ client, nobody should use it and so nobody will get hurt. The *external* client stays the same.

I've made all the necessary changes that might make the code uglier at this point, but the idea would be that we need to ditch fastify since:

1. We're using 1% of its features
2. Some of the TypeScript definitions are wrong
3. It's not friendly to the programming style we're using.

Most likely, we're going to use `micri` which proved to be small and barebone/close to the metal enough. In meantime, we'll live with this.


packages/core/src/types.ts
```diff
@@ -45,8 +45,7 @@ export interface IPrismInput<I> {
   validations: IPrismDiagnostic[];
 }
 
-export interface IPrismOutput<I, O> {
-  input: I;
+export interface IPrismOutput<O> {
```
StefanDywersant: Note to self: why?

packages/http/README.md
```diff
@@ -71,20 +71,18 @@ const prism = Prism.createInstance({
...
-  .process(
-    {
-      method: 'get',
-      url: {
-        path: '/todos',
-      },
-      headers: {
-        Accept: 'text/plain',
-      },
+return prism.process(
```
StefanDywersant:
```suggestion
return prism.request(
```

packages/http/README.md
```diff
@@ -71,20 +71,18 @@ const prism = Prism.createInstance({
...
     },
-    operations
-  )
-  .then(prismResponse => console.log(prismResponse.output));
+    headers: {
+      Accept: 'text/plain',
+    },
+  },
+  operations
+);
```
StefanDywersant:
```suggestion
)();
```
+ `TaskEither.fold` should be there?

packages/http/README.md
```diff
@@ -110,24 +108,22 @@ const Prism = require('@stoplight/prism-http');
...
 // Make a "GET /todos" request
-return prism
-  .process(
-    {
-      method: 'get',
-      url: {
-        path: '/facts',
-        baseUrl: 'https://cat-fact.herokuapp.com',
-      },
+return prism.process(
```
StefanDywersant:
```suggestion
return prism.request(
```

packages/http/README.md
```diff
@@ -110,24 +108,22 @@ const Prism = require('@stoplight/prism-http');
...
+    // We can override the default behaviour per request.
+    mock: {
+      dynamic: true,
     },
-    operations
-  )
-  .then(prismResponse => console.log(prismResponse.output));
+  },
+  operations
+);
```
StefanDywersant:
```suggestion
)();
```
.. and `TaskEither.fold` should be there?

packages/http/README.md
```diff
@@ -316,18 +310,16 @@ const Prism = require('@stoplight/prism-http');
...
+    url: {
+      path: '/facts',
+      baseUrl: 'https://cat-fact.herokuapp.com',
     },
-    operations
-  )
-  .then(prismResponse => console.log(prismResponse.validations.input));
+  },
+  operations
+);
```
StefanDywersant: ditto

packages/http/README.md
```diff
@@ -316,18 +310,16 @@ const Prism = require('@stoplight/prism-http');
...
 const prism = Prism.createInstance(config);
-prism
-  .process(
-    {
-      method: 'get',
-      url: {
-        path: '/facts',
-        baseUrl: 'https://cat-fact.herokuapp.com',
-      },
+prism.process(
```
StefanDywersant:
```suggestion
prism.request(
```

packages/http/README.md
```diff
@@ -141,15 +137,13 @@ const Prism = require('@stoplight/prism-http');
...
-    },
-  })
-  .then(prismResponse => console.log(prismResponse.output));
+return prism.request({
+  method: 'get',
+  url: {
+    path: '/facts',
+    baseUrl: 'https://cat-fact.herokuapp.com',
+  },
+});
```
StefanDywersant: ditto

packages/http-server/src/server.ts
```diff
@@ -60,72 +63,78 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
-        );
-        if (opts.errors && errorViolations.length > 0) {
-          throw ProblemJsonError.fromTemplate(
-            VIOLATIONS,
-            'Your request/response is not valid and the --errors flag is set, so Prism is generating this error for you.',
-            { validation: errorViolations }
+          const errorViolations = inputOutputValidationErrors.filter(
+            v => v.severity === DiagnosticSeverity[DiagnosticSeverity.Error]
           );
+          if (opts.errors && errorViolations.length > 0) {
```
StefanDywersant: I believe this code still needs some love and we will do it later?

packages/http/src/__tests__/http-prism-instance.spec.ts
```diff
@@ -65,43 +66,49 @@ describe('Http Client .request', () => {
...
 
     describe('baseUrl not set', () => {
-      it('ignores server validation and returns 200', async () => {
-        const result = await prism.request(
-          {
-            method: 'get',
-            url: {
-              path: '/pet',
+      it('ignores server validation and returns 200', () => {
+        assertResolvesRight(
```
StefanDywersant: Hello, my lovely helper function :)

packages/core/src/types.ts
StefanDywersant: I guess `Input` wasn't used, right?

XVincentX: Correct, it was not used at all

packages/http-server/src/server.ts
XVincentX: Yes, this was the minimal part to move it over. We'll keep iterating on this.

packages/http/src/__tests__/http-prism-instance.spec.ts
XVincentX: That was a good one! 👍

packages/http/README.md
XVincentX: Yes, I'll update the docs.

XVincentX:
> Code looks good, except the big nest of if's in server.ts, but I guess that is the part that will be reworked later.

Yes, we also need to dismantle this part of the application and make it easier.

> In order to this comment be more constructive: Maybe we should introduce two interfaces. Classic and FP?

This is happening already. We basically have two clients: the internal one (which is the one being modified by this Pull Request) and the public one (which is in the `client.ts` file); this one has not been modified at all. Does this clarify?

packages/http/README.md
```diff
@@ -97,6 +95,26 @@ Output
...
+        url: {
+          path: '/todos',
+        },
+        headers: {
+          Accept: 'text/plain',
+        },
+      },
+      operations
+    ),
+    TaskEither.fold(console.error, data => console.log(data));
```
StefanDywersant:
```suggestion
    TaskEither.fold(console.error, console.log);
```

XVincentX:
@StefanDywersant The client has a dedicated help page.

PR #796, opened 2019-11-09
Added text/html regex to plain serializer

### LINKED ISSUE #795: Cannot find serializer for text/html
**Describe the bug**
Unable to render `text/html` responses containing objects (per OAS 3.x, I can have objects within `text/html` block). Getting error:

```json
{
    "detail": "",
    "status": 500,
    "title": "Cannot find serializer for text/html",
    "type": "https://stoplight.io/prism/errors#UNKNOWN"
}
```
When making a valid request to schema path, having valid response within `text/html` content type, the response does not get parsed and error is returned.
 
**To Reproduce**
====spec====
```yaml
openapi: "3.0.1"
paths:
  "/path":
      get:
        responses:
          '200':
            content:
              text/text:
                schema:
                  properties:
                      status:
                        type: string
                      required:
                        - status
                  type: object
```

**Expected behavior**
I should be able to see the response in with the header type `text/html` in string if it have a valid response

 - Library version: 3.0.0
 - OS: MacOS

codeasashu:
Closes #795


packages/http-server/src/serialize.ts
```diff
@@ -19,7 +19,7 @@ const serializers = [
     serializer: (data: unknown) => (typeof data === 'string' ? data : xmlSerializer.parse({ xml: data })),
   },
   {
-    regex: /text\/plain/,
+    regex: /text\/(plain|html)\b/,
```
XVincentX: I think we should probably use `type-is` instead of an regex so we can safely forward all the text types

packages/http-server/src/__tests__/fixtures/content-server-example.oas3.yaml
```diff
@@ -0,0 +1,52 @@
...
+      enum:
+      - http
+      - https
+    host:
+      default: stoplight.io
+      enum:
+      - stoplight.io
+      - dev.stoplight.io
+    basePath:
+      default: api
```
XVincentX: We usually try to produce really minimal documents so that we can easily understand what this test is about

packages/http-server/src/__tests__/example-validation.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        method: 'GET',
+        url: '/pet/plain',
+        headers: {
+          Accept: '*/*',
+        },
+      });
+
+      expect(response.statusCode).toBe(200);
+      expect(response.payload).toMatch('Pet Example');
+    });
```
XVincentX: When it comes to these type of tests we usually write an harness one, whose usage is documented in the harness directory.

packages/http-server/src/serialize.ts
codeasashu: Sure,  but I am unaware of type-is lib. Can you point me to the documentation link for it?

packages/http-server/src/__tests__/fixtures/content-server-example.oas3.yaml
codeasashu: Agreed. I copy pasted the other fixture for this, I will get rid of unnecessary stuff and will produce a minimal one

packages/http-server/src/__tests__/example-validation.spec.ts
codeasashu: I will see and add the harness test for this. I am quite new to these tests hence will require some assistance

packages/http-server/src/serialize.ts
XVincentX: It is

packages/http-server/src/__tests__/example-validation.spec.ts
XVincentX: The readme should be pretty self explanatory, but feel free to ask any question in case something is not clear.

packages/http-server/src/serialize.ts
XVincentX: It's already a dependency so it shouldn't be a problem to use it.

codeasashu: Sorry, I am not sure what should `toString` method return? Because, I am not sure where it is being used. So should this work?

```js
regex: {
      test: (value: string) => !!typeIs.is(value, ['text/*']),
      toString: () => 'text/*',
    },
```

XVincentX: Put whatever you want — it does not matter.

codeasashu: Thanks, will do

codeasashu:
So, Instead of merging, I've rebased the master so I can easily squash and merge in the end.

XVincentX:
@codeasashu I do not think it's worth to change the serialiser signature just for that reason. Instead, I would probably just reword the message to be "Cannot serialise complex objects as text"

codeasashu:
@XVincentX do you still want to keep the error link to https://stoplight.io/prism/errors#NO_COMPLEX_OBJECT_TEXT_PLAIN ?

XVincentX:
Yeah let's keep it in that way for now please.

codeasashu:
I am unable to run harness tests locally because I somehow deleted the `packages/core/dist` directory and I do not know how to recreate it (any npm command for that?)

XVincentX:
@codeasashu Make sure to remove the package-lock file, we use yarn.

packages/http-server/src/__tests__/example-validation.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      const response = await server.fastify.inject({
+        method: 'GET',
+        url: '/pet/plain',
+        headers: {
+          Accept: '*/*',
+        },
+      });
+
+      expect(response.statusCode).toBe(200);
+      expect(response.payload).toMatch('Pet Example');
```
XVincentX: Remove this test, this is not necessary since we're going to write an harness one

packages/http-server/src/__tests__/fixtures/content-server-example.oas3.yaml
```diff
@@ -0,0 +1,52 @@
...
+      default: http
+      enum:
+      - http
+      - https
+    host:
+      default: stoplight.io
+      enum:
+      - stoplight.io
+      - dev.stoplight.io
+    basePath:
```
XVincentX: Remove this document, it's not required anymore

codeasashu: Fixed in 2fea3a0

packages/http-server/src/__tests__/example-validation.spec.ts
codeasashu: Fixed in 2fea3a0

XVincentX:
@StefanDywersant Can you take a look at this as well to make sure I did not forget anything?

codeasashu:
Do you want me to squash all the changes into single commit?

XVincentX:
@codeasashu We'll take care of it.

PR #802, opened 2019-11-12
fix: treat emtpy Accept header as */*

### LINKED ISSUE #500: OAS3 fails to generate text/plain response when no Accept header passed
**Describe the bug**

**To Reproduce**
```
====test====
Generate simple plain text response
====spec====
openapi: "3.0.1"
paths:
  "/path":
      get:
        responses:
          '200':
            content:
              text/plain:
                schema:
                  type: string
====server====
mock -p 4010
====command====
curl -i -H "Accept:" http://localhost:4010/path
====expect====
HTTP/1.1 200 OK
content-type: text/plain

string
```

**Actual**
body equals to ""

v3.0.3

XVincentX:
This PR modifies the negotiation in the way it looks for content in case the `Accept` header is invalid or not provided.

1. It will look for an _explicit_ `*/*` content type in the current response
2. If not found, it will look for an `application/json` one
3. If not found, it will take the first one on the list.

Closes #500

For some reason Prettier decided to reformat the whole file so please review from here: https://github.com/stoplightio/prism/pull/802/files?utf8=✓&diff=unified&w=1

### DIFF
--- a/packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
+++ b/packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
@@ -668,58 +668,59 @@ describe('NegotiatorHelpers', () => {
 
   describe('negotiateDefaultMediaType()', () => {
     describe('default content', () => {
-      it.each([['*/*', 'application/xml'], ['*/*', 'application/json'], ['application/json', 'application/xml']])(
-        'should return %s even when %s is avaiable',
-        (defaultMediaType, alternateMediaType) => {
-          const code = chance.string();
-          const partialOptions = {
+      it.each([
+        ['*/*', 'application/xml'],
+        ['*/*', 'application/json'],
+        ['application/json', 'application/xml'],
+      ])('should return %s even when %s is avaiable', (defaultMediaType, alternateMediaType) => {
+        const code = chance.string();
+        const partialOptions = {
+          code,
+          dynamic: chance.bool(),
+          exampleKey: chance.string(),
+        };
+
+        const contents: IMediaTypeContent[] = [alternateMediaType, defaultMediaType].map(mediaType => ({
+          mediaType,
+          examples: [],
+          encodings: [],
+        }));
+
+        const response: IHttpOperationResponse = {
+          code,
+          contents,
+          headers: [],
+        };
+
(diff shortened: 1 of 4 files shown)


packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -738,189 +739,230 @@ describe('NegotiatorHelpers', () => {
...
+      assertRight(actualOperationConfig, operationConfig => {
+        expect(operationConfig).toEqual(expectedConfig);
+      });
+    });
 
-        const negotiationResult = helpers.negotiateByPartialOptionsAndHttpContent(partialOptions, httpContent);
+    it('and example not exist should throw an error', () => {
+      const exampleKey = chance.string();
+      const partialOptions = {
+        code: chance.integer({ min: 100, max: 599 }).toString(),
```
StefanDywersant: I'm wondering whether this is good to have non-deterministic test input. If somehow at some point it causes failure then it will be a nightmare to track it down.

I'm curious about your opinion on that?

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -11,7 +11,7 @@ import * as Either from 'fp-ts/lib/Either';
 import { left, right } from 'fp-ts/lib/ReaderEither';
 import { assertRight, assertLeft } from '@stoplight/prism-core/src/__tests__/utils';
 import helpers from '../NegotiatorHelpers';
-import { IHttpNegotiationResult, NegotiationOptions } from '../types';
+import { IHttpNegotiationResult, NegotiationOptions, NegotiatePartialOptions } from '../types';
```
StefanDywersant:
```suggestion
import { IHttpNegotiationResult, NegotiationOptions } from '../types';
```

Unused import

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
XVincentX: Yes, I'm ok fixing the status code, although there are other pieces using chance everywhere in this file — I do not know what the original intentions were.

XVincentX:
@philsturgeon If you're ok with the logic, please approve and then merge

PR #803, opened 2019-11-13
fix: body and method support in forwarder

StefanDywersant:
## Checklist

Bug fix

Currently:
- proxy does not forward other http methods than `get`
- proxy does not forward `body`

This PR makes proxy forward above.

No

## Note
- I couldn't find a way of testing the body with `httpbin`. The problem is that httpbin returns body containing non-deterministic data. That data includes callee IP (it depends on environment) or curl version (minor/patch can change). I wrote traditional tests instead.
- inside `serializeBody` I'm not throwing when invalid JSON supplied (circular?). It is rather not possible because the payload has been parsed by prism. I know that it is possible to fail when `forwarder` is used in some other environment in the future.
- Not adding CHANGELOG line since it fixes unreleased stuff


packages/http/src/forwarder/index.ts
```diff
@@ -27,14 +28,48 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
...
+                Option.alt(() => Option.some(''))
+              )
+            )
+          )
+        ),
+        Option.alt(
+          () =>
+            pipe(
+              body,
+              Option.fromPredicate(body => typeof body === 'string')
```
XVincentX: Let's use a typeguard instead of a cast

```suggestion
              Option.fromPredicate((body): body is string => typeof body === 'string')
```

packages/http/src/forwarder/index.ts
```diff
@@ -27,14 +28,48 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
...
+            ) as Option.Option<string>
```
XVincentX:
```suggestion
            )
```

packages/http/src/forwarder/index.ts
```diff
@@ -27,14 +28,48 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
...
+        )
+      )
+    ),
+    Option.toUndefined
```
XVincentX: This means that likely the flow hasn't been set correctly.

XVincentX:
It would be really cool to add an harness test that would return back the JSON payload we're sending so we can verify it.

We could use the httpbin's service offers only https://httpbin.org/#/Anything that returns the sent payload, and use the JSON Schema to verify that the returned `json` property is the same as the sent request.

StefanDywersant:
> It would be really cool to add an harness test that would return back the JSON payload we're sending so we can verify it.
> 
> We could use the httpbin's service offers only https://httpbin.org/#/Anything that returns the sent payload, and use the JSON Schema to verify that the returned `json` property is the same as the sent request.

I looked at `anything` and the rest of httpbin endpoints. `anything` returns content wrapped in exactly the same way. Example below:

```
curl -v -H'Content-type: application/json' http://localhost/anything
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 80 (#0)
> GET /anything HTTP/1.1
> Host: localhost
> User-Agent: curl/7.54.0
> Accept: */*
> Content-type: application/json
>
< HTTP/1.1 200 OK
< Server: gunicorn/19.9.0
< Date: Thu, 14 Nov 2019 17:24:38 GMT
< Connection: keep-alive
< Content-Type: application/json
< Content-Length: 308
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
<
{
  "args": {},
  "data": "",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Content-Type": "application/json",
    "Host": "localhost",
    "User-Agent": "curl/7.54.0"
  },
  "json": null,
  "method": "GET",
  "origin": "172.17.0.1",
  "url": "http://localhost/anything"
}
* Connection #0 to host localhost left intact
(karol@MacBook-Pro-Karol ~)$
```

`origin` and `curl/7.54.0` are likely to be non-deterministic.

StefanDywersant:
Oh, I forgot the constructive part ;)

1. leave regular tests
2. replace httpbin
3. improve harness to test against part of the output

PR #804, opened 2019-11-13
Refactor and move security validations

XVincentX:
This PR will address most of the technical debt we had in the security validations. In particular:

* Moves the HTTP security validations to the HTTP package and not keep them in core. They didn't make sense since in the core package we do not know if we're handling HTTP stuff. This allowed me to remove a lot of `unknown` and weird `lodash.get` calls and use precise types
* Validate Security is now standardised and uses the same `ValidatorFn` shape we're using for `valdidateInput` and `validateOutput`
* The "securityHandler" look up function has been replaced with a static one which is type safer and probably even faster. The runtime option didn't make sense.
* Removed useless parameters that were not used, as well some optional that, well, were not optional
* Narrowed the types; security is only checking security, we do not need to pass the whole HTTP Request
* Tests and some minor bugs have been fixed because of the more precise typing

There are some other internal refactors that need to be done (although not urgent) because the current code is really hard to reason about. I'll tame this beast in the next airplane (tomorrow)


packages/http/src/validator/validators/security/handlers/bearer.ts
```diff
@@ -0,0 +1,21 @@
+import { fromNullable, getOrElse, map } from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { get, partial } from 'lodash';
+import { when } from './utils';
```
StefanDywersant: Do we still practice separating internal and external imports?

packages/http/src/validator/validators/security/handlers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (scheme.type === 'http') {
+    return fromNullable<IPrismDiagnostic>({
+      message: 'We currently do not support this type of security scheme.',
+      severity: DiagnosticSeverity.Warning,
+    })(securitySchemeHandlers[scheme.type][scheme.scheme]);
+  }
+  if (scheme.type === 'apiKey') {
+    return right(securitySchemeHandlers[scheme.type][scheme.in]);
+  }
+  return right(securitySchemeHandlers[scheme.type]);
```
StefanDywersant:
```suggestion
  return fromNullable<IPrismDiagnostic>({
      message: `We currently do not support this type of security scheme: ${scheme.type}`,
      severity: DiagnosticSeverity.Warning,
    })(securitySchemeHandlers[scheme.type]);
```

packages/http/src/validator/validators/security/handlers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    digest: httpDigest,
+    basic: httpBasic,
+    bearer,
+  },
+};
+
+export function findSecurityHandler(scheme: HttpSecurityScheme): Either<IPrismDiagnostic, ValidateSecurityFn> {
+  if (scheme.type === 'http') {
+    return fromNullable<IPrismDiagnostic>({
+      message: 'We currently do not support this type of security scheme.',
```
StefanDywersant:
```suggestion
      message: `We currently do not support this type of security scheme: http/${scheme.scheme}`,
```

packages/http/src/validator/validators/security/handlers/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      severity: DiagnosticSeverity.Warning,
+    })(securitySchemeHandlers[scheme.type][scheme.scheme]);
+  }
+  if (scheme.type === 'apiKey') {
+    return right(securitySchemeHandlers[scheme.type][scheme.in]);
```
StefanDywersant: ditto

packages/http/src/validator/validators/security/handlers/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  message: 'Invalid security scheme used',
+  code: 401,
+  tags: msg ? [msg] : [],
+});
+
+export function isScheme(shouldBeScheme: string, authScheme: string) {
+  return authScheme.toLowerCase() === shouldBeScheme;
+}
+
+export function when(condition: boolean, errorMessage: string | undefined): Either<IPrismDiagnostic, boolean> {
```
StefanDywersant: `boolean` seems to not be used anywhere, maybe `unknown` would be better?

packages/http/src/validator/validators/security/handlers/bearer.ts
XVincentX: yes, let me fix this.

packages/http/src/validator/validators/security/handlers/utils.ts
XVincentX: I'm not too sure I understand. The condition is evaluated in the following line

packages/http/src/validator/validators/security/handlers/index.ts
XVincentX: This cannot be null/undefined by definition — it's a completely static flow.

packages/http/src/validator/validators/security/handlers/index.ts
XVincentX: This cannot be null/undefined by definition — it's a completely static flow.

StefanDywersant: What if `scheme.in` is set to unsupported value? And the parser from `http-spec` won't catch it?

packages/http/src/validator/validators/security/handlers/utils.ts
StefanDywersant: Shit, wrong line.

StefanDywersant: I was referring to `boolean` in `Either<IPrismDiagnostic, boolean>`. I missed that `boolean` is also type of condition arg.

packages/http/src/validator/validators/security/handlers/index.ts
XVincentX: Good question — this cannot happen neither. Since the type is really strict and more importantly shared between Prism and http-spec, there is no way http-spec package can construct an object that's invalid.

There's a way, in case you'd do `const obj: any = {}` — but then I would say that we have a different problem.

I'll change it though, it's way better to have an error message rather than a crash :)

Issue #805, opened 2019-11-14
Proxy/Violations Feedback


philsturgeon:
1. Getting `{"location":["response"],"severity":"Error","message":"The received media type does not match the one specified in the document"}` is not very descriptive. I can see which media type is in the response, sure, but in logs it is even less descriptive. 

<img width="715" alt="image" src="https://user-images.githubusercontent.com/67381/68857962-ed79ac00-06db-11ea-8d09-57b705a7da3b.png">

Telling the user which media type was the problem would help for logs and for scraping sl-violations. 

2. I would not expect to see "The received media type does not match **the one** specified in the document" in a document where I have multiple media type objects defined. I had application/json and application/xml and it still said "the one". It could say "The received media type has not been defined"

3. Requesting invalid route returns error by default

```
$ http GET http://127.0.0.1:4010/nonsense
HTTP/1.1 404 Not Found
Connection: keep-alive
Date: Thu, 14 Nov 2019 12:45:45 GMT
access-control-allow-origin: *
content-length: 200
content-type: application/problem+json

{
    "detail": "The route /nonsense hasn't been found in the specification file",
    "status": 404,
    "title": "Route not resolved, no path matched",
    "type": "https://stoplight.io/prism/errors#NO_PATH_MATCHED_ERROR"
}
```

When running in log mode, prism proxy should be trucking along and making requests to the upstream server regardless of what happens, and reporting issues to logs and sl-violations. Requesting a totally invalid URL blows up with a problem+json. 

4. Proxy logs could tell me what upstream request is being made

I see this:

```
[HTTP SERVER] get /pet/123 ℹ  info      Request received
    [VALIDATOR] ✖  error     Violation: response The received media type does not match the one specified in the document
    [VALIDATOR] ✖  error     Violation: request Invalid security scheme used
```

It's rather tricky for me to now debug this, as I have to figure out the actual URL being made from copying and pasting various strings: `https://petstore.swagger.io/` from the command I wrote and `/pet/123` from the HTTP server. logging the full URL would make debugging easier.

```
[HTTP SERVER] get /pet/123 ℹ  info      Request received
    [PROXY] Making request to upstream server: https://petstore.swagger.io/pet/123
    [VALIDATOR] ✖  error     Violation: response The received media type does not match the one specified in the document
    [VALIDATOR] ✖  error     Violation: request Invalid security scheme used
```

[...]

PR #808, opened 2019-11-18
Keep proxying if resource is not found

XVincentX:
#805 point 3

[Review with no whitespaces](https://github.com/stoplightio/prism/pull/808/files?utf8=✓&diff=unified&w=1)

This PR will short circuit the error handler in case the `errors` flag is set to false and the proxy is on — so we can still proxy the request although all the validation checks will not be performed since we have no idea what resource the user is trying to hit.

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -84,7 +84,7 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
                     input: [
                       {
                         message:
-                          "The selected route hasn't been found and the errors is set false. Prims will proxy the request to the upstream server but no validation will happen",
+                          "The selected route hasn't been found and the errors is set false. Prism will proxy the request to the upstream server but no validation will happen",
                         severity: DiagnosticSeverity.Warning,
                       },
                     ],
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -76,7 +76,7 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
           error => {
             if (!config.errors && isProxyConfig(config)) {
               return pipe(
-                components.forward(input, config.upstream.href),
+                components.forward(input, config.upstream.href)(components.logger.child({ name: 'PROXY' })),
                 TaskEither.map<Output, IPrismOutput<Output>>(output => ({
                   input,
                   output,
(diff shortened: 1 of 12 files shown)


packages/core/src/factory.ts
```diff
@@ -71,25 +72,51 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
-        TaskEither.map(({ output, resource, inputValidations }) => {
-          const outputValidations = config.validateResponse
-            ? pipe(
-                Option.fromEither(Either.swap(components.validateOutput({ resource, element: output }))),
-                Option.getOrElse<IPrismDiagnostic[]>(() => [])
-              )
-            : [];
+        TaskEither.fold(
+          error => {
+            if (!config.errors && isProxyConfig(config)) {
```
StefanDywersant: I have a feeling that the code got complicated. I believe adding more separation between proxying and mocking would improve a lot here.

Below is kind of "work in progress" however, it shows clearly the idea of how to simplify the flow:
```
       return pipe(
        isProxyConfig(config)
          ? pipe(
            TaskEither.fromEither(components.route({ resources, input })),
            TaskEither.chain(resource => inputValidation(resource, input, config)),
            TaskEither.alt(error => config.errors ? TaskEither.left(error) : TaskEither.right({ inputValidations: [
                {
                  message:
                    "The selected route hasn't been found and the errors is set false. Prims will proxy the request to the upstream server but no validation will happen",
                  severity: DiagnosticSeverity.Warning,
                },
              ] })),
            TaskEither.chain(({ inputValidations }) => forward(input, config, inputValidations)),
          )
        : pipe(
            TaskEither.fromEither(components.route({ resources, input })),
            TaskEither.chain(resource => inputValidation(resource, input, config)),
            TaskEither.chain(({ resource, inputValidations }) => mock(resource, input, config, inputValidations)),
          ),
        TaskEither.map(outputValidation)
      );
```

This requires separating `forward` and `mock` and ofc some polishing around typing (I'll be happy to do that if you want).

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -15,6 +15,7 @@ function instantiatePrism2(operations: IHttpOperation[]) {
       validateRequest: true,
       validateResponse: true,
       mock: { dynamic: false },
+      errors: false,
```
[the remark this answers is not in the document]
XVincentX: > I'd lean towards the state when our 4 main components require minimal configuration on their external interfaces.

This is already happening. This is an internal object where all the properties are required; in the external client they're almost all optional and we provide sensible defaults.

packages/core/src/factory.ts
XVincentX: Yeah I agree the flow is getting complicated — I'm willing to try your solution. I'll get back on this and see if that will bring some improvements.

XVincentX: Ok I have tried and the resulting flow isn't really that simplified. I would temporarily keep it in this way but if you're willing to give it a spin — be my guest! Just open a PR starting from this brach and we'll go through together!

PR #825, opened 2019-11-18
Improved bootstrap of Prism instance in test harness

StefanDywersant:
## Checklist

_Refactor_

Test harness will fail after 3sec if prism instance has not been started correctly.

No

### DIFF
--- a/test-harness/index.ts
+++ b/test-harness/index.ts
@@ -13,7 +13,7 @@ import { parseSpecFile, xmlValidator } from './helpers';
 const glob = globFs({ gitignore: true });
 jest.setTimeout(5000);
 
-const WAIT_FOR_LINE = 'Server listening at';
+const WAIT_FOR_LINE = 'Prism is listening';
 const WAIT_FOR_LINE_TIMEOUT = 3000;
 
 describe('harness', () => {
--- a/test-harness/index.ts
+++ b/test-harness/index.ts
@@ -45,75 +45,77 @@ describe('harness', () => {
 
     const testText = `${file}${os.EOL}${parsed.test}`;
 
-    test(testText, async done => {
+    test(testText, done => {
       const [command, ...args] = parsed.command.split(/ +/).map(t => t.trim());
 
-      const prismMockProcessHandle = await startPrism(parsed.server, tmpFileHandle.name);
+      startPrism(parsed.server, tmpFileHandle.name, (error, prismMockProcessHandle) => {
+        if (error) {
+          throw error;
+        }
 
-      const clientCommandHandle = spawnSync(command, args, {
-        shell: true,
-        encoding: 'utf8',
-        windowsVerbatimArguments: false,
-      });
-      const output: any = parseResponse(clientCommandHandle.stdout.trim());
-      const expected: any = parseResponse((parsed.expect || parsed.expectLoose).trim());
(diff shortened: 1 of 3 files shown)


test-harness/index.ts
```diff
@@ -43,60 +45,79 @@ describe('harness', () => {
...
+    prismMockProcessHandle.stdout.pipe(split2()).on('data', (line: string) => {
+      if (line.includes(WAIT_FOR_LINE)) {
+        clearTimeout(timeout);
+        resolve(prismMockProcessHandle);
+      }
+    });
+  });
+}
+
+function shutdownPrism(processHandle: ChildProcess, done: (...args: any[]) => any) {
```
XVincentX:
```suggestion
function shutdownPrism(processHandle: ChildProcess, done: jest.DoneCallback) {
```

test-harness/index.ts
```diff
@@ -43,60 +45,79 @@ describe('harness', () => {
 
     const testText = `${file}${os.EOL}${parsed.test}`;
 
-    test(testText, done => {
+    test(testText, async done => {
```
XVincentX: In general mixing `async` and `done` callbacks together is not a good idea. We should either use one or the other system. In this case, let's stick with the `done` callback.

test-harness/index.ts
```diff
@@ -43,60 +45,79 @@ describe('harness', () => {
...
-            windowsVerbatimArguments: false,
-          });
-          const output: any = parseResponse(clientCommandHandle.stdout.trim());
-          const expected: any = parseResponse((parsed.expect || parsed.expectLoose).trim());
 
-          const isXml = xmlValidator.test(
-            get(output, ['header', 'content-type'], ''),
-            expected.body
-          );
+      const prismMockProcessHandle = await startPrism(parsed.server, tmpFileHandle.name);
```
XVincentX: …per the comment above, we should modify the `startPrism` function to take an additional done callback to continue the flow later.

test-harness/index.ts
StefanDywersant: It has to stay that way to be compatible both with jest call and our call (https://github.com/stoplightio/prism/pull/825/files#diff-d26ad5d0c95e0973686e7fa034e1283bR107).

test-harness/index.ts
StefanDywersant: Done 👍

test-harness/index.ts
StefanDywersant: You right, I updated it to use callback only.

test-harness/index.ts
XVincentX: Ah yeah ok, got it.

StefanDywersant:
> 2. When running `yarn test.harness` I get an empty output instead of the expected error

I checked locally. Works fine if no changes on WAIT_LINE. If you change WAIT_LINE it throws an exception after 3sec. Note that the check checks whether logline _includes_ WAIT_LINE. So if you shorten it then it can still work.

XVincentX:
> > 1. When running `yarn test.harness` I get an empty output instead of the expected error
> 
> I checked locally. Works fine if no changes on WAIT_LINE. If you change WAIT_LINE it throws an exception after 3sec. Note that the check checks whether logline _includes_ WAIT_LINE. So if you shorten it then it can still work.

Ah! let me try again.

PR #827, opened 2019-11-19
fix: improve error messages

XVincentX:
```diff
- Violation: response.query should be equal to one of the allowed values
+ Violation: request.query /status/0 should be equal to one of the allowed values: available, pending, sold
```

### DIFF
--- a/packages/http/src/validator/validators/utils.ts
+++ b/packages/http/src/validator/validators/utils.ts
@@ -4,14 +4,14 @@ import { getSemigroup } from 'fp-ts/lib/NonEmptyArray';
 import { getValidation } from 'fp-ts/lib/Either';
 import { option } from 'fp-ts/lib/Option';
 import { sequenceT } from 'fp-ts/lib/Apply';
-import * as Ajv from 'ajv';
+import { ErrorObject } from 'ajv';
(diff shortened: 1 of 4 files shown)


packages/http/src/validator/validators/utils.ts
```diff
@@ -29,9 +29,10 @@ export const validateAgainstSchema = (value: any, schema: JSONSchema, prefix?: s
     const validate = ajv.compile(schema);
     const valid = validate(value);
     if (!valid) {
-      return convertAjvErrors(validate.errors, DiagnosticSeverity.Error).map(error => {
+      const enhacedMessages = betterAjvErrors(schema, value, validate.errors, { format: 'js' });
+      return convertAjvErrors(validate.errors, DiagnosticSeverity.Error).map((error, index) => {
         const path = prefix ? [prefix, ...error.path] : [...error.path];
-        return Object.assign({}, error, { path });
+        return Object.assign({}, error, { path, message: enhacedMessages[index].error });
```
StefanDywersant:
```suggestion
        return { ...error, path, message: enhancedMessages[index].error };
```
Note `enhaced` -> `enhanced`

XVincentX:
Bad idea, better-ajv-errors does not work in the browser. Will find another way.

PR #828, opened 2019-11-19
feat: handle allowedParams

XVincentX:
PR to handle the `allowedValues` in the `params` object (if present) to print better messages.

### DIFF
--- a/packages/http/src/validator/validators/utils.ts
+++ b/packages/http/src/validator/validators/utils.ts
@@ -21,7 +21,7 @@ export const convertAjvErrors = (errors: Ajv.ErrorObject[] | undefined | null, s
     return {
       path: error.dataPath.split('.').slice(1),
       code: error.keyword || '',
-      message: `${error.message}${allowedParameters}` || '',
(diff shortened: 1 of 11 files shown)


packages/http/src/validator/validators/utils.ts
```diff
@@ -6,22 +6,25 @@ import { option } from 'fp-ts/lib/Option';
...
   }
 
-  return errors.map<IPrismDiagnostic & { path: Segment[] }>(error => ({
-    path: error.dataPath.split('.').slice(1),
-    code: error.keyword || '',
-    message: error.message || '',
-    severity,
-  }));
+  return errors.map<IPrismDiagnostic & { path: Segment[] }>(error => {
+    const allowedParameters = 'allowedValues' in error.params ? `: ${error.params.allowedValues.join(', ')}` : '';
```
StefanDywersant: Seems that `allowedValues` is set only when dealing with `enum` 👍 https://github.com/epoberezkin/ajv/search?q=allowedValues&unscoped_q=allowedValues

PR #830, opened 2019-11-19
refactor: make security validations easier to reason about

XVincentX:
The following PR brings some significant simplification to the security checks that were made in Prism.

It literally took me ages to understand what was going on and why some choices were made so far — I guess this part definitely went out of control.

Hopefully this PR, although not 100% complete (but honestly I'm so tired of working on this section) will bring some clarity and order so that in the next flight refactor it one more time will be easier.

### DIFF
--- a/packages/http/src/validator/validators/security/index.ts
+++ b/packages/http/src/validator/validators/security/index.ts
@@ -1,5 +1,6 @@
 import { IHttpOperation, HttpSecurityScheme } from '@stoplight/types';
 import * as Either from 'fp-ts/lib/Either';
+import * as Option from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/lib/pipeable';
 import { flatten } from 'lodash';
 import { set } from 'lodash/fp';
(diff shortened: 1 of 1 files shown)


packages/http/src/validator/validators/security/index.ts
```diff
@@ -2,109 +2,54 @@ import { IHttpOperation, HttpSecurityScheme } from '@stoplight/types';
...
-) {
-  const firstAuthErr = pipe(
-    firstAuthErrAsLeft,
-    Either.fold<IPrismDiagnostic, IPrismDiagnostic, IPrismDiagnostic>(identity, identity)
-  );
-
-  const invalidResultWithAuthHeader = gatherWWWAuthHeader([authResult], ['tags'], firstAuthErr);
-
-  return [Either.left(invalidResultWithAuthHeader)];
+function getWWWAuthHeader(authResults: NonEmptyArray<IPrismDiagnostic>) {
```
StefanDywersant: I have trouble understanding correlation between the name of the function and what it actually does.

It takes a list of diagnostics, merges the tags lists from them and writes that to the first diagnostic's tags field.

How is the above connected with `WWWAuthHeader`?

packages/http/src/validator/validators/security/index.ts
```diff
@@ -2,109 +2,54 @@ import { IHttpOperation, HttpSecurityScheme } from '@stoplight/types';
...
 
-function getAuthResults(securitySchemes: HttpSecurityScheme[][], input: Pick<IHttpRequest, 'headers' | 'url'>) {
+function getAuthResultsAND(securitySchemes: HttpSecurityScheme[][], input: HeadersAndUrl) {
   return securitySchemes.map(securitySchemePairs => {
-    const authResult = securitySchemePairs.map(securityScheme =>
+    const authResults = securitySchemePairs.map(securityScheme =>
       pipe(
         findSecurityHandler(securityScheme),
-        Either.chain(f => f(input, 'name' in securityScheme ? securityScheme.name : ''))
+        Either.chain(f => f(input, 'name' in securityScheme ? securityScheme.name : '')),
```
StefanDywersant:
```suggestion
        Either.chain(securityHandler => securityHandler(input, 'name' in securityScheme ? securityScheme.name : '')),
```

packages/http/src/validator/validators/security/index.ts
XVincentX: That's a very good point — this indeed produces *what* gets put in the WWWAuthHeader returned to the client.

PR #854, opened 2019-11-25
fix: put chalk as a dependency

### LINKED ISSUE #853: Add Chalk to dependencies list
Hello,
I have installed the version **3.2.0** of **@stoplight/prism-cli**

Executing the command `prism mock ./swagger.yml -p 8086` i receive the following error:
```
/Users/me/Documents/app/node_modules/@stoplight/prism-cli/dist/const/options.js:5
    CLI: { index: 0, color: chalk_1.default.bgWhiteBright },
                                            ^

TypeError: Cannot read property 'bgWhiteBright' of undefined
    at Object.<anonymous> (/Users/me/Documents/app/node_modules/@stoplight/prism-cli/dist/const/options.js:5:45)
```

My app uses **chalk** version **3.0.0** and i noticed that **prism-cli** have no **chalk** dependency defined in the `package.json` even if it uses it, this way it resolves using the globally installed version 3 which is not the version it needs.

I think adding the line `"chalk": "2.4.2"` to prism-cli package.json would resolve this little dependency problem

XVincentX:
Closes #853

### DIFF
--- a/packages/cli/src/const/options.ts
+++ b/packages/cli/src/const/options.ts
@@ -1,4 +1,4 @@
-import chalk from 'chalk';
+import * as chalk from 'chalk';
 
 export const LOG_COLOR_MAP = {
   CLI: { index: 0, color: chalk.bgWhiteBright },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 # Unreleased
 
+## Fixed
+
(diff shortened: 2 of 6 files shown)


packages/cli/src/const/options.ts
```diff
@@ -1,4 +1,4 @@
-import chalk from 'chalk';
+import * as chalk from 'chalk';
```
StefanDywersant: Is `* as ` needed for anything?

XVincentX: Yes, starting on 3.0 they have started to use default exports

PR #860, opened 2019-11-27
refactor: handle validation exception correctly

XVincentX:
Get rid of the throw statement and handle it correctly with the helper provided by the Option monad.

### DIFF
--- a/packages/http/src/validator/validators/__tests__/headers.spec.ts
+++ b/packages/http/src/validator/validators/__tests__/headers.spec.ts
@@ -3,6 +3,7 @@ import { query as registry } from '../../deserializers';
 import { HttpHeadersValidator } from '../headers';
 import * as validateAgainstSchemaModule from '../utils';
 import { assertRight, assertLeft } from '@stoplight/prism-core/src/__tests__/utils';
+import { some } from 'fp-ts/lib/Option';
 
 describe('HttpHeadersValidator', () => {
   const httpHeadersValidator = new HttpHeadersValidator(registry, 'header');
@@ -48,7 +49,7 @@ describe('HttpHeadersValidator', () => {
                 ])
               );
 
-              expect(validateAgainstSchemaModule.validateAgainstSchema).toReturnWith([]);
+              expect(validateAgainstSchemaModule.validateAgainstSchema).toReturnWith(some([]));
             });
           });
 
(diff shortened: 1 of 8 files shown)


packages/http/src/validator/validators/utils.ts
```diff
@@ -2,7 +2,7 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { DiagnosticSeverity, Segment } from '@stoplight/types';
 import { getSemigroup } from 'fp-ts/lib/NonEmptyArray';
 import { getValidation } from 'fp-ts/lib/Either';
-import { option } from 'fp-ts/lib/Option';
+import { option, tryCatch, Option } from 'fp-ts/lib/Option';
```
StefanDywersant: I would try to follow our code-style here and
```suggestion
import * as Option from 'fp-ts/lib/Option';
```

Especially when there are functions from different buckets.

packages/http/src/validator/validators/__tests__/headers.spec.ts
```diff
@@ -3,6 +3,7 @@ import { query as registry } from '../../deserializers';
 import { HttpHeadersValidator } from '../headers';
 import * as validateAgainstSchemaModule from '../utils';
 import { assertRight, assertLeft } from '@stoplight/prism-core/src/__tests__/utils';
+import { some } from 'fp-ts/lib/Option';
```
StefanDywersant:
```suggestion
import * as Option from 'fp-ts/lib/Option';
```

Although `some` is pretty typical for FP I would stick to our convention.

packages/http/src/validator/validators/__tests__/path.spec.ts
```diff
@@ -3,6 +3,7 @@ import { path as registry } from '../../deserializers';
 import { HttpPathValidator } from '../path';
 import * as validateAgainstSchemaModule from '../utils';
 import { assertLeft, assertRight } from '@stoplight/prism-core/src/__tests__/utils';
+import { some } from 'fp-ts/lib/Option';
```
StefanDywersant: Ditto

philsturgeon:
Not sure I have enough context to comment, as I float further away from code and more into product land I need screenshots or error messages that were fixed or something less fp-ish to comment on. :)

PR #862, opened 2019-11-29
Validation refactor pass 2

XVincentX:
This PR comes out because of an attempt of using `These` in Prism, resulting in some inconsistencies that were not making the thing possible:

* `validateBody` is not doing the `fromEither and fromOption` conversion anymore
* Correctly use `NonEmptyArray` where appropriate
* `validateAgainstSchema` function now returns `None` when no validation errors instead of an empty array — making the things a little bit weird.


packages/http/src/validator/validators/utils.ts
```diff
@@ -1,49 +1,47 @@
...
-        return Object.assign({}, error, { path });
-      });
-    }
-    return [];
-  });
-};
+export const validateAgainstSchema = (value: unknown, schema: JSONSchema, prefix?: string) =>
+  pipe(
+    Option.tryCatch(() => ajv.compile(schema)),
+    Option.chain(validate => {
```
StefanDywersant:
```suggestion
    Option.mapNullable(validate => {
    ...
    return validate.errors;
```

packages/http/src/validator/validators/body.ts
```diff
@@ -5,17 +5,16 @@ import * as Either from 'fp-ts/lib/Either';
 import * as Option from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/lib/pipeable';
 import { get } from 'lodash';
-import * as typeIs from 'type-is';
 import { JSONSchema } from '../../types';
 import { body } from '../deserializers';
 import { IHttpValidator } from './types';
 import { validateAgainstSchema } from './utils';
-import { fromArray } from 'fp-ts/lib/NonEmptyArray';
+import { NonEmptyArray, map } from 'fp-ts/lib/NonEmptyArray';
```
StefanDywersant:
```suggestion
import * as NonEmptyArray from 'fp-ts/lib/NonEmptyArray';
```

`map` is too common

packages/http/src/validator/validators/body.ts
```diff
@@ -91,42 +85,53 @@ export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent>
...
+            Either.mapLeft(diagnostics => applyPrefix(this.prefix, diagnostics))
+          )
+      )
     );
   }
 }
 
-function applyPrefix(prefix: string, diagnostics: IPrismDiagnostic[]): IPrismDiagnostic[] {
-  return diagnostics.map(d => ({ ...d, path: [prefix, ...(d.path || [])] }));
+function applyPrefix(prefix: string, diagnostics: NonEmptyArray<IPrismDiagnostic>): NonEmptyArray<IPrismDiagnostic> {
```
StefanDywersant:
```suggestion
function applyPrefix(prefix: string, diagnostics: NonEmptyArray.NonEmptyArray<IPrismDiagnostic>): NonEmptyArray.NonEmptyArray<IPrismDiagnostic> {
```

packages/http/src/validator/validators/body.ts
```diff
@@ -91,42 +85,53 @@ export class HttpBodyValidator implements IHttpValidator<any, IMediaTypeContent>
...
+  return pipe(
+    diagnostics,
+    map(d => ({ ...d, path: [prefix, ...(d.path || [])] }))
```
StefanDywersant:
```suggestion
    NonEmptyArray.map(d => ({ ...d, path: [prefix, ...(d.path || [])] }))
```

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -12,48 +12,42 @@ import { pipe } from 'fp-ts/lib/pipeable';
 import { generate as generateHttpParam } from '../generator/HttpParamGenerator';
 import { validateOutput } from '../../validator';
 import { parseResponse } from '../../utils/parseResponse';
-import withLogger from '../../withLogger';
 import { violationLogger } from '../../utils/logger';
 import { Logger } from 'pino';
 
-export function runCallback({
+export const runCallback = ({
```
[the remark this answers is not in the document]
XVincentX: Effectively nothing, I'll revert it back

PR #863, opened 2019-11-30
test: remove deprecated tests

XVincentX:
This PR removes some test that are not valid anymore (basically testing the same thing, since the default config object and the default config functions aren't concepts in Prism anymore) and removes the snapshot in favour or regular assertions.


packages/http-server/src/__tests__/getHttpConfigFromRequest.spec.ts
```diff
@@ -8,79 +8,43 @@ describe('getHttpConfigFromRequest()', () => {
...
       return expect(
         getHttpConfigFromRequest({
           method: 'get',
           url: { path: '/', query: { __dynamic: 'true' } },
         })
-      ).toMatchSnapshot();
-    });
-  });
-
-  describe('given default config function', () => {
```
XVincentX: We do not have a default config function anymore.

packages/http-server/src/__tests__/getHttpConfigFromRequest.spec.ts
```diff
@@ -8,79 +8,43 @@ describe('getHttpConfigFromRequest()', () => {
...
-      return expect(
-        getHttpConfigFromRequest({
-          method: 'get',
-          url: { path: '/', query: { __code: '400' } },
-        })
-      ).toMatchSnapshot();
-    });
-  });
-
-  describe('given default config object', () => {
```
XVincentX: We do not have a default config object anymore.

PR #875, opened 2019-12-02
Fix: schema generation can fail

XVincentX:
While debugging some JSON Ref issues I noticed that Schema Generation can effectively throw exceptions and would be treated that well.

This PR correctly wraps such situations in `tryCatch` statements.


packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -80,12 +83,13 @@ describe('JSONSchema generator', () => {
       };
 
       it('will have a string property matching the ip regex', () => {
-        const instance = generate(schema);
-        expect(instance).toHaveProperty('ip');
-        const ip = get(instance, 'ip');
+        assertRight(generate(schema), instance => {
```
StefanDywersant: Since `generate` return Either we should have a test of failure. Maybe a circular object will do the job?

StefanDywersant: ekhm :P

XVincentX: Ahi, I didn't see this. Will add.

PR #876, opened 2019-12-03
Add failing schema test

XVincentX:


packages/http/src/mocker/generator/__tests__/JSONSchema.spec.ts
```diff
@@ -93,6 +93,31 @@ describe('JSONSchema generator', () => {
       });
     });
 
+    describe('when used with a schema that is not valid', () => {
+      const schema: JSONSchema = {
```
StefanDywersant:
```suggestion
      const schema: JSONSchema = {
        type: 'object',
        properties: {
          _embedded: {
            $ref: '#/definitions/supermodelIoAdidasApiHAL',
          },
        },
      };
      it('will return a left', () => assertLeft(generate(schema)));
```

It gives exactly the same error and it is way shorter.

XVincentX: Fair! Fixed!

PR #892, opened 2019-12-13
fix: 500 only for output validation errors

XVincentX:
It turns out that we were returning a 500 even if there were just Input violations in the current request — but in that case the negotiation process is already giving us the response we want to return; the 500 transformation is only required for *output* violations.

Please approve but do not merge, so I can bundle this in a patch release.

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -72,16 +72,17 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
       TaskEither.chain(response => {
         const { output } = response;
 
-        const inputOutputValidationErrors = response.validations.output
-          .map(createErrorObjectWithPrefix('response'))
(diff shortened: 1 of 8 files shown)


packages/http-server/src/server.ts
```diff
@@ -72,16 +72,17 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
       TaskEither.chain(response => {
         const { output } = response;
 
-        const inputOutputValidationErrors = response.validations.output
-          .map(createErrorObjectWithPrefix('response'))
-          .concat(response.validations.input.map(createErrorObjectWithPrefix('request')));
+        const inputValidationErrors = response.validations.input.map(createErrorObjectWithPrefix('request'));
```
karol-maciaszek: I appreciate that additional line of code :)

PR #912, opened 2019-12-24
refactor: http match returns either

XVincentX:


XVincentX:
You are probably right on both points — but I do not want to do that right now here since I have no real bandwidth to test this. I just wanted, at this point, to avoid the function to throw exceptions.

I would merge this and address your points in a separate PR.

XVincentX:
I'll take care of them

PR #921, opened 2020-01-06
Simple change to remove hop-by-hop headers when relaying respon…

rogersolsvik:
…se from upstream

## Checklist

_Bug fix_

When running `prism proxy` it relays all headers from upstream source, even after adding new ones that might be incompatible with the original response. 

More specifically in my case, my application is sending the `Transfer-Encoding: chunked` header, but since Prism is buffering the whole response to do response validation, is also adds the `Content-Length` header. These headers are incompatible and causes certain clients to fail. For example, Postman fails with a «Parse error» without a decent cause description.

Read more about the `Transfer-Encoding` header [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding)

Should be no breaking changes with this PR as far as I can see


XVincentX:
Oh ok — interesting issue. The fix seems legit, however I would like to check it out deeply before I can merge and release this.

Thanks for the good investigation!

rogersolsvik:
No problem @XVincentX. It might be appropriate make this fix for other "hop-by-hop" headers as well, but did not investigate which other might cause an issue too.
Other "hop-by-hop" headers are:
```
Connection
Keep-Alive
Public
Proxy-Authenticate
Transfer-Encoding
Upgrade
```

XVincentX:
Yeah that is exactly my point — more investigation might be required to make this work properly in all the circumstances.

rogersolsvik:
Good. Let me know if I can help. I you want to remove all hop-by-hop headers as a part of this pull request, I can make the appropriate updates.

packages/http-server/src/server.ts
```diff
@@ -109,6 +109,10 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
           E.tryCatch(() => {
             if (output.headers) reply.headers(output.headers);
 
+            // Since we are buffering the whole response body, do not relay the "transfer-encoding" header if it exist.
+            // It will be combined with the "Content-Length" header which is not valid and will cause certain clients to fail(etc. Postman)
+            reply.removeHeader('transfer-encoding');
```
XVincentX: This should only be happening if the `proxy` option is set

packages/http-server/src/__tests__/header-params-validity.spec.ts
```diff
@@ -0,0 +1,116 @@
+import { createLogger } from '@stoplight/prism-core';
+import { IHttpOperation, HttpParamStyles } from '@stoplight/types';
+import * as fastify from 'fastify';
+import { createServer } from '..';
+import { IPrismHttpServer } from '../types';
+
+const logger = createLogger('TEST', { enabled: false });
+
+function instantiatePrism2(operations: IHttpOperation[]) {
```
XVincentX: Why the number `2`?

packages/http-server/src/server.ts
rogersolsvik: Did this by stripping the headers in the forwarder instead

packages/http-server/src/__tests__/header-params-validity.spec.ts
rogersolsvik: Some copy&paste "magic" in my editor :) . Have removed this test now.

rogersolsvik:
A little premature push. Will be fixing test soon

rogersolsvik:
I have not setup the harness tests locally. Any input on what's going wrong @XVincentX ?

rogersolsvik:
Did an update from master now which triggered a rebuild. The harness tests seem to work now.

XVincentX:
Thanks for checking this out — Ill review.

packages/http/src/forwarder/index.ts
```diff
@@ -55,3 +57,10 @@ function serializeBody(body: unknown) {
 
   return E.right(undefined);
 }
+
+const stripHopByHopHeaders = (response: IHttpResponse): TE.TaskEither<Error, IHttpResponse> => {
+  hopByHopHeaders.forEach(header => {
+    delete (response.headers as Dictionary<string, string>)[header];
```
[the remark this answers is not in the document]
rogersolsvik: Actually I was not able to make this work without casting it due to it being nullable. Blame it on my inexperience with TypeScript and functional programming 😃 
Excellent idea using `omit` instead 👍

packages/http/src/forwarder/index.ts
```diff
@@ -41,7 +42,8 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
         });
       }, E.toError)
     ),
-    TE.chain(parseResponse)
+    TE.chain(parseResponse),
+    TE.chain(stripHopByHopHeaders)
```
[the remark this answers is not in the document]
rogersolsvik: Fixed 👍

packages/http/src/forwarder/__tests__/index.spec.ts
```diff
@@ -2,6 +2,8 @@ import fetch from 'node-fetch';
 import forward from '../index';
 import { assertResolvesRight, assertResolvesLeft } from '@stoplight/prism-core/src/__tests__/utils';
 import { mapValues } from 'lodash';
+import { hopByHopHeaders } from '../resources';
+import _ = require('lodash');
```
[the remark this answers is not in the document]
rogersolsvik: Fixed 👍

XVincentX:
@rogersolsvik Thanks for the follow up. I'll check it out again but I think this might be good to merge now. Thanks!

XVincentX:
Thanks, @rogersolsvik

PR #927, opened 2020-01-08
Using Micri HTTP server

### LINKED ISSUE #790: Consider ditching Fastify for Micri
Fastify has been serving us fine so far with some high and lows, but I think it is time to remove it from the fastify http server and consider some alternatives.

## Motivations

* Because of the unique requirements of Prism (well to be frank it's more _because of the OpenAPI quirks_) we're using 1% of its features.
  * Fastify offers Error serialisation — oh ok I can’t use it
  * Fastify offers Logging out of the box — oh I can’t really use it
  * Fastify offers RadixTree route matching — oh I can’t use it
  * Fastify offers JSON Schema serialisation for payloads — oh I can’t use it
* Fastify is opinionated in the way it serialises the error — and so if you throw an Error with additional fields (required by the application/vnd.problem+json) fastify  will happily ignore all of them and pick only the fields it wants. This requires you to manually override the Error serialiser, cutting out a good chunk of Fastify itself.
* Fastify is too opinionated on the logging; there’s a log instance on the request object that effectively injects request details in the log messages which in theory is great; the problem is they do not get propagated correctly on the instance; you can’t retrieve them later in the inner parts of Prism. I’ve kind of “workarounded” the problem by looking at their source code, but I spent a significant chunk of time on that
* Some of the TypeScript definitions are wrong (it's not a big deal, but it is annoying sometimes)
* It is definitely not functional friendly, and that forces us to have a very bad looking request handler (see #782 for more details)

---

That said, I got a good experience with [micri](https://github.com/olliv/micri) — which is currently powering the hosted Prism version.

* Minimal utility on top on the raw http server nodejs offers
  * Generic error handler in case anything goes wrong (can't happen in Prism, but still…)
  * JSON and text parsing (which is effectively *ALL* we need from the parsing side)
  * Not opinionated, so we can just drop Prism's monads there and be happy
  * 300 lines of codes with 0 dependencies

---

This is not a priority, but still, I'd like to get this move forward when we have time.

karol-maciaszek:
Closes #790 

## Checklist

Refactor

This PR replaces fastify HTTP server with a smaller equivalent: [micri](https://github.com/OlliV/micri).
The motivation is described in the issue.

I hope it does not


packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { pipe } from 'fp-ts/lib/pipeable';
 import * as TE from 'fp-ts/lib/TaskEither';
 import * as E from 'fp-ts/lib/Either';
+import { MicriHandler } from 'micri/dist';
```
XVincentX:
```suggestion
import { MicriHandler } from 'micri';
```

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
-
-  server.addContentTypeParser('*', { parseAs: 'string' }, (req, body, done) => {
-    if (typeIs(req, ['application/*+json'])) {
-      try {
-        return done(null, JSON.parse(body));
-      } catch (e) {
-        return done(e);
-      }
-    }
+  const micriHandler: MicriHandler = async (request: IncomingMessage, reply: ServerResponse) => {
```
XVincentX: If we put the function type, we can avoid the types of the arguments

```suggestion
  const micriHandler: MicriHandler = async (request, reply) => {
```

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
 import { pipe } from 'fp-ts/lib/pipeable';
 import * as TE from 'fp-ts/lib/TaskEither';
 import * as E from 'fp-ts/lib/Either';
+import { MicriHandler } from 'micri/dist';
+
+function searchParamsToNameValues(searchParams: URLSearchParams): IHttpNameValues {
+  const params = {};
+  for (const key of searchParams.keys()) {
+    const values = searchParams.getAll(key);
+    params[key] = values.length === 1 ? values[0] : values;
```
XVincentX: Is this mandatory? Can't we use always an array?

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+  }
+  return params;
+}
+
+function addressInfoToString(address: AddressInfo | string | null) {
```
XVincentX: Given the successive casting, I guess we can just modify the signature to not accept the `string` as a type

```suggestion
function addressInfoToString(address?: AddressInfo) {
```

Even better, I would move the null check outside of this function

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+}
+
+function parseRequestBody(request: IncomingMessage) {
+  // if no body provided then return null instead of empty string
+  if (
+    request.headers['content-type'] === undefined
+    && request.headers['transfer-encoding'] === undefined
+    && (request.headers['content-length'] === '0' || request.headers['content-length'] === undefined)
+  ) {
+    return null;
```
XVincentX: `json` and `text` function are async, so I guess we should be returning 
```suggestion
    return Promise.resolve(null);
```

or mark the function as async

packages/http-server/src/server.ts
```diff
@@ -97,67 +110,80 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+        },
+      ),
+      Router.otherwise(micriHandler)
+    )
+  );
+
+  const prism = createInstance(config, components);
 
-  const prismServer: IPrismHttpServer = {
+  return {
```
XVincentX: I'm ok with this, however I recall I did this for a reason…but since I can't remember it, let's remove it.

packages/http-server/src/server.ts
```diff
@@ -97,67 +110,80 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+      Router.on.options(
+        () => opts.cors,
+        (req: IncomingMessage, res: ServerResponse) => {
+          res.setHeader('Access-Control-Allow-Origin', req.headers['origin'] || '*' as string);
+          res.setHeader('Access-Control-Allow-Credentials', 'true');
+          res.setHeader('Access-Control-Allow-Methods', 'GET,DELETE,HEAD,PATCH,POST,PUT');
+          send(res, 204);
+        },
+      ),
+      Router.otherwise(micriHandler)
```
XVincentX: Let's be agnostic as much has possible and call it `handler`?

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+  if (!address) return '';
+  const a = address as AddressInfo;
+  return `http://${a.address}:${a.port}`;
+}
+
+function parseRequestBody(request: IncomingMessage) {
+  // if no body provided then return null instead of empty string
+  if (
+    request.headers['content-type'] === undefined
+    && request.headers['transfer-encoding'] === undefined
```
XVincentX: Hmm, why do we need to check this header as well? Did you copy this from the fastify's source code?

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+    && (request.headers['content-length'] === '0' || request.headers['content-length'] === undefined)
```
XVincentX: Hmm this is weird — the `undefined` check has been done above — so I guess this can just become a single OR?

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
       url: {
-        path: (url || '/').split('?')[0],
-        query,
-        baseUrl: query.__server,
+        path: pathname,
+        baseUrl: searchParams.get('__server') || undefined,
+        query: searchParamsToNameValues(searchParams),
       },
-      headers,
+      headers: headers as IHttpNameValue,
```
XVincentX: I am afraid this cast is not legit, since the `IHttpNameValue` is just a string, while the IncomingHeaders *could* be an array, and we'll need to fix these.

packages/http-server/src/server.ts
```diff
@@ -97,67 +110,80 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
         inputOutputValidationErrors.forEach(validation => {
           const message = `Violation: ${validation.location.join('.') || ''} ${validation.message}`;
           if (validation.severity === DiagnosticSeverity[DiagnosticSeverity.Error]) {
-            request.log.error({ name: 'VALIDATOR' }, message);
+            components.logger.error({ name: 'VALIDATOR' }, message);
```
XVincentX: Nice, this is probably way better than our current code _anyway_

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -23,13 +23,14 @@ function instantiatePrism2(operations: IHttpOperation[]) {
 
 describe('body params validation', () => {
   let server: IPrismHttpServer;
+  let address: string;
 
-  afterAll(() => {
-    return server.fastify.close();
+  afterEach(() => {
+    return server.micri.close();
```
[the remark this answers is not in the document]
karol-maciaszek: The ideal name is actually `server.server` which is also not ideal :/
Since we are using only `.close()` I'd vote on wrapping it and have something like: `server.close()`.

I don't think we should expose the guts of our HTTP server implementation.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -23,13 +23,14 @@ function instantiatePrism2(operations: IHttpOperation[]) {
...
   });
 
   describe('http operation with body param', () => {
-    beforeEach(() => {
+    beforeEach(async () => {
       server = instantiatePrism2([
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -247,25 +248,24 @@ describe('body params validation', () => {
...
+          const response = await makeRequest('/json-body-no-request-content-type', {
+            method: 'POST',
+            body: JSON.stringify({ id: 'string' }),
+            headers: { 'content-type': 'application/json' },
           });
 
-          expect(response.statusCode).toBe(422);
-          expect(JSON.parse(response.payload)).toMatchObject({
+          expect(response.status).toBe(422);
+          expect(await response.json()).toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -280,20 +280,16 @@ describe('body params validation', () => {
...
+          const response = await makeRequest('/json-body-property-required', {
+            method: 'POST',
+            body: '{}',
+            headers: { 'content-type': 'application/json' },
           });
 
-          expect(response.statusCode).toBe(422);
-          expect(JSON.parse(response.payload)).toMatchObject({
+          expect(response.status).toBe(422);
+          expect(await response.json()).toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -303,28 +299,18 @@ describe('body params validation', () => {
...
-
       describe('when no body provided', () => {
         test('returns 422 & error message', async () => {
-          const response = await server.fastify.inject(operation);
-
-          expect(response.statusCode).toBe(422);
-          expect(JSON.parse(response.payload)).toMatchObject({
+          const response = await makeRequest('/json-body-required', { method: 'POST' });
+          expect(response.status).toBe(422);
+          expect(await response.json()).toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: Using Micri HTTP server by karol-maciaszek · Pull Request #927 · stoplightio_prism

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -356,15 +341,14 @@ describe('body params validation', () => {
...
+            const response = await makeRequest('/json-body-required', {
+              method: 'POST',
+              headers: { 'content-type': 'application/json' },
+              body: JSON.stringify({ status: 'string' }),
             });
 
-            expect(response.statusCode).toBe(422);
-            expect(JSON.parse(response.payload)).toMatchObject({
+            expect(response.status).toBe(422);
+            expect(await response.json()).toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: I know you are just scoring points on PR comments count

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -438,18 +422,23 @@ describe('body params validation', () => {
           security: [],
         },
       ]);
+      address = await server.listen(Math.ceil(30000 + Math.random() * 30000), '127.0.0.1');
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -438,18 +422,23 @@ describe('body params validation', () => {
...
     });
 
+    function makeRequest(url: string, init?: RequestInit) {
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -472,17 +461,17 @@ describe('body params validation', () => {
...
-          },
+          }),
+          headers: { 'content-type': 'application/json' },
         });
 
-        expect(response.statusCode).toBe(422);
-        const parsed = JSON.parse(response.payload);
+        expect(response.status).toBe(422);
+        const parsed = await response.json();
         expect(parsed).toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
XVincentX: Yeah, I agree with you. Let's go with that — you can easily copy the signature with TypeScript so we save all the overloads.

packages/http-server/src/server.ts
karol-maciaszek: 👍

packages/http-server/src/server.ts
karol-maciaszek: Good catch!

packages/http-server/src/server.ts
karol-maciaszek: 👍

packages/http-server/src/server.ts
karol-maciaszek: 👍

packages/http-server/src/server.ts
karol-maciaszek: 👍

packages/http-server/src/server.ts
karol-maciaszek: It is quite weird for me, but the typing says that node's ServerInstance.address can return AddressInfo, string or null. I couldn't make it return something other than AddresssInfo but the typings is the typings..

packages/http-server/src/server.ts
karol-maciaszek: Yes, it has been copied from there to mimic the behavior.

packages/http-server/src/server.ts
karol-maciaszek: I don't see where `content-length` has been checked already against `undefined`

packages/http-server/src/server.ts
karol-maciaszek: 👍

packages/http-server/src/server.ts
karol-maciaszek: I believe it is. The array is returned only when there are multiple name-value pairs: `a=1&a=2&a=3`. Prism relies on this behavior inside at least the validator.

It is also part of a bigger problem. I think then the parsing should be done inside validators (then we call it a parser). Then we don't need to guess whether it should be an array of string or whatever. There would be also a single place which does the parsing. We learned already that parsing might be very complicated due to encodings and other stuff.

packages/http-server/src/server.ts
karol-maciaszek: Good point. I need to dig into that, for now it seems that we don't support multiple headers with the same key inside Prism. While micri can provide such to us. The question is:
- shall we support multiple headers of the same name or
- override them in such case so it results with single string?

I will check how fastify was dealing with that.

karol-maciaszek: Fastify does not care about it: https://github.com/fastify/fastify/blob/master/fastify.d.ts#L34

karol-maciaszek: Ok, so Prism is wrong about headers; The duplicates are allowed in certain cases: https://nodejs.org/api/http.html#http_message_headers

karol-maciaszek: Looks like we've been supporting the happy-path where there is always a single string. For me, it seems like a bug and we should fix it in a separate task. @XVincentX agree?

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -333,15 +321,14 @@ describe('body params validation', () => {
...
+            const response = await makeRequest('/json-body-required', {
+              method: 'POST',
+              headers: { 'content-type': 'application/json' },
+              body: JSON.stringify({ id: 'string' }),
             });
 
-            expect(response.statusCode).toBe(422);
-            expect(JSON.parse(response.payload)).toMatchObject({
+            expect(response.status).toBe(422);
+            await expect(response.json()).resolves.toMatchObject({
```
XVincentX:
```suggestion
            expect(response.json()).resolves.toMatchObject({
```

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -356,15 +343,14 @@ describe('body params validation', () => {
...
+            const response = await makeRequest('/json-body-required', {
+              method: 'POST',
+              headers: { 'content-type': 'application/json' },
+              body: JSON.stringify({ status: 'string' }),
             });
 
-            expect(response.statusCode).toBe(422);
-            expect(JSON.parse(response.payload)).toMatchObject({
+            expect(response.status).toBe(422);
+            await expect(response.json()).resolves.toMatchObject({
```
XVincentX:
```suggestion
            expect(response.json()).resolves.toMatchObject({
```

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -472,18 +457,17 @@ describe('body params validation', () => {
...
-          },
+          }),
+          headers: { 'content-type': 'application/json' },
         });
 
-        expect(response.statusCode).toBe(422);
-        const parsed = JSON.parse(response.payload);
-        expect(parsed).toMatchObject({
+        expect(response.status).toBe(422);
+        await expect(response.json()).resolves.toMatchObject({
```
XVincentX:
```suggestion
        expect(response.json()).resolves.toMatchObject({
```

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -29,33 +29,39 @@ async function instantiatePrism(specPath: string) {
...
+
+  return {
+    close: server.close.bind(server),
+    address,
+  };
 }
 
 describe('GET /pet?__server', () => {
-  let server: IPrismHttpServer;
+  let server: { close: () => Promise<void>; address: string };
```
XVincentX: Ok, brace yourself. 

This is going to be a complete waste of your time but I wanted to do this because I'm old and stupid.

TypeScript has a type that infers the returned type from a function — so we can do

`let server: ReturnType<typeof instantiatePrism>` 

…but that will not be enough, because the type is now `Promise<{}>` — while we want what's _INSIDE_ the Promise.

It turns out we can do that using the infer feature in TypeScript 3.8 and the conditional types, so if we write this:

```ts
type ThenArg<T> = T extends Promise<infer U> ? U :
  T extends ((...args: any[]) => Promise<infer U>) ? U :
  T
```

(I think this should handle all the cases, but you never know)

then we can do

```ts
let server: ThenArg<ReturnType<typeof instantiatePrism>>;
```

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -29,33 +29,39 @@ async function instantiatePrism(specPath: string) {
...
 
-  beforeAll(async () => {
-    server = await instantiatePrism(resolve(__dirname, 'fixtures', 'templated-server-example.oas3.yaml'));
+  beforeEach(async () => {
+    server = await instantiatePrism(30002, resolve(__dirname, 'fixtures', 'templated-server-example.oas3.yaml'));
```
XVincentX: I vote to always use the same port — the reason is that this will help us discover cases of server handles not released/freed up quickly instead of blowing up CircleCI or our computers

packages/http-server/src/server.ts
XVincentX: You are right, I got confused from `content-type` above :)

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+      url,
+      method,
+      headers,
+      socket,
+    } = request;
 
-    if (typeIs(req, ['application/x-www-form-urlencoded'])) {
-      return done(null, body);
+    if (!url) {
+      return TE.left(new Error('Missing URL in request!'));
```
XVincentX: Ok this condition is never going to happen.

Follow up question: "Why is the URL then nullable?"

Because the same object `HttpServer.IncomingMessage` is used also for a ServerRequest which has not such property.

In our case — `HttpServer.IncomingMessage` is coming from a `ClientRequest` — which always has the URL property.

Source: https://nodejs.org/dist/latest-v12.x/docs/api/http.html#http_class_http_incomingmessage

packages/http-server/src/server.ts
XVincentX: Hmm I think in reality Fastify has been grouping the headers for us, and we should replicate the same thing here.

image

packages/http-server/src/server.ts
XVincentX: A string is returned if the server is listening on on a pipe or Unix domain socket; for TCP Socket is always going to be `AddressInfo`

packages/http-server/src/server.ts
```diff
@@ -10,63 +11,75 @@ import { IPrismDiagnostic } from '@stoplight/prism-core';
...
+    const body = await parseRequestBody(request);
 
-  const replyHandler: fastify.RequestHandler = (request, reply) => {
-    const {
-      req: { method, url },
-      body,
-      headers,
-      query,
-    } = request;
+    const { searchParams, pathname } = new URL(url, addressInfoToString(socket.address()));
```
XVincentX: Given all we want to do here is separate the searchParams and the pathname, we can use a fake url so we do not have to call `socket.address` for each request

```suggestion
    const { searchParams, pathname } = new URL(url, 'http://IDoNotCare');
```

Although I'm wondering if there's an alternative…

packages/http-server/src/server.ts
```diff
@@ -97,67 +110,86 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+              Object.entries(output.headers).forEach(([name, value]) => reply.setHeader(name, value));
 
-            reply
-              .code(output.statusCode)
-              .serializer((payload: unknown) => serialize(payload, reply.getHeader('content-type')))
-              .send(output.body);
+            send(
+              reply,
+              output.statusCode,
+              serialize(output.body, reply.getHeader('content-type') as string | undefined)
```
XVincentX: Hmm — the cast is effectively legit. I guess there is nothing we can do about it here.

packages/http-server/src/server.ts
karol-maciaszek: Aha! It is not Fastify who is grouping the headers, it is node's http. See here: https://nodejs.org/api/http.html#http_message_headers. There is list of rules for headers, Set-Cooke has special treatment, Cookie is always an array and custom headers are grouped with comma as in your example.

Fastify was doing physically nothing to headers (https://github.com/fastify/fastify/blob/master/fastify.d.ts#L34). It just made typing loose (`{ [key: string]: any }`) so Prism was not complaining.

To sum this up, this PR **does not introduce any bug regarding headers treatment**. But we definitely had bug before and still have one.

I think about separate PR because due to different treatment of different headers it might be not so simple to tackle.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -247,25 +248,24 @@ describe('body params validation', () => {
           security: [],
         },
       ]);
+      address = await server.listen(Math.ceil(30000 + Math.random() * 30000), '127.0.0.1');
```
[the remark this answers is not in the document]
karol-maciaszek: body-params and server.oas can run in parallel leading to ADDRINUSE errors. Before, we did not need to worry about this because of `fastify.inject`. I guess I can make the port the same across single test suite but not across different ones. I also saw somewhere a lib for dynamic assignment of ports (more clever than the Math.random solution).

Btw. how single port can help use identify the tests not-closing-handles?

packages/http-server/src/__tests__/body-params-validation.spec.ts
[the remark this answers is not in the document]
karol-maciaszek: 👍

packages/http-server/src/__tests__/server.oas.spec.ts
karol-maciaszek: You are crazy.

packages/http-server/src/server.ts
karol-maciaszek: Said some guy about 640kb of memory ;)

Anyway, I consider making req.url! with a proper comment, WDYT?

packages/http-server/src/server.ts
XVincentX: Ok that's fair. Let's put a comment and we can handle that in a different PR.

packages/http-server/src/server.ts
XVincentX: Fair!

packages/http-server/src/server.ts
XVincentX: Allright. Let's put a comment and figure it out in a following PR.

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -303,28 +301,18 @@ describe('body params validation', () => {
...
-
       describe('when no body provided', () => {
         test('returns 422 & error message', async () => {
-          const response = await server.fastify.inject(operation);
-
-          expect(response.statusCode).toBe(422);
-          expect(JSON.parse(response.payload)).toMatchObject({
+          const response = await makeRequest('/json-body-required', { method: 'POST' });
+          expect(response.status).toBe(422);
+          await expect(response.json()).resolves.toMatchObject({
```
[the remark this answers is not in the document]
karol-maciaszek: There must be a lot of magic behind the scenes of jest

packages/http-server/src/server.ts
karol-maciaszek: The alternative might be `urijs`. But it adds extra dependency vs. built-in URL&URLSearchParams.

I changed to `http://example.com` however, I feel like replacing magic with sorcery ;)

packages/http-server/src/__tests__/server.oas.spec.ts
karol-maciaszek: Test suites can't be on the same port because they run in parallel and will occasionally fail.

`    listen EADDRINUSE: address already in use 127.0.0.1:30000`

karol-maciaszek: Updated so the port inside test suite is the same

karol-maciaszek: I also commented on those ports. The Next Generation which will work on Prism might fall into hard to debug problem

packages/http-server/src/__tests__/server.oas.spec.ts
karol-maciaszek: Ok it works, now I need to learn why and how it works.`

packages/http-server/src/server.ts
karol-maciaszek: I would prefer to keep the function clean but this time it will put too much garbage into the place where it is called.

I would need to cast it like that: `addressInfoToString(server.address() as AddressInfo | undefined)` to keep compiler satisfied.

Are you ok with how it looks now or you have other preference?

PR #944, opened 2020-01-24
Use type-is for content type comparison

### LINKED ISSUE #943: Including "charset=UTF-8" in Content-Type header causes an incorrect 422 response
**Describe the bug**

A few of our client libraries automatically add "UTF-8" to it's Content-Type headers, like so `Content-Type: application/x-www-form-urlencoded; charset=UTF-8`. When querying these mocked endpoints with this extra field, we get back a 422 with the following body. When removed, we get back the expected mocked response.

```
{
    “type”: “https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY”,
    “title”: “Invalid request body payload”,
    “status”: 422,
    “detail”: “Your request is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.“,
    “validation”: [
        {
            “location”: [
                “body”
            ],
            “severity”: “Error”,
            “code”: “type”,
            “message”: “should be object”
        },
      ...
  ]
}
```

**To Reproduce**

Given this scenario file, see the failing error. The `x-www-urlencoded-ok.oas2.txt` is the same, minus the UTF-8 portion of the Content-Type header.

```
====test====
When I have a document with a Request with form-data body that should
be an array of strings (comma separated values)
And I send the correct values
And I include `charset=UTF-8` in my Content-Type
I should receive a 200 response
====spec====
swagger: '2.0'
paths:
  /path:
    post:
      produces:
        - text/plain
      consumes:
      responses:
        200:
          schema:
            type: string
      parameters:
        - in: formData
          type: array
          name: arr
          items:
            type: string
          collectionFormat: csv
====server====
mock -p 4010 ${document}
====command====
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/x-www-form-urlencoded; charset=UTF-8" --data-urlencode "arr=a,b,c"
====expect====
HTTP/1.1 200 OK
content-type: text/plain

string
```

**Expected behavior**

Valid defined response returned

 - Library version: 3.2.3 and `master` from source
 - OS: OSX Catalina & Mojave

**Additional context**

We noticed this in our 3.0.2 document, but also recreated in swagger 2.0 (shown above). If I was to take a wild guess, it might be [this bit here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/index.ts#L113) that's not handling the content-type correctly. But, I was struggling to re-build the binary to run the test-harness after tweaking some*. I couldn't find a relevant unit test that I could manipulate around this, but if you point me in the right direction, I can try investigating some more on my end.

Side note: I was able to `yarn build.binary` once, but when I tried to rebuild I got "Cannot write file '...' because it would overwrite input file." errors

XVincentX:
This PR will replace different checks we were doing in places to look for content so that instead of using regular string comparison we use the type-is library which will normalise content types with charset and so on

Closes #943

### DIFF
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -24,7 +24,7 @@ export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpO
 
 export function findBestHttpContentByMediaType(
   response: PickRequired<IHttpOperationResponse, 'contents'>,
-  mediaType: string[],
+  mediaType: string[]
 ): Option<IMediaTypeContent> {
   const bestType = accepts({
     headers: {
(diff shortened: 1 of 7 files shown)


packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -34,59 +34,56 @@ export function findBestHttpContentByMediaType(
...
     response.contents,
-    findFirst(content => content.mediaType === '*/*'),
+    findFirst(content => content.mediaType === '*/*')
   );
 }
 
 const byResponseCode = ord.contramap<number, IHttpOperationResponse>(ordNumber, response => parseInt(response.code));
 
-export function findLowest2xx(httpResponses: IHttpOperationResponse[]): Option<IHttpOperationResponse> {
+export function findLowest2xx(httpResponses: NonEmptyArray<IHttpOperationResponse>): Option<IHttpOperationResponse> {
```
karol-maciaszek: I'm aware of the discussion on http-spec typings. Just thinking of using IHttpOperation['responses'] as a type here to be consistent with what we could receive. Despite the fact of how bad the input things can be.

XVincentX: I would say these function will take a non empty array, we will validate this ahead of the time, most likely in the CLI.

If we're going to receive garbage input — we're going to make sure to validate it before it enters in Prism-Http.

karol-maciaszek: Sounds good!

karol-maciaszek: Well, almost good. I would be happier when we could rely on IHttpOperation for that.

XVincentX:
@karol-maciaszek The current version of http-spec is still returning a NonEmptyArray — when we will upgrade to the latest version (assuming it will make it in master) when we will upgrade the package we'll get type errors everywhere and we'll fix them.

PR #959, opened 2020-02-03
Client fixes

XVincentX:
* Adds documentation on partial urls
* Types the `CreatePrism` instances for a faster find next time I'll need to refactor them
* Moves the `errors` option (it was listed twice!) from the wrong place
* Small changes in the internal logging creation
* Removes `Error.captureStackTrace(this, ProblemJsonError);` — that function is V8 specific and does not work in other browsers

### DIFF
--- a/docs/guides/http-client.md
+++ b/docs/guides/http-client.md
@@ -81,3 +81,15 @@ For the shortcut methods (since the only mandatory option is intrinsic in the fu
 ```ts
 client.get('https://google.it', { validateRequest: false }).then(response => console.log(response));
 ```
+
+You can also use relative links when doing requests. In such case you won't be able to use the proxy and the server validation will be disabled:
+
+```ts
+client.get('/users/10', { validateRequest: false }).then(response => console.log(response));
+```
+
+…or you can also set the base path in the options object:
+
+```ts
+client.get('/users/10', { baseUrl: 'https://api.stoplight.io/' }).then(response => console.log(response));
+```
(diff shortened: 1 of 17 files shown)


packages/http-server/src/types.ts
```diff
@@ -16,6 +15,4 @@ export interface IPrismHttpServer {
   listen: (port: number, address?: string, backlog?: number) => Promise<string>;
 }
 
-export type ThenArg<T> = T extends Promise<infer U> ? U :
-  T extends ((...args: any[]) => Promise<infer U>) ? U :
-    T;
+export type ThenArg<T> = T extends Promise<infer U> ? U : T extends (...args: any[]) => Promise<infer U> ? U : never;
```
karol-maciaszek: When I saw it I was afraid that you did some changes to that line. Fortunately, it is linter who did it ;)

docs/guides/http-client.md
```diff
@@ -81,3 +81,15 @@ For the shortcut methods (since the only mandatory option is intrinsic in the fu
 ```ts
 client.get('https://google.it', { validateRequest: false }).then(response => console.log(response));
 ```
+
+You can also use relative links when doing requests. In such case you won't be able to use the proxy and the server validation will be disabled:
```
philsturgeon: are these relative or root-relative? https://mor10.com/html-basics-hyperlink-syntax-absolute-relative-and-root-relative/

XVincentX: Root relative. `/pet/10`

PR #961, opened 2020-02-04
Make matchBaseUrl exception-safe

### LINKED ISSUE #960: Make sure matchBaseUrl does not throw
In Prism we try to never throw exceptions, since the Error is part of its domain model.

We still have few places where this is happening, and we want to fix that.

https://github.com/stoplightio/prism/blob/master/packages/http/src/router/matchBaseUrl.ts

It would be great to come up with a way to not throw an exception anymore and return an `Either` structure — and then propagate it back to the user with the relative `.chain` method.

marcelltoth:
Closes #960


packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,55 @@
...
+import { pipe } from 'fp-ts/lib/pipeable';
+
 import { INodeVariable, IServer } from '@stoplight/types';
 import { MatchType } from './types';
 
 const variableRegexp = /{(.*?)}/g;
 
-export function matchBaseUrl(server: IServer, baseUrl: string): MatchType {
-  const templateMatchResult = matchRequestUrlToTemplateUrl(baseUrl, server.url, server.variables);
+export function matchBaseUrl(server: IServer, baseUrl: string) {
```
XVincentX: Let's always try to be explicit with the return types, especially when they're monads

```suggestion
export function matchBaseUrl(server: IServer, baseUrl: string): E.Either<Error, MatchType> {
```

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,55 @@
...
+          matches => (matches.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE)
+        )
+      )
+    ),
+    E.fold(
+      e => {
+        throw e;
+      },
+      matchType => matchType
+    )
```
XVincentX: Don't do this — the idea of the monad is to propagate it back as it is unless you really need to handle the error. Just return the either and then we can use the `chain` method to use it back

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,55 @@
...
 const variableRegexp = /{(.*?)}/g;
 
-export function matchBaseUrl(server: IServer, baseUrl: string): MatchType {
-  const templateMatchResult = matchRequestUrlToTemplateUrl(baseUrl, server.url, server.variables);
+export function matchBaseUrl(server: IServer, baseUrl: string) {
+  return pipe(
+    convertTemplateToRegExp(server.url, server.variables),
+    E.map(regex =>
+      pipe(
+        O.fromNullable(regex.exec(baseUrl)),
```
XVincentX: Either has a `fromNullable` method as well — and then you can use `E.orElse` to return a fallback value so that it gets transformed in a `right`

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,55 @@
...
 
-export function convertTemplateToRegExp(urlTemplate: string, variables?: { [name: string]: INodeVariable }) {
-  const regexp = !variables
-    ? urlTemplate
-    : urlTemplate.replace(variableRegexp, (_match, variableName) => {
-        const variable = variables[variableName];
+    function replaceStringUnsafe(input: string): string {
+      return input.replace(variableRegexp, (_match, variableName) => {
+        const variable = vars[variableName];
         if (!variable) {
```
XVincentX: I think this is the place where we should be using `E.tryCatch()` — if we can convert this into two regular `E.right` and `E.left` would even be a bonus point

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,55 @@
...
+    ),
+    E.map(regexString => new RegExp(`^${regexString}$`))
+  );
 
-  return templateMatchResult.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE;
-}
+  function replaceString(vars: { [name: string]: INodeVariable }, input: string) {
+    return E.tryCatch<Error, string>(
+      () => replaceStringUnsafe(input),
+      e => (e instanceof Error ? e : new Error('unknown error'))
```
XVincentX: Why wouldn't this be an error?

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
```diff
@@ -55,8 +56,8 @@ describe('matchServer.ts', () => {
               host: { default: 'www.example.com' },
             },
           },
-          'http://stoplight.io/v1',
-        ),
+          'http://stoplight.io/v1'
+        )
```
XVincentX: Did you change this or was prettier reformatting the code?

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
```diff
@@ -136,7 +137,7 @@ describe('matchServer.ts', () => {
   describe('convertTemplateToRegExp()', () => {
     test('given no variables should resolve to the original string', () => {
       const regexp = convertTemplateToRegExp('{a}');
-      expect(regexp).toEqual(/^{a}$/);
+      expect(regexp).toEqual(E.right(/^{a}$/));
```
XVincentX: We have a dedicated helper for this — check out the `assertRight` function

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
```diff
@@ -156,17 +157,19 @@ describe('matchServer.ts', () => {
...
-      expect(() =>
-        convertTemplateToRegExp('{a}', {
-          b: {
-            default: 'vb',
-          },
-        }),
-      ).toThrow(`Variable 'a' is not defined, cannot parse input.`);
+    test('given single variable and no matching variable should return Left', () => {
+      expect(
+        E.isLeft(
```
XVincentX: ditto, `assertLeft` is the function you want to use

packages/http/src/router/matchBaseUrl.ts
marcelltoth: This - and the second one - was actually intentional. I was looking to commit this in 2 phases, first just changing the internals, but keeping the behavior unchanged, then - if you are OK with the implementation - go ahead and bubble the signature change up. So this change is comming.:) (See **Status**.)

packages/http/src/router/matchBaseUrl.ts
marcelltoth: Got you, see comment above.

packages/http/src/router/matchBaseUrl.ts
marcelltoth: I'll try that, thanks for the tip.

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
marcelltoth: It surely wasn't intentional, I might have accidentally fired some formatting tool on it though. I'm going to check what happens if I put it back and I fire prettier on it, and we'll see.

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
marcelltoth: Oh, that's what I was looking for... Thank you, I'll clean those up.

packages/http/src/router/matchBaseUrl.ts
marcelltoth: This was adopted from the example in the `fp-ts` documentation. In this case I do know that it will be an `Error`, so I could really skip the check, but even then, `e` is non-generically typed as `unknown`. How do I solve it in a simpler way without a typecast?

XVincentX: Using the `toError` function from fp-ts :trollface: 

```suggestion
      E.toError
```

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
marcelltoth: Double checked, running prettier removes them.

XVincentX: No that's fine, I know that we upgrader prettier some time ago and it's refactoring the code as we change it.

XVincentX: @marcelltoth We should have an hook on TS files that should reformat the source code even if you format in your own way — that's kind of weird.

packages/http/src/router/matchBaseUrl.ts
marcelltoth: This library starts to feel like lodash. _There's a function for everything_

XVincentX: It's battle tested, they've gone through these use cases all the times. You were right on the approach though, which is good.

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,45 @@
...
+      () => E.right(urlTemplate),
+      vars => replaceString(vars, urlTemplate)
+    ),
+    E.map(regexString => new RegExp(`^${regexString}$`))
+  );
 
-  return templateMatchResult.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE;
-}
+  function replaceString(vars: { [name: string]: INodeVariable }, input: string) {
+    return E.tryCatch<Error, string>(() => replaceStringUnsafe(input), E.toError);
```
XVincentX:
```suggestion
    return E.tryCatch(() => replaceStringUnsafe(input), E.toError);
```

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,45 @@
...
+    O.fold(
+      () => E.right(urlTemplate),
+      vars => replaceString(vars, urlTemplate)
+    ),
+    E.map(regexString => new RegExp(`^${regexString}$`))
+  );
 
-  return templateMatchResult.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE;
-}
+  function replaceString(vars: { [name: string]: INodeVariable }, input: string) {
```
XVincentX:
```suggestion
  function replaceString(vars: Dictionary<INodeVariable>, input: string): E.Either<Error, string> {
```

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,45 @@
...
+  variables?: { [name: string]: INodeVariable }
+): E.Either<Error, RegExp> {
+  return pipe(
+    O.fromNullable(variables),
+    O.fold(
+      () => E.right(urlTemplate),
+      vars => replaceString(vars, urlTemplate)
+    ),
+    E.map(regexString => new RegExp(`^${regexString}$`))
+  );
```
XVincentX: Try to stick to the same monad whenever is possible:

```ts
  return pipe(
    variables,
    E.fromNullable(new Error('no variables')),
    E.fold(() => E.right(urlTemplate), vars => replaceString(vars, urlTemplate)),
    E.map(regexString => new RegExp(`^${regexString}$`))
  );
```

However since the `E.fromNullable` here is doing nothing else than discriminating the starting `Either`, we can also write

```ts
  return pipe(
    variables ? replaceString(variables, urlTemplate) : E.right(urlTemplate),
    E.map(regexString => new RegExp(`^${regexString}$`))
  );

```

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,45 @@
...
+    )
+  );
+}
 
-  if (!templateMatchResult) {
-    return MatchType.NOMATCH;
-  }
+export function convertTemplateToRegExp(
+  urlTemplate: string,
+  variables?: { [name: string]: INodeVariable }
```
XVincentX:
```suggestion
  variables?: Dictionary<INodeVariable>
```

packages/http/src/router/matchBaseUrl.ts
```diff
@@ -1,23 +1,45 @@
...
+export function matchBaseUrl(server: IServer, baseUrl: string): E.Either<Error, MatchType> {
+  return pipe(
+    convertTemplateToRegExp(server.url, server.variables),
+    E.chain(regex =>
+      pipe(
+        regex.exec(baseUrl),
+        E.fromNullable(new Error('No matches')),
+        E.map(matches => (matches.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE)),
+        E.orElse(() => E.right<Error, MatchType>(MatchType.NOMATCH))
+      )
```
XVincentX:
```ts
  return pipe(
    convertTemplateToRegExp(server.url, server.variables),
    E.map(regex => regex.exec(baseUrl)),
    E.map(matches => matches ? (matches.length > 1 ? MatchType.TEMPLATED : MatchType.CONCRETE) : MatchType.NOMATCH),
  );
```

packages/http/src/router/matchBaseUrl.ts
XVincentX: Actually please use the latter version

```ts
  return pipe(
    variables ? replaceString(variables, urlTemplate) : E.right(urlTemplate),
    E.map(regexString => new RegExp(`^${regexString}$`))
  );
```

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
```diff
@@ -156,17 +168,17 @@ describe('matchServer.ts', () => {
...
+    test('given single variable and no matching variable should return Left', () => {
+      assertLeft(
         convertTemplateToRegExp('{a}', {
           b: {
             default: 'vb',
           },
-        }),
-      ).toThrow(`Variable 'a' is not defined, cannot parse input.`);
+        })
+      );
```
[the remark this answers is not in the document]
marcelltoth: Missed it, my bad.

packages/http/src/router/matchBaseUrl.ts
marcelltoth: Makes sense, thank you.

packages/http/src/router/__tests__/matchBaseUrl.spec.ts
```diff
@@ -8,21 +9,31 @@ describe('matchServer.ts', () => {
         {
           url: 'http://www.example.com/',
         },
-        'http://www.example.com/',
+        'http://www.example.com/'
       );
 
-      expect(serverMatch).toEqual(MatchType.CONCRETE);
+      assertRight(serverMatch, result => result === MatchType.CONCRETE);
```
XVincentX:
```suggestion
      assertRight(serverMatch, result => expect(result).toBe(MatchType.CONCRETE));
```

XVincentX: The same with all the other assertions

karol-maciaszek:
@marcelltoth congrats! 👍

marcelltoth:
Thank you guys! First functional PR 💪

karol-maciaszek:
Please accept your first Fourier transform badge:

# ℱ

XVincentX:
One day I'll make a tattoo on my arm with this `f`.

PR #965, opened 2020-02-06
refactor: remove double error definition

XVincentX:
I figured that we have defined the error class twice, for no reason. This PR removes the duplicate.


packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -24,6 +22,7 @@ import {
   hasContents,
 } from './InternalHelpers';
 import { IHttpNegotiationResult, NegotiatePartialOptions, NegotiationOptions } from './types';
+import { ProblemJsonError } from '@stoplight/prism-http';
```
marcelltoth: This is currently a TS error, as it cannot find it. I'm not yet familiar with the TS setup here, are we supposed to import from "ourselves" like this?
```
import { ProblemJsonError } from 'http/src'
```
fixes the problem.

XVincentX: Oh I made a mistake here, fixed.

PR #969, opened 2020-02-09
fix: disambiguate on methods when server validation is not enabled

### LINKED ISSUE #968: Prism returning different status codes and data types than expected
**Describe the bug**
My development team uses Prism to test our API contracts during the build process for our React app. During our most recent build, several API contract tests failed due to Prism returning different status codes as well as data types that differ from the OpenAPI spec. Our tests were passing locally, but we were able to recreate the issue by removing the node modules and doing a fresh npm install.

These are the errors we are experiencing:

* Prism returning an object instead of an array (as specified in OpenAPI spec)

* Expected: 204 Received: 200

* Request failed with status code 422

**To Reproduce**

**Prism is returning the response for the first call under a path (responses are being interchanged)**

Using the following example: 
```
paths:
  /api/users:
    post:
      summary: "Create a user"
      operationId: "dal.users.create_and_get"
      requestBody:
        description: "Parameters to define a new user."
        content:
          application/json:
            schema:
              x-body-name: "user"
              $ref: "#/components/schemas/UserCreateEntity"
            example:
              name: "test_user"
      responses:
        200:
          description: "User created successfully"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/UserCreateResponseEntity"
              example:
                id: "e274bf3ebe3841a88ade1630515624c6"
                name: "test_user"
        409:
          description: "Duplicate user name"
    get:
      summary: "List all users"
      operationId: "dal.users.get_all"
      responses:
        200:
          description: "Get a list of all Users."
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/UserCollectionEntity"
```

If I make a `get` call to `/api/users`, I am expecting to receive an array of users. Prism is actually returning the response for the `post` call, and is responding with a `UserCreateResponseEntity` instead of a `UserCollectionEntity`.

We have observed that if there are multiple CRUD operations under a path, Prism responds with the first operation by default. We have observed this across multiple OpenAPI Spec files and tests. 

**Expected behavior**
1. Prism should return the correct data type based on OpenAPI Spec
2. Prism should not be returning 422 because there is no request body for that call
3. Prism should return the status code as defined in the OpenAPI Spec

**Screenshots**

 - Library version: `@stoplight/prism-cli": "^3.2.5`,
 - OS: MacOS Mojave 
 - Browser: Chrome 79.0.3945.130

**Additional context**

**We feel that this issue is linked to a dependency in Prism. We pinned Prism's dependency versions (removed ^), and were able to resolve the issue.**

XVincentX:
Closes #968 

Adding tests in a while…

### DIFF
--- a/test-harness/specs/routing/select-correct-method.oas3.txt
+++ b/test-harness/specs/routing/select-correct-method.oas3.txt
@@ -2,17 +2,11 @@
 When multiple actions are mapped to a single endpoint, selects the correct response based on the incomming HTTP method
 ====spec====
 openapi: 3.0.0
-info:
-  title: bug
-  version: '1.0'
 paths:
   /test:
     post:
-      summary: Create a user
-      operationId: dal.users.create_and_get
       responses:
         '200':
-          description: User created successfully
           content:
             application/json:
               schema:
@@ -21,8 +15,6 @@ paths:
                   response-for-post:
                     type: string
     get:
-      summary: List all users
-      operationId: dal.users.get_all
       responses:
         '200':
           content:
@@ -32,8 +24,6 @@ paths:
                 properties:
                   response-for-get:
                     type: string
-    parameters: []
-
 ====server====
 mock -p 4010 ${document}
 ====command====
--- a/test-harness/specs/routing/select-correct-method.oas3.txt
+++ b/test-harness/specs/routing/select-correct-method.oas3.txt
@@ -1,0 +1,45 @@
+====test====
+When multiple actions are mapped to a single endpoint, selects the correct response based on the incomming HTTP method
+====spec====
+openapi: 3.0.0
+info:
+  title: bug
+  version: '1.0'
+paths:
+  /test:
+    post:
+      summary: Create a user
+      operationId: dal.users.create_and_get
+      responses:
+        '200':
+          description: User created successfully
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response-for-post:
+                    type: string
+    get:
+      summary: List all users
+      operationId: dal.users.get_all
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response-for-get:
+                    type: string
+    parameters: []
+
+====server====
+mock -p 4010 ${document}
+====command====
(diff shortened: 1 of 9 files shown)


marcelltoth:
I added a harness test to check for the exact scenario described in the issue.

Also verified that the exact same test fails on current master.

test-harness/specs/routing/select-correct-method.oas3.txt
```diff
@@ -0,0 +1,45 @@
+====test====
+When multiple actions are mapped to a single endpoint, selects the correct response based on the incomming HTTP method
+====spec====
+openapi: 3.0.0
+info:
+  title: bug
+  version: '1.0'
```
XVincentX: Let's remove this.

test-harness/specs/routing/select-correct-method.oas3.txt
```diff
@@ -0,0 +1,45 @@
...
+paths:
+  /test:
+    post:
+      summary: Create a user
+      operationId: dal.users.create_and_get
```
XVincentX: This is not necessary

test-harness/specs/routing/select-correct-method.oas3.txt
```diff
@@ -0,0 +1,45 @@
...
+      responses:
+        '200':
+          description: User created successfully
```
XVincentX: This is not necessary

test-harness/specs/routing/select-correct-method.oas3.txt
```diff
@@ -0,0 +1,45 @@
...
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  response-for-get:
+                    type: string
+    parameters: []
```
XVincentX: This is not necessary

XVincentX:
Will merge and release tomorrow.

marcelltoth:
> 
> 
> Looks good. The only thing is that we try to use super minimal gas documents so that it's easy to understand what the original issue was. @marcelltoth can you remove all the unnecessary parts?

Got it. I see you already completed this one, but I'll try to be perfectly minimal next time.

PR #975, opened 2020-02-10
Fix "Cannot write file ... because it would overwrite..." errors

marcelltoth:
`yarn build` would fail after the first invocation, because it tries to compile the - already existing - output files as well.

Fixed that by explicitly excluding the _dist_ folders from the typescript build.

### DIFF
--- a/packages/tsconfig.build.json
+++ b/packages/tsconfig.build.json
@@ -1,5 +1,5 @@
 {
-  "exclude": ["**/__tests__"],
+  "exclude": ["**/__tests__", "*/dist"],
   "include": [],
   "extends": "./tsconfig.json",
   "references": [


XVincentX:
Nicely done, but we must test this carefully before we merge this in. I have been struggling with the same thing back in the time and removing such directory was giving problems in VSCode — and the problem is in the TypeScript incremental compiler feature. i do not know if they fixed the bug somehow, but I'll pull down the branch and play with it tomorrow.

marcelltoth:
Alright, let me know. I only added the changes to the `tsconfig.build.json` so I'm hoping it's gonna pass.

XVincentX:
It didn't really fix the issue

image

Indeed, the `dist` directory is already excluded by the compilation by default because TypeScript infers that from the "outDir" option (`"outDir": "dist"`)

I still recall this is a problem of the incremental compiler, and it's not about directory and stuff. Moreover, the `tsconfig.build.json` is just a holder to compile the 4 projects simultaneously — the real files to modify (in case we'll be able to) are the singular one in the projects.

Even modifying these — the issue is still there.

I think the reason why it works on your computer is because you've rebuilt a project that didn't have no difference (aka compiling twice the same files) and so no file was effectively written. Next time, try to compile first, change something in any typescript file, and then you recompile. You should get the error.

I'm closing this, but feel free to reopen in case you want to dig into more. Do not spend too much time on it though, I lost nights and I ended up digging in the TypeScript's source code to understand the root of the problem.

PR #979, opened 2020-02-11
chore: autopublish from master

philsturgeon:


philsturgeon:
Ah yeah this is github flow not git flow. In spectral, master = released. Erf. 

I guess I don't want documentation improvements to be held up by the release timeline, but I could always hit Publish in Studio.

PR #984, opened 2020-02-13
Prefer header

### LINKED ISSUE #966: Support the Prefer header
We currently support the selection of a specific response with the `__code` query parameter and a specific example with `__example` query parameter.

However, instead of polluting the query string with such thing we should probably be using the [Prefer header](https://tools.ietf.org/html/rfc7240)

We could do something like
`Prefer: code=201, example=named_example, dynamic=true`

I would also suggest, once this feature is merged to declare the query parameters thing deprecated.

XVincentX:
* Replace our homemade query string validation with an appropriate io-ts decoder
* Converting the process of decoding in an Either, pushing it in the pipe list we already have
* Modify the tests
* Implement options lookup in headers, fallback to query for compatibility

Closes #966


XVincentX:
I am still not super super happy with the shape, and I would like to avoid to do copy pasta of stuff to test this properly, since the difference here is merely the data source.

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -1,26 +1,37 @@
...
+      example: t.string,
+    },
+    'Preferences'
+  ),
+]);
 
-  if (!query) {
-    return {};
-  }
+type requestPreference = Partial<Omit<IHttpOperationConfig, 'mediaType'>>;
```
karol-maciaszek:
```suggestion
type RequestPreference = Partial<Omit<IHttpOperationConfig, 'mediaType'>>;
```

Lower-case type name may mix with the variable name. At least during review on GitHub.

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -1,26 +1,37 @@
...
 
-  const { __code, __dynamic, __example } = query;
+export const getHttpConfigFromRequest = (req: IHttpRequest): E.Either<Error, requestPreference> => {
+  const preferences =
+    req.headers && req.headers['prefer']
```
karol-maciaszek: Thinking about two things here:
1. if `Prefer` header exists then all values passed via query are ignored. So:
`
GET /?__dynamic=false
Prefer: code=202
`
Will result in `dynamic=true; code=202`.

So, there are few questions. Do we want to force-move people to Prefer header? If so, then leaving it as is have educational value. If not, then we should support mixture of query and header with header preference.


2. I believe we should be a bit more careful about this header. Our power-api-expert-enterprise users may already use Prefer header and have it defined in their API descriptions (e.g. with regex pattern). Our headers can interfere with theirs leading to unexpected 422s.

I would extend the approach with stripping "our" preferences from header and if it is empty after that, removing it at all. That should happen before further processing in Prism.

XVincentX: 1. This is correct — if somebody specifies the options in the header, they will take the full preference and and the query is going to be ignored. I'll update the documentation so that this stands clear.

2. That is really really unlikely to be happening. The `Prefer` header is a multi key value and multi purpose header and the changes that some other proxies (Cloudflare, Netlify, Fastly or whatever is fancy these days) are using/modifying it is a real possibility; this means that this is one of the few headers that you can't really validate for real if you do not parse it first — thing that almost nobody is doing. My take is that for now we keep it as it is (no strip, no parse) and if somebody will start to complain, we will figure out a solution for it.

karol-maciaszek: > This is correct — if somebody specifies the options in the header, they will take the full preference and and the query is going to be ignored. I'll update the documentation so that this stands clear.

It is minor, but if you find a suitable place for this information it would be good to add it.

karol-maciaszek: > That is really really unlikely to be happening. The Prefer header is a multi key value and multi purpose header and the changes that some other proxies (Cloudflare, Netlify, Fastly or whatever is fancy these days) are using/modifying it is a real possibility; this means that this is one of the few headers that you can't really validate for real if you do not parse it first — thing that almost nobody is doing. My take is that for now we keep it as it is (no strip, no parse) and if somebody will start to complain, we will figure out a solution for it.

I agree. We are not closing the doors for fixing it in the future.

PR #992, opened 2020-02-19
Access-Control-Allow-Credentials header

mtfuji:
- Add Access-Control-Allow-Credentials header to othrerwise methods.
- for non-preflight request environment.


XVincentX:
Thanks.

PR #993, opened 2020-02-19
fix: returning actual body

karol-maciaszek:
Prism's http-client is no longer returning `{}` when the body is empty.


karol-maciaszek:
> I am wondering why I did this back in the time.

Together with @marcelltoth, we asked the same question when doing git blame :]

PR #1005, opened 2020-02-24
chore: convert all examples to yaml

XVincentX:
There is currently no reason to have the petstore json example and the oas3 one in JSON. Remove and unify.


packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -206,7 +206,7 @@ describe.each([['petstore.no-auth.oas2.yaml', 'petstore.no-auth.oas3.yaml']])('s
     });
 
     // oas2 does not support overriding servers and named examples
-    if (file === 'petstore.oas3.json') {
+    if (file === 'petstore.no-auth.oas3.yaml') {
```
marcelltoth: I think you'd need to update lines 178-179 to _.no-auth_ too, don't you?

XVincentX: I've put it in a constant so we stop with these magic strings.

XVincentX:
> Where do these *.no-auth.oas3.yaml files come from?

We have them in the repo:

https://github.com/stoplightio/prism/blob/master/packages/http-server/src/__tests__/fixtures/petstore.no-auth.oas3.yaml

https://github.com/stoplightio/prism/blob/master/packages/http-server/src/__tests__/fixtures/petstore.no-auth.oas2.yaml

marcelltoth:
> 
> 
> > Where do these *.no-auth.oas3.yaml files come from?
> 
> We have them in the repo:

I don't know what I was looking at 😅

PR #1009, opened 2020-02-27
refactor: move createFromString and resource from Prism Http to CLI 

### LINKED ISSUE #1008: Remove createClientFromSpec method from Http
The `createClientFromSpec` method forces Prism http to have a reference to some packages using the file system — that complicates the game when using it in the browser and non node environments in general.

We are going to remove such method and either move it in the CLI package or get rid of it completely so that browser bundling of Prism will be a breeze.

XVincentX:
This PR is a proposal of moving the methods `createClientFromResource` and `createClientFromSpec` from the `prism-http` package to `prism-cli` package

The main rationale behind this is that the dependencies of such feature (json-ref-resolver, json-ref-readers) have an hard dependency on the `fs` package that makes prism-http a little bit harder to embed in the browser (you need to configure web pack to mock the `fs` module to empty)

---

This is technically a breaking change since we're removing a function although we have never "marketed" our client API — I'm summoning @philsturgeon to get his opinion.

If we're going to take this as a breaking change (Prism 4), then I'd take the occasion to also

1. Change our artefacts directory from `dist` to `lib` — it makes more sense and does not make you feel bad when importing stuff // cc @marcelltoth — we had such discussion some time ago
2. Rename some of the client methods to something that makes a little bit more sense

Closes #1008 
Closes stoplightio/elements/issues/261


docs/guides/http-client.md
```diff
@@ -2,19 +2,33 @@
 
 Prism includes a fully-featured HTTP Client that you can use to seamlessly perform requests to both a real server and a mocked document. The client is modeled after Axios so it may feel familiar.
 
-### Create from a filename or http resource
+### Create From Manual Http Operations
```
philsturgeon:
```suggestion
### Create From Manual HTTP Operations
```

docs/guides/http-client.md
```diff
@@ -2,19 +2,33 @@
...
+    },
+  ],
+  { mock: true, validateRequest: true, validateResponse: true }
+);
 ```
 
-### Create from OpenAPI string
+To create the required operations array you can use two utilities function defined in the `@stoplight/prism-cli` package:
+
+### Create from a filename or http resource
```
philsturgeon:
```suggestion
### Create from file or HTTP resource
```

packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { createLogger } from '@stoplight/prism-core';
-import { getHttpOperationsFromResource } from '@stoplight/prism-http';
+import { getHttpOperationsFromResource } from '@stoplight/prism-cli/src/getHttpOperations';
```
philsturgeon: Are we happy with HTTP Server pulling from the CLI package? I think the HTTP Server is very "HTTP Server for the CLI" so these two are already a bit co-dependant, but this feels a bit like an arrow going the wrong way.

docs/guides/http-client.md
```diff
@@ -2,19 +2,33 @@
...
+);
 ```
 
-### Create from OpenAPI string
+To create the required operations array you can use two utilities function defined in the `@stoplight/prism-cli` package:
+
+### Create from file or HTTP resource
 
 ```ts
+const { default, getHttpOperationsFromResource } = require('@stoplight/prism-cli/dist/getHttpOperations')
```
karol-maciaszek: I thought about two alternatives:
1. `const { getHttpOperations, getHttpOperationsFromResource } = require('@stoplight/prism-cli/dist/operations')`
2. `const { getHttpOperations, getHttpOperationsFromResource } = require('@stoplight/prism-cli')` - why dig into internals at all?

Also, the default export might be misleading. I would name the `getHttpOperations` explicitly.

docs/guides/http-client.md
```diff
@@ -25,27 +39,8 @@ paths:
...
 
-const client = await createClientFromString(descriptionDoc, {
-  mock: true,
-  validateRequest: true,
-  validateResponse: true,
-});
-```
-
-### Create From Manual Http Operations
+const operations = await default(descriptionDoc);
```
karol-maciaszek:
```suggestion
const operations = await getHttpOperations(descriptionDoc);
```

It's a documentation, I propose not to overload people with puzzles.

packages/http-server/src/__tests__/server.oas.spec.ts
karol-maciaszek: It's not that bad. It is only dev dependency.

packages/http/package.json
```diff
@@ -30,6 +25,7 @@
     "faker": "^4.1.0",
     "fp-ts": "^2.1.1",
     "json-schema-faker": "0.5.0-rc23",
+    "node-fetch": "^2.6.0",
```
karol-maciaszek: Good catch 👍

packages/http-server/src/__tests__/server.oas.spec.ts
philsturgeon: Yep but we often talk about splitting prism into multiple packages, and I have suggested we hold off until things are in the right place. This would be moving a thing to the wrong place for that to happen. CLI should require HTTP Server, but HTTP Server should not require CLI. Right? We want to continously boy scout this code until we're in a place where we could split the repos, but this would make that a little harder, even if its only 1%.

XVincentX: You're correct @philsturgeon, effectively this the cli is overflowing in the http package which is should not happen. Keep in mind this is in POC state to get an idea/feedback — and in any case this is only happening in the test.

The solution would be to duplicate the code here and in CLI (I do not think it's worth to make a separate package just for this function)

XVincentX:
Since we're going to ship this as a breaking change I think I will tweak a little bit the function names (the current one might not exactly make that much sense) as well as maybe rearrange the parameters. I'll look into this later today and put something online.

marcelltoth:
> Change our artefacts directory from dist to lib — it makes more sense and does not make you feel bad when importing stuff // cc @marcelltoth — we had such discussion some time ago

Just my two cents here. I think if we are going to need a major bump we should take the time to really define what our public API is. Merely renaming the directory to `lib` and saying it's OK to import from there is disastrous in my opinion. From that point on **changing any export in any file** will technically be a major break. I'm sure `http\dist\mocker\negotiator\internalHelpers` shouldn't be part of a public library distribution. But probably `http-server/getHttpConfigFromRequest` neither.

Let's take a step back, decide what the goal is for each of the packages (in terms of external use), and what are the key APIs we'd like to offer to the public to achieve them.

Then put these files into _somewhere_ (lib folder, index.ts export / top level files / whatever we decide on), keep the rest where it is, and clearly document what is part of our API and what is not.

XVincentX:
@marcelltoth I understand where you're coming from and I would love to do that, but I do not think we can now. The public JS API will probably be done (if done) once we'll gather from more feedback from hosted Prism, Mocking Tab usage and all these product using the package.

On the other hand, we need to make Prism Http browser-compatible as soon as possible since it's blocking some stuff we're doing — and I took the occasion of the breaking change to do propose some other reorganisation — among these the idea of using `lib` directory as output instead of `dist` because it just makes us feel sick. I'm fine deferring this if it's an issue for anybody though.

marcelltoth:
@XVincentX Then I propose to keep it that way and import from `dist` as before.

That pushes the "risk" to the consuming packages if they import from `dist`. Then we sort it out in the next major version, and everyone's happy.

However if we publish a major version where everything is public, how are we going to support that? Do we release a new major every 2 days? Do we break semver? In my eyes either of these are worse than just keeping it in `dist` and surviving with those ugly imports until we have the time to create an actual long term solution.

XVincentX:
We'll keep the `dist` directory as it is at the moment — I am fine deferring the discussion when we'll have the time.

That said, we still need this change for our platform and so apparently this change is unenviable unless we publish this as a 4 beta and use it internally, which might be ok with me and with platform.

XVincentX:
This is not even more important since the bundle of prism-http is big.

packages/http/src/mocker/index.ts
```diff
@@ -20,7 +20,7 @@ import { map } from 'fp-ts/lib/Array';
 import { Do } from 'fp-ts-contrib/lib/Do';
 import { isNumber, isString, keyBy, mapValues, groupBy, get } from 'lodash';
 import { Logger } from 'pino';
-import * as typeIs from 'type-is';
+import { is } from 'type-is';
```
karol-maciaszek, line 23:
```suggestion
import { is: typeIs } from 'type-is';
```

Just a suggestion, `is` may be too generic as it may be accidentally used like that `is(Vincenzo, listeningToGigiDagostino)`.

marcelltoth: I was thinking about making exactly the same suggestion but dropped it 😄

XVincentX: Yes, it's a good idea

packages/http/src/client.ts
```diff
@@ -179,4 +167,4 @@ export type PrismHttp = {
   trace: IRequestFunctionWithMethod;
 };
 
-export { createClientFromResource, createClientFromString, createClientFromOperations };
+export { createClientFromOperations };
```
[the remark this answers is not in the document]
XVincentX: Makes sense, will change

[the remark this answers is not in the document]
XVincentX: Yes, I think this is a good idea.

XVincentX:
@marcelltoth 

I'll address all the comments here:

> Publicly teaching users that it is OK to import stuff from /dist is not great. It'd be OK if we bundled stuff but as we are not, this makes it very difficult to define correct semver versioning.

That is very correct, and it was a mistake, I admit it. We aren't going to release this breaking change soon; most likely we'll be still on 3.x for a while. The idea is to accumulate all the breaking changes (included your very good ideas on defining a public API) and then release. So at the moment, no harms has been done

> importing a package called cli feels weird. A cli feels like it is intended to be an end-user thing that you run, not something you use in other packages, right?

It is a bit weird, but I really do not see any good reason to create a `prism-utils` package. If we can find another part where we can paste this, I'm all for it. Maybe `http-spec`? I have no clue

> Is /src/ in the bundle? Should it?

Eheheh ok this is complicated — essentially is a trick to use TypeScript in VSCode without having to compile the code first. I agree that's not great, but 

- It's in the tests — not shipped code
- We can fix that by exporting stuff from the main package (which unfortunately though has some side effects, so not viable right now)

PR #1019, opened 2020-03-03
fix: do not install all deps just for building

XVincentX:


XVincentX:
Open to alternatives! I think the Dockerfile has been changing a bit and maybe now we can change some of the things — the idea was to speedup the build process by installing only what's required to transpile the TypeScript code; if it causes troubles I am ok leaving it like this.

marcelltoth:
I will investigate on Monday

XVincentX:
Do not lose time for this, we have bigger fishes to fry.

PR #1031, opened 2020-03-10
fix: detect preflight requests

### LINKED ISSUE #1020: CORS Headers Options verb
**User story.**
As a frontend developer, I want to connect to the local mock server, so that I can see mocked responses in my browser.

I can not connect to the mock server via a secure browser, as the [CORS Headers](https://developer.mozilla.org/de/docs/Web/HTTP/CORS) are not set up correctly to allow any origin and any verb (OPTIONS, GET, POST, PATCH, etc.)

The stoplight client mock server is sending out widely set CORS headers to allow any origin and any verb - or it allows me to configure it myself.

**Additional context**
Screenshot 2020-03-03 at 09 55 02

XVincentX:
Closes #1020


Primajin:
Thank you this helps us a lot in our development of data driven UIs 😄

packages/http-server/src/server.ts
```diff
@@ -155,28 +155,38 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+
+          if (!!req.headers['origin'] && !!req.headers['access-control-request-method']) {
+            // This is a preflight request, so we'll respond with the appropriate CORS response
+            res.setHeader(
+              'Access-Control-Allow-Methods',
+              req.headers['access-control-request-method'] || 'GET,DELETE,HEAD,PATCH,POST,PUT,OPTIONS'
+            );
+
+            res.setHeader('Vary', 'origin');
+            res.setHeader('Content-Length', '0');
```
marcelltoth: This is not required:
>    1.  Any response to a HEAD request and any response with a 1xx
>       (Informational), 204 (No Content), or 304 (Not Modified) status
>       code is always terminated by the first empty line after the
>       header fields, regardless of the header fields present in the
>       message, and thus cannot contain a message body.

Source: https://tools.ietf.org/html/rfc7230#section-3.3.3

packages/http-server/src/server.ts
```diff
@@ -155,28 +155,38 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+          setCommonCORSHeaders(req.headers, res);
+
+          if (!!req.headers['origin'] && !!req.headers['access-control-request-method']) {
+            // This is a preflight request, so we'll respond with the appropriate CORS response
+            res.setHeader(
+              'Access-Control-Allow-Methods',
+              req.headers['access-control-request-method'] || 'GET,DELETE,HEAD,PATCH,POST,PUT,OPTIONS'
+            );
+
+            res.setHeader('Vary', 'origin');
```
marcelltoth: Good question is, whether we want this.
On one hand, `prism` sends the same reply regardless of `Origin`, so this is technically a(n innocent) lie.
On the other, I know most servers will actually include a `Vary` header in response to a preflight request, so I can totally understand why we'd want `prism` to do the same.

packages/http-server/src/server.ts
XVincentX: This is required because of Safari. For some reason, even though we're sending a 204, Safari will check the content length anyway and have trouble if there's not.

https://github.com/pallets/werkzeug/issues/1025#issuecomment-390469386

packages/http-server/src/server.ts
XVincentX: This is coming from the previous middleware (fastify-cors) — I'm going to keep this for the time being

packages/http-server/src/server.ts
MonteKarol: Got it, how about commenting this line?

karol-maciaszek: Got it, how about commenting this line?

packages/http-server/src/server.ts
```diff
@@ -155,28 +155,41 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
...
+
+            // This should not be required since we're responding with a 204, which has no content by definition. However
+            // Safari does not really understand that and throws a Network Error. Explicit is better than implicit.
```
karol-maciaszek: 👍

packages/http-server/src/server.ts
marcelltoth: > 
> 
> This is required because of Safari. For some reason, even though we're sending a 204, Safari will check the content length anyway and have trouble if there's not.
> 
> [pallets/werkzeug#1025 (comment)](https://github.com/pallets/werkzeug/issues/1025#issuecomment-390469386)

Doesn't the quoted comment say the exact opposite?
> the **unwanted Content-Length header** currently makes Mac OS Safari throw a network error and **fail, when a 204 is returned**
 (emphasis mine)

XVincentX: I am sorry I pointed you to the wrong resource.

https://github.com/fastify/fastify-cors/blob/master/index.js#L88-L89

And yes I confirm such behaviour in Safari 13.0.5

PR #1035, opened 2020-03-13
feat: add custom error for 2**

XVincentX:
Create a specific error for Prism when it cannot find a suitable 2XX response so we can differentiate with something that's not under our control

### DIFF
--- a/docs/guides/errors.md
+++ b/docs/guides/errors.md
@@ -239,17 +239,7 @@ paths:
     get:
       responses:
         300:
-          content:
-            text/plain:
-              schema:
-                type: object
-                properties:
-                  name:
-                    type: string
-                    example: Clark
-                  surname:
-                    type: string
-                    example: Kent
+          description: 'Desc'
 ```
 
 `curl -X POST http://localhost:4010/ -A 'Accept: text/plain'`
--- a/docs/guides/errors.md
+++ b/docs/guides/errors.md
@@ -226,7 +226,7 @@ paths:
 
 ### NO_SUCCESS_RESPONSE_DEFINED
 
-**Message: No 2** response defined, cannot mock\*\*
+**Message: No response in the range 200-299 defined**
 **Returned Status Code: `500`**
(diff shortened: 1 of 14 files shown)


packages/http-server/src/server.ts
```diff
@@ -83,7 +84,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
         ? TE.right(false)
         : pipe(
             getHttpConfigFromRequest(input),
-            E.map(operationSpecificConfig => Object.assign(opts.config.mock, operationSpecificConfig)),
```
marcelltoth: What does this change do, and why don't I see a corresponding test?

packages/http/src/mocker/errors.ts
```diff
@@ -18,6 +18,12 @@ export const NOT_FOUND: Omit<ProblemJson, 'detail'> = {
   status: 404,
 };
 
+export const NO_SUCCESS_RESPONSE_DEFINED: Omit<ProblemJson, 'detail'> = {
+  type: 'NO_SUCCESS_RESPONSE_DEFINED',
+  title: 'No 2** response defined, cannot mock',
```
marcelltoth: Maybe "No response with code 200-299 defined..."?

While we totally understand the "2** response" version, I'm afraid it might not be as clear for users less experienced in the domain.

XVincentX: @marcelltoth "No response in the range 200-299 defined" ?

docs/guides/errors.md
```diff
@@ -224,6 +224,36 @@ paths:
...
+            text/plain:
+              schema:
+                type: object
+                properties:
+                  name:
+                    type: string
+                    example: Clark
+                  surname:
+                    type: string
+                    example: Kent
```
karol-maciaszek: How about something simpler? We just want to emphasize that there is no 2xx response.

```suggestion
      responses:
        300:
          description: 'Desc'
```

XVincentX: Yes!

packages/http/src/mocker/errors.ts
marcelltoth: I'm cool with that.

PR #1044, opened 2020-03-16
fix: upgrade http-spec

### LINKED ISSUE #1036: Order of processing for securitySchemes
**Describe the bug**
When defining security for an API. it appears that the _order_ of the declarations is important (though it shouldn't be). Which results in unexpected `401`, and at other times getting a `200`.

**To Reproduce**

```
paths:
  /ping:
    description: a test
    summary: ping test
    get:
      description: get test
      summary: pingTest
      tags:
        - test
      operationId: pingTest
      security:
        - bearer: []
        - basic: []
      responses:
        '200':
          description: ping response
          content:
            text/plain:
              schema:
                type: string
                example:
                  ack!
components:
  securitySchemes:
    basic:
      type: http
      scheme: Basic
    bearer:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: http://foo/
```
2. Run 
```
curl --location --request GET 'http://localhost:4010/ping' \
--header 'Authorization: Bearer fasfsadfsdafsfadsfsafsdfs'
```
returns a `200 OK`.

3. Run 
```
curl --location --request GET 'http://localhost:4010/ping' \
--header 'Authorization: Basic MTpk'
```
returns a `401` with `WWW-Authenticate: OAuth2` header

Now reverse the order of the security declarations:
```
      security:
        - basic: []
        - bearer: []
```
and rerun the above `curl` commands. This time, both of the above requests return a `200 OK`

***Related***
In the second case (`basic` first), I noticed this entry in the console, in case that helps triage:
```
[11:17:56 AM] ›     [VALIDATOR] ⚠  warning   Violation: request We currently do not support this type of security scheme: http/Basic
```
**Expected behavior**
* `200 OK` in both cases
* even in the case of failure (e.g. I sent `digest`), I would have expected 2 `WWW-Authenticate` headers, `OAuth2` and `Basic <realm>`

**Screenshots**

 - prism-cli version: 3.3.1

XVincentX:
Needs the latest Http Spec package

Closes #1036

### DIFF
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -49,7 +49,7 @@ export function findDefaultContentType(
 
 const byResponseCode = ord.contramap<number, IHttpOperationResponse>(ordNumber, response => parseInt(response.code));
 
-export function findLowest2xx(httpResponses: NonEmptyArray<IHttpOperationResponse>): Option<IHttpOperationResponse> {
+export function findLowest2xx(httpResponses: IHttpOperationResponse[]): Option<IHttpOperationResponse> {
   const generic2xxResponse = () =>
     pipe(
       findResponseByStatusCode(httpResponses, '2XX'),
@@ -67,7 +67,7 @@ export function findLowest2xx(httpResponses: NonEmptyArray<IHttpOperationRespons
 }
 
 export function findResponseByStatusCode(
-  responses: NonEmptyArray<IHttpOperationResponse>,
+  responses: IHttpOperationResponse[],
   statusCode: string
 ): Option<IHttpOperationResponse> {
   return pipe(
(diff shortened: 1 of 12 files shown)


packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -49,7 +49,7 @@ export function findDefaultContentType(
 
 const byResponseCode = ord.contramap<number, IHttpOperationResponse>(ordNumber, response => parseInt(response.code));
 
-export function findLowest2xx(httpResponses: NonEmptyArray<IHttpOperationResponse>): Option<IHttpOperationResponse> {
+export function findLowest2xx(httpResponses: IHttpOperationResponse[]): Option<IHttpOperationResponse> {
```
karol-maciaszek: `NonEmptyArray`, I'll miss U 💔

XVincentX: Yeah I really thought this was something cool.

PR #1056, opened 2020-03-21
refactor: minor improvements

XVincentX:
Because of this CoronaVirus and raining here in Austin I basically had nothing better to do.

1. Removed different `any` in Prism that we know, are not ok. I've gone through the code base and adjusted the types so that we use the most generic one adding runtime checks where necessary.
2. Update to prettier 2 and latest version of eslint parser
3. Set the lint process on the CI


packages/cli/src/util/paths.ts
```diff
@@ -44,7 +44,15 @@ function generateParamValue(spec: IHttpParam): E.Either<Error, unknown> {
     E.chain(value => {
       switch (spec.style) {
         case HttpParamStyles.DeepObject:
-          return E.right(serializeWithDeepObjectStyle(spec.name, value));
+          return pipe(
+            value,
+            E.fromPredicate(
+              (value: unknown): value is string | Dictionary<unknown, string> =>
+                typeof value === 'string' || typeof value === 'object',
```
karol-maciaszek: 👍

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+<!-- markdown-link-check-disable -->
```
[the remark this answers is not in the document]
XVincentX: It's not really failing, but with all these checks on GitHub after a while it starts responding with 429 (too many requests) making the check fail. I have no other way to make this work.

packages/http/src/validator/deserializers/style/deepObject.ts
```diff
@@ -12,15 +12,13 @@ export class DeepObjectStyleDeserializer implements IHttpQueryParamStyleDeserial
...
-        .replace(/\[/g, '\\[')
-        .replace(/\]/g, '\\]');
+    function constructArray(currentPath: string[], items: unknown): object[] {
+      const path = resolve(currentPath).replace(/\[/g, '\\[').replace(/\]/g, '\\]');
 
       const regexp = new RegExp(`^${path}\\[([0-9]+)\\]`);
 
       const indexes = Object.keys(parameters).reduce((list, k) => {
-        const matches = k.match(regexp);
+        const matches = regexp.exec(k);
```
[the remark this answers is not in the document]
XVincentX: This is suggested by ESLint — and I did the same on other parts of the codebase in the past. I think the reason is exactly the one you described.

XVincentX:
> As a next step, we could think about parsing (io-ts) some input values. 

Yes, it's really on the to do. There are different places where we can technically use io-ts to remove the unknown and return earlier in the process if the payload does not match what we're looking for.

package.json
```diff
@@ -14,8 +14,7 @@
     "lint.fix": "yarn lint --fix",
     "build": "ttsc --build ./packages/tsconfig.build.json",
     "posttest": "yarn lint",
-    "test": "jest --maxWorkers=2",
-    "test.watch": "yarn test --watchAll",
+    "test": "jest",
```
marcelltoth: Why did we lose `test.watch`? Also did you measure that it is faster without `maxWorkers`? What is the rationale?

XVincentX: 1. nobody was using it and you can still use It and do `yarn test --watchAll`
2. On a MacBook Jest runs with 4 workers. This value was previously set to 2 because of CircleCI: because of some virtualisation quirks jest detects 8 workers but in reality there are only two available. I've moved such limit in the CI command as you can see.

marcelltoth: All make sense, cool

PR #1072, opened 2020-03-31
fix: properly declare all the dependencies

XVincentX:
Explicitly declare the dependencies of the packages instead of relying on the module resolution


XVincentX:
Nope, we're actually making sure that if I install only prism-core for some reason I get all its dependencies necessary to run. If now you install it and try to require it you'll get an error because fp-ts (direct dependency) isn't present.

XVincentX:
I used an eslint plugin to detect them so the list was kind of automatically generated.

PR #1074, opened 2020-04-01
fix: do not override mock options

### LINKED ISSUE #1071: Prefer header kept for subsecuent requests
**Describe the bug**
I am following the following documentation to select what example I want when making a request:
https://stoplight.io/p/docs/gh/stoplightio/prism/docs/guides/01-mocking.md#response-examples

When I specify the `Prefer: example=<my example name>;` header it returns the requested example, but prism caches the header and all the subsequent requests without it will try to get that example.

**To Reproduce**
If I have following 2 paths:
```
/my-path
  examples:
    example1: <my example here>
    example2: <my example here>
/my-other-path
  examples:
    another-example-name: <my example here>
```

Making a call to `/my-path` with header `Prefer: example=example1;` it will return the specified example with no issues.

If I make a call to `/my-other-path` **without** the prefer header, it will return this:
```
{"type":"https://stoplight.io/prism/errors#NOT_FOUND","title":"The server cannot find the requested content","status":404,"detail":"Response for contentType: application/json and exampleKey: example1 does not exist."}
```

It's like after making a call with the Prefer header, all the subsequent calls without it, will assume that I want the `example1`.

**Expected behavior**
I am expecting that when I don't specify the `Prefer` header, it returns the first example, as it is specified in the documentation.

This used to work with version 3.2.9

**Environment:**
 - Library version: 3.3.2
 - OS: Alpine Linux v3.11, macOS

**Additional context**
I am using your docker containers and locally in my mac, both fail

### LINKED ISSUE #1073: Can't mock error response after an example is chosen
**Describe the bug**
`Prefer: code=400` doesn't work after a request with `Prefer: example=elon` is sent.
It works if no example has been chosen previously.

**To Reproduce**

1. Given the attached OpenAPI document
2. Run this CLI command: `prism mock openapi.yaml -h 0.0.0.0`
3. Execute the following requests:

`curl -H "Prefer: example=elon" http://localhost:14080/client-api/v2/test`
`curl -H "Prefer: code=400" http://localhost:14080/client-api/v2/test`

**Expected behavior**
I expect a 400 error response in the second request, I get the same response as the first request instead.

**Additional context**
Might be related to https://github.com/stoplightio/prism/issues/1071
[openapi.txt](https://github.com/stoplightio/prism/files/4416397/openapi.txt)

XVincentX:
Closes #1073
Closes #1071

### DIFF
--- a/packages/http-server/src/__tests__/server.oas.spec.ts
+++ b/packages/http-server/src/__tests__/server.oas.spec.ts
@@ -111,7 +111,7 @@ describe('Prefer header overrides', () => {
       });
     });
 
-    describe('and I send a request with Prefer header to false', () => {
+    describe('and I send a request with Prefer header selecting a specific example', () => {
       describe('and then I send a second request with no prefer header', () => {
         let payload: unknown;
         let secondPayload: unknown;
--- a/packages/http-server/src/__tests__/server.oas.spec.ts
+++ b/packages/http-server/src/__tests__/server.oas.spec.ts
@@ -81,7 +81,7 @@ describe('GET /pet?__server', () => {
   }
 });
 
-describe('dynamic flag preservation', () => {
+describe('Prefer header overrides', () => {
   let server: ThenArg<ReturnType<typeof instantiatePrism>>;
 
   beforeAll(async () => {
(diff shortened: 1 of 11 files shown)


packages/http-server/src/__tests__/server.oas.spec.ts
```diff
@@ -110,6 +110,26 @@ describe('dynamic flag preservation', () => {
         it('shuold return two different objects', () => expect(payload).not.toStrictEqual(secondPayload));
       });
     });
+
+    describe('and I send a request with Prefer header to false', () => {
```
marcelltoth, line 114: "Prefer header to false"? What does that mean?

Issue #1075, opened 2020-04-03
Prism proxy does not proxy headers


nhuray:
**Describe the bug**

I'm testing prism in proxy mode to validate API contracts directly from my application. 

My API support the media types `application/json` and `application/vnd.api+json` (JSON:API).

My application is making the following request (I cleaned unnecessary headers):

```
curl 'http://localhost:3000/api/v1/labels?lang=en&ns=default' -H 'Accept: application/vnd.api+json'
```

When I make that request to my API (with prism as proxy), prism returns the following error:

```
[4:43:43 p.m.] ›     [VALIDATOR] ✖  error     Violation: response The received media type "application/json; charset=utf-8" does not match the one specified in the current response: application/vnd.api+json
```

I investigated the headers received by the API when I make requests directly to the API or using Prism: 

- When I make the request directly to the API, the `Accept` header is `application/vnd.api+json` (what I'm passing).

- When I make the request directly through Prism, the `Accept` header is `application/json, text/plain, */*`.

**Expected behavior**

Prism should not alter the requests or responses headers and should just proxy them.

**Environment :**
 - Library version: 3.3.3
 - OS: Mac OS High Sierra Version 10.13.6
 - Browser: Chrome Version 80.0.3987.132

PR #1076, opened 2020-04-03
Forward request to server without altering accept header

nhuray:
Hi everyone,

This PR is to fix the issue #1075. 

The issue is in the `forwarder` because the request created override the `Accept` header by `application/json, text/plain, */*`.

That PR fix that issue preserving the header if it has been set and add to `application/json, text/plain, */*` if wasn't set.

Nicolas


XVincentX:
Thanks

PR #1082, opened 2020-04-06
Keeps raw body

blacktemplar:
When using prism as proxy it may happen that it modifies the json body since it parses it to array and then encodes it again. For example, if the body contained whitespaces between json values they will be removed. This will be a problem if the request contains a digest for the body string.

To avoid this we remember the rawBody string and use it in the forwarded request.


XVincentX:
Hey,

thanks for the PR, but I do not think this is the best way to move forward. @blacktemplar I would suggest you to open a detailed issue with how to reproduce the error, and we'll go from there and try to fix it.

PR #1084, opened 2020-04-12
use transpile only in cli

XVincentX:
Faster iterations when needing to try the CLI

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,8 +39,8 @@
     "url": "https://github.com/stoplightio/prism.git"
   },
   "scripts": {
-    "cli": "node -r ts-node/register/transpile-only src/index.ts",
-    "cli:debug": "node -r ts-node/register/transpile-only --inspect-brk src/index.ts"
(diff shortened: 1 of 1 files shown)


packages/cli/package.json
```diff
@@ -39,8 +39,8 @@
     "url": "https://github.com/stoplightio/prism.git"
   },
   "scripts": {
-    "cli": "node -r ts-node/register -r tsconfig-paths/register src/index.ts",
-    "cli:debug": "node -r ts-node/register -r tsconfig-paths/register --inspect-brk src/index.ts"
+    "cli": "node -r ts-node/register -r ts-node/register/transpile-only src/index.ts",
+    "cli:debug": "node -r ts-node/register -r ts-node/register/transpile-only --inspect-brk src/index.ts"
```
marcelltoth, line 43: Why is that double indirection in there in the first place?

```suggestion
    "cli": "node -r ts-node/register/transpile-only src/index.ts",
    "cli:debug": "node -r ts-node/register/transpile-only --inspect-brk src/index.ts"
```

This is probably even faster.

XVincentX:
On my computer is 2.87 s vs 11.6s — pretty noticeable!

PR #1107, opened 2020-04-25
Weekly fixes for Prism

### LINKED ISSUE #1103: Error output missing on docker logs for >=3.2.4
**Describe the bug**
Parsing the OAS document does not output errors for Docker images tagged 3.2.4 or above.
Last working version found was 3.2.3
**To Reproduce**

1. Give the wrong yaml/json name for the spec
2. docker run --init -P stoplight/prism:3.3.3 mock -h 0.0.0.0 api.oas2.ym**o**l
3. No error outputted

**Expected behavior**
1. Give the wrong yaml/json name for the spec
2. docker run --init -P stoplight/prism:3.2.3 mock -h 0.0.0.0 api.oas2.ym**o**l
3. Error outputted as` ✖  fatal     ENOENT: no such file or directory, open 'api.oas2.ym**o**l'`

 - Library version: [e.g. 3.2.4]
 - OS: [e.g. Windows 10]
 - Docker version: 19.03.6, build 369ce74a3c

**Additional context**
This applies to parsing errors as well.

### LINKED ISSUE #1104: Unable to run docker image as the documentation instructions on MacOsX
**the bug**
docker image for prism just does not start. I followed all the instructions as per the documentation.

**To Reproduce**
I am on Mac OS X Catalina and I am executing the following command 
docker run --init --rm -it -v $(pwd):/tmp -P stoplight/prism:3 mock -h 0.0.0.0 "/tmp/XXX-Service.v1.json"
[10:24:56 AM] › [CLI] …  awaiting  Starting Prism…

and it just exits and the container did not start.
But when I use the Prism CLI, it just works. There's definitely something wrong with prism3 image.

**Expected behavior**
I expected the container to behave the same as CLI .
However, when I use the prism2 image it works.

**Environment**
 - Library version:  3.0.0
 - OS: Mac OS X and Ubuntu 18

### LINKED ISSUE #1106: Can not catch server startup errors
**Describe the bug**
Currently, `createServer` calls `server.listen` but this method always resolves, even if there is a problem such as the port being taken. This means ultimately these exceptions land as unhandled exceptions as there is no way to handle them.

**To Reproduce**

1. Start prism-cli on a certain port.
2. In another terminal start one on the same port.

**Expected behaviour**
I'd expect the promise to reject, which would allow those of us using `createServer` directly to handle the scenario. It would also allow prism-cli to respond gracefully without causing an unhandled exception.

**Screenshots**

 - Library version: 3.3.3
 - OS: OSX Catalina

XVincentX:
Closes #1106
Closes #1103
Closes #1104


packages/cli/src/util/createServer.ts
```diff
@@ -37,27 +39,22 @@ const createMultiProcessPrism: CreatePrism = async options => {
...
   const logStream = new PassThrough();
   const logInstance = createLogger('CLI', cliSpecificLoggerOptions, logStream);
   pipeOutputToSignale(logStream);
 
-  try {
-    return createPrismServerWithLogger(options, logInstance);
-  } catch (e) {
-    logInstance.fatal(e.message);
-  }
+  return createPrismServerWithLogger(options, logInstance).catch(e => logInstance.fatal(e.message));
```
karol-maciaszek, line 57: Not sure whether `logInstance.fatal` returns any value, probably not. But that: https://github.com/stoplightio/prism/blob/master/packages/cli/src/util/runner.ts#L11 relies on this function returning a server instance or undefined. So if it gets a thing it will assume it is a server instance. I would:
```suggestion
  return createPrismServerWithLogger(options, logInstance).catch(e => { logInstance.fatal(e.message); });
```

I would also investigate the option of returning rejected promise. It would remove some magic from the code.

packages/http/src/getHttpOperations.ts
```diff
@@ -27,7 +27,9 @@ const httpAndFileResolver = new Resolver({
 
 export async function getHttpOperationsFromResource(file: string): Promise<IHttpOperation[]> {
   const isRemote = /^https?:\/\//i.test(file);
-  const fileContent = isRemote ? await fetch(file).then(d => d.text()) : fs.readFileSync(file, { encoding: 'utf8' });
+  const fileContent = await (isRemote
```
karol-maciaszek, line 30: Out of curiosity, does this change fix anything?

XVincentX: No, it just uses the fs promise API instead of the sync one

packages/cli/src/util/createServer.ts
XVincentX: Good point. After the log I'll rethrow the exception.

karol-maciaszek: The same should be done here: https://github.com/stoplightio/prism/pull/1107/files#diff-986568f87b3f6dba35678ad3af18cf90R45 ?

Also, I have a bad feeling that this will break something. Somebody intentionally did not used promises :/

XVincentX: @karol-maciaszek Can you point to the specific line? I can't see it

karol-maciaszek: Yes, line 45 same file. `createMultiProcessPrism()` seems to be doing exactly the same.

sgrmgj-quotient:
@XVincentX I am still not able to execute this as per the documentation -
docker run --init --rm -it -v "$(pwd):/tmp" -P stoplight/prism:3 mock -h 0.0.0.0 /tmp/openapi.v1.json
[11:50:23 AM] › [CLI] …  awaiting  Starting Prism…

It just breaks, still not able to use the prism image

XVincentX:
@sgrmgj-quotient We have not released yet a new Prism version. You would need to build the docker image by yourself first:

1. Clone the source code
2. `docker built -t prism .`

sgrmgj-quotient:
okay cool I will do that thanks @XVincentX

PR #1141, opened 2020-04-30
fix: update mock server diagram

XVincentX:
- Update the mock server diagram
- Validate input only when `--errors` is true when proxying


XVincentX:
Now we do.

test-harness/specs/proxy/proxy-invalid-request-no.errors.txt
```diff
@@ -0,0 +1,23 @@
+====test====
+Sending an invalid request through the proxy server without the --errors flag returns a validation error
```
marcelltoth, line 2: Does **not** return a validation error, right?

PR #1159, opened 2020-05-08
Compare content type parameters separately

### LINKED ISSUE #1152: Issue in findBestHttpContentByMediaType due to inconsistent behaviour between `accepts` and `type-is` libraries
**Describe the bug**
Our API provides versioning information in the `Accept` header with this structure `application/vnd.archa.api+json; version=1` but trying to use this structure as the content type seems to result in one of two errors depending on how it's specified:

If we specify a `Content-Type` of `application/vnd.archa.api+json; version=1` we get an error saying `Violation: response The received media type "application/vnd.archa.api+json; version=1" does not match the one specified in the current response: application/vnd.archa.api+json; version=1` (though the request returns).
- This seems to be the result of a behaviour in the "type-is" library where it doesn't deal with the `version` value (https://github.com/jshttp/type-is/issues/14)

Or, if we specify a `Content-Type` of `application/vnd.archa.api+json` we get an error saying `Request terminated with error: https://stoplight.io/prism/errors#NOT_ACCEPTABLE: The server cannot produce a representation for your accept header`.
- This seems to be because the "accepts" library seems to do an exact comparison of the media type

**To Reproduce**

```
openapi: 3.0.0
info:
  title: httpbin
  version: '1.0'
paths:
  /get:
    get:
      parameters:
        in: header
        name: Accept
        required: true
        schema:
          type: string
        example: application/vnd.archa.api+json; version=1
      responses:
        '200':
          description: Created
          content:
            application/vnd.archa.api+json:
              schema:
                required:
                  - value
                properties:
                  value:
                    type: string
                    example: "hello"
```

2. Run this CLI command
`prism mock api.yaml `

3. See error
After sending a Get request with the correct `Accept` header (see below) you receive the `Request terminated with error: https://stoplight.io/prism/errors#NOT_ACCEPTABLE: The server cannot produce a representation for your accept header` error.
```
curl --location --request GET 'http://127.0.0.1:8001/get' \
--header 'Accept: application/vnd.archa.api+json; version=1'
```

**Expected behavior**
Providing an `Accept` header of `application/vnd.archa.api+json; version=1` should find a response with a `Content-Type` of `application/vnd.archa.api+json`.

 - Library version: 3.3.4
 - OS: OSX 10.15.2

------

**Question**
Which is the expected `Content-Type` in this case? (`application/vnd.archa.api+json; version=1` or `application/vnd.archa.api+json`)

------

**Potential Fix 1**
I've done something wrong and change my spec 😛

**Potential Fix 2**
One work-around I've found is to update the `findBestHttpContentByMediaType` method to use the "type-is" library, this way it matches the `application/vnd.archa.api+json; version=1` `Accept` header to the `application/vnd.archa.api+json` `Content-Type`.
```
function findBestHttpContentByMediaType(response, mediaType) {
    const req = {
        headers: {
            ['content-type']: mediaType.join(',')
        }
    };
    return pipeable_1.pipe(response.contents, Array_1.findFirst(content => typeIs.is(req, [content.mediaType])));
}
```

**Potential Fix 3**
An option which seems pretty hacky to me would be to remove anything after a `;` in the media type so we can continue to use the "accepts" library.
```
function findBestHttpContentByMediaType(response, mediaType) {
    const bestType = accepts({
        headers: {
            accept: mediaType.map(m => m.split(";")[0]).join(','),
        },
    }).type(response.contents.map(c => c.mediaType));
    
    return pipeable_1.pipe(response.contents, Array_1.findFirst(content => content.mediaType === bestType));
}
```

XVincentX:
This pull request implements a backup strategy to ignore parameters in content type when is possible. Since they're not standard (apart for the `q` value) we won't assume anything about it and drop them in some cases

Closes #1152

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*       @XVincentX @marcelltoth @karol-maciaszek
+*       @XVincentX @karol-maciaszek
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -1,8 +1,10 @@
 import { IHttpContent, IHttpOperationResponse, IMediaTypeContent } from '@stoplight/types';
 // @ts-ignore
 import * as accepts from 'accepts';
+import * as contentType from 'content-type';
+import * as O from 'fp-ts/lib/Option';
 import { filter, findFirst, head, sort } from 'fp-ts/lib/Array';
-import { NonEmptyArray } from 'fp-ts/lib/NonEmptyArray';
+import { NonEmptyArray, fromArray } from 'fp-ts/lib/NonEmptyArray';
 import { alt, map, Option } from 'fp-ts/lib/Option';
 import { ord, ordNumber } from 'fp-ts/lib/Ord';
 import { pipe } from 'fp-ts/lib/pipeable';
@@ -24,17 +26,26 @@ export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpO
 
 export function findBestHttpContentByMediaType(
   contents: IMediaTypeContent[],
-  mediaType: string[]
+  mediaTypes: string[]
 ): Option<IMediaTypeContent> {
-  const bestType = accepts({
-    headers: {
-      accept: mediaType.join(','),
-    },
-  }).type(contents.map(c => c.mediaType));
+  const bestType: string | false = accepts({ headers: { accept: mediaTypes.join(',') } }).type(
+    contents.map(c => c.mediaType)
+  );
 
   return pipe(
-    contents,
-    findFirst(content => content.mediaType === bestType)
+    bestType,
+    O.fromPredicate((bestType): bestType is string => !!bestType),
+    O.chain(bestType => findFirst<IMediaTypeContent>(content => content.mediaType === bestType)(contents)),
+    O.alt(() =>
+      pipe(
+        mediaTypes
+          .map(mt => contentType.parse(mt))
+          .filter(mt => Object.keys(mt.parameters).length > 0)
+          .map(mt => mt.type),
+        fromArray,
+        O.chain(mediaTypesWithNoParameters => findBestHttpContentByMediaType(contents, mediaTypesWithNoParameters))
+      )
+    )
   );
 }
 
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -23,17 +23,17 @@ export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpO
 }
 
 export function findBestHttpContentByMediaType(
-  response: PickRequired<IHttpOperationResponse, 'contents'>,
+  contents: IMediaTypeContent[],
   mediaType: string[]
 ): Option<IMediaTypeContent> {
   const bestType = accepts({
     headers: {
       accept: mediaType.join(','),
     },
-  }).type(response.contents.map(c => c.mediaType));
+  }).type(contents.map(c => c.mediaType));
 
   return pipe(
-    response.contents,
+    contents,
     findFirst(content => content.mediaType === bestType)
   );
 }
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -4,6 +4,7 @@ import * as accepts from 'accepts';
 import * as contentType from 'content-type';
 import * as O from 'fp-ts/lib/Option';
 import { filter, findFirst, head, sort } from 'fp-ts/lib/Array';
+import { pick } from 'lodash';
 import { NonEmptyArray, fromArray } from 'fp-ts/lib/NonEmptyArray';
 import { alt, map, Option } from 'fp-ts/lib/Option';
 import { ord, ordNumber } from 'fp-ts/lib/Ord';
@@ -37,12 +38,13 @@ export function findBestHttpContentByMediaType(
     O.fromPredicate((bestType): bestType is string => !!bestType),
     O.chain(bestType => findFirst<IMediaTypeContent>(content => content.mediaType === bestType)(contents)),
     O.alt(() =>
-      // Since media type parameters are not standardised (apart from the quality value), we're going to try again ignoring them all.
+      // Since media type parameters are not standardised (apart from the quality value), we're going to try again ignoring them all but q.
       pipe(
         mediaTypes
           .map(mt => contentType.parse(mt))
           .filter(mt => Object.keys(mt.parameters).length > 0)
-          .map(mt => mt.type),
+          .map(({ type, parameters }) => ({ type, parameters: pick(parameters, 'q') }))
+          .map(mt => contentType.format(mt)),
         fromArray,
         O.chain(mediaTypesWithNoParameters => findBestHttpContentByMediaType(contents, mediaTypesWithNoParameters))
       )
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -37,6 +37,7 @@ export function findBestHttpContentByMediaType(
     O.fromPredicate((bestType): bestType is string => !!bestType),
     O.chain(bestType => findFirst<IMediaTypeContent>(content => content.mediaType === bestType)(contents)),
     O.alt(() =>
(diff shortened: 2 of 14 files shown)


packages/http/src/validator/index.ts
```diff
@@ -94,8 +95,20 @@ const findResponseByStatus = (responses: IHttpOperationResponse[], statusCode: n
 
 const validateMediaType = (contents: NonEmptyArray<IMediaTypeContent>, mediaType: string) =>
   pipe(
-    contents,
-    findFirst(c => !!typeIs(mediaType, [c.mediaType])),
+    O.fromNullable(mediaType),
+    O.map(mediaType => contentType.parse(mediaType)),
```
marcelltoth, line 99: `O.map(contentType.parse)`?

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -155,7 +152,7 @@ const helpers = {
         });
       }
 
-      if (mediaTypes) {
+      if (mediaTypes && isNonEmpty(mediaTypes)) {
```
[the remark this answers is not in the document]
XVincentX: Yes, in general the entire block can probably be refactored to be a little bit better. I'll get the tests back to green somehow and see if there are good opportunities to make that look better. Thanks for spotting this!

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -289,19 +291,19 @@ const helpers = {
             pipe(
               previous,
               O.alt(() => {
-                logger.trace(`Unable to find a ${statusCodes[index]} response definition`);
+                logger.debug(`Unable to find a ${statusCodes[index]} response definition`);
```
XVincentX, line 294: trace is not caught by our signals logger, so I moved it back to debug as it should be

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -354,13 +356,9 @@ const helpers = {
                 });
               } else {
                 return pipe(
-                  findEmptyResponse(
-                    response,
-                    response.headers || [],
-                    get(contentWithExamples, 'mediaType', get(responseWithSchema, 'schema')) || ['*/*']
```
XVincentX, line 360: `contentWithExamples` is always undefined in this branch, that's why I've removed it

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*       @XVincentX @marcelltoth @karol-maciaszek
+*       @XVincentX @karol-maciaszek
```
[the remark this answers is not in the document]
marcelltoth: 💔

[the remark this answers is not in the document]
karol-maciaszek: :weary:

packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -23,18 +26,29 @@ export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpO
...
-    findFirst(content => content.mediaType === bestType)
+    bestType,
+    O.fromPredicate((bestType): bestType is string => !!bestType),
+    O.chain(bestType => findFirst<IMediaTypeContent>(content => content.mediaType === bestType)(contents)),
+    O.alt(() =>
+      // Since media type parameters are not standardised (apart from the quality value), we're going to try again ignoring them all but q.
+      pipe(
+        mediaTypes
+          .map(mt => contentType.parse(mt))
+          .filter(mt => Object.keys(mt.parameters).filter(k => k !== 'q').length > 0)
```
karol-maciaszek, line 45:
```suggestion
          .filter(({ parameters }) => Object.keys(parameters).find(k => k !== 'q'))
```
or for more clarity:
```suggestion
          .filter(({ parameters }) => !!Object.keys(parameters).find(k => k !== 'q'))
```

packages/http/src/mocker/negotiator/__tests__/InternalHelpers.unit.ts
```diff
@@ -10,7 +10,7 @@ describe('InternalHelpers', () => {
       };
 
       it('should respect the q parameter', () => {
-        const possibleResponse = findBestHttpContentByMediaType(avaiableResponses, [
+        const possibleResponse = findBestHttpContentByMediaType(avaiableResponses.contents, [
```
karol-maciaszek, line 13:
```suggestion
        const possibleResponse = findBestHttpContentByMediaType(availableResponses.contents, [
```
Will need updates in the code above.

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -740,6 +752,48 @@ describe('NegotiatorHelpers', () => {
...
+    });
+
+    describe('when avaiable content types has the Q and a non standard parameter', () => {
+      it('should return an unparametrised version', () => {
+        assertSome(
+          findBestHttpContentByMediaType([{ mediaType: 'application/json; version=1; q=0.6' }], ['application/json'])
+        );
+      });
+
+      describe('multiple media types avaiable', () => {
```
karol-maciaszek, line 771:
```suggestion
      describe('multiple media types available', () => {
```

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -740,6 +752,48 @@ describe('NegotiatorHelpers', () => {
...
+        it('will still give preference with the q parameter', () => {
+          assertSome(
+            findBestHttpContentByMediaType(
+              [
+                { mediaType: 'application/json; version=1; q=1' },
+                { mediaType: 'application/json; version=1; q=0.6' },
```
karol-maciaszek, line 777:
```suggestion
                { mediaType: 'application/json; version=1; q=0.6' },
                { mediaType: 'application/json; version=1; q=1' },
```

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -283,3 +283,17 @@ describe('HttpValidator', () => {
...
 });
+
+describe('validateMediaType()', () => {
+  describe('when avaiable content type does not have parameters', () => {
+    const content: IMediaTypeContent = {
+      mediaType: 'application/vnd.archa.api+json',
+    };
+
+    describe('and the requeste media type matches, but has parameter', () => {
+      const result = validator.validateMediaType([content], 'application/vnd.archa.api+json; version=1');
```
karol-maciaszek, line 294: I have bad experience in tracking errors outside `it()` functions. Anything against moving to `it()`?

XVincentX:
> You could add a changelog if this is something that is user-visible. 

@marcelltoth I added the changelog line indeed.

packages/http/src/mocker/negotiator/InternalHelpers.ts
XVincentX: Yeah good point. @karol-maciaszek how about this: bbb4c73

packages/http/src/mocker/negotiator/__tests__/InternalHelpers.unit.ts
XVincentX: bb632bf

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
XVincentX: 99de7b7

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
XVincentX: cbdadf3

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -283,3 +283,18 @@ describe('HttpValidator', () => {
     });
   });
 });
+
+describe('validateMediaType()', () => {
+  describe('when avaiable content type does not have parameters', () => {
```
karol-maciaszek, line 288:
```suggestion
  describe('when available content type does not have parameters', () => {
```

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -283,3 +283,18 @@ describe('HttpValidator', () => {
...
+    const content: IMediaTypeContent = {
+      mediaType: 'application/vnd.archa.api+json',
+    };
+
+    describe('and the requeste media type matches, but has parameter', () => {
```
karol-maciaszek, line 293:
```suggestion
    describe('and the request media type matches, but has parameter', () => {
```

Issue #1007, opened 2020-02-26
Prism is not able to interpolate hyphenated parameters


XVincentX:
Considering this OAS3 file:

```yaml
openapi: 3.0.0
info:
  title: httpbin
  version: '1.0'
paths:
  /get:
    get:
      responses:
        '200':
          description: OK
      parameters:
        - schema:
            type: number
          in: query
          name: q-param
          required: true
```

### Expected
Prism should be able to correctly interpolate it and propose a query parameter example in the CLI

### Result

<img width="608" alt="image" src="https://user-images.githubusercontent.com/1416224/75390301-ab046580-58ad-11ea-9625-327e04d9bc7f.png">

Prism is not crashing (which is good) and it's using the fallback behaviour of displaying the URI Template

The problem is most likely caused in the `createQueryUriTemplate` which is ignoring some part of the RFC spec (for example, hyphens are not allowed and so they should be properly encoded.

stephenlprice:
This behavior is also affecting path parameters. If I remove the hyphen from the names of my path parameters I can can generate valid mocked responses. However if path parameter names are still hyphened Prism will present the behavior seen in these images.

![requestMakerError](https://user-images.githubusercontent.com/58522233/78914246-c7b8bf00-7a4f-11ea-9864-f8c130aaa252.png)

![consoleError](https://user-images.githubusercontent.com/58522233/78914260-cd160980-7a4f-11ea-814a-8bb0e2ad834a.png)

XVincentX:
@stephenlprice I need to triple check but I am pretty sure that is a different issue. 

Hyphens are not a valid URI Template character, so the OpenAPI document you have is invalid. Spectral could probably emit a warning/error about it.

stephenlprice:
@XVincentX thanks for the fast response. I just checked Spectral in Studio and CLI for this case and it did not return any related errors.
Here is the example for your reference:

```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "testAPI",
    "version": "1.0"
  },
  "servers": [
    {
      "url": "http://localhost:3000"
    }
  ],
  "paths": {
    "/testPath/{subscriber-id}/{destination-id}/{application-id}/{doc-id}": {
      "get": {
        "summary": "getTest",
        "tags": [],
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "foo": {
                      "type": "string"
                    },
[...]

PR #1189, opened 2020-05-19
fix: handle hyphens in path and query parameters

chohmann:
Issue: #1007 

## Checklist

_Bug Fix_

This PR allows there to be hyphens in path and query parameters.

No


packages/cli/src/util/paths.ts
```diff
@@ -31,9 +31,28 @@ export function createExamplePath(
...
+
+      const cleanedExpandedPath = URI.expand(cleanedTemplate, cleanedValues);
+
+      // replace cleaned values with real values
+      // i.e put "-"" back in
+      let finalExpandedPath = cleanedExpandedPath;
+      for (const realValue in realValues) {
+        const cleanedValue = realValue.replace(/-/g, '');
+        finalExpandedPath = cleanedExpandedPath.replace(cleanedValue, realValue);
+      }
```
chohmann, line 52: @XVincentX this is not a very good way to get the real query/path param names back into the expanded url. For example, if a query param with hyphen removed exists in other parts of the path the wrong part of the string will be replaced.

Example: 
- path of `/path-path`
- query param of `p-a`
- `cleanedExpandedPath` will be `/path-path?pa=test`
- the query param with hyphen removed is `pa`
- you'll end up doing `/path-path?pa=test`.replace('pa', 'p-a')` which will replace the wrong part of the path and not actually replace the query param

Any suggestions for a more targeted, safer replacement of the path/query params in the `cleanedExpandedPath`?

chohmann: I added a more targeted approach in 6fd670e that uses `urijs` package to do a more targeted replace. This fixes the issue but introduces another issue. It auto-encodes so we're now getting `+` instead of `%20`. 

I also added 2 tests in 28cba6a that will fail if we stick with the original unsafe replace approach to illustrate the issue further and test different approaches

chohmann: urijs has a function to not escape `%20` to `+` called `escapeQuerySpace(false)`. This wasn't available to use until I added `"@types/urijs": "^1.19.8"` as a dev dependency. However this devDependency is messing with another packages `uri` namespace, causing the build to fail. Looking into ways to fix this.

chohmann:
@XVincentX @karol-maciaszek this is ready for review! See https://github.com/stoplightio/prism/pull/1189#discussion_r428742146 about build failures.

packages/cli/src/operations.ts
```diff
@@ -3,52 +3,24 @@ import {
...
-  const { result: resolvedContent, errors } = await httpAndFileResolver.resolve(parsedContent, { baseUri });
-
-  if (errors.length) {
-    const uniqueErrors = uniq(errors.map(error => error.message)).join(EOL);
-    throw new Error(
-      `There's been an error while trying to resolve external references in your document: ${uniqueErrors}`
-    );
-  }
+export async function getHttpOperationsFromSpec(specFilePathOrObject: string | object): Promise<IHttpOperation[]> {
+  const result = await dereference(specFilePathOrObject);
```
chohmann, line 18: the `json-schema-ref-parser` package [dereference method](https://apitools.dev/json-schema-ref-parser/docs/ref-parser.html#dereferenceschema-options-callback) expects either a schema object or a file path string. We were passing in the stringified schema object or file contents before. Updated to take either a string --> filePath, or object --> schema object and pass that directly to `dereference` method.

packages/cli/src/util/__tests__/getHttpOperations.spec.ts
```diff
@@ -4,54 +4,44 @@ describe('getHttpOperationsFromSpec()', () => {
...
-          })
-        )
-      ).rejects.toThrow(
-        /^There's been an error while trying to resolve external references in your document: Error: EISDIR: illegal operation on a directory, read$/
-      );
+        getHttpOperationsFromSpec({
+          openapi: '3.0.0',
+          paths: { $ref: 'abc://' },
+        })
+      ).rejects.toThrow('Unable to resolve $ref pointer "abc://"');
```
chohmann, line 11: updated these tests to not stringify the schema object and just pass in the object as the dereference method expects. See comment https://github.com/stoplightio/prism/pull/1189/files#r428930281.

packages/http/src/validator/index.ts
```diff
@@ -143,8 +144,21 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+function getPathParams(path: string, template: string, realParams: IHttpPathParam[]) {
+  // replace "-" in template path and query params
+  const cleanedTemplate = template.replace(/(?<=\{.*)(?=.*\})(-)/g, '');
+
+  const pathParams = new URI.Template(cleanedTemplate).match(path);
+
+  // get pathParams back into real format (i.e put "-" back)
+  const finalPathParams = {};
+  for (const realParam of realParams) {
+    const cleanedName = realParam.name.replace(/-/g, '');
```
XVincentX, line 156:
```suggestion
    const cleanedName = realParam.replace(/-/g, '');
```

packages/http/src/validator/index.ts
```diff
@@ -143,8 +144,21 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+    if (pathParams[cleanedName]) {
+      finalPathParams[realParam.name] = pathParams[cleanedName];
```
XVincentX, line 158:
```suggestion
      finalPathParams[realParam] = pathParams[cleanedName];
```

packages/cli/src/util/paths.ts
```diff
@@ -31,9 +32,35 @@ export function createExamplePath(
...
-    );
+    .return(({ pathData, queryData }) => {
+      // replace "-" in template path and query params
+      const cleanedTemplate = queryData.template.replace(/(?<=\{.*)(?=.*\})(-)/g, '');
+      const realValues = transformValues({ ...pathData.values, ...queryData.values });
+      const cleanedValues = {};
+      for (const realValue in realValues) {
+        const cleanedValue = realValue.replace(/-/g, '');
+        cleanedValues[cleanedValue] = realValues[realValue];
+      }
```
XVincentX, line 43: I can see this is the same logic that is used in the validator file. I suggest to encapsulate this in a function and export it from the package so we have a single point to modify in the future, if required.

packages/cli/src/util/paths.ts
```diff
@@ -31,9 +32,35 @@ export function createExamplePath(
...
+
+      const cleanedExpandedPath = URI.expand(cleanedTemplate, cleanedValues);
+      const uri = new URIJS(cleanedExpandedPath).escapeQuerySpace(false);
```
XVincentX, line 46: Why is the `escapeQuerySpace` required to be set to false?

packages/http/src/validator/index.ts
```diff
@@ -143,8 +144,21 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
+    }
+  }
+  return finalPathParams;
```
XVincentX, line 161: This function is fundamentally ignoring the parameter in case it cannot be found in the "real" list. I think we should communicate we haven't found a parameter to the caller, so I think we need to employ the `Option` monad here, and then the caller will decide what to do?

packages/http/src/validator/index.ts
```diff
@@ -143,8 +144,21 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
...
-  return new URI.Template(template).match(path);
+function getPathParams(path: string, template: string, realParams: IHttpPathParam[]) {
+  // replace "-" in template path and query params
+  const cleanedTemplate = template.replace(/(?<=\{.*)(?=.*\})(-)/g, '');
+
+  const pathParams = new URI.Template(cleanedTemplate).match(path);
+
+  // get pathParams back into real format (i.e put "-" back)
+  const finalPathParams = {};
+  for (const realParam of realParams) {
```
XVincentX, line 155: In general in Prism we do not iterate, we map over objects, so that we do not have side effects or indeterministic functions.

packages/cli/src/util/paths.ts
```diff
@@ -31,9 +32,35 @@ export function createExamplePath(
...
+        uri.path(uri.path().replace(new RegExp(`;${cleanedPathParam}`, 'g'), `;${realPathParam}`));
+      }
+
+      // add real query param names back
+      for (const realQueryParam in queryData.values) {
+        const cleanedQueryParam = realQueryParam.replace(/-/g, '');
+        uri.query(uri.query().replace(new RegExp(cleanedQueryParam, 'g'), realQueryParam));
+      }
+
+      return uri.normalize().valueOf();
```
XVincentX, line 62: Although they do the exact same thing, `toString()` is way more readable.

```suggestion
      return uri.normalize().toString();
```

packages/http/src/validator/__tests__/index.spec.ts
```diff
@@ -135,6 +135,35 @@ describe('HttpValidator', () => {
...
+                id: '1',
+                request: {
+                  path: [
+                    { name: 'a-id', style: HttpParamStyles.Simple },
+                    { name: 'b-id', style: HttpParamStyles.Matrix },
+                  ],
+                },
+                responses: [{ code: '200' }],
+              },
+              element: { method: 'get', url: { path: '/a-path/1/b/;b=2' } },
```
XVincentX, line 156: Is this even valid? Isn't matrix only possible for query strings?

chohmann: Yes Matrix parameters can be in the path. Also, this is basically the same test as the one above it that was already there. I just added a hyphen in the param names. 

BUT, now that I'm looking at this, I did need to update the `element` to use `;b-id=2` instead of `;b=2` to make this a real use case.

Updated in 066041f.

packages/cli/src/util/paths.ts
chohmann: Done. 6dd4f44

packages/cli/src/util/paths.ts
chohmann: Otherwise urijs turns `%20` to `+`. Some of the tests were expecting `%20` instead of `+` in query params, so this kept the output the same as before.

packages/http/src/validator/index.ts
```diff
@@ -143,8 +144,21 @@ const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ resource,
   );
 };
 
-function getPathParams(path: string, template: string) {
-  return new URI.Template(template).match(path);
+function getPathParams(path: string, template: string, realParams: IHttpPathParam[]) {
```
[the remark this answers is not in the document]
chohmann: "real" means as they are in the spec, with the hyphen in it. 

I renamed to specParamNames and typed the function to Dictionary<string>.

779a7f6 and 2a272cf

packages/http/src/validator/index.ts
chohmann: done. 779a7f6

packages/http/src/validator/index.ts
chohmann: done 779a7f6

packages/http/src/validator/index.ts
chohmann: So this was building an object from an array, not iterating through an object.

Anyways, changed to use reduce to build the object from the array, although I think it's waaaaay less readable now. 

Done in 779a7f6.

packages/cli/src/util/paths.ts
chohmann: It is similar but not the same.

The one in Validator file is going from cleaned (no hyphens) --> real (with hyphens).

This one is going from real values (with hyphens) --> cleaned (no hyphens).

But I will update this to not use `for... in` syntax too, since you requested that for the Validator part, and instead build the object from the array using the reduce syntax. Even though I think it's way less readable.

chohmann: yeah we can still abstract it to a method... duh 🤦‍♀️ . I'll do that.

chohmann: haha nvm again. The things we're iterating over to build the object are of different types. I don't think it's worth to try to make one function that works for both types, so I'm not going to do that. If you feel really strongly about doing this, let's discuss.

packages/http/src/validator/index.ts
chohmann: well the real list contains params from the spec itself. So if it's not in there, of course it won't be added. I don't really see an issue with having this being driven from the params that are actually in the spec.

But, if we can't find a "real" spec param in the list of params generated by `uri-template-lite` after cleaning it's name the same way we did to build the list of params with `uri-template-lite` then we should probably throw an error saying something like `Error getting path param {real_param_name} from the path provided`.

So I don't know if we really need to do what you're proposing with the Option monad (but to be honest I don't really know what you mean by Option monad 😬 )

packages/cli/src/operations.ts
```diff
@@ -3,52 +3,24 @@ import {
...
 
-  const transformOperations = detectTransformOperationsFn(parsedContent);
+  const transformOperations = detectTransformOperationsFn(result);
   if (!transformOperations) throw new Error('Unsupported document format');
 
-  return transformOperations(resolvedContent);
+  return transformOperations(result as any);
```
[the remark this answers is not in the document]
chohmann: done. 
34e0b7f

packages/cli/src/util/paths.ts
chohmann: see https://github.com/stoplightio/prism/pull/1189#discussion_r428742146

packages/cli/src/operations.ts
```diff
@@ -3,52 +3,24 @@ import {
...
-});
-
-export async function getHttpOperationsFromResource(file: string): Promise<IHttpOperation[]> {
-  const isRemote = /^https?:\/\//i.test(file);
-  const fileContent = await (isRemote
-    ? fetch(file).then(d => d.text())
-    : fs.promises.readFile(file, { encoding: 'utf8' }));
-
-  return getHttpOperationsFromSpec(fileContent, isRemote ? file : resolve(file));
+export async function getHttpOperationsFromResource(specFilePathOrObject: string | object): Promise<IHttpOperation[]> {
```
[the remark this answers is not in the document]
chohmann: then I'll just leave it as is for now 🙃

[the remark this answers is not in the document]
XVincentX: #1191

packages/http/src/validator/index.ts
XVincentX: I understand it's probably not that readable, but it's now a correct instance of a Monoid, as it should be. 👍 

There's still another step to make this even better but I do not want to overload you with this now :)

packages/cli/src/util/paths.ts
```diff
@@ -31,9 +32,35 @@ export function createExamplePath(
...
+      }
+
+      const cleanedExpandedPath = URI.expand(cleanedTemplate, cleanedValues);
+      const uri = new URIJS(cleanedExpandedPath).escapeQuerySpace(false);
+
+      // add real path param names back
+      // only need to do for matrix style since that's only style names remain in paths
+      for (const realPathParam in pathData.values) {
+        const cleanedPathParam = realPathParam.replace(/-/g, '');
+        // matrix will have ; in front of it
```
XVincentX, line 52: Damn matrix parameters.

PR #1192, opened 2020-05-22
Hyphenless followup

### LINKED ISSUE #1191: Remove the getHttpOperationsFromResource
I would say it's time to remove this function since it's doing the same thing of the other one that lives close by.

This function is documented and it is used in different places, so it might take some effort.

XVincentX:
Update the library and make it great again with dashes

P.S: Damn OpenAPI 3.

Closes #1191


packages/cli/src/util/paths.ts
```diff
@@ -16,11 +16,10 @@ import * as E from 'fp-ts/lib/Either';
 import * as A from 'fp-ts/lib/Array';
 import { pipe } from 'fp-ts/lib/pipeable';
 import { Do } from 'fp-ts-contrib/lib/Do';
-import { get, identity, fromPairs } from 'lodash';
+import { get, identity, fromPairs, mapKeys, invert, flow } from 'lodash';
```
karol-maciaszek, line 19:
```suggestion
import { get, identity, fromPairs } from 'lodash';
```

PR #1255, opened 2020-06-23
Attempt to clarify code + example headers usage

clemblanco:
https://github.com/stoplightio/studio/issues/460


clemblanco:
I don't fully agree. I looked at `mocking.md` but it doesn't mention anything about `code` for example.

In my opinion, this is where I was expecting to find it in the docs when browsing them.

If you want to move this bit into `mocking.md` I feel like there is more to be copied/moved then. I'm happy to do it as long as you give me some guidance around what bits are missing.

Curious to know what @philsturgeon thinks about it.

XVincentX:
Let me revisit the documentation a bit as a whole and we'll find the better place to put this.

---
Hasta pronto,

V.

clemblanco:
Nice one 👍 

I take it you decided it could go over here too.

XVincentX:
Yes, I read the whole documentation and it just made sense where you put it originally.

Great job.

PR #1308, opened 2020-07-24
fix: resolve $ref when generating static response

lottamus:


packages/http/src/mocker/index.ts
```diff
@@ -51,7 +52,9 @@ const mock: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IMockH
   input,
   config,
 }) => {
-  const payloadGenerator: PayloadGenerator = config.dynamic ? generate : generateStatic;
+  const payloadGenerator: PayloadGenerator = config.dynamic
+    ? generate
+    : (schema: JSONSchema) => generateStatic(schema, resource);
```
XVincentX, line 57: This could effectively work, but the idea is that this would only work if the reference is within the same resource. If for some reason it is not, then the process would break.

Does our RefResolver guarantees this?

PR #1357, opened 2020-08-21
Remove lib prefix in fp-ts

XVincentX:
fp-ts 2.8.0 has a mechanism to detect if require the es5 or the es6 version. Let it do the work for us.

This should also reduce the bundle size because bundlers will take the es6 version when possible.


karol-maciaszek (approved):
👍

PR #1358, opened 2020-08-24
Upgrade package mayor versions

XVincentX:
* Upgrade major versions of all our dependencies whenever is possible
* Latest ESLint version has become way more strict about the stuff. Prism is not passing a lot of these checks. I took some time to fix some of these, but that ain't enough yet.


karol-maciaszek (approved):
LGTM 👍

PR #1359, opened 2020-08-24
Upgrade with 3.3.7 changelog

XVincentX:
Add changelog from 3.3.7 — the fix about the compressed resource was really needed in the wild.


karol-maciaszek (approved):
LGTM :+1:

PR #1368, opened 2020-08-25
4.0.0

XVincentX:
4.0 publish


CHANGELOG.md
```diff
@@ -18,6 +20,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 * **BREAKING**: The `getHttpOperationsFromSpec` has been moved from the HTTP Package to the CLI package. If you're using Prism programmatically, this might require some code changes on your side. `getHttpOperationsFromResource` has been removed. [#1009](https://github.com/stoplightio/prism/pull/1009), [#1192](https://github.com/stoplightio/prism/pull/1192)
 * **BREAKING**: The `createClientFromOperations` is now exported as `export function` instead of exporting an object. If you're using Prism programmatically, this might require some code changes on your side [#1009](https://github.com/stoplightio/prism/pull/1009)
+* **BREAKING**: Prism does **NOT** support Node 8 and 10 anymore; the miminal runtime is now 12
```
karol-maciaszek, line 23: What is the reason for dropping support on old node engines?
Is it because of dependencies or it was blocking us from something?

I bet there are many corporate users bound to old versions.

karol-maciaszek (approved):
LGTM, just a question on dropping legacy node support.

CHANGELOG.md
XVincentX: Node 8 and 10 are not supported by the NodeJS organisation either (their LTS is terminated) and they're lagging behind significantly in terms of ES6 and modules support.

For people that can't upgrade we still offer executables as well as Docker Images.

PR #1391, opened 2020-09-07
Fix IHttpConfig 

### LINKED ISSUE #1390: href undefined
Maybe I'm using this wrong, but I'm getting the following error with the following code running in a jest test.

```(bash)
TypeError: Cannot read property 'href' of undefined
.../node_modules/@stoplight/prism-core/dist/factory.js:41:107)
.../node_modules/@stoplight/prism-core/dist/factory.js:58:141
.../node_modules/fp-ts/lib/Either.js:189:75
.../node_modules/fp-ts/lib/Task.js:137:44
```
```(javascript)
describe('GET /users', () => {
  it('should satisfy OpenAPI spec', async () => {
    const operations = await getHttpOperationsFromSpec(
      'reference/users.v1.yml'
    );
    const client = createClientFromOperations(operations, {
      baseUrl: 'https://dev.api.example.com',
      mock: false,
      validateRequest: true,
      validateResponse: true,
      checkSecurity: false,
      errors: true,
    });

    client
      .request('/users?email=user@example.com', {
        method: 'get',
      })
      .then((response) => console.log(response))
      .catch((error) => console.error(error));
  });
});
```

XVincentX:
Closes #1390

The code that was causing the bug is now returning a type error, as it should be:

image

Lesson learned: If you want to design unions without using tags, just don't, it's not gonna work.


karol-maciaszek (approved):
Pretty challenging to understand all those changes. Looks way better 👍

PR #1401, opened 2020-09-07
bug: fix upstream url

michaelmerrill:
URL is not defined in the types file for the upstream property. You may want this to be a string, but by import URL it fixed the error.

I also updated the documentation to include the 'upstream' property.


packages/core/src/types.ts
```diff
@@ -5,6 +5,8 @@ import { ReaderTaskEither } from 'fp-ts/ReaderTaskEither';
 import { TaskEither } from 'fp-ts/TaskEither';
 import { Logger } from 'pino';
 import { NonEmptyArray } from 'fp-ts/NonEmptyArray';
+import { URL } from 'url';
```
XVincentX, line 8: This is not required, are you receiving any error?

XVincentX (changes requested):
Thanks for the documentation changes.

I've put a small comment, I guess I do not understand why it's required to import URL; can you show me the specific error you're receiving?

packages/core/src/types.ts
michaelmerrill: You're right, it's not required. I guess the issue I was running into was that without the URL type in the module typescript seems to think that `upstream` is an `any` type and was not warning me when I passed in a string as a url.  Maybe this is an issue with my tsconfig or vscode?

Screen Shot 2020-09-09 at 4 55 52 PM
Screen Shot 2020-09-09 at 4 57 08 PM

XVincentX: I'll take a look into. Can you remote the import in meantime?

XVincentX: I've done that for you. This is good to merge.

XVincentX (approved):
Good to go, thanks for the change

PR #1425, opened 2020-09-15
Remove fp-ts-contrib

XVincentX:
fp-ts 2.4 has built-in support for `bind` notation — allowing us to remove `fp-ts-contrib` and refactor some code.

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -62,18 +62,18 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
     config: Config,
     validations: IPrismDiagnostic[]
   ): TE.TaskEither<Error, ResourceAndValidation & { output: Output }> => {
+    const mockCall = components.mock({
+      resource,
+      input: { data, validations },
+      config: config.mock,
+    });
+
     const produceOutput = isProxyConfig(config)
       ? components.forward(
           { validations: config.errors ? validations : [], data },
           config.upstream.href
         )(components.logger.child({ name: 'PROXY' }))
-      : TE.fromEither(
-          components.mock({
-            resource,
-            input: { data, validations },
-            config: config.mock,
-          })(components.logger.child({ name: 'NEGOTIATOR' }))
-        );
+      : TE.fromEither(mockCall(components.logger.child({ name: 'NEGOTIATOR' })));
 
     return pipe(
       produceOutput,
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -68,11 +68,11 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
       config: config.mock,
     });
 
+    const forwardCall = (config: IPrismProxyConfig) =>
+      components.forward({ validations: config.errors ? validations : [], data }, config.upstream.href);
+
     const produceOutput = isProxyConfig(config)
-      ? components.forward(
-          { validations: config.errors ? validations : [], data },
-          config.upstream.href
-        )(components.logger.child({ name: 'PROXY' }))
+      ? forwardCall(config)(components.logger.child({ name: 'PROXY' }))
       : TE.fromEither(mockCall(components.logger.child({ name: 'NEGOTIATOR' })));
 
     return pipe(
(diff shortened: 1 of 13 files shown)


packages/core/src/factory.ts
```diff
@@ -62,18 +62,18 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
     validations: IPrismDiagnostic[]
   ): TE.TaskEither<Error, ResourceAndValidation & { output: Output }> => {
+    const mockCall = components.mock({
+      resource,
+      input: { data, validations },
+      config: config.mock,
+    });
+
     const produceOutput = isProxyConfig(config)
       ? components.forward(
```
karol-maciaszek, line 72: If the intention was to increase readability then I'd vote on doing also `const forwardCall = ...`. Or leaving both. as they were.

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -93,13 +85,15 @@ function assembleBody(request?: IHttpOperationRequest): O.Option<{ body: string;
...
-    O.mapNullable(request => request.headers),
+const assembleHeaders = (request?: IHttpOperationRequest, bodyMediaType?: string): O.Option<Dictionary<string>> =>
+  pipe(
+    O.fromNullable(request?.headers),
     O.chain(params =>
       traverseOption(params, param =>
-        doOption.bind('value', generateHttpParam(param)).return(({ value }) => [param.name, value])
+        pipe(
+          generateHttpParam(param),
+          O.map(value => [param.name, value])
```
karol-maciaszek, line 95: Good catch

packages/http/src/validator/index.ts
```diff
@@ -96,21 +96,20 @@ const findResponseByStatus = (responses: IHttpOperationResponse[], statusCode: n
...
-              isMatch(parsedMediaType.parameters, parsedSelectedContentMediaType.parameters)
-            );
-          })
-        )
+    O.fromNullable(mediaType),
+    O.map(contentType.parse),
+    O.chain(parsedMediaType =>
+      pipe(
+        contents,
+        findFirst(c => {
```
karol-maciaszek, line 104: Totally minor, but if we're around then maybe:
```suggestion
        A.findFirst(c => {
```

packages/http/src/utils/runtimeExpression.ts
```diff
@@ -111,16 +108,14 @@ export function resolveRuntimeExpression(
...
-        .bind('body', O.fromNullable(body))
-        .bind(
-          'path',
-          pipe(
-            lookup(2, parts),
-            O.chain(part => O.tryCatch(() => pointerToPath('#' + part)))
-          )
+      O.fromNullable(body),
+      O.bindTo('body'),
+      O.bind('path', _ =>
```
karol-maciaszek, line 113:
```suggestion
      O.bind('path', () =>
```

Not sure here, but I feel that this argument is commonly skipped so there is no reason to emphasize that it exists.

packages/http/src/utils/runtimeExpression.ts
```diff
@@ -73,11 +72,9 @@ export function resolveRuntimeExpression(
 
   function tryRequestQuery() {
     return pipe(
-      doOption
-        .do(isPart(1, 'query'))
-        .bind('query', O.fromNullable(request.url.query))
-        .bind('part', lookup(2, parts))
-        .done(),
+      isPart(1, 'query'),
+      O.bind('query', _ => O.fromNullable(request.url.query)),
```
karol-maciaszek, line 76: Same as below

karol-maciaszek (approved):
Minor comments, one lib less, good stuff! :)

packages/http/src/validator/index.ts
XVincentX: 13f478e

packages/http/src/utils/runtimeExpression.ts
XVincentX: ca7fccd

packages/core/src/factory.ts
XVincentX: 2297a1b

PR #1428, opened 2020-09-17
Fix: types

XVincentX:
* Add types for json schema types
* Remove tests for what TypeScript can ensure us

--- 

Next: remove the `supports` method and make it a static map.


pytlesk4 (approved):
🔥

PR #1429, opened 2020-09-17
passthrough binary content

### LINKED ISSUE #1321: Passthrough Binary Request/Response Bodies
## User story.

As a developer, I can proxy requests/responses with binary bodies, so that I can still proxy all requests through prism-cli proxy without having to write reverse proxy rules to ignore certain routes.

## Is your feature request related to a problem?

I'm using apache as a reverse proxy for our app and I'm required to write additional proxy rules for routes which have requests/responses containing binary body data so that prism doesn't distort the payloads.

## Describe the solution you'd like

For me, prism would ideally ignore body validation when `Content-Type` is set to `application/octet-stream` so that basic endpoint validation can still be confirmed i.e. "does the endpoint we're hitting exist in the openapi.yaml file?" 

## Additional context

I've read through some issues which show that binary bodies are not a priority but it would be helpful if there was a way to transparently passthrough the data so these routes could still be used with the prism proxy.

XVincentX:
Closes #1321


packages/http-server/src/server.ts
```diff
@@ -77,7 +81,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
       body,
     };
 
-    components.logger.info({ input }, 'Request received');
+    components.logger.info({ input: omit(input, 'body') }, 'Request received');
```
XVincentX, line 84: body could be an octet stream now, so we're not going to log it anymore, because we cannot serialise it to a string.

Well we can, but it's not worth it.

packages/http-server/src/server.ts
```diff
@@ -126,7 +130,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
           }
         });
 
-        return TE.fromIOEither(() =>
+        return IOE.fromEither(
```
XVincentX, line 133: In case you didn't notice, see that we have a function less now

pytlesk4: Interesting

pytlesk4 (commented):
Looks good, just have a couple questions.

packages/http-server/src/server.ts
XVincentX: I think this was not really interesting anyway — this was logging the body sent by the user — and the user is aware of it anyway I would say.

XVincentX:
Note to self: missing tests.

packages/http-server/src/server.ts
pytlesk4: Cool, ping me when you add tests and I can look at this again.

XVincentX: We got it

XVincentX:
Closing for lack of interest by the stakeholder.

Issue #1135, opened 2020-04-25
Validation Process rethink, part 3


XVincentX:
It turns out that validation is a very complicated part of Prism and, although [we](https://github.com/stoplightio/prism/pull/664) have [reworked](https://github.com/stoplightio/prism/pull/731) it [several times](https://github.com/stoplightio/prism/pull/862) we still haven't got that right.

With the recent change in the way the proxy works for invalid requests in case the `--errors` flag is provided, we can really likely extract the inputValidation parts from proxy and mock (they are now duplicated, with almost the same code) and make the inputValidation step return a real `Either<NonEmptyArray<IPrismDiagnostic>,InputPart>` and halt the request processing there.

It's a very good opportunity to make the code easier to read and simplify some things in Prism.

The only big problem I found while trying to tame this on my own is that the Core package does not allow for any manipulation of stuff because it's too generic. The benefits of having the core package is starting to be a little bit questionable now and it might make sense to break that and transfer all the code in prism-http directly.

karol-maciaszek:
> Either<NonEmptyArray<IPrismDiagnostic>,InputPart>
That makes it impossible to pass through the erroneous request in proxy mode, right?

> The only big problem I found while trying to tame this on my own is that the Core package does not allow for any manipulation of stuff because it's too generic. The benefits of having the core package is starting to be a little bit questionable now and it might make sense to break that and transfer all the code in prism-http directly.

This is actually a big question. Do we have plans to support other protocols? IMO it would be so cool to do support.

XVincentX:
> That makes it impossible to pass through the erroneous request in proxy mode

Most likely we can still make it happen by returning a `Right` in some special cases, that part does not concern me.

> Do we have plans to support other protocols? IMO it would be so cool to do support.

I think we do, but when's this happening?

* I think the support for another protocol is not happening anytime soon (>= 1year)
* Duplication is cheaper than having the wrong abstraction, and core might be a wrong abstraction.

Let's think about it without committing to anything specific yet and we'll go from there.

karol-maciaszek:
We could take a rough look at other protocols to know what we are dealing with. I have a feeling that we have gone too far with core being compatible with HTTP. That may answer the question of whether there is a good abstraction.

XVincentX:
Definitely more to ponder!

philsturgeon:
[...]

PR #1430, opened 2020-09-17
Break the deserialiser apart

XVincentX:
This PR is the first part of a kind of a big effort for #1135

In particular this is breaking apart the deserialisers to make sure they do not use the classes anymore favouring the functions (which is what we do use in Prism in general) and using types whenever is possible to avoid runtime tests instead.


packages/http/src/validator/index.ts
```diff
@@ -20,20 +20,16 @@ import { validateSecurity } from './validators/security';
...
 import { findOperationResponse } from './utils/spec';
 import { HttpBodyValidator, HttpHeadersValidator, HttpQueryValidator } from './validators';
 import { NonEmptyArray } from 'fp-ts/NonEmptyArray';
 import { HttpPathValidator } from './validators/path';
 
 export const bodyValidator = new HttpBodyValidator('body');
-export const headersValidator = new HttpHeadersValidator(headerDeserializerRegistry, 'header');
-export const queryValidator = new HttpQueryValidator(queryDeserializerRegistry, 'query');
-export const pathValidator = new HttpPathValidator(pathDeserializerRegistry, 'path');
+export const headersValidator = new HttpHeadersValidator(header, 'header');
```
XVincentX, line 30: This is the next guy I'll dismantle.

packages/http/src/validator/validators/__tests__/headers.spec.ts
```diff
@@ -35,24 +34,6 @@ describe('HttpHeadersValidator', () => {
 
       describe('header is present', () => {
         describe('schema is present', () => {
-          describe('deserializer not available', () => {
```
karol-maciaszek, line 38: Why this test is disabled? The behaviour has changed?

packages/http/src/validator/validators/body.ts
```diff
@@ -28,11 +28,11 @@ export function deserializeFormBody(
...
       if (encoding && encoding.style) {
-        const deserializer = body.get(encoding.style);
-        if (deserializer && schema.properties) {
-          const propertySchema = schema.properties[property];
-          deserialized[property] = deserializer.deserialize(property, decodedUriParams, propertySchema as JSONSchema);
-        }
+        const deserializer = body[encoding.style];
+        const propertySchema = schema.properties?.[property];
+
+        if (propertySchema && typeof propertySchema !== 'boolean')
```
karol-maciaszek, line 34: Does this fix the difference between JSONSchema versions? AFAIK one of them can have schema equal to bool false.

karol-maciaszek (commented):
Feels way lighter! Added a couple of comments, most important about named exports and some removed tests.

packages/http/src/validator/deserializers/style/delimited.ts
```diff
@@ -1,40 +1,27 @@
...
     if (Array.isArray(value)) {
       // last query param is taken into account
       value = value[value.length - 1];
     }
 
-    return value ? value.split(this.separator) : '';
+    return value ? value.split(separator) : '';
   }
+
+  return function deserialize(name: string, parameters: IHttpNameValues, schema?: JSONSchema, explode?: boolean) {
```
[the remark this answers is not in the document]
XVincentX: Absolutely. Done in 4e79cd0

packages/http/src/validator/deserializers/style/__tests__/form.spec.ts
```diff
@@ -1,114 +1,92 @@
...
-        describe('query param is an array', () => {
-          it('returns unmodified value', () => {
-            expect(
-              formStyleDeserializer.deserialize('key', { key: ['val1', 'val2'] }, { type: 'array' }, true)
-            ).toEqual(['val1', 'val2']);
-          });
+  describe('schema type is an array', () => {
+    describe('explode is set', () => {
+      describe('query param is an array', () => {
+        it('returns unmodified value', () => {
```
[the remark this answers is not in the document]
XVincentX: a63665e

packages/http/src/validator/deserializers/style/__tests__/matrix.spec.ts
```diff
@@ -1,111 +1,81 @@
-import { HttpParamStyles } from '@stoplight/types';
-
-import { MatrixStyleDeserializer } from '../matrix';
+import matrix from '../matrix';
```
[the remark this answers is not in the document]
XVincentX: 4db588e

packages/http/src/validator/validators/__tests__/headers.spec.ts
XVincentX: We do not have the registry anymore where you can dynamically query for a deserialiser. They are all fixed and the Type System makes sure we're selecting only the one we effectively have.

packages/http/src/validator/validators/body.ts
XVincentX: The type system claims this is the definition of `propertySchema`:


image

with such check, I am excluding it's a boolean so it's at least a JSONSchema (the version is yet to determine)

packages/http/src/validator/deserializers/index.ts
```diff
@@ -1,29 +1,17 @@
...
-export const path = new HttpParamDeserializerRegistry([
-  new SimpleStyleDeserializer(),
-  new LabelStyleDeserializer(),
-  new MatrixStyleDeserializer(),
-]);
+import { deserializeSimple as simple } from './style/simple';
+import { deserializeForm as form } from './style/form';
+import { deserializeDeepObject as deepObject } from './style/deepObject';
+import { deserializeLabel as label } from './style/label';
+import { deserializeMatrixStyle as matrix } from './style/matrix';
```
karol-maciaszek, line 5: Lines 1-4 are not postfixed with "Style" while "Matrix" is, is it intended?

karol-maciaszek (changes requested):
Minor inconsistency in naming

packages/http/src/validator/validators/body.ts
karol-maciaszek: 👍

packages/http/src/validator/deserializers/index.ts
XVincentX: It was not. Fixed in 8e13d24

packages/http/src/validator/validators/__tests__/headers.spec.ts
karol-maciaszek: There is a possibility that deserializer is not found: https://github.com/stoplightio/prism/pull/1430/files#diff-ec5f393de032d67a4f08a12ecde5456cR43

What I mean is that despite the typing system it is still possible that http-spec will sneak some garbage. Our code will deal correctly with the garbage ^ but we don't have a test for such behavior. Ideally, we should just rely on http-spec doing a good job but then comes reality.

XVincentX: Ok I'll add the test back, but I'll still try tore move this — if Http Spec is terrible, we'll fix it there.

karol-maciaszek (approved):
Good stuff!

PR #1431, opened 2020-09-18
Use faker everywhere

XVincentX:
We already have a random data generator library, so we can get rid of chance

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -20,6 +20,7 @@ import { merge } from 'lodash/fp';
 import { pipe } from 'fp-ts/pipeable';
 import * as TE from 'fp-ts/TaskEither';
 import * as E from 'fp-ts/Either';
+import * as IOE from 'fp-ts/IOEither';
 
 function searchParamsToNameValues(searchParams: URLSearchParams): IHttpNameValues {
   const params = {};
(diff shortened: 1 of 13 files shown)


packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -850,7 +847,7 @@ describe('negotiateByPartialOptionsAndHttpContent()', () => {
         dynamic: true,
       };
       const httpContent: IMediaTypeContent = {
-        mediaType: `${chance.string(chanceOptions)}/${chance.string(chanceOptions)}`,
+        mediaType: faker.system.mimeType(),
```
pytlesk4, line 850: Nice

packages/http-server/src/server.ts
```diff
@@ -90,7 +91,7 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
     pipe(
       TE.fromEither(requestConfig),
       TE.chain(requestConfig => prism.request(input, operations, requestConfig)),
-      TE.chain(response => {
+      TE.chainIOEitherK(response => {
```
pytlesk4, line 94: What's the difference between `chain` and `chainIOEitherK`?

pytlesk4 (approved):
Just a small question, but love the use of faker.

packages/http-server/src/server.ts
XVincentX: TE.chain requires the function to return a `TaskEither`; in this part we're internally returning an `Either` instead and so a transformation was required: `return TE.fromIOEither(r)`

fp-ts 2.4 introduced the ability of returning an Either from a TaskEither chain using the dedicated function, so we are not handing the transformation anymore.

PR #1433, opened 2020-09-21
Break the validator apart

XVincentX:
…followup #1430 for #1135 

Better review with no whitespaces: https://github.com/stoplightio/prism/pull/1433/files?w=1

🎉  Note: when this is merged, Prism will be 🥳  _class-free_ 🥳  🎉 (apart from the ProblemJSON error, but that's inevitable as fas as I know)


packages/http/src/validator/index.ts
```diff
@@ -66,12 +59,10 @@ const validateInput: ValidatorFn<IHttpOperation, IHttpRequest> = ({ resource, el
...
-          request.path
-            ? pathValidator.validate(getPathParams(element.url.path, resource.path), request.path)
-            : E.right(undefined)
+          request.body ? tryValidateBody(request.body, body, mediaType) : E.right(undefined),
+          request.headers ? validateHeaders(element.headers || {}, request.headers) : E.right(undefined),
+          request.query ? validateQuery(element.url.query || {}, request.query) : E.right(undefined),
+          request.path ? validatePath(getPathParams(element.url.path, resource.path), request.path) : E.right(undefined)
         )
     ),
     E.map(() => element)
```
karol-maciaszek, line 68: Looks like a candidate to `.bindTo`. Ofc that's totally irrelevant for this PR.

packages/http/src/validator/index.ts
```diff
@@ -141,4 +132,4 @@ function getPathParams(path: string, template: string): Dictionary<string> {
   return new URI.Template(template).match(path);
 }
 
-export { validateInput, validateOutput, validateSecurity };
+export { validateInput, validateOutput, validateSecurity, validateMediaType };
```
karol-maciaszek, line 135: Do we need that line? We usually export stuff directly in place where it is defined.

packages/http/src/validator/validators/index.ts
```diff
@@ -1,4 +1,8 @@
...
-export { HttpQueryValidator } from './query';
-export { IHttpValidator } from './types';
+import { validate as validateBody } from './body';
+import { validate as validateHeaders } from './headers';
+import { validate as validateQuery } from './query';
+import { validate as validatePath } from './path';
+
+const obj = { validateQuery, validatePath, validateHeaders, validateBody };
+
+export = obj;
```
karol-maciaszek, line 8:
```suggestion
export { validate as validateBody } from './body';
export { validate as validateHeaders } from './headers';
export { validate as validateQuery } from './query';
export { validate as validatePath } from './path';
```

^ if I'm not mistaken about the syntax..

packages/http/src/validator/validators/security/__tests__/security.spec.ts
```diff
@@ -9,7 +9,7 @@ const baseRequest: IHttpRequest = {
 };
 
 describe('validateSecurity', () => {
-  const token = new Buffer('test:test').toString('base64');
+  const token = Buffer.from('test:test').toString('base64');
```
karol-maciaszek, line 12: Perfectionist.

karol-maciaszek (changes requested):
Minor comment to address and we're good to go 👍

packages/http/src/validator/validators/index.ts
XVincentX: I tried that as well but that breaks jest spy because direct exports do not have the setter defined (and that's required to spy the function)

image

I was looking for a way to remove these tests and rewrite them, but it was too much to do at this point

packages/http/src/validator/index.ts
XVincentX: We do not. Removed in 2c6607d

packages/http/src/validator/validators/security/__tests__/security.spec.ts
XVincentX: I was receiving a deprecation alert

packages/http/src/validator/index.ts
XVincentX: Not exactly, bindTo and bind work on the current object returned from the previous chains; here we're mapping to return the original untouched element.

karol-maciaszek: Ah, I need to learn a bit about this topic.

karol-maciaszek (approved):
LGTM 👍

Issue #1190, opened 2020-05-20
Multiple Set-Cookie headers get aggregated


Michael-Dawkins:
**Describe the bug**
HTTP response that set multiple cookies that pass through prism proxy see their set-cookie headers merged, breaking the client understanding of them.

**To Reproduce**

1. Make a web server that sets 2 or more cookies, (Nodejs and Java servlet both set them as separate Set-Cookie headers in the same HTTP response)
2. fetch the server through prism proxy on a web page
3. The HTTP response obtained by the browser (or POSTMAN shows it aswell) contains only one Set-Cookie, with the two cookies aggregated and separated by a space
Both Chrome and Firefox only aknoledge the first cookie and ignore the second, which, in my case, breaks authentication.

**Expected behavior**
I expect the multiple Set-Cookie headers to be preserved.

**Environment (remove any that are not applicable):**
 - Library version: 3.3.4
 - OS: Windows and MacOS
 - Browser: Chrome 83 and Firefox

philsturgeon:
Could I ask, what API authentication requires cookies?

Michael-Dawkins:
It is our own APIs, called by the frontend. Just curious, is it a surprising use case?
By using cookies with "HTTP ONLY", we can make sure the frontend has no read access to the tokens, to protect against XSS attacks.
But I am open to suggestions.

philsturgeon:
It's not hugely common, but it is done. 

https://developer.atlassian.com/cloud/jira/platform/jira-rest-api-cookie-based-authentication/

We should not be combining cookies regardless, prism shouldn't mess with headers for no reason.

@Michael-Dawkins could you give us an example OpenAPI and curl request so that we can clearly see exactly what is happening? Reproducible bugs are solved a lot quicker. :)

Lidbetter:
Hi, I am also running into this problem, here are the 2 sets of raw headers returned (according to firefox) from my backend directly and through prism proxy (docker stoplight/prism:4.1.0)

Direct (works as expected):
```
HTTP/1.1 200 OK
Server: nginx/1.17.0
Content-Type: application/json;charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Set-Cookie: app2=432dfe8a3addeb0cbbdca5d65437aeb9; path=/; domain=.app2.net; HttpOnly; SameSite=Lax
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
[...]

PR #1489, opened 2020-10-20
combine multiple headers using comma

Lidbetter:
This addresses issue #1190 (Multiple Set-Cookie headers get aggregated)

https://tools.ietf.org/html/rfc2616#section-4.2
> It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma.

I searched newer RFCs and as far as I can tell the above still holds true

One thing to note, is this:
https://tools.ietf.org/html/rfc6265#section-3
> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.

With that in mind (and some individual header spesific things outlined in other RFCs) it may be better to rework how these headers flow through the application so that multiple headers do not need to be serialized.

For now joining with a comma fixes the `set-cookie` issue and seems to be _more correct_ than joining with a space


packages/http/src/utils/parseResponse.ts
```diff
@@ -19,7 +19,7 @@ export const parseResponseBody = (
   );
 
 export const parseResponseHeaders = (headers: Dictionary<string[]>): Dictionary<string> =>
-  mapValues(headers, hValue => hValue.join(' '));
+  mapValues(headers, hValue => hValue.join(','));
```
XVincentX, line 22: This is probably the good change — but I'd say we only need to apply this on the `Set-Cookie` header instead of doing it for all of them. `mapValues` is likely offering an overload to also get the key and react accordingly

packages/http/src/utils/__tests__/parseResponse.spec.ts
```diff
@@ -78,7 +78,7 @@ describe('parseResponseHeaders()', () => {
     expect(parseResponseHeaders({ h1: ['a b'], h2: ['c'], h3: ['a', 'b'] })).toEqual({
       h1: 'a b',
       h2: 'c',
-      h3: 'a b',
+      h3: 'a,b',
```
XVincentX, line 81: We'll also have to write a more specific test to make sure it only happens only under specific conditions.

XVincentX (changes requested):
Thanks a lot for the PR. I've put a couple of comments but we're definitely on the way to fix this!

packages/http/src/utils/parseResponse.ts
Lidbetter: The reason I didn't initially only special case `set-cookie` was because of the wording in rfc2616, as far as I can tell joining with a comma is right thing to do by default - a space would be the exception to the general rule.

I could be wrong, do you have a link to more/newer info about the correct way to multiple header fields into one?

XVincentX: I need to take a deeper look into; the reason is that it seems like such rule applies only for "state" headers that is Cookies, Caching and some other that now I do not remember.

You can leave it as it is. I'll check it out later or tomorrow and then get back to you.

XVincentX (approved):
Ok I've reviewed the entire thing now. While it won't fix a probable issue with `Set-Cookie` with commas, it should still do the job for now. Thanks

PR #1529, opened 2020-11-18
Standardized Docker commands

hpatoio:
From the Docker documentation https://docs.docker.com/engine/reference/commandline/run/

> - P : Publish all exposed ports to random port

much better to specify the port to expose using `-p` as in the last example.


XVincentX (approved):
Thanks!

PR #1531, opened 2020-11-19
Remove 2xx limitation

### LINKED ISSUE #1520: Allow absence of success reponses
## User story.

As a tester of an API (client or server), I may need to validate that a certain endpoint always returns an error code, so that I can make sure the endpoint is never used. I would then need the mock to respond with an error code.

## Is your feature request related to a problem?

This is a relevant use case for standardized APIs, where the designers want to prevent that the implementer of the API introduces a custom usage of endpoint within API.

## Describe the solution you'd like

Instead of returning the server error (as documented in [link](https://meta.stoplight.io/docs/prism/docs/guides/errors.md#no_success_response_defined)), the mock would return the 4xx or 5xx as documented in the contract (e.g. OpenAPI).

## Additional context

An example of API endpoints always returning error codes is available at [link](https://forge.etsi.org/swagger/editor?url=https://forge.etsi.org/jenkins/job/NFV%20-%20Network%20Functions%20Virtualisation/job/sol005-master/23/artifact/build/SOL005-NSLifecycleManagement-API.yaml/*view*/).

XVincentX:
This PR refactors the code a bit and remove the limitation of only looking for 2xx responses; they will still be privileged, but if nothing's found we'll take the first response we find.

I'm afraid that this, technically, is a breaking change. Shall we call this Prism 5? @philsturgeon 

Closes #1520


philsturgeon:
I don’t think it’s breaking in any way that matters to a user. It would blow up in an opinionated way and return an error, now it’ll return whatever is defined which is also probably an error (because it couldn’t find a successy response) or it’ll return an error (because no responses are defined) so it seems like it’s not breaking.

XVincentX:
Very good, in that case I'll mark this as a regular bug fix. @karol-maciaszek this is pending your review now.

chohmann (approved):
This LGTM! May want another pair of eyes on it as I'm not the most experienced with Prism.

packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -77,13 +70,17 @@ export function findLowest2xx(httpResponses: IHttpOperationResponse[]): Option<I
     head
   );
 
-  return pipe(first2xxResponse, alt(generic2xxResponse));
+  return pipe(first2xxResponse, O.alt(generic2xxResponse));
+}
+
+export function findFirstResponse(httpResponses: IHttpOperationResponse[]): O.Option<IHttpOperationResponse> {
+  return pipe(httpResponses, head);
```
karol-maciaszek:
```suggestion
  return pipe(httpResponses, A.head);
```

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -429,22 +429,20 @@ describe('NegotiatorHelpers', () => {
...
+    it('given no 2xx response should return the first response', () => {
       const desiredOptions = { dynamic: false };
       jest.spyOn(helpers, 'negotiateOptionsBySpecificResponse');
 
-      const negotiationResult = helpers.negotiateOptionsForDefaultCode(httpOperation, desiredOptions)(logger);
-      assertLeft(negotiationResult, e => {
-        expect(e.name).toBe('https://stoplight.io/prism/errors#NO_SUCCESS_RESPONSE_DEFINED');
-      });
+      const negotiationResult = helpers.negotiateOptionsForUnspecifiedCode(httpOperation, desiredOptions)(logger);
+      assertRight(negotiationResult, res => expect(res.code).toBe('300'));
```
karol-maciaszek, line 445: This emphasizes how severe this change is. Are we still sure that it is not a major change?
// @philsturgeon

karol-maciaszek (approved):
LGTM 👍

PR #1537, opened 2020-11-20
Use new Decoders

XVincentX:
Use the new io-ts decoding system and remove the types, since we can use the `parse` method.


chohmann:
@XVincentX what issue is this for? What the problem we're solving here?

XVincentX:
There is not a specific issue. V1 decoders in io-ts are deprecated. I updated them last week in the platform and since I was solving another issue here in Prism I decided to move them to the latest version as well

chohmann (approved):

PR #1538, opened 2020-11-21
Remove conditions and use fp-ts

XVincentX:
Small refactor and changes that will make the code base better hopefully

### DIFF
--- a/packages/http/src/mocker/negotiator/InternalHelpers.ts
+++ b/packages/http/src/mocker/negotiator/InternalHelpers.ts
@@ -3,21 +3,21 @@ import { IHttpContent, IHttpOperationResponse, IMediaTypeContent } from '@stopli
 import * as accepts from 'accepts';
 import * as contentType from 'content-type';
 import * as O from 'fp-ts/Option';
-import { filter, findFirst, head, sort } from 'fp-ts/Array';
+import * as A from 'fp-ts/Array';
 import { pick } from 'lodash';
-import { NonEmptyArray, fromArray } from 'fp-ts/NonEmptyArray';
+import * as NEA from 'fp-ts/NonEmptyArray';
 import { ord, ordNumber } from 'fp-ts/Ord';
 import { pipe } from 'fp-ts/pipeable';
 import { ContentExample } from '../../';
 
-export type IWithExampleMediaContent = IMediaTypeContent & { examples: NonEmptyArray<ContentExample> };
+export type IWithExampleMediaContent = IMediaTypeContent & { examples: NEA.NonEmptyArray<ContentExample> };
 
 export function findFirstExample(httpContent: IHttpContent) {
-  return pipe(O.fromNullable(httpContent.examples), O.chain(fromArray), O.chain(head));
+  return pipe(O.fromNullable(httpContent.examples), O.chain(NEA.fromArray), O.chain(A.head));
 }
 
 export function findExampleByKey(httpContent: IHttpContent, exampleKey: string) {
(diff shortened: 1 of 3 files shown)


packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -6,29 +6,24 @@ import * as O from 'fp-ts/Option';
...
 import { pipe } from 'fp-ts/pipeable';
-import { ContentExample, PickRequired } from '../../';
+import { ContentExample } from '../../';
 
 export type IWithExampleMediaContent = IMediaTypeContent & { examples: NonEmptyArray<ContentExample> };
 
-export function findBestExample(httpContent: IHttpContent) {
-  return httpContent.examples && httpContent.examples[0];
+export function findFirstExample(httpContent: IHttpContent) {
+  return pipe(O.fromNullable(httpContent.examples), O.chain(fromArray), O.chain(head));
```
karol-maciaszek, line 16:
```suggestion
  return pipe(O.fromNullable(httpContent.examples), O.chain(A.fromArray), O.chain(A.head));
```

packages/http/src/mocker/negotiator/InternalHelpers.ts
```diff
@@ -6,29 +6,24 @@ import * as O from 'fp-ts/Option';
...
 }
 
 export function findExampleByKey(httpContent: IHttpContent, exampleKey: string) {
-  return httpContent.examples && httpContent.examples.find(example => example.key === exampleKey);
-}
-
-export function hasContents(v: IHttpOperationResponse): v is PickRequired<IHttpOperationResponse, 'contents'> {
-  return !!v.contents;
+  return pipe(O.fromNullable(httpContent.examples), O.chain(findFirst(example => example.key === exampleKey)));
```
karol-maciaszek, line 20:
```suggestion
  return pipe(O.fromNullable(httpContent.examples), O.chain(A.findFirst(({ key }) => key === exampleKey)));
```

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -33,63 +32,48 @@ const createEmptyResponse = (code: string, headers: IHttpHeaderParam[], mediaTyp
...
+          findFirstExample(httpContent),
+          O.map(bodyExample => ({ code, mediaType, bodyExample })),
+          O.alt(() =>
+            pipe(
+              O.fromNullable(httpContent.schema),
+              O.map<JSONSchema, BodyNegotiationResult>(schema => ({ schema, code, mediaType }))
+            )
+          ),
+          O.getOrElse<BodyNegotiationResult>(() => ({ code, mediaType }))
+        )
```
karol-maciaszek, line 75: Is it possible to put here:
`O.fold(() => E.right({ code, mediaType }), E.right)`

and remove

`return E.right(` ?

karol-maciaszek (changes requested):
It feels so good to see fp-ts code after a long break.

Added minor suggestions, good chunk of improvements 👍

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: It can, but it would require additional explicit typings because of the sad shape of our data:

image

karol-maciaszek: Uh

karol-maciaszek (approved):
LGTM 👍

PR #1542, opened 2020-11-23
fix: don’t allow strings in code property

chohmann:
Issue: https://github.com/stoplightio/platform-internal/issues/4687

Changes type of `code` in `IHttpOperationConfig` to number. This is to disallow `default` as the requested code.

**Screenshot**
image


packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -6,11 +6,15 @@ import * as D from 'io-ts/lib/Decoder';
 import * as parsePreferHeader from 'parse-prefer-header';
 
 const BooleanFromString = D.parse<string, boolean>(s =>
-  s === 'true' ? D.success(true) : s === 'false' ? D.success(false) : D.failure(s, 'Not a boolean')
+  s === 'true' ? D.success(true) : s === 'false' ? D.success(false) : D.failure(s, 'a boolean')
```
chohmann, line 9: The error message when this failed was saying `should be Not a boolean`, so updated so that the error message would read `should be a boolean`.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -206,16 +206,16 @@ const helpers = {
...
     desiredOptions: NegotiationOptions,
-    code: string
+    code: number
   ): RE.ReaderEither<Logger, Error, IHttpNegotiationResult> {
     // find response by provided status code
     return pipe(
       withLogger(logger =>
         pipe(
-          findResponseByStatusCode(httpOperation.responses, code),
+          findResponseByStatusCode(httpOperation.responses, code.toString()),
```
XVincentX, line 215: I think these functions should still take a number; the conversion should be done internally.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
```diff
@@ -206,16 +206,16 @@ const helpers = {
...
           O.alt(() => {
             logger.info(`Unable to find a ${code} response definition`);
-            return createResponseFromDefault(httpOperation.responses, code);
+            return createResponseFromDefault(httpOperation.responses, code.toString());
```
XVincentX, line 218: ditto

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
```diff
@@ -302,15 +302,15 @@ describe('NegotiatorHelpers', () => {
...
       const fakeResponse = {
-        code,
+        code: code.toString(),
         contents: [],
         headers: [],
       };
       const desiredOptions = { dynamic: false };
       const fakeOperationConfig: IHttpNegotiationResult = {
-        code,
+        code: code.toString(),
```
XVincentX, line 313: If we're doing this change, I guess it would be a good idea to also make sure the returned result is a number.

XVincentX (changes requested):
Do not forget to add a change log line

XVincentX:
Can you also change the message of the error? It's saying "Invalid request body payload" but in reality it's a generic invalid request, nothing to do with the body itself.

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -6,11 +6,15 @@ import * as D from 'io-ts/lib/Decoder';
...
+);
+
+const IntegerFromString = D.parse<string, number>(s =>
+  !isNaN(parseInt(s)) ? D.success(parseInt(s)) : D.failure(s, 'a number')
```
rainum, line 13:
```suggestion
  !isNaN(parseInt(s, 10)) ? D.success(parseInt(s, 10)) : D.failure(s, 'a number')
```

rainum (commented):
Added some comments. I agree with @XVincentX, would be nice to have tests to make sure the returned result is a number.

packages/http-server/src/getHttpConfigFromRequest.ts
XVincentX: Do not parse the integer twice. Save the result in a variable and reuse it both for the test and for the returned value.

chohmann: Done. c00bfcb

chohmann:
> Do not forget to add a change log line

@XVincentX what is the process/format for adding a changelog line? Do I put something under Unreleased?

XVincentX:
> > Do not forget to add a change log line
> 
> 
> 
> @XVincentX what is the process/format for adding a changelog line? Do I put something under Unreleased?

Correct. In this case, let's use the "Changed" section.

Then when releasing, I usually put the release number with the date.

chohmann:
@XVincentX changelog added. 1758de3

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
chohmann: @XVincentX These functions took a string before, so this would be changing them to take a number, and do the conversion internally. But that can be done.

If I do that I have to change [here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L272), [here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L277), and [here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/NegotiatorHelpers.ts#L269) to convert the `first` and `current` variables to a number. Or trace all the way back to where those are passed in and change those to numbers along the way. Worth it?

XVincentX: Generally speaking, I want Prism to have a consistent interface, no matter the cost.

I'm obviously willing to sacrifice it in case it'd be such a pain, but if so it's usually because of an underlying design issue that should be fixed somehow.

Let me take a look into the code and I'll get back to you.

chohmann: ok cool. I'll look into all the things we'd need to change too.

chohmann: [this](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/negotiator/InternalHelpers.ts#L62) is gonna keep us from changing the code param to a number for the `findResponseByStatusCode`.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
chohmann: Resolving this convo as the outcome of the convo [here](https://github.com/stoplightio/prism/pull/1542#discussion_r529062453) will apply here too.

packages/http/src/mocker/negotiator/NegotiatorHelpers.ts
XVincentX: I figured, and that function looks wrong. Check https://github.com/stoplightio/prism/pull/1544 out.

chohmann: done. ea2ae13

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
chohmann: I'm not sure what you mean. All the functions called in this test and the function `negotiateOptionsBySepcificCode` return an object where code is a string not a number.

Can you elaborate on what you mean?

chohmann: or do you just mean that the code is parseInt-able?

chohmann:
> Can you also change the message of the error? It's saying "Invalid request body payload" but in reality it's a generic invalid request, nothing to do with the body itself.

@XVincentX do you mean for this error entirely, so change it [here](https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/errors.ts#L5)? Or just for our use of this error in `getHttpConfigFromRequest` so change it https://github.com/stoplightio/prism/blob/master/packages/http-server/src/getHttpConfigFromRequest.ts#L31?

XVincentX:
@chohmann Yes, I mean to change this: https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/errors.ts#L5

packages/http/src/mocker/negotiator/__tests__/NegotiatorHelpers.spec.ts
XVincentX: I have checked this out and unfortunately we'll have to leave this as is at the moment, because of the @stoplight/types package.

chohmann: womp womp 😞

chohmann:
@XVincentX ready for re-review!

examples/Default-Reproduction.v1.yaml
```diff
@@ -0,0 +1,663 @@
...
+        default:
+          description: Error
+          content:
+            application/json:
+              schema: *ref_0
+      operationId: post-api-devices
+      description: >-
+        <span style="font-size: 1.5em; color: #5D535E;">
+          <i class="fas fa-robot fa-xs"></i> newDevice
+        </span>
```
XVincentX, line 374: @chohmann Leftover file?

chohmann: yep! Accidentally hit stage all in Tower. thanks. Removed cf79216

XVincentX (approved):

PR #1543, opened 2020-11-24
Reduce function requirements and ProblemJson function name

XVincentX:
This PR reduces the arguments the HTTP Request parameter extraction requires to do its job as well as renaming the function in ProblemJson to be more communicative.

This is feedback I gathered while doing some upgrades of Prism in Stoplight's platform.

P.S: This PR has been created exclusively from an iPad Air 4 using GitHub CodeSpaces and their native iOS app. :trollface:

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -140,13 +140,13 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
         );
       }),
       TE.mapLeft((e: Error & { status?: number; additional?: { headers?: Dictionary<string> } }) => {
-        if (!reply.finished) {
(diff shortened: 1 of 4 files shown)


packages/http-server/src/server.ts
```diff
@@ -140,13 +140,13 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
         );
       }),
       TE.mapLeft((e: Error & { status?: number; additional?: { headers?: Dictionary<string> } }) => {
-        if (!reply.finished) {
+        if (!reply.writableEnded) {
```
karol-maciaszek, line 143: 👍

karol-maciaszek (approved):
All good 👍

PR #1547, opened 2020-11-25
Remove Map Statement

XVincentX:
There is no reason to map, we can reconstruct the context in the required function directly.


karol-maciaszek (approved):
👍

PR #1550, opened 2020-11-26
fix(http): dynamic generation don't follow constraints

### LINKED ISSUE #1549: min/maxItems not being respected in dynamic responses
## Describe the bug

The `minItems` / `maxItems` attributes for an array are not being respected in dynamically-generated responses.

## To Reproduce

Using this API `test.yaml`:

```yaml
openapi: 3.0.0
info:
  title: Orange v1.0
  version: "1.0"
  description: Orange you glad I didn't say banana! Hah!
  contact:
    name: Stoplight
    url: stoplight.io
    email: support@stoplight.io
servers:
  - url: "http://localhost:3001"
    description: Development
paths:
  /stuff:
    get:
      summary: Retrieve API stuff
      tags: []
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                type: array
                maxItems: 5
                minItems: 0
                items:
                  type: string
                  minLength: 1
                  maxLength: 20
                  x-faker: lorem.sentences
              examples: {}
      description: |-
        Stuff
      parameters: []
      operationId: stuff
    parameters: []
components:
  schemas: {}
```

Then run mock server:

```sh
prism mock test.yml
```

And generate dynamic response:

```sh
curl -sL -H "prefer: dynamic=true" localhost:4010/stuff | jq .
```

## Expected behavior

Per the spec, I should expect to see an array with:

- 0 - 5 elements
- each element having a max of 20 characters

## Actual Behavior

20 entries are included in the array:

```
$ curl -sL -H "prefer: dynamic=true" localhost:4010/stuff | jq .
[
  "Architecto facere la",
  "Alias neque dolores ",
  "Rem dolor ipsum aute",
  "Non excepturi qui es",
  "Rerum aut quibusdam.",
  "Ut omnis laborum. As",
  "Aut dolore officia i",
  "Veritatis nemo rerum",
  "Eum quia quis. Dolor",
  "Velit voluptas facer",
  "Quam voluptate numqu",
  "Repudiandae est nemo",
  "Est non deserunt. La",
  "Perspiciatis in inve",
  "Eligendi neque omnis",
  "Voluptas maxime plac",
  "Ipsum dolor sunt dis",
  "Deserunt a eum autem",
  "Sint voluptatibus qu",
  "Quaerat quaerat quae"
]
```

 - Prism 4.1.1

IGassmann:
This fixes #1549.

Dynamic generations wouldn't follow `maxItems` and `maxLength` JSON
schema constraints due to the configured `json-schema-faker` options:

> maxItems — Override maxItems if it's greater than this value
> maxLength — Override maxLength if it's greater than this value

### DIFF
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -15,8 +15,6 @@ jsf.option({
   optionalsProbability: 1,
   fixedProbabilities: true,
   ignoreMissingRefs: true,
-  maxItems: 20,
-  maxLength: 100,
 });
 
 export function generate(source: JSONSchema): Either<Error, unknown> {


XVincentX:
Hey @IGassmann,

thanks a lot for the PR.

I'll clone the branch and try this out, but I am not 100% sure how this is going to fix the issue.

According to the documentation, the override will happen only if the value that's coming from the schema is greater than the default option value.

If I look at the issue, that is not what's happening (provided:5, option: 20).

Can you clarify?

XVincentX:
@IGassmann I took another look at this — while this effectively fixes the thing, it's not the best fix. 

The real problem is in JSON Schema Faker that's incorrectly interpreting the `minItems: 0` as a false value, making a fallback to out option.

The real fix is https://github.com/json-schema-faker/json-schema-faker/pull/619

We'll update the dependency when ready.

PR #1619, opened 2021-01-06
feat: readOnly and writeOnly support

### LINKED ISSUE #1142: Read-Only and Write-Only doesn't respect the scope.
**Describe the bug**
The required properties in schema don't respect the scope if we take a look at the OpenAPI Specification [documentation](https://swagger.io/docs/specification/data-models/data-types/) there is the following explanation:

> **Read-Only and Write-Only Properties**
If a readOnly or writeOnly property is included in the required list, required affects just the relevant scope – responses only or requests only. That is, read-only required properties apply to responses only, and write-only required properties – to requests only.

Seems to me that Prism doesn't apply this approach as you guys can see below.

**To Reproduce**

```yaml
openapi: 3.0.0
info:
  title: API
  description: API.
  version: v1
  contact:
    email: email@email.com
    name: Contato
tags:
  - name: items
servers:
  - url: 'http://127.0.0.1:4010'
    description: Prism
paths:
  /items:
    post:
      operationId: create-item
      summary: summary
      description: description
      tags:
        - items
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              required:
                - name
                - id
              properties:
                id:
                  description: Item unique identifier
                  type: string
                  format: uuid
                  readOnly: true
                name:
                  description: Project name
                  type: string
                  example: Item One
      responses:
        '201':
          description: Created
          headers:
            Location:
              description: '/items/{id}'
              schema:
                type: string

```

2. Run this CLI command 
```bash
prism mock api.yaml -d
```
3. See error
After sending a `POST` request, the request has a 400 status code, informing that `id` is required.
**Expected behavior**
Return a `201` status code, because `id` is required but is `readOnly` so must be required just on the response scope.

**Screenshots**

1 - Swagger Request

image

2 - Swagger Response 

image

3 - Console Output

image

- Library version: 3.3.3
- OS: Ubuntu 18.04
- Browser: All browsers

**Additional context**
I did some tests using the same spec file sending the rest to a [Mock Bin](http://mockbin.org/) and everything looks good.

marcelltoth:
Resolves #1142 

I'm not familiar with prism and FP much so code style, placement and naming might be off. Yet I think it fixes both the issue above, both the related `writeOnly` support. Hope it's not that terrible. 😄 

Also how would you go about testing this? I was messing with `http/src/validator/__tests__/index.spec.ts` but the actual validator return value is always mocked out, so doesn't verify much. Is there a test suite that tests the actual schema validation logic?

### DIFF
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -6,7 +6,7 @@ import * as sampler from 'openapi-sampler';
 import { Either, tryCatch, toError, right } from 'fp-ts/Either';
 import * as O from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/pipeable';
-import { stripWriteOnly } from 'http/src/utils/jsonSchema';
+import { stripWriteOnly } from '../../utils/jsonSchema';
 
 jsf.extend('faker', () => faker);
 
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -1,10 +1,12 @@
 import * as faker from 'faker/locale/en_US';
-import { cloneDeep } from 'lodash';
 import { JSONSchema } from '../../types';
 
 import * as jsf from 'json-schema-faker';
 import * as sampler from 'openapi-sampler';
-import { Either, tryCatch, toError } from 'fp-ts/Either';
+import { Either, tryCatch, toError, right } from 'fp-ts/Either';
+import * as O from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/pipeable';
+import { stripWriteOnly } from 'http/src/utils/jsonSchema';
 
 jsf.extend('faker', () => faker);
 
@@ -20,7 +22,13 @@ jsf.option({
 });
 
 export function generate(source: JSONSchema): Either<Error, unknown> {
-  return tryCatch(() => jsf.generate(cloneDeep(source)), toError);
+  return pipe(
+    stripWriteOnly(source),
+    O.fold(
+      () => right(undefined),
+      schema => tryCatch(() => jsf.generate(schema), toError)
+    )
+  );
 }
 
 export function generateStatic(resource: unknown, source: JSONSchema): Either<Error, unknown> {
--- a/packages/http/src/utils/jsonSchema.ts
+++ b/packages/http/src/utils/jsonSchema.ts
@@ -1,10 +1,11 @@
 import { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
 import * as O from 'fp-ts/lib/Option';
 import { pipe } from 'fp-ts/pipeable';
-import { mapValues, intersection } from 'lodash/fp';
+import { mapValues, intersection, pickBy } from 'lodash/fp';
 
 type RequiredSchemaSubset = {
   readOnly?: boolean;
+  writeOnly?: boolean;
   properties?: Record<string, JSONSchema6 | JSONSchema7 | JSONSchema4 | boolean>;
   required?: string[] | false;
 };
@@ -27,7 +28,8 @@ const buildSchemaFilter = (predicate: (schema: RequiredSchemaSubset) => boolean)
             O.toUndefined
           );
         })
-      )
+      ),
+      O.map(pickBy(val => val !== undefined))
     );
 
     const strippedPropertyKeys = pipe(
@@ -53,3 +55,4 @@ const buildSchemaFilter = (predicate: (schema: RequiredSchemaSubset) => boolean)
 };
 
 export const stripReadOnly = buildSchemaFilter(schema => schema.readOnly !== true);
+export const stripWriteOnly = buildSchemaFilter(schema => schema.writeOnly !== true);
--- a/packages/http/src/utils/jsonSchema.ts
+++ b/packages/http/src/utils/jsonSchema.ts
@@ -9,39 +9,47 @@ type RequiredSchemaSubset = {
   required?: string[] | false;
 };
 
-export function stripReadOnly<S extends RequiredSchemaSubset>(inputSchema: S): O.Option<S> {
-  if (inputSchema.readOnly) {
-    return O.none;
-  }
+type SchemaFilterFunc = <S extends RequiredSchemaSubset>(schema: S) => O.Option<S>;
 
-  const strippedProperties = pipe(
-    O.fromNullable(inputSchema.properties),
-    O.map(
-      mapValues(val => {
-        return pipe(
-          O.some(val),
-          O.chain(val => (typeof val === 'boolean' ? O.none : stripReadOnly(val))),
-          O.toUndefined
-        );
-      })
-    )
-  );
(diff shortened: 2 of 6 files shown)


packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -1,10 +1,12 @@
 import * as faker from 'faker/locale/en_US';
-import { cloneDeep } from 'lodash';
 import { JSONSchema } from '../../types';
 
 import * as jsf from 'json-schema-faker';
 import * as sampler from 'openapi-sampler';
-import { Either, tryCatch, toError } from 'fp-ts/Either';
+import { Either, tryCatch, toError, right } from 'fp-ts/Either';
```
XVincentX, line 6: We usually import everything 
```suggestion
import * as E from 'fp-ts/Either';
```

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -1,10 +1,12 @@
...
+import * as O from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/pipeable';
```
XVincentX, line 8: Don't import from pipeable
```suggestion
import { pipe } from 'fp-ts/function';
```

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -20,7 +22,13 @@ jsf.option({
 });
 
 export function generate(source: JSONSchema): Either<Error, unknown> {
-  return tryCatch(() => jsf.generate(cloneDeep(source)), toError);
+  return pipe(
+    stripWriteOnly(source),
+    O.fold(
+      () => right(undefined),
+      schema => tryCatch(() => jsf.generate(schema), toError)
+    )
```
XVincentX, line 30: We never fold, we rather propagate the error and handle it on the edges. This should be chained

packages/http/src/utils/jsonSchema.ts
```diff
@@ -0,0 +1,58 @@
+import { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
+import * as O from 'fp-ts/lib/Option';
+import { pipe } from 'fp-ts/pipeable';
```
XVincentX, line 3: import from function

packages/http/src/utils/jsonSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+  required?: string[] | false;
+};
+
+type SchemaFilterFunc = <S extends RequiredSchemaSubset>(schema: S) => O.Option<S>;
+
+const buildSchemaFilter = (predicate: (schema: RequiredSchemaSubset) => boolean): SchemaFilterFunc => {
+  const filter = <S extends RequiredSchemaSubset>(inputSchema: S): O.Option<S> => {
+    if (!predicate(inputSchema)) {
+      return O.none;
+    }
```
XVincentX, line 19: Use `O.fromPredicate`

packages/http/src/utils/jsonSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const strippedProperties = pipe(
+      O.fromNullable(inputSchema.properties),
+      O.map(
+        mapValues(val => {
+          return pipe(
```
XVincentX, line 25: Incorrect piping

packages/http/src/utils/jsonSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+            O.some(val),
+            O.chain(val => (typeof val === 'boolean' ? O.none : filter(val))),
+            O.toUndefined
```
XVincentX, line 28: folding — this should not happen

packages/http/src/utils/jsonSchema.ts
```diff
@@ -0,0 +1,58 @@
...
+          );
+        })
+      ),
+      O.map(pickBy(val => val !== undefined))
+    );
+
+    const strippedPropertyKeys = pipe(
+      strippedProperties,
+      O.map(Object.keys),
+      O.getOrElse(() => [] as string[])
```
XVincentX, line 38: folding — this should not happen

packages/http/src/validator/validators/body.ts
```diff
@@ -82,13 +83,21 @@ function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema,
...
+
+export const validate: validateFn<unknown, IMediaTypeContent> = (target, specs, mediaType, context = 'none') => {
   const findContentByMediaType = pipe(
     O.Do,
     O.bind('mediaType', () => O.fromNullable(mediaType)),
     O.bind('contentResult', ({ mediaType }) => findContentByMediaTypeOrFirst(specs, mediaType)),
     O.alt(() => O.some({ contentResult: { content: specs[0] || {}, mediaType: 'random' } })),
-    O.bind('schema', ({ contentResult }) => O.fromNullable(contentResult.content.schema))
+    O.bind('schema', ({ contentResult }) =>
+      pipe(O.fromNullable(contentResult.content.schema), O.chain(schemaProcessorMap[context]))
```
XVincentX, line 99: Incorrect piping

XVincentX (changes requested):
I gave it a quick look; this might solve the issue but the implementation is a little bit off.

I am also not 100% sure if this is something that Prism should handle directly or if AJV supports this in some degrees.

I suggest to punt this solution for know and move forward when a full investigation has done.

I've left some comments (mostly for personal purposes)

philsturgeon:
@XVincentX only OpenAPI has knowledge of "request" or "response", JSON Schema does not. This makes an awkward difference between "OpenAPI tooling" and "JSON Schema tooling".

[openapi-sampler](https://github.com/Redocly/openapi-sampler) supports this nicely:

image

json-schema-faker was not particularly interested in supporting this last time we talked about it:

https://github.com/json-schema-faker/json-schema-faker/issues/289#issuecomment-521951812

Then AJV would also need to have support added. Whilst we could certainly spend some time convincing the authors that this is a good feature, and help them implement it, we might be better off handling it ourselves for now.

It's not that weird if we do it ourselves because we are an "application" and that has been mentioned specifically in the way readOnly/writeOnly work.

> These keywords can be used to assist in user interface instance generation. In particular, an application MAY choose to use a widget that hides input values as they are typed for write-only fields. 

https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.4

Seeing as JSON Schema (and therefore OpenAPI v3.1) consider readOnly and writeOnly to be annotations only, the "application" can decide what to do with them. Prism as an application is choosing to strip them before passing them into JSON Schema specific tooling using our own helper logic. When the dependencies have a simple switch to do the same thing, we wont need to have our own logic to do the thing, but either way its basically the same.

philsturgeon:
@XVincentX misclick?

XVincentX:
Yeah — although the implementation is a no go anyway and will have to be refactored.

philsturgeon:
@XVincentX no worries, maybe we can leave this one up and somebody can take it over sometime?

XVincentX:
I do not mind keeping this open as a reminder!

radzserg:
This issue has been addressed in https://github.com/stoplightio/prism/pull/1853

PR #1620, opened 2021-01-06
chore: remove autoapprove

philsturgeon:


XVincentX (commented):
Why this change? I have been using this for a while to automate dependencies updates

philsturgeon:
Ignore me, I confused this with automerge which is a core feature now. I'm sure at some point this too will be unnecessary but not worth thinking about now.

XVincentX:
Yes! I am aware that we'll have to migrate but there are still some minor features that are not supported yet.

This has been a very good timesaver for me, so I'd keep it as much as possible and figure out something later.

PR #1623, opened 2021-01-07
chore:Use else if to return early

XVincentX:
Unify conditions and use `else if` to stop evaluation early.


karol-maciaszek (approved):
All fine!

I considered `switch` statement for a moment but it would result in more lines.

PR #1782, opened 2021-04-20
Fix: Support nullable validations

chohmann:
I've been receiving validation errors when using nullable: true in my OpenAPI schema.
This extends @nulltoken 's work here #1734 -- with the option described here.

Manually testing this locally appears to work as previously failing validations now pass 🥳 -- This unblocks us from using Prism in our E2E test suite.

### DIFF
--- a/packages/http/src/validator/validators/utils.ts
+++ b/packages/http/src/validator/validators/utils.ts
@@ -7,8 +7,14 @@ import type { ErrorObject } from 'ajv';
 import { JSONSchema } from '../../';
 import * as AjvOAI from 'ajv-oai';
 
-const ajv = new AjvOAI({ allErrors: true, messages: true, schemaId: 'auto' });
-const ajvNoCoerce = new AjvOAI({ allErrors: true, messages: true, schemaId: 'auto', coerceTypes: false });
+const ajv = new AjvOAI({ allErrors: true, nullable: true, messages: true, schemaId: 'auto' });
+const ajvNoCoerce = new AjvOAI({
+  allErrors: true,
+  nullable: true,
+  messages: true,
+  schemaId: 'auto',
+  coerceTypes: false,
+});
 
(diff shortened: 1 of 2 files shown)


philipbjorge:
Thanks @chohmann -- Excited to get access to this functionality!

Amjcraft (approved):

packages/http/src/validator/validators/utils.ts
```diff
@@ -7,8 +7,14 @@ import type { ErrorObject } from 'ajv';
 import { JSONSchema } from '../../';
 import * as AjvOAI from 'ajv-oai';
 
-const ajv = new AjvOAI({ allErrors: true, messages: true, schemaId: 'auto' });
-const ajvNoCoerce = new AjvOAI({ allErrors: true, messages: true, schemaId: 'auto', coerceTypes: false });
+const ajv = new AjvOAI({ allErrors: true, nullable: true, messages: true, schemaId: 'auto' });
```
XVincentX, line 10: @chohmann 

I still keep receiving notifications — I hope you do not mind me chiming in.

The only reason we are using Ajv OAI is for the formats that it includes, not for the transformation capabilities. All the OpenAPI Specific schemas should be transformed by Http Spec, and that has been correctly (as far as I remember) happening until https://github.com/stoplightio/http-spec/pull/128 landed in.

The idea was not to let OAS-specific data to leak in Prism; Http Spec should be the piece of software handling that part, and that's where the bug should be fixed.

Additionally, I have received an email from NPM with an attempt to republish the fix; not too sure what happened, but the version currently published _might not_ be the one you intended to.

chohmann: @XVincentX your concerns are already being addressed in https://github.com/stoplightio/prism/pull/1783. 

>Additionally, I have received an email from NPM with an attempt to republish the fix; not too sure what happened, but the version currently published might not be the one you intended to.

🤔  Hmm that is strange. Well the binaries and source code attached to 4.1.3 tag all look correct, and the 4.1.3 version in NPM wasn't published until we published those, so I'm thinking it's correct.

Issue #1743, opened 2021-03-24
Loading time can take up to one hour


thinkingserious:
## Describe the bug

There is a bug in the v8 engine that was to be resolved in the latest version. After upgrading, the bug persists.

Please see: https://github.com/stoplightio/prism/issues/1479#issue-722791366 for further details

## To Reproduce

Please see: https://github.com/stoplightio/prism/issues/1479

## Expected behavior

Please see: https://github.com/stoplightio/prism/issues/1479#issue-722791366

## Additional context
Creating a new issue based on the last interaction at #1479

marbemac:
Long time no talk @thinkingserious!

@karol-maciaszek looking at https://github.com/stoplightio/prism/issues/1479#issuecomment-735983921 - why are we including / using query parameters when matching route in the first place? Usually the only bit relevant to routing is the path.

marbemac:
And another q re routing @karol-maciaszek - had we ever considered using a battle hardened lib like https://github.com/pillarjs/path-to-regexp rather than rolling our own?

X3per:
> And another q re routing @karol-maciaszek - had we ever considered using a battle hardened lib like https://github.com/pillarjs/path-to-regexp rather than rolling our own?



> And another q re routing @karol-maciaszek - had we ever considered using a battle hardened lib like https://github.com/pillarjs/path-to-regexp rather than rolling our own?

`uri-template-lite` is being used for this regex generation. `path-to-regex` might help here but only if it produces different regex for the same input. The main problem may still reside inside node@15 engine.

I'd try two things:
- re-do the reproduction code provided by. @XVincentX here: https://github.com/stoplightio/prism/issues/1479#issuecomment-735983921
- try to replace `uri-template-lite` with `path-to-regex` (don't like it, we do use `uri-template-lite` in other places, we'll end up with two libs doing the same)

karol-maciaszek:
> And another q re routing @karol-maciaszek - had we ever considered using a battle hardened lib like https://github.com/pillarjs/path-to-regexp rather than rolling our own?



> And another q re routing @karol-maciaszek - had we ever considered using a battle hardened lib like https://github.com/pillarjs/path-to-regexp rather than rolling our own?

`uri-template-lite` is being used for this regex generation. `path-to-regex` might help here but only if it produces different regex for the same input. The main problem may still reside inside node@15 engine.

I'd try two things:
[...]

PR #1816, opened 2021-04-30
Use encodeURI param names to avoid performance issues on start up.

radzserg:
Addresses [#1743]

Use encodeURI param names to avoid performance issues on startup.

See more details in the  https://github.com/stoplightio/prism/issues/1743#issuecomment-830021600 thread

### DIFF
--- a/packages/cli/src/util/__tests__/paths.spec.ts
+++ b/packages/cli/src/util/__tests__/paths.spec.ts
@@ -274,7 +274,7 @@ describe('createExamplePath()', () => {
       );
     });
 
-    it('generates when params have special characters', () => {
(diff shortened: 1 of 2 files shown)


rainum (approved):
Whoa, what a difference! From 7.389s to 0.459ms, good job 🚀 

To be honest I'm quite surprised that `uri-template-lite` package that we're using for URI template parsing has so much trouble with parsing not URL safe characters. According to https://tools.ietf.org/html/rfc6570#section-2.1 not encoded `<` and `>` are not allowed in the templates either, they should be ignored. Anyway I'm glad that issue is fixed 👌

pytlesk4 (approved):

PR #1829, opened 2021-05-17
v4.2.3

wmhilton:
Bump for `sanitize-http` dep


chohmann (changes requested):
@wmhilton I don't see a release created with the various clients and source code:

image

Did something fail when you ran `yarn release`?

chohmann (approved):

PR #1831, opened 2021-05-18
Feat/releasing docs

radzserg:
Adds notes on how to release the new Prism version.


chohmann (changes requested):
These docs shouldn’t be in here as people outside of stoplight should not be releasing new versions of prism.

PR #1833, opened 2021-05-21
build bins for all os'

ksoviero-stoplight:


.circleci/config.yml
```diff
@@ -87,7 +87,7 @@ jobs:
           command: yarn lerna publish from-git --create-release=github --yes
       - run:
           name: Create CLI binaries
-          command: npx pkg --out-path ./cli-binaries --targets linux,macos ./packages/cli/
+          command: npx pkg --out-path ./cli-binaries ./packages/cli/
```
rainum, line 90: This should be the harmless change, since those are default options according to https://github.com/vercel/pkg#targets

rainum (commented):
I'm curious why did you delete the `azure-pipelines.yml`? Seems like this is not related to the current PR

rainum (approved):
🚢 it!

Issue #1785, opened 2021-04-20
Prism does not resolve URIs of resource-oriented services with custom methods correctly


kkaravitis:
## Describe the bug
I have the following services that I want to mock:

POST /users/{username}:activate
POST /users/{username}:block

The URI format is the proposed from Google resource-oriented services URI format with custom methods
https://google.aip.dev/136

Prism mocks correctly the POST /users/{username}:activate service
![success](https://user-images.githubusercontent.com/16665277/115473768-f82b5f80-a244-11eb-9946-aa0cced69d30.PNG)
but when I make a POST request to the second path ( /users/{username}:block ) I receive the following error:
 **error     Violation: request.path should have required property 'username'**
![error](https://user-images.githubusercontent.com/16665277/115473773-f8c3f600-a244-11eb-9ac5-b8c9e301e42b.PNG)

It seems like Prism does not recognize the difference between /users/user32:activate and /users/user32:block because when I make a POST request to the path

  /users/user32:block:activate  

then it returns a 200-OK response. 
It seems like Prism recognized the user32:block as the username in the path and because I put the :activate custom method segment at the end of the path, then it returned the response.

## To Reproduce

1. Given this OpenAPI document https://app.swaggerhub.com/apis/kkaravitis/prism-resource-oriented-services/1.0
2. Run this CLI command : prism mock
3. Execute curl -X POST "http://localhost:4010/users/user32:block" -H  "accept: */*" -d ""
3. See error

## Expected behavior

I would expect Prism to resolve URIs of the format of /resources/{resourceId}:customMethod correctly.

PR #1841, opened 2021-06-03
Feature/support colons in paths

jamesvillarrubia:
Addresses #1785 

**Summary**

First attempt at resolving the "custom method" standard that Google advises around REST endpoints (meeting the need for composable RPC calls within a REST framework).  This typically looks like this:

```
POST /users/{username}:activate
POST /users/{username}:block
```
This PR modifies the http/router/matchPath function to split fragments based on either a slash or a colon but does make some presumptions around trailingSlashes not being compatible.

It also allows for parameterized verbs such as:
```
POST /users/{username}:{action}
```
**Checklist**

- The basics
- Tests
  - [X] Added & updated in .spec files
    - any concrete path with colon should match an equal concrete path
    - none request path should match path with less fragments
    - none request path with colons should match path with less fragments
    - none request path with a colon should not match equivalent slash path
    - none request path should match concrete path with more fragments
    - none request path with colons should match path with more fragments
    - request path should match a templated path and resolve variables
    - request path should match a template path and resolve undefined variables
    - none path should match templated operation with more path fragments
  - [X] Added in test-harness files
    - When a path has a colon in it, it is selected first
    - When I send a request to an operation with a colon in it, return 200
  - [ ] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A

**Additional context**

I updated the faker references to use the non-deprecated `faker.datatype` instead of the deprecated `faker.random` call, but only in the files I touched.

I also noted a few tests that I suspect aren't actually testing what they're supposed to be... comparing random string vs random string instead of two similar structures.  `x/y/z` will never equal `a/b/c` or `a/b`.  I did not copy that test structure in my tests, but I'm open to modifications.  

Also thanks to @bziggz for assisting.

### DIFF
--- a/packages/http/src/router/matchPath.ts
+++ b/packages/http/src/router/matchPath.ts
@@ -27,7 +27,7 @@ export function matchPath(requestPath: string, operationPath: string): E.Either<
     return E.right(MatchType.NOMATCH);
   }
 
-  // a/b:c
+  // a/b:c = a/b:c
   const operationPathColonFragments = fragmentarizeWithColon(operationPath);
   const requestPathColonFragments = fragmentarizeWithColon(requestPath);
   if (
--- a/packages/http/src/router/matchPath.ts
+++ b/packages/http/src/router/matchPath.ts
@@ -5,12 +5,18 @@ function fragmentarize(path: string): string[] {
   return path.split('/').slice(1);
 }
 
+function fragmentarizeWithColon(path: string): string[] {
+  return path.split(/[/:]/).slice(1);
+}
+
 function getTemplateParamName(pathFragment: string) {
   const match = /{(.*)}/.exec(pathFragment);
   return match && match[1];
 }
 
 export function matchPath(requestPath: string, operationPath: string): E.Either<Error, MatchType> {
+  // a/b/c = a/b/c
+
   const operationPathFragments = fragmentarize(operationPath);
   const requestPathFragments = fragmentarize(requestPath);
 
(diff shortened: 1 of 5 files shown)


packages/http/src/router/matchPath.ts
```diff
@@ -21,10 +27,20 @@ export function matchPath(requestPath: string, operationPath: string): E.Either<
     return E.right(MatchType.NOMATCH);
   }
 
+  // a/b:c = a/b:c
+  const operationPathColonFragments = fragmentarizeWithColon(operationPath);
+  const requestPathColonFragments = fragmentarizeWithColon(requestPath);
+  if (
```
radzserg, line 33: Such a check doesn't check the order in separators. 
```
  test('it does not match if separators are not equal', () => {
    assertRight(matchPath('a:b/c', 'a/b:c'), e => expect(e).toEqual(MatchType.NOMATCH));
  });
```

I'd try to simplify this method. 2 kinds of separations a bit confusing. Considering that at the end we will use only `fragmentarizeWithColon`.  We could check if separators are equal in the operation and request.

```
import { isEqual } from 'lodash';

const pathSeparatorsRegex = /[/:]/g;

function fragmentarize(path: string): string[] {
  return path.split(pathSeparatorsRegex).slice(1);
}

function getTemplateParamName(pathFragment: string) {
  const match = /{(.*)}/.exec(pathFragment);
  return match && match[1];
}

function isSeparationEqual(path1: string, path2: string): boolean {
  return isEqual(path1.match(pathSeparatorsRegex), path2.match(pathSeparatorsRegex));
}

export function matchPath(requestPath: string, operationPath: string): E.Either<Error, MatchType> {
  if (!isSeparationEqual(requestPath, operationPath)) {
    return E.right(MatchType.NOMATCH);
  }

  const operationPathFragments = fragmentarize(operationPath);
  const requestPathFragments = fragmentarize(requestPath);
  if (operationPathFragments.length !== requestPathFragments.length) {
    return E.right(MatchType.NOMATCH);
  }

  const params = [];
  while (requestPathFragments.length) {
    const requestPathFragment = requestPathFragments.shift();
    const operationPathFragment = operationPathFragments.shift();

    const paramName = getTemplateParamName(operationPathFragment as string);

    if (paramName === null && operationPathFragment !== requestPathFragment) {
      // if concrete fragment and fragments are different return false
      return E.right(MatchType.NOMATCH);
    } else if (paramName !== null) {
      params.push({
        name: paramName,
        value: requestPathFragment,
      });
    }
  }

  return E.right(params.length ? MatchType.TEMPLATED : MatchType.CONCRETE);
}

```

radzserg (changes requested):
Great job. Thanks for your contribution. 

But we need to make sure this test pass to complete it.
```
 test('it does not match if separators are not equal', () => {
    assertRight(matchPath('a:b/c', 'a/b:c'), e => expect(e).toEqual(MatchType.NOMATCH));
  });
```

radzserg:
@jamesvillarrubia can you make the change to fix 
```
assertRight(matchPath('a:b/c', 'a/b:c'), e => expect(e).toEqual(MatchType.NOMATCH));
```
that I described above?

PR #1845, opened 2021-06-13
Fix/remove deprecation

jbl428:
**Summary**

Replace following deprecated usage to non-deprecated one

- faker.random.number
- faker.random.uuid
- faker.random.boolean
- E.either
- E.getValidation
- A.array
- O.option
- E.stringifyJSON

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

### DIFF
--- a/packages/http/src/forwarder/index.ts
+++ b/packages/http/src/forwarder/index.ts
@@ -6,7 +6,7 @@ import * as NEA from 'fp-ts/NonEmptyArray';
 import * as E from 'fp-ts/Either';
 import * as TE from 'fp-ts/TaskEither';
 import * as RTE from 'fp-ts/ReaderTaskEither';
-import * as JSON from 'fp-ts/json';
+import * as J from 'fp-ts/Json';
 import { defaults, omit } from 'lodash';
 import { format, parse } from 'url';
 import { posix } from 'path';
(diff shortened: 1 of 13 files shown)


packages/http/src/forwarder/index.ts
```diff
@@ -6,6 +6,7 @@ import * as NEA from 'fp-ts/NonEmptyArray';
 import * as E from 'fp-ts/Either';
 import * as TE from 'fp-ts/TaskEither';
 import * as RTE from 'fp-ts/ReaderTaskEither';
+import * as JSON from 'fp-ts/json';
```
rainum, line 9: You shouldn't use `JSON` keyword, it is reserved by JavaScript. Better is to use `import * as J from 'fp-ts/Json'` instead.

packages/http/src/mocker/callback/callbacks.ts
```diff
@@ -77,7 +78,7 @@ function assembleBody(request?: IHttpOperationRequest): O.Option<{ body: string;
     O.bind('body', ({ content }) => generateHttpParam(content)),
     O.chain(({ body, content: { mediaType } }) =>
       pipe(
-        E.stringifyJSON(body, () => undefined),
+        JSON.stringify(body),
```
rainum, line 81: Same as above ⬆️

rainum (changes requested):
Hey @jbl428, sorry for the long delay with the review. You did a great job on this PR, thank you! Just address couple minor issues and we're good to go.

Looks like CI's build if failing due to `Cannot find module 'fp-ts/json'` error. Correct module name is 'fp-ts/Json' (Json is capitalized). Update it, this should fix build issues.

rainum (approved):
Everything seems to be in order. Good job @jbl428 and thanks for you contribution!

PR #1873, opened 2021-08-20
fix: decode path before matching it

rainum:
**Summary**

You have a spec with `/some path` in it and you're trying to mock this path with mocker. If you'll send an encoded path (`/some%20path`) to the mocker it wont be able to find a corresponding path in the spec at will throw `NO_PATH_MATCHED_ERROR` which is not correct.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

### DIFF
--- a/packages/http/src/router/matchPath.ts
+++ b/packages/http/src/router/matchPath.ts
@@ -1,18 +1,29 @@
 import { MatchType } from './types';
 import * as E from 'fp-ts/Either';
 
-function fragmentarize(path: string): string[] {
+function fragmentize(path: string): string[] {
   return path.split('/').slice(1);
 }
 
-function getTemplateParamName(pathFragment: string) {
-  const match = /{(.*)}/.exec(pathFragment);
+// Attempt to decode path fragment. Decode should not do any harm since it is
+// decoding only URI sequences which are previously created by encodeURIComponent
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent
+function decodePathFragment(pathFragment?: string) {
+  try {
+    return pathFragment && decodeURIComponent(pathFragment);
+  } catch (_) {
+    return pathFragment;
+  }
+}
+
+function getTemplateParamName(pathFragment?: string) {
+  const match = typeof pathFragment === 'string' && /{(.*)}/.exec(pathFragment);
   return match && match[1];
 }
 
 export function matchPath(requestPath: string, operationPath: string): E.Either<Error, MatchType> {
-  const operationPathFragments = fragmentarize(operationPath);
-  const requestPathFragments = fragmentarize(requestPath);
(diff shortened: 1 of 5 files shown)


packages/http/src/router/matchPath.ts
```diff
@@ -1,18 +1,29 @@
 import { MatchType } from './types';
 import * as E from 'fp-ts/Either';
 
-function fragmentarize(path: string): string[] {
+function fragmentize(path: string): string[] {
```
rainum, line 4: I decided to update working a bit

packages/http/src/router/matchPath.ts
```diff
@@ -1,18 +1,29 @@
...
-function getTemplateParamName(pathFragment: string) {
-  const match = /{(.*)}/.exec(pathFragment);
+// Attempt to decode path fragment. Decode should not do any harm since it is
+// decoding only URI sequences which are previously created by encodeURIComponent
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent
+function decodePathFragment(pathFragment?: string) {
+  try {
+    return pathFragment && decodeURIComponent(pathFragment);
+  } catch (_) {
+    return pathFragment;
```
radzserg, line 15: when `decodeURIComponent` can throw an error?

packages/http/src/router/matchPath.ts
```diff
@@ -1,18 +1,29 @@
...
+  }
+}
+
+function getTemplateParamName(pathFragment?: string) {
+  const match = typeof pathFragment === 'string' && /{(.*)}/.exec(pathFragment);
```
radzserg, line 20: I feel like `mathPath` lacks a functional approach. How to turn 2 linkes into 11 lines ;)

```
  return pipe(
    O.fromNullable(pathFragment),
    O.map(pathFragment =>
      pipe(
        pathFragment,
        O.fromPredicate(pathFragment => typeof pathFragment === 'string'),
        O.map((pathFragment: string) => /{(.*)}/.exec(pathFragment)),
        O.map(m => m && m[1])
      )
    )
  );

```

packages/http/src/router/matchPath.ts
rainum: It'll throw if you provide any invalid sequence, for example: `decodeURIComponent('%1')`. I just want to make sure that we can handle some weird URLs.

packages/http/src/router/matchPath.ts
rainum: Yeah, it definitely is. In this particular case I just fixed type issue, let's plan a functional overhaul of `matchPath.ts` for another time 😉

radzserg (approved):
LGTM 👍

Issue #1224, opened 2020-06-03
Referential integrity in dynamic mocks


jedateach:
**User story.**
As a developer, I can define relationships between dynamic data, so that I can produce mocks that exercise more than a single model.

**Is your feature request related to a problem?**

I make use of the JSONAPI spec: (https://jsonapi.org), which is a very flat structure, which uses IDs to create relationships between models, which can sometimes be "included" (side-loaded) in a single call.

Apparently json-schema-faker has the concept of jsonPath to hook up references, but I wasn't able to get this working.

**Describe the solution you'd like**

I'd like dynamic mocks that produce JSONAPI data with valid references.

**Additional context**

This may have cross over with plans for persistence. I don't think jsonPath would be a viable solution across multiple API calls where data should be related.

XVincentX:
Hey @jedateach — what you're asking looks interesting, and it might make sense in Prism.

I would need to get more details though; could you construct an hypothetical OAS2/3 document and show me what a response to a particular request would look like?

taras:
I came here looking for the same. There is precedent for this. An obvious one is [Mirage.js](https://miragejs.com/) which is effectively an ORM behind an in-browser server. [Frontside team](https://github.com/thefrontside/) created a data store that we use to simulate a GraphQL server which provides us referential transparency for queries and mutations. We created it to allow us to create high fidelity simulations of a GraphQL schema that is not yet available. We would like to do the same thing with OpenAPI. 

This is easier to do with systems that have a concept of a model. Mirage.js for example has [models](https://miragejs.com/docs/main-concepts/models/) and [factories](https://miragejs.com/docs/main-concepts/factories/) which allows to create a simulated server by creating handlers that mutate in memory data store. The ORM allows to express relationships which can be used to construct complex data graph which is serialized to response payload. This is necessary for JSONAPI spec.

I'm guessing that it would be possible to wire Mirage ORM behind Prism but it wouldn't be my first choice for a long term solution. There is quite a bit of overhead to maintaining the Mirage simulator because it treats models and factories as separate things. Mirage's APIs for creating related data require quiet a bit of wiring. 

The Frontside team previously created an immutable in-memory data store with referential transparency and self materializing relationships using [Microstates.js](https://github.com/microstates). There might be a spike someplace. 

We're going to continue thinking about this and looking for opportunities to create a high fidelity simulator. I would love to contribute this work to Prism if you're interested in collaborating.

philsturgeon:
[...]

Issue #1647, opened 2021-01-19
Allow end user to configure global faker options


marbemac:
## User story.

As a Prism user, I can do configure global faker/json-schema-faker options.

## Is your feature request related to a problem?

There are lots of useful global json schema faker options, but no way to configure them. We set hard coded defaults here - https://github.com/stoplightio/prism/blob/master/packages/http/src/mocker/generator/JSONSchema.ts#L11-L20.

## Describe the solution you'd like

Open to suggestions, but the simplest and most consistent option that I can think of is to add support for a root level `x-faker` object prop in the openapi document.

```yaml
openapi: 3.0.0

x-faker:
  # pass through faker options (there is only one relevant one I think?)
  locale: de

  # pass through json-schema-faker options to override defaults. in this case always generating one array item
  minItems: 1
  maxItems: 1

# ... rest of document
```

## Additional context

Related issues: #1139 #1160

chohmann:
Another ticket asking for the same thing: https://github.com/stoplightio/prism/issues/1781

chrisjallen:
> Another ticket asking for the same thing: #1781

Hey Im sure this is the same,  #1647 is around global config, say I wanted to configure the parameters singularly for a faker generator method and pass in arguments specific to that method??

eg on faker website, the params for email generator
faker.internet.email(firstName?: string, lastName?: string, provider?: string): string

But on yaml we can only do 

```
        someProperty:
          type: string
          x-faker: internet.email
```


Q: is there a way to pass these in? I know its possible via json-schema-faker 
https://github.com/json-schema-faker/json-schema-faker/blob/master/docs/USAGE.md#more-examples
Notice :

```
  "properties": {
    "cash": {
      "type": "string",
      "faker": {
        "finance.amount": [100, 10000, 2, "$"]
      }
    }
  },
```

EdVinyard:
> Hey Im sure this is the same, #1647 is around global config, say I wanted to configure the parameters singularly for a faker generator method and pass in arguments specific to that method??

@chrisjallen it is possible to configure Faker per-property, but we didn't have an example of that, either in tests or documentation.  I've opened https://github.com/stoplightio/prism/pull/2012 to fix that.

In the future, please feel free to open a new issue with your questions or concerns.  We often miss comments on closed issues.

chrisjallen:
ah incredible thank you @EdVinyard

PR #1898, opened 2021-09-25
feat: Resolve values using JSON Path in examples

shrink:
There's a reference to this configuration option in #1224 but otherwise, I can't find any explanation for why it is not enabled. The option is very helpful to ensure that examples are contextually correct, e.g: a valid `JSON:API` response may need to provide both a `Resource Identifier Object`  _and_ included data for that Resource. 

> [**Available options**](https://github.com/json-schema-faker/json-schema-faker/tree/v0.5.0-rcv.34/docs#available-options)
> ...
> `resolveJsonPath` — If enabled, it will expand jsonPath keywords on all generated objects (default: `false`)

### DIFF
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -21,6 +21,7 @@ jsf.option({
   ignoreMissingRefs: true,
   maxItems: 20,
   maxLength: 100,
+  resolveJsonPath: true,
 });
 
 export function generate(bundle: unknown, source: JSONSchema): Either<Error, unknown> {


radzserg:
Hey @shrink, thank you for your contribution. Would you mind to add some OAS schema and test so we can see jsonPath resolution in action.

radzserg:
We can address this in https://github.com/stoplightio/prism/issues/1647

shrink:
> We can address this in #1647

Ah! I didn't see this ticket when looking for a way to achieve this. I am happy to close this Pull Request and instead submit one to solve #1647 if that's something that would be helpful?

radzserg:
Sure, go for it 👍

Issue #1781, opened 2021-04-19
more control over faker datatypes and generated mocks


Nilegfx:
## User story.

As a developer, I can pass faker function arguments, so that I can further customize my mocks.

## Is your feature request related to a problem?

No. Actually, I would like to thank the team behind prism and ask for feature that gives developers more control over the generated mocks

## Describe the solution you'd like

for now, I read the docs and went through the code quickly, I couldn't find a way to customize the faker datatypes, for example, paragraph with 5 sentences or firstName and gender as female.. etc

## Additional context

I would suggest having something along the lines of `x-faker: name.firstName female`

chohmann:
@Nilegfx Thanks for your input! We actually have another issue describing this same feature request:

https://github.com/stoplightio/prism/issues/1647

We'll use that one to track this feature request, and I'll close this one.

Feel free to add any additional info or ideas to that ticket.

Issue #1160, opened 2020-05-10
Faker locale


aibar:
How to set locale for faker extension?

XVincentX:
Unfortunately it's not possible at this moment. We might want to make this configurable, but we are going to need to find a reasonable way to do that.

xvilo:
I would also like to see such an ability.

philsturgeon:
We've got an issue talking about this over here: https://github.com/stoplightio/prism/issues/1647

Issue #1139, opened 2020-04-28
Control Number of Records in Collection


philsturgeon:
**User story.**
As a CLI or HTTP user, I can do request a particular number of records be returned from the mock server, so that I can test how my frontend application looks with multiple records, odd numbers of records, and a few different values.

**Is your feature request related to a problem?**

Users need more than just a single item in a collection.

> I am +1 on having more dynamic data generated and having that controlled by configuration and/or command line switches. I use the mock server when working on our SDKs and I'm alarmingly confident at how well they work with one record returned! **-- @lornajane**

Let's let users pick, regardless of if they're using hosted mocking or CLI.

**Describe the solution you'd like**

Default behavior continues to function exactly as it does now. 

If `Prefer: records=X` appears, the first time an `items: {}` is found it will repeat the number of records in that array to match the number provided.

There may well be multiple `items: {}`, but we pick the "first", which in JSON would be closest to the root, and also in top to bottom order.

```json
{
  "links": {
    "self": "http://example.com/articles/1"
  },
  "data": [ ... ],
  "related": [ ... ],
}
```

With a schema covering this, it's the data that would be used.

This isn't going to work for everyone, but it will help a lot of people until a more comprehensive idea can be thought up. 

#20 was outstanding for a year, so let's do the quick and dirty approach now to serve those who need it sooner rather than later.

**Additional context**

A few ideas were discussed over in https://github.com/stoplightio/prism/issues/20 but this issue is to track the chosen implementation.<br><br><i>(related to <a href='https://stoplight.zendesk.com/agent/tickets/5164'>Zendesk ticket #5164</a>)<br>gz#5164</i>

philsturgeon:
Heads up, HAL puts some of _links into an array:

```
{
    "_links": {
        "self": { "href": "/orders" },
        "curies": [{ "name": "ea", "href": "http://example.com/docs/rels/{rel}", "templated": true }],
        "next": { "href": "/orders?page=2" },
        "ea:find": {
            "href": "/orders{?id}",
            "templated": true
        },
        "ea:admin": [{
            "href": "/admins/2",
            "title": "Fred"
        }, {
            "href": "/admins/5",
            "title": "Kate"
        }]
    },
    "currentlyProcessing": 14,
    "shippedToday": 20,
    "_embedded": {
        "ea:order": [{
            "_links": {
                "self": { "href": "/orders/123" },
                "ea:basket": { "href": "/baskets/98712" },
[...]

PR #1899, opened 2021-09-27
feat: Allow JSON Schema Faker configuration in specification

shrink:
Related to...

- [x] #1647 
- [x] #1781 
- [x] #1898 
- [x] #1160

**Summary**

I, and others, would like to be able to configure the JSON Schema Faker behaviour for our mock server. JSON Schema Faker supports a variety of options, and the underlying Faker instance does too. I am specifically interested in the `resolveJsonPath` option because it'll allow me to create contextually accurate examples.

```yaml
openapi: 3.1.0

x-json-schema-faker:
  locale: de
  min-items: 2
  max-items: 10
  resolve-json-path: true
```

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Additional context**

I don't _love_ the implementation -- _should_ configuration options for a mock server live in the API specification? -- but it's what was suggested in the canonical ticket for this issue, so that's what I've done! I figured that if that was the wrong approach then there'd have been some feedback in #1647, but if this is a bad approach, I'm happy to adjust.

I explicitly named the member `x-json-schema-faker` _instead of_ `x-faker` because this applies to JSON Schema Faker, _not_ Faker, and there's no relationship between the configuration options of each -- that is to say, `minItems` has no meaning in the context of Faker. An argument could be made that `json-schema-faker` _already_ uses `x-faker` so it would be more consistent to use `x-faker` _but_ `x-faker` is used by JSON Schema Faker to communicate a Faker option, so I think the inconsistency in the member name is justified to represent the different purpose.

I chose not to do any sort of validation against the `x-json-schema-faker` options because JSON Schema Faker doesn't do any validation itself, it accepts arbitrary values into a `registry` and then pulls them as needed -- no failure on bad input. I think it's important that this feature can enable configuration of JSON Schema Faker even as it changes, so while it'd be ideal to validate options -- automatically against some sort of configuration exposed by JSON Schema Faker -- it's not possible to do reliably.

I chose not to adjust the specification request logic, so both `getHttpOperationsFromSpec` and `configureExtensionsFromSpec` will make a request. I... am conflicted, on the one hand, the second lookup is wasteful but on the other, I don't want to mess with `getHttpOperationsFromSpec` because it's used in quite a few places -- so on balance, I decided two lookups is better than changing core logic and tests... but if someone feels strongly that the second request isn't acceptable I can make some changes (as far as I know `json-schema-ref-parser` doesn't do any caching so it'd be the responsibility of Prism to cache?).

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -17,6 +17,7 @@
     "chalk": "^4.1.2",
     "chokidar": "^3.5.2",
     "fp-ts": "^2.11.5",
+    "json-schema-faker": "0.5.0-rcv.40",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.5",
     "pino": "^6.13.3",
--- a/packages/cli/tsconfig.json
+++ b/packages/cli/tsconfig.json
@@ -4,7 +4,8 @@
     "paths": {
       "@stoplight/prism-core": ["./core/src"],
       "@stoplight/prism-http": ["./http/src"],
-      "@stoplight/prism-http-server": ["./http-server/src"]
+      "@stoplight/prism-http-server": ["./http-server/src"],
+      "json-schema-faker": ["../node_modules/@types/json-schema-faker/index.d.ts"]
     }
   }
 }
--- a/test-harness/specs/dynamic-responses/dynamic_response_json_schema_faker_config.txt
+++ b/test-harness/specs/dynamic-responses/dynamic_response_json_schema_faker_config.txt
@@ -1,0 +1,49 @@
+====test====
+Given json-schema-faker configuration in the specification,
+When I send a request to an operation,
+Then the json-schema-faker configuration should influence the response.
+====spec====
+openapi: "3.0.2"
+x-json-schema-faker:
+  resolveJsonPath: true
+  useDefaultValue: true
+tags:
+  - name: example-tag
+info:
+  version: "0"
+  title: JSON Schema Faker test
+  description: JSON Schema Faker test
+  contact:
+    email: support@stoplight.io
+servers:
+  - url: http://api.example.com
+paths:
+  /widget:
+    get:
+      description: widget details
+      operationId: widgetDetails
+      tags: 
+        - example-tag
+      responses:
+        "200":
+          description: widget details response
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  name:
+                    type: string
+                    default: "Super Widget"
+                  alias:
+                    type: string
+                    jsonPath: $..name
+====server====
+mock -p 4010 -d ${document}
+====command====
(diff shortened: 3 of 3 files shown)


philsturgeon:
Wow thank you @shrink, for the great pull request and for taking the time to write up your reasoning.

@ryotrellim what do you think of the approach here? 

I'm usually somewhat against configuration in the mock for many things, but letting the author of the spec define how to build things seems a little helpful.

I'm not sure it exactly covers #1139 because its a global setting for all endpoints, unchangeable by the caller to cover tricky situations like pagination, but it does seem handy.

shrink:
@philsturgeon I agree. For context, I am designing a [JSON API Resource Revisioning extension](https://github.com/scenarios-com/json-api-ext-revisions) and need `resolveJsonPath` enabled in JSON Schema Faker in order to provide a good developer experience with the (to be) bundled OpenAPI Specification + Prism mocking workflow. My immediate need would be resolved by this Pull Request or my previous [feat: Resolve values using JSON Path in examples](https://github.com/stoplightio/prism/pull/1898).

Context is important for a Mock to be a great Mock: currently, Prism provides great mock data when sourcing from an explicit specification-defined example (correct shape and contextually correct values) but the dynamic mock data is only guaranteed to be correct in shape (sometimes random lipsum _is_ contextually correct, but not always). 

From a Prism product perspective ("...a best-in-class Mock Server...") I don't believe this change (exposing the JSON Schema Faker configuration in the specification) is right, but I do think there are big gains to be made towards the product's stated goal. A best-in-class Mock Server is a Mock Server that is indistinguishable from the real thing to a client, and so dynamic mock data that is contextually correct is important. Being able to use `jsonPath` in examples is probably the single highest impact + lowest-cost way to get closer to contextually correct data and hope to see that in the short term, but long term I'm sure there are even more helpful/interesting ways to achieve that goal.

So in summary, as a user with an immediate need to provide contextually correct dynamic mock data, I'd like to see this merged so I can use `jsonPath` in my examples. As a person thinking about Prism's stated purpose and Prism 5 years from now, this change (introducing `x-json-schema-faker` configuration into the specification) feels wrong and I'd rather see #1898 merged. I probably should have shared these thoughts in that Pull Request instead before jumping to create this one!

I've removed #1139 from the list of related tickets, good catch, thank you.

rainum:
Sorry guys, I'll be absent next couple of weeks. Assigning this PR to @philsturgeon to unblock @shrink.

ryotrellim:
@philsturgeon I'm fine with this approach.

lukasz-kuzynski-11sigma:
Generally looks fine. Please rebase with master and check tests.

Issue #1917, opened 2021-10-18
Example not found error when asking of example with code 200 with invalid request body


lukasz-kuzynski-11sigma:
## Describe the bug

Prism returns wrong error if asking for example for code 200 with an example key but with invalid body.

## To Reproduce
1. Given this OpenAPI document 
```yaml
openapi: 3.1.0
info:
  title: test
  version: '1.0'
servers:
  - url: 'http://localhost:3000'
paths:
  /test:
    post:
      summary: ''
      operationId: post-test
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  name:
                    type: string
              examples:
                200-test:
                  value:
                    name: 200-test
        '422':
          description: Unprocessable Entity (WebDAV)
          content:
            application/json:
              schema:
                type: object
                properties:
                  name:
                    type: string
              examples:
                422-test:
                  value:
                    name: 422-test
                422-test2:
                  value:
                    name: 422-test2
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                test:
                  type: string
              required:
                - test
```
2. Run this CLI command 
Note missing content-type
```shell
curl --request POST \
  --url https://localhost:3000/test \
  --header 'Prefer: code=422, example=422-test2'
```
3. See error
```json
{"name":"https://stoplight.io/prism/errors#NOT_FOUND","status":404,"detail":"Response for contentType: application/json and exampleKey: 200-test does not exist.","additional":{"workspaceSlug":"lukasz-11sigma","projectSlug":"test-repo","url":"/test","code":200,"exampleKey":"200-test","dynamic":false,"mediaTypes":["*/*"]}}
```

## Expected behavior

`Example` should be ignored in that case and first respo

## Additional context

**Screenshots**

**Environment (remove any that are not applicable):**
 - Prism version: 4.4.0

PR #1919, opened 2021-10-18
Fix/1917 ignore example key on invalid body

ghost:
Addresses #1917

**Summary**

Ignoring example key for if it doesn't match requested code.

NOTE!
This change implies a subtle change in logic where exampleKey should always be accompanied by preferred response code.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A


lukasz-kuzynski-11sigma:
@philsturgeon How do you feel about a new suggestion/behavior to expect people to provide response code for example key since otherwise might be ignored?

Issue #1813, opened 2021-04-29
Support 415 responses for unsupported media types in requests


savage-alex:
## Describe the bug

It appears that content type validation is not working as I can POST a body with an unacceptable content type (vs what the API definition states is allowed) and it is accepted without error
## To Reproduce

Run prism and with an operation that accepts a post body, send a body but with a "content-type" header that is not in the openAPI definition

## Expected behavior

A http 415 is returned informing the user which content types are accepted

## Notes
This works great for accept headers when as a consumer I ask for something in a format that is not supported. This bug is to protect against consumers sending bad content types in their request bodies. 

**Screenshots**
![image](https://user-images.githubusercontent.com/25430683/116577025-02fe8800-a908-11eb-86a7-98e01473c0a4.png)

PR #1935, opened 2021-11-03
feat: 1813 start using 415 code for invalid content-types instead constantly inferring it

lukasz-kuzynski-11sigma:
Addresses #1813

**Summary**

Replace this with something that explains what this PR is for and why it matters.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A


packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
```diff
@@ -0,0 +1,45 @@
...
+    O.fold(
+      () => false,
+      ({ a, b }) => {
+        return a.type === b.type && a.subtype === b.subtype;
+      }
+    )
+  );
+}
+
+const REGEXP = /^([-\w.]+)\/(?:([-\w.]+)\+)?([-\w.]+)$/;
```
EdVinyard, line 19: 1. I think there are some characters allowed by the standard that are not included by `[-\w.]`, such as `!`, `#` and a few others.  See [RFC-6838 §4.2](https://datatracker.ietf.org/doc/html/rfc6838#section-4.2).
1. Is this intended to be more liberal than the media-type registries for the top-level type?  That is, are completely custom top-level types allowed?  If not, I think the list of top-level types could be included exhaustively (based on "Registries included below" [here](https://www.iana.org/assignments/media-types/media-types.xhtml).
2. I think it should be possible to include the "parameter" portion of a media type following a semicolon (e.g., `charset=UTF-8`).  Is that possible here?  I wasn't able to figure out if that had been removed prior to this code running.

packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
```diff
@@ -0,0 +1,45 @@
...
+        extension: hasExtension ? match[2] : undefined,
+      });
+    })
+  );
+}
+
+interface ParsedContentType {
+  type: string;
+  subtype: string;
+  extension?: string;
```
EdVinyard, line 44: This is probably just pedantic, but I think the standard refers to this third part as "suffix" rather than "extension".

packages/http/src/mocker/index.ts
```diff
@@ -153,7 +154,7 @@ export function createInvalidInputResponse(
 ): R.Reader<Logger, E.Either<ProblemJsonError, IHttpNegotiationResult>> {
   const securityValidation = failedValidations.find(validation => validation.code === 401);
```
radzserg, line 155: securityValidation is not used anymore, let's delete it.

packages/http/src/mocker/index.ts
```diff
@@ -169,15 +170,41 @@ export function createInvalidInputResponse(
...
-            ? createUnauthorisedResponse(securityValidation.tags)
-            : createUnprocessableEntityResponse(failedValidations);
+          return createResponseForViolations(failedValidations);
         })
       )
     )
   );
 }
 
+function getExpectedCodesForViolations(failedValidations: NonEmptyArray<IPrismDiagnostic>): NonEmptyArray<number> {
```
radzserg, line 180:
```
const findValidationByCode = (validations: NonEmptyArray<IPrismDiagnostic>, code: number) =>
  validations.find(validation => validation.code === code);
```
feels like we can add a function and use it in 4 places ;)

packages/http/src/validator/index.ts
```diff
@@ -34,6 +34,22 @@ const checkBodyIsProvided = (requestBody: IHttpOperationRequestBody, body: unkno
     )
   );
 
+const isMediaTypeDefinedInContents = (mediaType?: string, contents?: IMediaTypeContent[]): boolean =>
```
radzserg, line 37: I'd call it `isMediaTypeSupportedInContents` otherwise 
```
    O.fromNullable(mediaType),
    O.fold(
      () => true.     // it's not really defined right 
```

packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
radzserg, line 19: let's give some meaningful name pls `const REGEXP = //`

radzserg (approved):
Looks good to me. I've added a few suggestions though.

packages/cli/src/extensions.ts
```diff
@@ -7,8 +7,12 @@ export async function configureExtensionsFromSpec(specFilePathOrObject: string |
   const result = decycle(await dereference(specFilePathOrObject));
 
   forOwn(get(result, 'x-json-schema-faker', {}), (value: any, option: string) => {
-    if (option === 'locale') return jsf.locate('faker').setLocale(value);
+    if (option === 'locale') {
+      // @ts-ignore
```
rainum, line 11: Is it possible to get rid of this ts-ignore by defining types for the `json-schema-faker` locally? If not would be great to have a comment with an explanation why do we need it here.

packages/http/src/__tests__/memory-leak-prevention.spec.ts
```diff
@@ -38,6 +38,6 @@ describe('Checks if memory leaks', () => {
     }
 
     global.gc();
-    expect(process.memoryUsage().heapUsed).toBeLessThanOrEqual(baseMemoryUsage * 1.02);
+    expect(process.memoryUsage().heapUsed).toBeLessThanOrEqual(baseMemoryUsage * 1.03);
```
rainum, line 41: I like magic numbers 😬 This is 3% as far as I understand?

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -10,8 +10,10 @@ import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/lib/Either';
 import { stripWriteOnlyProperties } from '../../utils/filterRequiredProperties';
 
+// @ts-ignore
```
rainum, line 13: Same as above ⬆️

packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
rainum: This is a good point! @EdVinyard I found a regex at Stackoverflow that based on the intersection of the grammar from [RFC 2045 Section 5.1](https://www.rfc-editor.org/rfc/rfc2045#section-5.1) and [RFC 7231 Section 3.1.1.1](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.1). They both related to the [RFC-6838](https://datatracker.ietf.org/doc/html/rfc6838#section-4.2) as far as I can tell.

This regex should be able to captures type, subtype, and parameters: 

```js
/(application|audio|font|example|image|message|model|multipart|text|video|x-(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+))/([0-9A-Za-z!#$%&'*+.^_`|~-]+)((?:[ \t]*;[ \t]*[0-9A-Za-z!#$%&'*+.^_`|~-]+=(?:[0-9A-Za-z!#$%&'*+.^_`|~-]+|\"(?:[^\"\\\\]|\\.)*\"))*)/
```

Looks a bit scary to be honest haha.

rainum (commented):
Everything looks good except bunch of `@ts-ignore`s. Please remove those if possible or add some comments with explanation.

You did a great job @lukasz-kuzynski-11sigma 👏

packages/http/src/mocker/index.ts
lukasz-kuzynski-11sigma: Fixed

packages/http/src/validator/index.ts
lukasz-kuzynski-11sigma: Done

packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
lukasz-kuzynski-11sigma: Done

packages/http/src/validator/utils/wildcardMediaTypeMatch.ts
lukasz-kuzynski-11sigma: Thanks for regexp. However I don't need this since first part of parsing content-type over here is done by `content-type` package that already has this regexp, therefore I don't need to repeat it. What I need to do is to extract suffix and subtype. 

However @EdVinyard is right that I didn't cover enough characters here. Updated Regexp

lukasz-kuzynski-11sigma: Fixed

rainum (approved):
This one is ready to go, good stuff @lukasz-kuzynski-11sigma 👍

rainum:
@lukasz-kuzynski-11sigma seems like diagram new image is cropped for some reason: https://raw.githubusercontent.com/stoplightio/prism/f5c13f6e6602809a8acfd5390feb8538e65f65ac/packages/http/docs/images/mock-server-dfd.jpeg

Issue #1064, opened 2020-03-30
Postman plus Prism Proxy causes errors


lornajane:
**Describe the bug**
I use Postman (https://www.postman.com/) with Prism, with good results. When I put Prism into proxy mode, Postman thinks it didn't get a response (but curl to the same endpoint works fine).

Prism shows no error messages, the output from Prism is identical whether I make the request through Postman or curl

Postman's console says "incorrect header check"

curl shows that Prism is returning a header `content-encoding: gzip`, which I don't get when I request the live API directly rather than via the Prism proxy.

**To Reproduce**

1. Given this OpenAPI document (actually I'm not sure it matters but I'm using https://github.com/Nexmo/api-specification/blob/master/definitions/verify.yml and hitting the `search/` endpoint
2. Run this CLI command `prism proxy verify.yml https://api.nexmo.com/verify`
3. Make a request to the Search endpoint using Postman

**Expected behavior**
The API response

**Environment (remove any that are not applicable):**
 - Library version: 3.2.9
 - OS: Ubuntu 18.04
 - Postman 7.21.1
 - curl: 7.58.0

**Additional context**
I managed to get things working by adding an additional header to Postman `Accept-Encoding` and leaving it empty - I think Postman sends this header by default. However curl does not send `Accept-Encoding` by default and I still see that Prism returns a `content-encoding: gzip` to curl so while I'm not sure Postman is sending correct headers, I think Prism is also adding something here that doesn't help!

XVincentX:
I am not experiencing this behaviour @lornajane — can you check out if I am reproducing correctly?

1. Run Prism mock on the nexmo API Server

<img width="1078" alt="image" src="https://user-images.githubusercontent.com/1416224/78037192-d35e0480-7330-11ea-8f12-5e31c6e899d7.png">


2. Use Postman to run a request:

<img width="1164" alt="image" src="https://user-images.githubusercontent.com/1416224/78037210-dbb63f80-7330-11ea-94d9-8739999fcc84.png">

3. Get Missing API Key error

<img width="600" alt="image" src="https://user-images.githubusercontent.com/1416224/78037235-e2dd4d80-7330-11ea-8aec-c0cc499ea7ab.png">

XVincentX:
@lornajane Feel free to reopen once you have more informations for us.

hilary:
Thanks @lornajane, your workaround is working for us, too.

Hi @XVincentX. We're also having the same problem. To reproduce, run

```
prism proxy ...
[...]

PR #1949, opened 2021-11-19
Removes `content-encoding` response header since we decompressed response body

radzserg:
Addresses #1064

**Summary**

In proxy mode, when a client requests an endpoint with "Accept-Encoding: gzip" we fetched compressed response but node-fetch decompress content automatically so we can validate response format. We send back a decompressed response but at the same time we pass back original headers that contain "Content-Encoding: gzip". In the end the client gets `Content-Encoding: gzip` header and tries to decompress the content and gets an error since the content is already decompressed.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [X] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A

**Steps to reproduce**

[prism_testing.zip](https://github.com/stoplightio/prism/files/7569403/prism_testing.zip)

- import Postman collection from the attached files
- run Prism in proxy mode `yarn cli  proxy ../../prismtest.oas2.json http://reqres.in -p 4014`
- Run a request from Postman
- Make sure that Postman gets a response successfully

**Screenshots**

Screenshot 2021-11-19 at 13 04 08

**Additional context**

Perhaps we could go even further and return a compressed response (as the original server does), but I don't think that is necessary, at least for the moment.

### DIFF
--- a/packages/http/src/forwarder/index.ts
+++ b/packages/http/src/forwarder/index.ts
@@ -20,63 +20,65 @@ import { UPSTREAM_NOT_IMPLEMENTED } from './errors';
 
 const { version: prismVersion } = require('../../package.json'); // eslint-disable-line
 
-const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig>['forward'] = (
-  { data: input, validations }: IPrismInput<IHttpRequest>,
-  baseUrl: string,
-  resource
-): RTE.ReaderTaskEither<Logger, Error, IHttpResponse> => logger =>
-  pipe(
-    NEA.fromArray(validations),
-    TE.fromOption(() => undefined),
-    TE.map(failedValidations => {
-      const securityValidation = failedValidations.find(validation => validation.code === 401);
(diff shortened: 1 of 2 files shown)


packages/http/src/forwarder/index.ts
```diff
@@ -20,63 +20,65 @@ import { UPSTREAM_NOT_IMPLEMENTED } from './errors';
...
-  { data: input, validations }: IPrismInput<IHttpRequest>,
-  baseUrl: string,
-  resource
-): RTE.ReaderTaskEither<Logger, Error, IHttpResponse> => logger =>
-  pipe(
-    NEA.fromArray(validations),
-    TE.fromOption(() => undefined),
-    TE.map(failedValidations => {
-      const securityValidation = failedValidations.find(validation => validation.code === 401);
+const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHttpConfig>['forward'] =
```
radzserg, line 23: It's just prettier updates

EdVinyard (approved):

Issue #1920, opened 2021-10-20
Prism cannot recognize readOnly property nested in allOf 


radzserg:
## Describe the bug

Prim cannot recognize readOnly property nested in allOf. Here's an example of a failing test. 

```
  it('strips nested in allOf readOnly property', () => {
    const schema: JSONSchema = {
      type: 'object',
      properties: {
        name: { type: 'string' },
        description: { type: 'string', writeOnly: true },
        title: { allOf: [{ readOnly: true }, { type: 'string' }] },
      },
      required: ['name', 'description', 'title'],
    };

    assertSome(stripReadOnlyProperties(schema), schema => {
      expect(schema.required).toEqual(['name', 'description']);
      expect(schema.properties).toEqual({
        name: expect.any(Object),
        description: expect.any(Object),
      });
    });
  });
```

## To Reproduce

```
openapi: 3.0.0
paths:
  /test:
    post:
      responses:
        201:
          content:
            application/json:
              schema:
                type: object
                properties:
                  result:
                    type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Post'

components:
  schemas:
    Post:
      title: Post
      properties:
        id:
          type: string
        author:
          allOf:
            - readOnly: true
            - $ref: '#/components/schemas/User'
      required:
        - id
        - author

    User:
      title: User
      type: object
      properties:
        id:
          type: string

```

```sh
 curl -X POST http://127.0.0.1:4010/test  -v -H "Content-Type: application/json" -d '{"id": "dqdd"}'

...
{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request","status":422,"detail":"Your request is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["body"],"severity":"Error","code":"required","message":"must have required property 'author'"}]}
```

## Expected behavior

> must have required property 'author'

Prism should not treat "author" as a required field since it's readOnly.

## Proposed Fix:
1. in https://github.com/stoplightio/prism/blob/master/packages/http/src/utils/filterRequiredProperties.ts#L92 update the stripReadOnly method to first "resolve" the schema to flatten the combiners using `json-schema-merge-allOf` package and then check for readOnly property.

languitar:
What happened to the commit that closed this issue? We are facing a similar issue.

radzserg:
[...]

PR #1957, opened 2021-11-29
for input validation, readOnly properties inside allOf are omitted from normalized schema

EdVinyard:
Addresses #1920 

## Summary

When validating input, `readOnly` properties inside `allOf` should be omitted from normalized schema.  That is, it should never be required to supply a `readOnly` property in input.  Prior to this change, if the `readOnly` attribute was used within an `allOf`, it was not correctly removed from the "normalized" schema against which we'll validate. 

## Checklist

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

## Additional context

Some open questions exist; see comments inline in changes/diffs.

### DIFF
--- a/packages/http-server/src/__tests__/body-params-validation.spec.ts
+++ b/packages/http-server/src/__tests__/body-params-validation.spec.ts
@@ -422,7 +422,12 @@ describe('body params validation', () => {
         test('returns 422 & error message', async () => {
           const response = await makeRequest('/json-body-circular-property-required', {
             method: 'POST',
-            body: JSON.stringify({ id: 123, self: {} }),
+            body: JSON.stringify({
+              id: 123,
+              self: {
+                // missing id property here
+              },
+            }),
             headers: { 'content-type': 'application/json' },
           });
 
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@stoplight/json": "^3.13.7",
-    "@stoplight/json-schema-merge-allof": "^0.7.7",
+    "@stoplight/json-schema-merge-allof": "^0.7.8",
     "@stoplight/json-schema-sampler": "0.2.2",
     "@stoplight/prism-core": "^4.6.1",
     "@stoplight/types": "^12.2.0",
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@stoplight/json": "^3.13.7",
-    "@stoplight/json-schema-merge-allof": "^0.7.8",
+    "@stoplight/json-schema-merge-allof": "0.7.8",
     "@stoplight/json-schema-sampler": "0.2.2",
     "@stoplight/prism-core": "^4.6.1",
     "@stoplight/types": "^12.2.0",
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -30,6 +30,7 @@
     "faker": "^5.5.3",
     "fp-ts": "^2.11.5",
     "json-schema-faker": "0.5.0-rcv.40",
+    "json-schema-merge-allof": "^0.8.1",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.5",
     "pino": "^6.13.3",
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@stoplight/json": "^3.13.7",
+    "@stoplight/json-schema-merge-allof": "^0.7.7",
     "@stoplight/json-schema-sampler": "0.2.2",
     "@stoplight/prism-core": "^4.6.1",
     "@stoplight/types": "^12.2.0",
@@ -30,7 +31,6 @@
     "faker": "^5.5.3",
     "fp-ts": "^2.11.5",
     "json-schema-faker": "0.5.0-rcv.40",
-    "json-schema-merge-allof": "^0.8.1",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.5",
     "pino": "^6.13.3",
--- a/packages/http/src/validator/validators/__tests__/body.spec.ts
+++ b/packages/http/src/validator/validators/__tests__/body.spec.ts
@@ -1,6 +1,6 @@
 import { HttpParamStyles, IMediaTypeContent } from '@stoplight/types';
 import { JSONSchema } from '../../..';
-import { validate, findContentByMediaTypeOrFirst, ForUnitTesting } from '../body';
+import { validate, findContentByMediaTypeOrFirst } from '../body';
 import { assertRight, assertLeft, assertSome } from '@stoplight/prism-core/src/__tests__/utils';
 import { ValidationContext } from '../types';
 
@@ -169,6 +169,111 @@ describe('validate()', () => {
       assertRight(validate({ name: 'Item One', title: 'title' }, specs, ValidationContext.Output, 'application/json'));
     });
   });
+
+  describe('merge allOf', () => {
+    it('nested below top-level', () => {
+      // Arrange
+      const schemas: IMediaTypeContent[] = [
+        {
+          mediaType: 'application/json',
+          schema: {
+            type: 'object',
+            required: ['level1'],
+            properties: {
+              level1: {
+                type: 'object',
+                required: ['level2'],
+                properties: {
+                  level2: {
+                    allOf: [
+                      { description: 'a description' },
+                      { type: 'string' },
+                    ]
+                  },
+                },
+              },
+            },
+          },
+        },
+      ];
+
+      // Act
+      const actual = validate(
+        { level1: { level2: 'abc' } },
+        schemas,
+        ValidationContext.Output,
+        'application/json');
+
+      // Assert
+      assertRight(actual);
+    });
+    it('does NOT require writeOnly params in output', () => {
+      // Arrange
+      const schemas: IMediaTypeContent[] = [
+        {
+          mediaType: 'application/json',
+          schema: {
+            type: 'object',
+            required: ['name', 'writeOnlyProperty'],
+            properties: {
+              name: {
+                type: 'string',
+              },
+              writeOnlyProperty: {
+                allOf: [
+                  { writeOnly: true },
+                  { type: 'string' }
+                ]
+              },
+            },
+          },
+        },
+      ];
+
+      // Act
+      const actual = validate(
+        { name: 'Ann' },
+        schemas,
+        ValidationContext.Output,
+        'application/json');
+
+      // Assert
+      assertRight(actual);
+    });
+    it('does NOT require readOnly params in input', () => {
+      // Arrange
+      const schemas: IMediaTypeContent[] = [
+        {
+          mediaType: 'application/json',
+          schema: {
+            type: 'object',
+            required: ['name', 'readOnlyProperty'],
+            properties: {
+              name: {
+                type: 'string',
+              },
+              readOnlyProperty: {
+                allOf: [
+                  { readOnly: true },
+                  { type: 'string' }
+                ]
+              },
+            },
+          },
+        },
+      ];
+
+      // Act
+      const actual = validate(
+        { name: 'Ann' },
+        schemas,
+        ValidationContext.Input,
+        'application/json');
+
+      // Assert
+      assertRight(actual);
+    });
+  });  
 });
 
 describe('findContentByMediaTypeOrFirst()', () => {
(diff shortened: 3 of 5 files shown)


packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -422,7 +422,12 @@ describe('body params validation', () => {
         test('returns 422 & error message', async () => {
           const response = await makeRequest('/json-body-circular-property-required', {
             method: 'POST',
-            body: JSON.stringify({ id: 123, self: {} }),
+            body: JSON.stringify({
+              id: 123,
+              self: {
+                // missing id property here
```
EdVinyard, line 428: I only added this comment to clarify the test's expectations.

packages/http/package.json
```diff
@@ -30,6 +30,7 @@
     "faker": "^5.5.3",
     "fp-ts": "^2.11.5",
     "json-schema-faker": "0.5.0-rcv.40",
+    "json-schema-merge-allof": "^0.8.1",
```
EdVinyard, line 33: Although Stoplight has a fork of `json-schema-merge-allof` it introduces some new behavior ([discarding `$ref`s](https://github.com/stoplightio/json-schema-merge-allof/blob/85cc8c7ee7d944aa117f173c11b0bc1087c36dae/src/index.js#L609-L611)) that breaks an existing test.  I believe that is a "true" failure that reveals an undesirable change.  This prompts a couple questions:

1. Is it necessary that we use the Stoplight branch?
2. If so, is it possible to use the Stoplight branch without the changes?
3. If the custom behaviors can't be disabled, do I need to resolve references (via `json-schema-ref-parser` or similar) prior to using Stoplight's merge-allOf?

packages/http/src/validator/validators/body.ts
```diff
@@ -11,6 +11,8 @@ import { JSONSchema } from '../../types';
 import { body } from '../deserializers';
 import { validateAgainstSchema } from './utils';
 import { ValidationContext, validateFn } from './types';
+// @ts-ignore no typings
+import * as mergeAllOf from 'json-schema-merge-allof';
```
EdVinyard, line 15: `@ts-ignore no typings` feels wrong.  Is there a better way?  I couldn't find any types for the Stoplight branch of `json-schema-merge-allof` or the current version of `json-schema-merge-alllof`.

packages/http/src/validator/validators/body.ts
```diff
@@ -86,20 +88,31 @@ function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema,
...
+      // const newSchema = normalizer(schema);
       cache.set(schema, newSchema);
       return newSchema;
     }
     return cached;
   };
 }
 
-type SchemaNormalizer = (schema: JSONSchema) => O.Option<JSONSchema>;
+export type SchemaNormalizer = (schema: JSONSchema) => O.Option<JSONSchema>;
```
EdVinyard, line 115: I don't see a way to avoid adding an export for types, similar to what I did below with `ForUnitTesting`.  Is there a better solution to keeping this type name relatively private?

packages/http/src/validator/validators/body.ts
```diff
@@ -86,20 +88,30 @@ function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema,
...
+  });
+}
+
 function memoizeSchemaNormalizer(normalizer: SchemaNormalizer): SchemaNormalizer {
   const cache = new WeakMap<JSONSchema7, O.Option<JSONSchema7>>();
   return (schema: JSONSchema7) => {
     const cached = cache.get(schema);
     if (!cached) {
-      const newSchema = normalizer(schema);
+      const newSchema = normalizer(withoutAllOf(schema));
```
EdVinyard, line 106: @rainum , I made this change here rather than in [`stripReadOnlyProperties()](https://github.com/stoplightio/prism/blob/8888233a91415284de9151f5cb4e79dc98248628/packages/http/src/utils/filterRequiredProperties.ts#L91) because I think we want symmetrical behavior for `writeOnly` properties when validating output.  Is that correct?

packages/http/package.json
rainum: I'm not sure that I'm the right person to answer this question, but I'll try:

1. We use `json-schema-merge-allof` as well in `public-api` and `studio` packages. Both are using our fork.
2. I'd say we should use the same version everywhere for consistency.
3. This is a good question. The schema should be dereferenced by the `getHttpOperationsFromSpec` method, I'm not sure this is required in this particular case.

packages/http/src/validator/validators/body.ts
rainum: > I think we want symmetrical behavior for writeOnly properties when validating output.

Perfect 👌 You're completely right

packages/http/package.json
EdVinyard: @P0lip, after talking with @rainum today, my tentative plan is to rely on Stoplight's fork of the _json-schema-merge-allof_ library here instead of the "original".  However, the recently added behavior in our fork that [discards $refs](https://github.com/stoplightio/json-schema-merge-allof/blob/85cc8c7ee7d944aa117f173c11b0bc1087c36dae/src/index.js#L609-L611) is not desirable in this situation, so I think I'll need to add some kind of switch/configuration to our fork.

How would you approach this?  Is what I propose a reasonable solution?

P0lip: You can provide a custom $ref resolver, i.e. https://github.com/stoplightio/json-schema-tree/blob/master/src/mergers/mergeAllOf.ts

EdVinyard: Thank you, @P0lip !  That's perfect.

EdVinyard:
@P0lip , I've updated with PR following our conversation this morning:

- `packages/http-server/src/__tests__/body-params-validation.spec.ts` includes `$ref` tests
- added test to confirm removing the `deep: true` option to json-schema-merge-allof doesn't break anything
- I'll start a follow-up discussion regarding impossible schema.  For now, at least Prism logs errors for incorrect output and warnings for incorrect input, even if it doesn't explain what's wrong.

Additionally, I've changed my tests to exercise `validate()` rather than the schema normalizer directly.  That should make it a little safer/easier to refactor the module internals later (which will be necessary for impossible-schema warnings).

packages/http/src/validator/validators/body.ts
```diff
@@ -86,12 +88,27 @@ function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema,
   );
 }
 
+/**
+ * Given a schema, return an equivalent schema that does not include `allOf`.
+ */
+function withoutAllOf(s: JSONSchema): JSONSchema {
+  try {
+    return mergeAllOf(s, { ignoreAdditionalProperties: true });
```
P0lip, line 96: https://github.com/stoplightio/json-schema-merge-allof/blob/master/src/index.js#L477

```suggestion
    return mergeAllOf(s, { deep: false, ignoreAdditionalProperties: true });
```
`deep` is true by default

The bad news is that flipping it off makes some of the tests fail.

This is because `allOf` isn't at the top-level, but rather nested below

```json
{
  "type": "object",
  "required": [
    "name",
    "readOnlyProperty"
  ],
  "properties": {
    "name": {
      "type": "string"
    },
    "readOnlyProperty": {
      "allOf": [
        {
          "readOnly": true
        },
        {
          "type": "string"
        }
      ]
    }
  }
}
```

packages/http/src/validator/validators/__tests__/body.spec.ts
```diff
@@ -169,6 +169,111 @@ describe('validate()', () => {
       assertRight(validate({ name: 'Item One', title: 'title' }, specs, ValidationContext.Output, 'application/json'));
     });
   });
+
+  describe('merge allOf', () => {
```
P0lip, line 173: should we add a test for mutually exclusive constraints imposed by `allOf` members?

packages/http/src/validator/validators/body.ts
P0lip:
```suggestion
    if ('allOf' in s) {
      return withoutAllOf(mergeAllOf(s, { deep: false, ignoreAdditionalProperties: true }));
    }

    if (isPlainObject(s.properties)) {
      return {
        ...s,
        properties: Object.entries(s.properties).reduce((result, [key, value]) => {
          result[key] = typeof value === 'boolean' ? value : withoutAllOf(value);
          return result;
        }, {}),
      };
    }
```

this should fix the issue tho.

P0lip: ^ note that this should be prob using fp-ts

packages/http/src/validator/validators/__tests__/body.spec.ts
EdVinyard: I made [some notes about mutually exclusive `allOf`](https://app.zenhub.com/workspaces/team-platinum-falcons-5eb9bc4a1fddfe30b94f78b1/issues/stoplightio/prism/1920#issuecomment-984065953), including a draft test that would fail.

I brought this issue up with Team P.F. this morning during stand up.  @ryotrellim seems OK with solving this problem in some other way, addressing it elsewhere in the workflow.  To me, it seems odd to include input/output validation messages that concern the _schema against which we're validating_, rather than the relationship between the input/output and the schema.

What do you think, @P0lip ?

ryotrellim: Admittedly I'm not fully comprehending what's happening here so lmk if we're introducing some bad UX or excessive tech debt in the context of this bug fix.

P0lip: tl;dr let's not worry about it too much now. We can do it later. We have here is already good enough.

> To me, it seems odd to include input/output validation messages that concern the schema against which we're validating

I totally agree. We should validate it earlier before we even get to this point.

@EdVinyard @ryotrellim 

>  or excessive tech debt in the context of this bug fix.

I'm not a Prism expert, so I'm not quite sure where that logic should live and cannot quite say whether it'd result in a tech debt, but I'd hope it's a rather not an overly complicated thing to address in the future.

>  introducing some bad UX

I'm mostly concerned about that part, but frankly speaking, it really boils down to how we want to handle corner cases (that unfortunately oftentimes exist in those large enterprise-grade documents).

Consider the following
```yaml
openapi: 3.0.0
paths:
  /test:
    post:
      responses:
        201:
          content:
            application/json:
              schema:
                type: object
                properties:
                  result:
                    type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Post'

components:
  schemas:
    Post:
      title: Post
      properties:
        id:
          type: string
        author:
          allOf:
            - readOnly: true
            - $ref: '#/components/schemas/Viewer'
            - $ref: '#/components/schemas/Guest'
      required:
        - id
        - author

    Viewer:
      title: Viewer
      type: object
      properties:
        id:
          type: integer

    Guest:
      title: Guest
      type: object
      properties:
        id:
          type: number
```

Running the very same command that Sergey posted in the ticket leads to the same error

```
 curl -X POST http://127.0.0.1:4010/test  -v -H "Content-Type: application/json" -d '{"id": "dqdd"}'

{"type":"https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY","title":"Invalid request","status":422,"detail":"Your request is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.","validation":[{"location":["body"],"severity":"Error","code":"required","message":"must have required property 'author'"}]}* Closing connection 0
```

and it might be not apparent why it still doesn't work.

The library itself prints such an error
image
but we don't expose it to the end-user.

Either way, that's rather a nice to have kind of a thing rather than a huge defect that hampers the UX.

packages/http/src/validator/validators/body.ts
EdVinyard: @P0lip , what's the difference between your recursive allOf-replacement and calling `mergeAllOf()` with `deep: true`?  I know you've mentioned performance problems when using `deep: true`; are those problems caused by something other than the arbitrarily-deep recursion?

P0lip: Mine doesn't traverse as deeply.
json-schema-merge-allof, when deep mode is enabled,  traverses the _entire_ schema, from the very top to the very bottom.

Consider the following scenario

Input schema
```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "foo": {
      "type": "object",
      "properties": {
        "bar": {
          "allOf": [
            {
              "type": "array",
              "items": {
                "allOf": [
                  {
                    "type": "object"
                  }
                ]
              }
            }
          ]
        }
      }
    }
  }
}
```

Output
- deep (the existing code)
   ```json
   {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "foo": {
          "type": "object",
          "properties": {
            "bar": {
              "type": "array",
              "items": {
                "type": "object"
              }
            }
          }
        }
      }
    }
   ```
   
 - my sample
   ```json
    {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "foo": {
          "type": "object",
          "properties": {
            "bar": {
              "type": "array",
              "items": {
                "allOf": [
                  {
                    "type": "object"
                  }
                ]
              }
            }
          }
        }
      }
    }
    ```

packages/http/src/validator/validators/body.ts
```diff
@@ -86,12 +88,34 @@ function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema,
...
 
+/**
+ * Given a schema, return an equivalent schema that does not include `allOf`.
+ */
+function withoutAllOf(s: JSONSchema): JSONSchema {
+  try {
+    return mergeAllOf(s, {
+      // `deep` traverses the *whole* schema.  If this becomes a performance
+      // problem, see the discussion and alternate implementation at
+      // https://github.com/stoplightio/prism/pull/1957/files#r760950082
```
EdVinyard, line 99: After discussing this with @P0lip further, we've decided to leave the `deep` allOf merging option enabled for now.  I've added a note in the code that points to this discussion should poor performance become a problem later.

packages/http/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@stoplight/json": "^3.13.7",
+    "@stoplight/json-schema-merge-allof": "^0.7.8",
```
P0lip, line 20:
```suggestion
    "@stoplight/json-schema-merge-allof": "0.7.8",
```
perhaps let's lock the version, as according to semantic versioning, 0.x.x is considered as "under active development" and the public API may change anytime. I'm not saying it'll change, but at the same time I'd not be particularly surprised if we introduced some subtle breaking changes that could affect the work you did here.

P0lip (approved):
:+1: on the merge-allof thing.

PR #1959, opened 2021-12-04
fix: remove deprecated usage of parse

jbl428:
**Summary**

Replace depreacted parse function to WHATWG URL.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

### DIFF
--- a/packages/http/src/__tests__/client.spec.ts
+++ b/packages/http/src/__tests__/client.spec.ts
@@ -43,10 +43,10 @@ describe('User Http Client', () => {
         jest.spyOn(client, 'request');
       });
 
-      afterAll(() => jest.clearAllMocks());
-
       describe('when calling with no options', () => {
         beforeAll(() => client.get('/pet'));
+
+        afterAll(() => jest.clearAllMocks());
 
         test('shall call the mocker with the default options', () =>
           expect(mock.default).toHaveBeenCalledWith({
@@ -61,6 +61,8 @@ describe('User Http Client', () => {
 
       describe('when calling a method with overridden url', () => {
         beforeAll(() => client.get('/pet', { baseUrl: 'https://www.google.it' }));
+
+        afterAll(() => jest.clearAllMocks());
 
         test('should call the mocker with the replaced full url', () => {
           expect(mock.default).toBeCalledWith({
@@ -80,6 +82,8 @@ describe('User Http Client', () => {
 
       describe('when calling a method with a full url', () => {
         beforeAll(() => client.get('https://www.google.it/pet'));
+
+        afterAll(() => jest.clearAllMocks());
 
         test('should call the mocker with the replaced full url', () => {
           expect(mock.default).toBeCalledWith({
--- a/packages/http/src/client.ts
+++ b/packages/http/src/client.ts
@@ -48,7 +48,7 @@ export function createClientFromOperations(resources: IHttpOperation[], defaultC
       const httpUrl: IHttpUrl = {
         baseUrl,
         path: parsedUrl.pathname,
-        query: parseQueryString(parsedUrl.search.substring(1)) as IHttpNameValues,
+        query: parseQueryString(parsedUrl.search?.substring(1) || '') as IHttpNameValues,
       };
 
       return pipe(
--- a/packages/http/src/client.ts
+++ b/packages/http/src/client.ts
@@ -2,11 +2,11 @@ import { IPrismOutput } from '@stoplight/prism-core';
 import { IHttpOperation } from '@stoplight/types';
 import { defaults } from 'lodash';
 import { parse as parseQueryString } from 'querystring';
-import { parse as parseUrl } from 'url';
 import { createInstance, IHttpNameValues } from '.';
 import { IHttpConfig, IHttpRequest, IHttpResponse, IHttpUrl } from './types';
 import { fold } from 'fp-ts/TaskEither';
 import * as Task from 'fp-ts/Task';
+import * as O from 'fp-ts/Option';
 import { pipe } from 'fp-ts/function';
 import * as pino from 'pino';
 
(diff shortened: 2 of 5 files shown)


Nezteb:
FYI @chohmann I added TC as a reviewer. This PR is old/stale, so I just wanted to make sure it's on your radar!

packages/http/src/__tests__/client.spec.ts
```diff
@@ -42,11 +42,11 @@ describe('User Http Client', () => {
         jest.spyOn(client, 'request');
       });
 
-      afterAll(() => jest.clearAllMocks());
```
daniel-white, line 45: why was this moved into the inner scopes?

packages/http/src/client.ts
```diff
@@ -30,16 +30,25 @@ export function createClientFromOperations(resources: IHttpOperation[], defaultC
...
+        O.tryCatch(() => new URL(url)),
+        O.getOrElseW(() => new URL(url, 'https://stoplight.io'))
+      );
 
       const httpUrl: IHttpUrl = {
-        baseUrl: parsedUrl.host ? `${parsedUrl.protocol}//${parsedUrl.host}` : mergedConf.baseUrl,
+        baseUrl,
         path: parsedUrl.pathname,
-        query: parseQueryString(parsedUrl.query || '') as IHttpNameValues,
+        query: parseQueryString(parsedUrl.search.substring(1)) as IHttpNameValues,
```
daniel-white, line 51: is `search` guaranteed not to be undefined?

packages/http/src/forwarder/__tests__/index.spec.ts
```diff
@@ -4,11 +4,12 @@ import { assertResolvesRight, assertResolvesLeft } from '@stoplight/prism-core/s
 import { keyBy, mapValues } from 'lodash';
 import { hopByHopHeaders } from '../resources';
 import { DiagnosticSeverity, Dictionary } from '@stoplight/types';
+import { format } from 'url';
```
daniel-white, line 7: i'm not seeing any functional changes here

packages/http/src/forwarder/index.ts
```diff
@@ -41,9 +41,11 @@ const forward: IPrismComponents<IHttpOperation, IHttpRequest, IHttpResponse, IHt
       TE.chainEitherK(() => serializeBody(input.body)),
       TE.chain(body =>
         TE.tryCatch(async () => {
-          const partialUrl = parse(baseUrl);
+          const partialUrl = new URL(baseUrl);
           const url = format({
             ...partialUrl,
+            host: partialUrl.host,
```
daniel-white, line 47: why include this if is already spread in from line 46?

daniel-white (changes requested):

packages/http/src/forwarder/__tests__/index.spec.ts
jbl428: It is my mistake :)

packages/http/src/__tests__/client.spec.ts
jbl428: I thought that previous mock's call history should be cleared before each `describe` started .

If second `describe` pass, the third one always pass because both  have same `test` body. 

```ts
      describe('when calling a method with overridden url', () => {
        beforeAll(() => client.get('/pet', { baseUrl: 'https://www.google.it' }));

        afterAll(() => jest.clearAllMocks());

        test('should call the mocker with the replaced full url', () => {
          expect(mock.default).toBeCalledWith({ // <-- same body
            resource: expect.anything(),
            input: expect.objectContaining({
              data: expect.objectContaining({
                url: expect.objectContaining({
                  baseUrl: 'https://www.google.it',
                  path: '/pet',
                }),
              }),
            }),
            config: expect.anything(),
          });
        });
      });

      describe('when calling a method with a full url', () => {
        beforeAll(() => client.get('https://www.google.it/pet'));

        afterAll(() => jest.clearAllMocks());

        test('should call the mocker with the replaced full url', () => {
          expect(mock.default).toBeCalledWith({  // <-- same body
            resource: expect.anything(),
            input: expect.objectContaining({
              data: expect.objectContaining({
                url: expect.objectContaining({
                  baseUrl: 'https://www.google.it',
                  path: '/pet',
                }),
              }),
            }),
            config: expect.anything(),
          });
        });
      });
    })
```

packages/http/src/client.ts
jbl428: Yes, I found that it is always string even though there are no query string in URL. 
but I think it is better to make code more defensive 😄

packages/http/src/forwarder/index.ts
jbl428: Since `partialUrl.host` is a getter not property. It is not copied properly. 

Come to think of it, it would be better to use `Object.assign` instead of spread. :)

docs/guides/01-mocking.md
```diff
@@ -146,7 +146,7 @@ Testing against the exact same piece of data over and over again is not the best
 
 Dynamic mode solves this by generating a random value for all the properties according to their type, and other information like `format` or even the all-powerful `x-faker` extension.
 
-All the random properties are generated using [Faker.js](https://github.com/marak/Faker.js) under the hood, via [JSON Schema Faker](https://github.com/json-schema-faker/json-schema-faker). The `x-faker` keyword is optional, but when present allows for a specific Faker API method to be used (of which there [are a lot](https://github.com/marak/Faker.js#api-methods)) so you get a lot of control over the response.
+All the random properties are generated using [Faker.js](https://fakerjs.dev) under the hood, via [JSON Schema Faker](https://github.com/json-schema-faker/json-schema-faker). The `x-faker` keyword is optional, but when present allows for a specific Faker API method to be used (of which there [are a lot](https://fakerjs.dev/guide)) so you get a lot of control over the response.
```
jbl428, line 149: I have changed the link to pass CI temporary.

chohmann:
The code looks good! Can you resolve the docs conflicts? We'll be good to go then!

daniel-white (approved):

brendarearden (approved):

PR #1990, opened 2022-02-16
allow for empty request bodies to skip validation, fixes #1939

### LINKED ISSUE #1939: Prism 4.6 Returning "Bad Request" when Content-Type header is set
## Describe the bug

After upgrading Prism from 4.5 to 4.6, requests that specify a content-type header return with the following error message: `[6:24:06 AM] ›     [VALIDATOR] ✖  error     Violation: request Supported content types: `

## To Reproduce

1. Given this OpenAPI document (see attached)
[swagger.txt](https://github.com/stoplightio/prism/files/7502490/swagger.txt)

2. Run Prism
3. Make a request with `Content-Type: application/json`
4. The request will return a 400

## Expected behavior

The request should return a 200

## Additional context
We're running this on Docker with the public images. Rolling back to 4.5 works as expected (and thank you for such a nice product!)

**Screenshots**

 - Library version: 4.6
 - OS: Mac OS (although Prism is running in the Docker images provided by Stoplight
 - Browser: Curl
 - Platform: 

gz#8029<br><br><i>(related to <a href='https://stoplight.zendesk.com/agent/tickets/8029'>Zendesk ticket #8029</a>)</i>

daniel-white:
Addresses #1939

**Summary**

inspects the request content-length header for making the decision to enforce the content-type on the request. if it is >1 then the content-type header must be one of the request body types.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

### DIFF
--- a/packages/http/src/validator/index.ts
+++ b/packages/http/src/validator/index.ts
@@ -16,7 +16,7 @@ import { sequenceOption, sequenceValidation } from '../combinators';
 import { pipe } from 'fp-ts/function';
 import { inRange, isMatch } from 'lodash';
 import { URI } from 'uri-template-lite';
-import { IHttpRequest, IHttpResponse } from '../types';
+import { IHttpRequest, IHttpResponse, IHttpNameValue } from '../types';
 import { findOperationResponse } from './utils/spec';
 import { validateBody, validateHeaders, validatePath, validateQuery } from './validators';
 import { NonEmptyArray } from 'fp-ts/NonEmptyArray';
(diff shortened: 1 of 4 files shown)


test-harness/specs/validate-body-params/enforce-content-type-if-body-included.oas3_1.txt
```diff
@@ -0,0 +1,26 @@
+====test====
+Given form data expected in request body
```
daniel-white, line 2: guess i should update these

packages/http/src/validator/index.ts
```diff
@@ -68,29 +68,50 @@ const tryValidateInputBody = (
...
     E.chain(() => {
+      const headersNormalized = caseless(headers || {});
+
+      const contentLength = parseInt(headersNormalized.get('content-length')) || 0;
+      if (contentLength === 0) {
+        // generously allow this content type if there isn't a body actually provided
+        return E.right(body);
+      }
+
+      const mediaType = headersNormalized.get('content-type');
```
EdVinyard, line 84: Now that I've seen how you did this, I like that it's inside `validateInputIfBodySpecIsProvided()`.

EdVinyard (approved):

packages/http/src/validator/index.ts
chohmann: samesies! 🙌  Nice work!

chohmann (approved):

rngtng:
thanks for the fix. Could it be that with this an empty PSOT request body doesn't trigger exception when a body payload is required? In my case I somehow can't get an exception triggered on missing body...!?

I quick drafted a potential test case here:
```
====test====
Given a request declaring a content-type but has no body although it is required
then return 400
====spec====
openapi: '3.1.0'
paths:
  /path:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: object
              required:
                - foo
               properties:
                 foo:
                   type: string
      responses:
        200:
          content:
            text/plain:
              example: ok
       400:
          content:
            text/plain:
              example: body required
====server====
mock -p 4010 ${document}
====command====
curl -i -X POST http://localhost:4010/path -H "Content-Type: application/json"
====expect====
HTTP/1.1 400 Bad Request
content-type: text/plain

body required

Issue #2022, opened 2022-03-27
Use node 14 as base engine version


jasonbarry:
Node 12's [end-of-life](https://nodejs.org/en/about/releases/) is approaching on 2022-04-30. We should upgrade to the next LTS version which is node 14. 

## Context
Keeping this repo up-to-date with supported node releases. 

## Current Behavior
The repo doesn't build in CI if any dependencies specify an engine of >= 14. 

## Expected Behavior
Things should work the exact same, but will require node >= 14 to run.

chohmann:
closed because we merged the PR upgrading to node 16.

PR #2023, opened 2022-03-27
Upgrading minimum node engine to 16

jasonbarry:
Addresses #2022 

**Summary**
This PR upgrades the minimum node engine version from 12 to 16, as node version 12 is [approaching end-of-life](https://nodejs.org/en/about/releases/) on 04-30-2022. Specifying the minimum version of 16 will allow us to use dependencies that require a minimum version of 16. 

**Checklist**

- The basics
  - [ ] I tested these changes manually in my local or dev environment
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Screenshots**
N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,10 +2,10 @@ version: 2.1
 executors:
   docker-node:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:14
   docker-harness:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:14
       - image: kennethreitz/httpbin
         name: httpbin.org
   macos:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,10 +2,10 @@ version: 2.1
 executors:
   docker-node:
     docker:
-      - image: circleci/node:14
+      - image: circleci/node:16
   docker-harness:
     docker:
-      - image: circleci/node:14
+      - image: circleci/node:16
       - image: kennethreitz/httpbin
         name: httpbin.org
   macos:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12 as compiler
+FROM node:14 as compiler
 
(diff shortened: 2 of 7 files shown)


daniel-white:
hey @jasonbarry thanks for this! ill take a look, but in the mean time, can you bump this to node16 - that gives us more time and apple silicon support.

jasonbarry:
@daniel-white just updated references to point to v16.

daniel-white:
hey @jasonbarry - sorry for the delay - can you rebase to pickup the actions changes?

jasonbarry:
@daniel-white sure I can do that, should I also change it so that it's a range instead of a min version? e.g. `>=12` could avoid needing to make a potentially breaking change.

daniel-white:
@jasonbarry lets keep it as a range that 14 or 16 is the minimum. i think were good for potential breaking changes as we are planning to upgrade a bunch of dependencies that may introduce breaking changes as well.

jasonbarry:
@daniel-white oh I was thinking of a different PR, this one is already set to `>=16`. Just merged in changes from latest master.

daniel-white (approved):

deefactorial:
I'm not happy with this change, it's a breaking change that didn't change the major version. this even breaks previous versions as it tries to install the package `@stoplight/prism-core` at `4.10.0` I have an issue with upgrading to node 16 with another package so my only solution is to remove this package until I can find a way to upgrade to node 16. this breaking change could have been managed differently.

daniel-white:
hey @deefactorial - i'm sorry that this was a breaking change for you. we discussed this internally and #2085 was written as a result. to summarize some of outcomes:

1. short term, we'll mark `@stoplight/prism-cli` to be explicit with its versions it depends on to avoid that another prism package would be upgraded
2. we'll be discussing and documenting our decisions on versions we support and when breaking changes will be declared (or not)

PR #2041, opened 2022-04-19
Added alphabetical ordering to generated objects 

stoplightAndrew:
Ordering by the original JSON schema would be optimal but this seems better than random?

Addresses #2038 

**Summary**

Json parameters are returned in (random?) order from dynamic mocking.  Ugly at best, hard to find what you are looking for.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

**Screenshots**

image
image

**Additional context**

Please overly scrutinize this, but also be gentle ;) -  I'd love to be helpful but don't want my help to be detrimental.  And I'm pretty new to TypeScript generally.

Obvious concerns are recursion, typing, unintended consequences - I don't know the project context.

It also seems like https://github.com/json-schema-faker/json-schema-faker might be the place to really address some of this?

### DIFF
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -33,7 +33,7 @@ jsf.option({
 });
 
 export function generate(bundle: unknown, source: JSONSchema): Either<Error, unknown> {
-  var test = pipe(
+  return pipe(
     stripWriteOnlyProperties(source),
     E.fromOption(() => Error('Cannot strip writeOnly properties')),
     E.chain(updatedSource =>
@@ -42,8 +42,6 @@ export function generate(bundle: unknown, source: JSONSchema): Either<Error, unk
       tryCatch(() => sortSchemaAlphabetically(jsf.generate({ ...cloneDeep(updatedSource), __bundled__: bundle })), toError)
     )
   );
-
-  return test
 }
 
 //sort alphabetically by keys
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -44,14 +44,14 @@ export function generate(bundle: unknown, source: JSONSchema): Either<Error, unk
 
 //sort alphabetically by keys
 export function sortSchemaAlphabetically(source: any): any {
-  if (Array.isArray(source)) {
+  if (source && Array.isArray(source)) {
     for (const i of source) {
       if (typeof source[i] === 'object') {
         source[i] = sortSchemaAlphabetically(source[i]);
       }
     }
     return source;
-  } else if (typeof source === 'object') {
+  } else if (source && typeof source === 'object') {
     Object.keys(source).forEach((key: string) => {
       if (typeof source[key] === 'object') {
         source[key] = sortSchemaAlphabetically(source[key]);
--- a/packages/http/src/mocker/generator/JSONSchema.ts
+++ b/packages/http/src/mocker/generator/JSONSchema.ts
@@ -1,5 +1,5 @@
 import * as faker from 'faker';
-import { cloneDeep } from 'lodash';
+import { cloneDeep, forEach, isString, keys, values } from 'lodash';
 import { JSONSchema } from '../../types';
 
 import * as jsf from 'json-schema-faker';
@@ -9,6 +9,11 @@ import { IHttpOperation } from '@stoplight/types';
 import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/lib/Either';
 import { stripWriteOnlyProperties } from '../../utils/filterRequiredProperties';
+import { JSONSchema7Type } from 'json-schema';
+import { right } from 'fp-ts/lib/EitherT';
+import { sort } from 'fp-ts/lib/ReadonlyNonEmptyArray';
+import { isFalsy } from 'utility-types';
+// import obj from 'http/src/validator/validators';
 
 // necessary as workaround broken types in json-schema-faker
 // @ts-ignore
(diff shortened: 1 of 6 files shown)


packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,6 +9,11 @@ import { IHttpOperation } from '@stoplight/types';
 import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/lib/Either';
 import { stripWriteOnlyProperties } from '../../utils/filterRequiredProperties';
+import { JSONSchema7Type } from 'json-schema';
```
chohmann, line 12: I don't see this being used anywhere. Can remove.

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -34,11 +39,33 @@ export function generate(bundle: unknown, source: JSONSchema): Either<Error, unk
...
       // necessary as workaround broken types in json-schema-faker
       // @ts-ignore
-      tryCatch(() => jsf.generate({ ...cloneDeep(updatedSource), __bundled__: bundle }), toError)
+      tryCatch(() => sortSchemaAlphabetically(jsf.generate({ ...cloneDeep(updatedSource), __bundled__: bundle })), toError)
     )
   );
 }
 
+//sort alphabetically by keys
+export function sortSchemaAlphabetically(source: any): any {
```
chohmann, line 48: Can you write some unit tests for this function?

Probably with a test case for each of the following:
1. not an object or array --> leaves source untouched
2. array of non-objects (i.e. strings) --> leaves the array untouched
3. array of objects --> sorts each object
4. simple object (i.e. each property has just a string value) --> properties are sorted and values stay with the keys
5. simple complex object (i.e. some properties are just strings, others are objects) --> top level object has properties sorted as well as sub object(s)
6. Maybe a performance-ish test of a bunch of nested objects and arrays and maybe the object has a shit ton of properties to sort?
7. other cases I haven't thought of...?

I think it would also be good to write a [test harness test](https://github.com/stoplightio/prism/tree/master/test-harness#adding-a-new-test) for this so we have an E2E test of this working.

chohmann (changes requested):
I think this looks good! Just a few unused imports to clean up.

But we do need some tests around this. I commented what I think are good things to test for, but don't treat that as a comprehensive list if you can think of others.

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -9,6 +9,11 @@ import { IHttpOperation } from '@stoplight/types';
 import { pipe } from 'fp-ts/function';
 import * as E from 'fp-ts/lib/Either';
 import { stripWriteOnlyProperties } from '../../utils/filterRequiredProperties';
+import { JSONSchema7Type } from 'json-schema';
+import { right } from 'fp-ts/lib/EitherT';
+import { sort } from 'fp-ts/lib/ReadonlyNonEmptyArray';
+import { isFalsy } from 'utility-types';
```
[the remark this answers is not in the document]
stoplightAndrew: Yup!  I think VS Code might've added things as I was messing around, will pay attention to it!

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -32,13 +32,38 @@ export function generate(bundle: unknown, source: JSONSchema): Either<Error, unk
...
+        toError
+      )
     )
   );
 }
 
+//sort alphabetically by keys
+export function sortSchemaAlphabetically(source: any): any {
+  if (Array.isArray(source)) {
+    for (const i of source) {
```
daniel-white, line 48: a traditional `for` loop would be easier and more defensive

test-harness/index.ts
```diff
@@ -83,6 +83,10 @@ describe('harness', () => {
         }
         if (parsed.expect) {
           expect(output.body).toStrictEqual(expected.body);
+        } else if (parsed.expectKeysOnly){
+          const jsonOutput = JSON.parse(output.body);
+          const jsonExpected = JSON.parse(expected.body);
+          expect(Object.keys(jsonOutput)).toStrictEqual(Object.keys(jsonExpected));
```
daniel-white, line 89: interesting that this works with iterators

daniel-white (changes requested):

test-harness/index.ts
chohmann: it doesn't. This is meant only to work for json objects. Can probably update the readme to make that clear.

packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -32,13 +32,38 @@ export function generate(bundle: unknown, source: JSONSchema): Either<Error, unk
...
+    }
+    return source;
+  } else if (typeof source === 'object') {
+    Object.keys(source).forEach((key: string) => {
+      if (typeof source[key] === 'object') {
+        source[key] = sortSchemaAlphabetically(source[key]);
+      }
+    });
+    return Object.fromEntries(Object.entries(source).sort());
+  }
```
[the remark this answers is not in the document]
chohmann: fixed. [0de8272](https://github.com/stoplightio/prism/pull/2041/commits/0de82721da89e3601cf0de688fcb5a97e570acdb)

test-harness/index.ts
chohmann: [ac63648](https://github.com/stoplightio/prism/pull/2041/commits/ac6364871b702aaf0f11ddd3915f251629334804)

packages/http/src/mocker/generator/JSONSchema.ts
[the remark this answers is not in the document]
chohmann: enter a whole debate on why it's there (because the spec says son) and why people believe that's a mistake: https://stackoverflow.com/questions/18808226/why-is-typeof-null-object 😂

daniel-white (approved):

Issue #2010, opened 2022-03-11
Address 


chohmann:
- URIJS --> 1.19.10
- node-fetch --> 2.6.7
- glob-parent --> 5.1.2

- marked (through postman collection) --> 4.0.10 own ticket

- tar --> node-gyp

chohmann:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/-Team-Tacocat-5eb9bc4a1fddfe30b94f78b1/issues/stoplightio/prism/2010?planning-poker) with ZenHub @EdVinyard @daniel-white @brendarearden

PR #2075, opened 2022-06-22
update deps

chohmann:
Addresses #2010 

**Summary**

Bump dependencies to fix critical + high security vulnerabilities.

There was only one high security vulnerability that we can't fix completely and that's the [ansi-regex](https://github.com/stoplightio/prism/security/dependabot/31) one. I was able to at least get off of the really only version that caused [this one](https://github.com/stoplightio/prism/security/dependabot/32). But some of our deps that use 5.0.0 are on their latest version so they haven't updated to not use 5.0.0 version of ansi-regex yet.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A


package.json
```diff
@@ -54,11 +54,11 @@
     "http-string-parser": "^0.0.6",
     "husky": "^4.3.8",
     "jest": "^27.2.5",
-    "lerna": "^4.0.0",
+    "lerna": "5.1.4",
```
chohmann, line 57: this was the only change that scared me a little since it's a big jump. But I think we only use lerna for releasing, so not sure how to test this prior to trying to release a new version of Prism 😬

daniel-white: looks like we wouldn't be affected: https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes

daniel-white (approved):

package.json
chohmann: nice! Hadn't gotten around to looking at the docs for this just yet. That's promising!

Issue #2081, opened 2022-06-27
Self referencing schema causing an error


brian-geoghegan:
This particular piece of self referencing and reflects the actual API response.
## Context

`    content:
      type: object
      properties:
        text:
          type: string
          description: Text content
        type:
          type: string
          description: Content type (e.g. text, link, image)
        metadata:
          type: object
          description: Content metadata (e.g. href, src, alt)
        children:
          type: array
          description: sub-content
          items:
            $ref: '#/components/schemas/content'
`

## Current Behavior

`{
    "type": "https://stoplight.io/prism/errors#UNKNOWN",
    "title": "Error: Prop not found: paths (#/paths/~1coreapi~1recent-news~1v1~1search/get/responses/200/content/application~1json; charset=utf-8/schema/allOf/1/properties/articles/items/properties/body/items)\n    at Object.getLocalRef (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:932:17)\n    at recursiveUtil.resolveSchema (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2252:25)\n    at traverse (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2079:16)\n    at Object.arrayType [as array] (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:1567:25)\n    at traverse (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2155:44)\n    at /home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2188:52\n    at Array.forEach (<anonymous>)\n    at traverse (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2184:27)\n    at Object.objectType [as object] (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:1851:14)\n    at traverse (/home/node/app/node_modules/json-schema-faker/dist/main.cjs.js:2155:44) in /properties/articles/items/0/properties/body/items/0/properties/children",
    "status": 500,
    "detail": ""
}`

## Expected Behavior

Expect prism to create the mock request and response here

## Steps to Reproduce

1. Create a self referencing component
2. Witness the bug

## Environment

- Version used:
- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your environment/workspace/project:

daniel-white:
[...]

PR #2086, opened 2022-07-06
adds tests to validate #2081 recursive thru an array

daniel-white:
Adds tests for #2081 to cover the broken case for the OP. i was unable to reproduce locally.


brendarearden (approved):
looks good!

Issue #2089, opened 2022-07-13
Configure Dependabot to create it's own PRs


ryotrellim:
See: https://github.com/stoplightio/prism/security/dependabot
Change this: https://github.com/stoplightio/prism/blob/master/.github/dependabot.yml
Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

PR #2095, opened 2022-08-03
resetting sensible defaults for dependabot

daniel-white:
Addresses #2089

from what i can tell our config was too strict and was forcing to use a different NPM registry. since this is an open source package, the public one should be fine.


fcasad (approved):
Not sure how to verify this locally, so merge away!

Issue #2102, opened 2022-08-04
Prism 4.10 Returning "Internal Server Error" when Content-Type header is set but body is empty/null


acolombier:
_This issue seems related to #1939_

## Context

I'm using Prism as a validator proxy under AppSync HTTP VTL (AWS services), which must pass a `Content-Type` (and a `Content-Lenght` computed from the body length) even when using `GET`.

## Current Behavior

For the following request:

```
curl --request GET \
  --url http://localhost:4010/my/endpoint \
  --header 'Content-Length: 0' \
  --header 'Content-Type: text/plain'
```
You would get this error:

```
< HTTP/1.1 500 Internal Server Error
...
{
  "type": "TypeError",
  "title": "Request with GET/HEAD method cannot have body",
  "status": 500,
  "detail": ""
}
```
(_note that there is similar result when using `application/json`_)

This impacts both the last image (`stoplight/prism:4` and also `master`)

## Expected Behavior

The request should be forward `as-is` , without any validation error. Potentially, an `sl-violations` should flag unexpected headers if the `GET` endpoint spec doesn't specify acceptance for this `Content-Type (this is arguable since the content is explicitly said to be null there)

## Possible Workaround/Solution

(The issue seems to come from [fetch](https://github.com/stoplightio/prism/blob/9eac3340f097c7ee289106df0d17cf3968f6b57a/packages/http/src/forwarder/index.ts#L65), which denies bodies to be passed for GET request and seem to suffer from a ~drama~disagreement from contributor to remove the limitation)

I was able to fix the problem by applyinmg the following patch

```
diff --git a/packages/http-server/src/server.ts b/packages/http-server/src/server.ts
index 5339e428..e0f4ba87 100644
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -40,9 +40,11 @@ function addressInfoToString(addressInfo: AddressInfo | string | null) {
 function parseRequestBody(request: IncomingMessage) {
   // if no body provided then return null instead of empty string
   if (
-    request.headers['content-type'] === undefined &&
-    request.headers['transfer-encoding'] === undefined &&
-    (request.headers['content-length'] === '0' || request.headers['content-length'] === undefined)
+    (request.headers['content-type'] === undefined &&
+      request.headers['transfer-encoding'] === undefined &&
+      request.headers['content-length'] === undefined) ||
+    // If the body size is null, it means the body itself is null so the promise can resolve with a null value
[...]

PR #2103, opened 2022-08-06
fix(http-server): discard request body if the content-length header i…

acolombier:
Addresses #2102

**Summary**

This PR changes the condition of the HTTP server resolving the request body. It assumes that the body is null in case of an explicit `Content-Length: 0`.

Since in #1990 the decision was made to assume no body in case of no explicit `Content-Length` other than zero (omitting will assume no body - which seems matching [the HTTP/1.1 spec](https://httpwg.org/specs/rfc7230.html#header.content-length)), however I didn't follow the same approach as it could be seen as a minor changes (as apposite to a patch).

Let me know if you would like me to default it to 0 in case if missing as well!

**Checklist**

- The basics
  - [X] I tested these changes manually in my local or dev environment (Running + Docker)
- Tests
  - [X] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [X] I reported errors and followed the error reporting guidelines
  - [ ] N/A

**Screenshots**

_Before_
image

_After_
image

**Additional context**

N/A


.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
acolombier, line 1: I have taken the freedom to add a [NVM](https://github.com/nvm-sh/nvm) RC file. Let me know if you would like this out

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -716,16 +731,9 @@ describe('body params validation', () => {
...
-              severity: 'Error',
-              code: 'required',
-              message: "must have required property 'id'",
-            },
-            {
-              location: ['body'],
               severity: 'Error',
               code: 'required',
-              message: "must have required property 'status'",
+              message: 'Body parameter is required',
```
chohmann, line 736: It might be better to update this test to have one of the required properties (either id or status) so that the spirit of the test is still the same: throws error if there's a missing required property in the body. Otherwise this test now changes to testing the behavior if the body is missing completely.

acolombier: I believe there is a problem:
- The body used to be optional (thus change at line 689)
- There [`body required` check](https://github.com/acolombier/prism/blob/03b46ff021b84e17d5f19844a8385b2395b7d5b7/packages/http/src/validator/index.ts#L74) and the [content assertion](https://github.com/acolombier/prism/blob/03b46ff021b84e17d5f19844a8385b2395b7d5b7/packages/http/src/validator/index.ts#L110) happens at two different places

My understanding of the `describe`/`it` is that this unit test might not have been covering the part it says it does.

I have made a change which I think should reflect a bit better the usecase this test was meant to test. The [`body required` check](https://github.com/acolombier/prism/blob/03b46ff021b84e17d5f19844a8385b2395b7d5b7/packages/http/src/validator/index.ts#L74) seems to already be covered by [this test](https://github.com/acolombier/prism/blob/a8c787f893109a966b290d3ab2c8fe909fdfa5b2/packages/http-server/src/__tests__/body-params-validation.spec.ts#L569), but happy to add an other one if you think it should have thorough testing.

chohmann (approved):
Thanks for updating the test! This looks good to me. Thank you for your contribution!

chohmann:
This has been released in version 4.10.2 🎉

childish-sambino:
@chohmann This looks to be causing a breakage where requests are accepted when no body is provided, but the body is defined as having one or more required parameters.

PR #2126, opened 2022-08-26
Add the ability to disable sl-violations header

thewoods96:
**Summary**
When parsing CLI output in proxy mode to validate specifications the sl-violations header may not be required. In some cases where there is a significant delta between the specification and response this header can exceed size limitations and cause failures.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

### DIFF
--- a/packages/cli/src/commands/proxy.ts
+++ b/packages/cli/src/commands/proxy.ts
@@ -36,6 +36,12 @@ const proxyCommand: CommandModule = {
             'If an http proxy is required to reach upstream, formatted as "{protocol}://[{user}[:{password}]@]{host}[:{port}]". eg "http://myUser:myPassword@proxy.example.com:1234"',
           string: true,
         },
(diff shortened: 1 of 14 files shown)


EdVinyard:
@thewoods96 , we really appreciate you bringing this problem to our attention, and also submitting a possible fix.  

After talking with the Prism team, we'd like to solve this problem in a way that doesn't require an additional command line option.  That is, we'd like Prism to behave in a way that won't break proxies and clients automatically.  To that end, I opened https://github.com/stoplightio/prism/issues/2129.  If you're so inclined, feel free to submit a PR in response to that Issue.  If not, we'll work on getting it prioritized ourselves.

Again, thank you; we didn't even know this problem existed.

Issue #2138, opened 2022-09-15
mock fails with `TypeError: Cannot read properties of undefined (reading 'isMaster')`


mitsos1os:
## Context
You cannot start `mock` server command

## Current Behavior

Running `mock` command on the prim cli using either directly NodeJS source code or Dockerfile throws the following error:
```
TypeError: Cannot read properties of undefined (reading 'isMaster')
  at createMultiProcessPrism (/usr/src/prism/packages/cli/dist/util/createServer.js:27:32)
  at Object.runPrismAndSetupWatcher (/usr/src/prism/packages/cli/dist/util/runner.js:8:12)
  at Object.handler (/usr/src/prism/packages/cli/dist/commands/mock.js:27:25)
  at Object.runCommand (/usr/src/prism/node_modules/yargs/build/index.cjs:446:48)
  at Object.parseArgs [as _parseArgs] (/usr/src/prism/node_modules/yargs/build/index.cjs:2697:57)
  at Object.get [as argv] (/usr/src/prism/node_modules/yargs/build/index.cjs:2651:25)
  at Object.<anonymous> (/usr/src/prism/packages/cli/dist/index.js:15:26)
  at Module._compile (node:internal/modules/cjs/loader:1126:14)
  at Object.Module._extensions..js (node:internal/modules/cjs/loader:1180:10)
  at Module.load (node:internal/modules/cjs/loader:1004:32)
```

## Expected Behavior

I would expect the command to work successfully

## Possible Workaround/Solution

The problem seems to be that when compiling **prism** using `yarn build` command, the output of the file `packages/cli/dist/util/createServer.js` regarding the `node:cluster` module seems to be wrong. The related lines are:
**L7:** `const node_cluster_1 = require("node:cluster");`
**L27:** `if (node_cluster_1.default.isMaster) {`

**CAUSE**
The `default` attribute on `node_cluster_1` is undefined leading to the `TypeError` mentioned above. As far as I know `.default` was never an attribute of the node **cluster** module. So I guess this is due to some compilation or types error of the module

## Steps to Reproduce

1. `yarn && yarn build`
2. `node . mock <api-spec-file>`

Exactly the same with Dockerfile using latest `stoplight/prism` image

## Environment

- Version used: **4.latest** as of today 15/9/2022
- Environment name and version (e.g. Chrome 39, node.js 5.4): NodeJS 16.17 and latest docker image `stoplight/prism:4`

alexisjanvier:
[...]

PR #2143, opened 2022-09-30
2138: Fix mock command regression

fcasad:
Addresses #2138

**Summary**

This fixes the mock command regression introduced in 4.10.4 by reverting [this commit ](https://github.com/stoplightio/prism/pull/2119/commits/f6c4e9f6929f65f40a7c8d1f35dc0c3ca691b80c).  The reason these breaking code changes were made was partly due to an issue with the type definitions for the core node cluster module.  See [this issue](https://github.com/stoplightio/prism/pull/2119/commits/f6c4e9f6929f65f40a7c8d1f35dc0c3ca691b80c).  I've added an .eslintignore as well because the lint errors were distracting when scanning the build output.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Additional context**

Node 16.17 is in LTS as of today, and I'm not sure we should bump to 18 yet.  @types/node generally should line up with the version of node we are on.  However due to the issue linked above I've left those @13.1 as before.

### DIFF
--- a/packages/http/src/__tests__/memory-leak-prevention.spec.ts
+++ b/packages/http/src/__tests__/memory-leak-prevention.spec.ts
@@ -33,14 +33,12 @@ describe('Checks if memory leaks', () => {
 
     for (let i = 0; i < 5000; i++) {
       round(client);
-      if (i % 100 === 0 && global.gc) {
+      if (i % 100 === 0) {
         global.gc();
       }
     }
 
-    if (global.gc) {
-      global.gc();
-    }
+    global.gc();
     expect(process.memoryUsage().heapUsed).toBeLessThanOrEqual(baseMemoryUsage * 1.03);
   });
 });
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+node_modules
+dist
+logs
+coverage
+cli-binaries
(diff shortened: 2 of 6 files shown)


packages/http/src/__tests__/memory-leak-prevention.spec.ts
```diff
@@ -33,14 +33,12 @@ describe('Checks if memory leaks', () => {
 
     for (let i = 0; i < 5000; i++) {
       round(client);
-      if (i % 100 === 0 && global.gc) {
+      if (i % 100 === 0) {
```
EdVinyard, line 36: Help me out: I guess this test will fail if `global.gc()` doesn't exist, but why was it optional before?  (This might be a good place for a comment.)

EdVinyard (approved):
The reversions look reasonable to me, but I don't understand the garbage collection change in the test.  I'll go ahead and approve it, since I guess this is causing problems, but if you can help me understand the GC change, I'd appreciate it.

packages/http/src/__tests__/memory-leak-prevention.spec.ts
fcasad: Apparently that was another change made to satisfy the @types/node@18 typings.

`global.gc` could not be undefined in @types/node@13 but it can in v18

PR #2167, opened 2022-11-02
fix #1895

### LINKED ISSUE #1895: Respect `Prefer` header with validation proxy when server returns 501
## Summary

I was able to use `Prefer` header to control returned mock response when using `mock-server` (`Prefer: example=example-key`). Then I switched to `validation-proxy` , hard-coded my upstream sever to return 501 for all requests and send a request to the `validation-proxy` with `Prefer` header as I did with `mock-server`. Surprisingly, it didn't work. I always returned the first example.

Is this how the `validation-proxy` is supposed to work?

## Additional context

Prism: 4.4.0 (get via docker)

nursanamar:
Addresses #1895

**Summary**

**Checklist**

- The basics
- Tests
  - [X] N/A

### DIFF
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -6,8 +6,10 @@ import { pipe } from 'fp-ts/function';
 import { defaults } from 'lodash';
 import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, IPrismProxyConfig, IPrismOutput } from './types';
 import { getSemigroup, NonEmptyArray } from 'fp-ts/NonEmptyArray';
-import { DiagnosticSeverity } from '@stoplight/types';
+import { DiagnosticSeverity, IHttpRequest } from '@stoplight/types';
+import { IHttpNameValue } from "@stoplight/prism-http/src/types";
 import { identity } from 'fp-ts/function';
+import { getHttpConfigFromRequest } from '@stoplight/prism-http-server/src/getHttpConfigFromRequest';
 
 const eitherSequence = A.sequence(E.getApplicativeValidation(getSemigroup<IPrismDiagnostic>()));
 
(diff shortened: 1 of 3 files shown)


packages/core/src/factory.ts
```diff
@@ -62,11 +63,23 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
     config: Config,
     validations: IPrismDiagnostic[]
   ): TE.TaskEither<Error, ResourceAndValidation & { output: Output }> => {
+    
+    const input = {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
daniel-white, line 68: what typescript errors are these hiding? is there another way to express what you need this to?

packages/http-server/src/getHttpConfigFromRequest.ts
```diff
@@ -21,13 +21,21 @@ const PreferencesDecoder = D.partial({
 
 type RequestPreferences = Partial<Omit<IHttpOperationConfig, 'mediaType'>>;
 
-export const getHttpConfigFromRequest = (
+export const getRequestPrefrences = (
```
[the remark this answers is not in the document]
daniel-white: also what is this doing other than adding an extra function?

nursanamar:
@daniel-white thank you for the reviews, about the test I don't really sure how to test this fix.

nursanamar:
Add some tests

packages/core/src/factory.ts
```diff
@@ -6,8 +6,10 @@ import { pipe } from 'fp-ts/function';
 import { defaults } from 'lodash';
 import { IPrism, IPrismComponents, IPrismConfig, IPrismDiagnostic, IPrismProxyConfig, IPrismOutput } from './types';
 import { getSemigroup, NonEmptyArray } from 'fp-ts/NonEmptyArray';
-import { DiagnosticSeverity } from '@stoplight/types';
+import { DiagnosticSeverity, IHttpRequest } from '@stoplight/types';
+import { IHttpNameValue } from "@stoplight/prism-http/src/types";
 import { identity } from 'fp-ts/function';
+import { getHttpConfigFromRequest } from '@stoplight/prism-http-server/src/getHttpConfigFromRequest';
```
daniel-white, line 12: we do not want a circular dependency between core and http. could we change the call sites to `mockOrForward` to include calculation of the mocking config?

packages/core/src/factory.ts
```diff
@@ -62,12 +64,30 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
     config: Config,
     validations: IPrismDiagnostic[]
   ): TE.TaskEither<Error, ResourceAndValidation & { output: Output }> => {
-    const mockCall = () =>
-      components.mock({
+    const mockCall = () => {
```
daniel-white, line 67: in the event `mockOrForward`'s call sites include the mock config, this should be all reverted

packages/core/src/factory.ts
```diff
@@ -62,12 +64,30 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
...
+
+      // Construct input
+      const input = {
+        url: {
```
daniel-white, line 71: this also should support the `__prefer` query string options as well.

daniel-white (changes requested):

PR #2183, opened 2022-11-30
Replaced old image with Mermaid

pamgoodrich:
Created decision tree using Mermaid to increase readability.


chohmann (approved):
love it!

heitortsergent (approved):

PR #2208, opened 2023-01-20
Expose the http server on prism server creation

MarianU:
With this PR I am exposing the underlying micri http server. The goal and intention here is to have at our disposal all events from the server itself to be able to hook into **request** , **response** or any other event to be able to fine tune the behavoir of the server.

**Checklist**

- The basics
- Tests
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Additional context**

Main ask was to be able to change the response (headers and body) based on some received headers or some specific data set in the body of the request. This is not only to be used as some extra specific validators of data but also as a error injection mechanism to conduct fine grained testing.

### DIFF
--- a/packages/http-server/src/server.ts
+++ b/packages/http-server/src/server.ts
@@ -209,6 +209,10 @@ export const createServer = (operations: IHttpOperation[], opts: IPrismHttpServe
       return components.logger;
     },
 
+    get server() {
+      return server;
+    },
+
     close() {
       return new Promise((resolve, reject) =>
         server.close(error => {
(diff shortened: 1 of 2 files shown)


chohmann:
@MarianU we are going to close this PR as we don't feel customer example chooser logic belongs in Prism. We've crafted an example of how we recommend you achieve custom response chooser logic:

https://github.com/stoplightio/ExampleChooserPrismProxy

Please follow along with that example, and let us know if that doesn't work for you for some reason.

PR #2214, opened 2023-01-26
chore(docs): restore SVG images

wmhilton:
Part of https://github.com/stoplightio/platform-internal/issues/14766


chohmann:
@pamgoodrich can you take a look at this? I think you recently removed the demo part of this so probably don't want it added back right? I'll let you review instead of me.

pamgoodrich (approved):

wmhilton:
@chohmann Pam removed it because SVG images were broken, but now that it's fixed we're adding it back.

chohmann:
ahhh ok sounds good. Thanks @wmhilton !

Issue #773, opened 2019-11-06
Verbose logging


karol-maciaszek:
**User story.**
As a Prism CLI user, I can supply --verbose (or --debug) CLI option, so that I can see more verbose logs.

**Is your feature request related to a problem?**
There are occasions when I do not have easy access to request or response payload. For those cases, it would be helpful if I could see the actual request and response payloads together with headers. That should help me debug possible problems with spec or request data.

**Describe the solution you'd like**
A `-d|--debug` or `--verbose` CLI option which enables the above functionality.

**Additional context**
This could be also super helpful in debugging callbacks, which are performing asynchronous request based on previous req/res pair.

XVincentX:
Aha! I can understand the frustration while debugging the thing.

I think we have two steps here to do:

1. Logging the things that we need
2. Forward the filtering capabilities to Pino

Right now we have already several log levels that come bundled with [pino](https://getpino.io/#/docs/api?id=logger)

> trace, debug, info, warn, error, and fatal.

On top of these, we add some other that are used just for beautifying purposes:

> note, success, start

I've prepared #780 to clean up some stuff that effectively does not make much sense anymore and inject the custom logging logic in the CLI instead of keeping them in the main package.

Once such PR is merged we can log the stuff you think it's important, review the levels and then the filtering capability.

philsturgeon:
In progress https://github.com/stoplightio/prism/pull/801

TylerNielsen:
Hey all, wanted to check to see if this feature is still in progress? It looks like it had some momentum during the winter but fell off. Is this still planned? Would love to be able to use this.

XVincentX:
Yeah we have paused working on this for two reasons:

1. We are busy on doing something else and this feature is not exactly that critical
2. We are generally rethinking a little bit the way we do logging and it is kind of probably that we are going to change the way we output messages (specifically speaking, changing the abstraction from `ReaderTaskEither` to a `Writer<TaskEither>>` — but we still need to check out if it's worth doing and makes sense for real.

Logging correctly is hard — that's for sure.

philsturgeon:
@XVincentX how's this one come along? Is it going to happen or should we close it down?

XVincentX:
I do not think this is gonna happen anytime soon, but the content is still valid.

philsturgeon:
@ryotrellim hey would you like to close this or prioritise it? It’s been around a while but could be helpful.

rattrayalex:
[...]

PR #2231, opened 2023-03-05
Feature: Prism CLI verbose logging options

ilanashapiro:
Addresses #773

**Summary**
Prism CLI now has the "--verboseLevel" or "--v" (alias) flag which turns on request and response logging for mocking, proxy and callbacks. This is based on the never-completed PR #801. I incorporated the comments from there requesting that the client be able to specify the log level in the request (i.e. --verbose 'trace', --verbose 'info', etc all the way to --verbose 'silent'). The log level specified in the request must be chosen from Pino's log levels ("fatal" | "error" | "warn" | "info" | "debug" | "trace"). Custom log levels like "start" and "success"  that already exist in the codebase are now set in between the severity ranks of Pino's "info" and "warn" levels. Thus, they will log whenever "info" or any lower level severity rank is indicated. "info" is also the default log level if the client doesn't override it in the request. The logging for 'trace' isn't yet implemented (i.e. there's no further logging than there would be for the next level down of 'debug') -- I wasn't totally sure of the kinds of things you'd like logged at the 'trace' level and it would be great if I could discuss ideas with a member of the team before implementing the trace level logs!

NOTE: originally, custom log levels were set to the same log severity level as Pino's "info" level. However, when I did this all such levels ("info" and all custom levels) would end up all logging as whatever custom log name last overrode the severity rank value. For instance, if we had "info", "success", and "start" all set to 30, all logs at this level would show up as "start." I originally approached it this way so that we support an arbitrary number of custom log levels showing up when the debug level is "info", rather than limiting it to 9 (i.e. levels at 31-39 until we hit "warn" at 40). I did some digging and apparently this is a known bug in Pino: https://github.com/pinojs/pino/issues/222#issuecomment-472212376. Therefore, at the time being I am keeping custom logs at their own level with the knowledge that limited custom levels can be supported with the current scheme.

**Checklist**

- The basics
- Tests
  - [X] Added or updated 
  - [ ] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A

**Screenshots**
The response for each of these is {"type":"https://stoplight.io/prism/errors#NO_PATH_MATCHED_ERROR","title":"Route not resolved, no path matched","status":404,"detail":"The route /pets hasn't been found in the specification file"}
Different levels of verbose logging are demonstrated below

image

image

image

The response here is [{"name":"string","tag":"string","id":-9223372036854776000}], and different levels of verbose logging are demonstrated below (this one has debug logs unlike the previous)

image

image

The following example demonstrates the debug logs for a validation error (here, a security error).

image


ilanashapiro:
@rattrayalex How does this look? Were you looking for trace level logging as well?

brendarearden (approved):
I pulled the changes locally and tested them out - everything is looking great! I have opened a separate issue to tackle trace logs - that will allow the team to discuss where those should go and keep this PR focused on the original issue.

ilanashapiro:
@brendarearden @rattrayalex Please see PR #2246 for a bug fix due to a bug I caught in this PR just after it was merged

PR #2246, opened 2023-03-17
Remove conflicting log levels due to Pino bug

ilanashapiro:
Addresses bug introduced in PR #2231

**Summary**

There is a bug in my recently accepted pull request #2331 for verbose logging options. In PR #2331, I set custom log levels to the same log severity level as Pino's "info" level. However, when I did this all such levels ("info" and all custom levels) would end up all logging as whatever custom log name last overrode the severity rank value. For instance, if we had "info", "success", and "start" all set to 30, all logs at this level would show up as "start." I originally approached it this way so that we support an arbitrary number of custom log levels showing up when the debug level is "info", rather than limiting it to 9 (i.e. levels at 31-39 until we hit "warn" at 40). I did some digging and apparently this is a known bug in Pino: https://github.com/pinojs/pino/issues/222#issuecomment-472212376. Therefore, to deal with this Pino bug and maintain the correct behavior of the logger, I am submitting this PR to keep custom logs at their own level with the knowledge that limited custom levels can be supported with the current scheme.

I also added to the verbose commands tests to ensure that 'info' is being used as the default log level.

**Checklist**

- The basics
- Tests
  - [X] Added or updated
  - [ ] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A

**Screenshots**

Before bugfix:

image

Note how all info-level log labels say "success" when only some should

After bugfix:

image

### DIFF
--- a/packages/cli/src/commands/__tests__/commands.spec.ts
+++ b/packages/cli/src/commands/__tests__/commands.spec.ts
@@ -31,6 +31,7 @@ describe.each<{ 0: string; 1: string; 2: unknown }>([
         document: '/path/to',
         multiprocess: false,
         errors: false,
+        verboseLevel: 'info',
       })
     );
   });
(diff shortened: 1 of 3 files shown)


chohmann (approved):
Looks good!

PR #2252, opened 2023-03-27
Don't parse response body as json on 204

skarimo:
Addresses ----

**Summary**

Currently, the response is parsed as json when content-type of `application/json`.  However in the case of `204` responses, this should not happen as the body would be empty and will raise an exception.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Additional context**

We recently ran into an issue where an endpoint returns a response with header `Content-type: application/json` on a `204 No Content` status code. This was raising an exception as the body was getting parsed as json. As it seems the behavior is to read body as text on all non `application/json` response types, I think it should be safe to do the same for 204 responses as well since a valid json body is not expected.


chohmann (approved):
Thank you for your contribution!

Issue #2255, opened 2023-04-04
Prism should not error due to extra content-type header


ryotrellim:
## Context
SPS Commerce has reached out to us with this ask:
"We are the process of rolling out Prism in proxy mode to validate our APIs specs in services, in logging mode to evaluate traffic.
That is working quite well, except one MAJOR issue related to the stoplight platform functionality for “Send API Requests”.

Specifically, with the Prism Proxy in place with a command such as:
`prism proxy design/v4.oas.yml http://localhost:5000/`

If I then make a curl request to the service (notice I’ve included content-type for this GET request):

```
curl --request GET \
 --url '[http://localhost:4010/examples`](http://localhost:4010/examples) --header 'Content-Type: application/json'
```

This will fail on the Prism side as follows:
```
{
    "type": "TypeError",
    "title": "Request with GET/HEAD method cannot have body",
    "status": 500,
    "detail": ""
}
```

## Expected Outcome:
1. Prism ignores content-type header if no request body

shartman2003:
Got this from SPS Commerce: "Our teams are well integrated with Stoplight for sending requests to our API, and so rolling out Prism in front of it has been halted for the time being until we can fix either the Prism side or the Stoplight side."

ryotrellim:
Corresponding upstream UI issue: https://github.com/stoplightio/elements/issues/2351

travisgosselin:
Between this noted issue and the issue in stoplight elements (https://github.com/stoplightio/elements/issues/2351), it makes it pretty much impossible to use Prism alongside elements in the same API. Glad to see a priority of `high` on this issue given the impact of how this can break Prism even in log mode.

PR #2260, opened 2023-04-12
feat: adds more standard compliant request body handling

daniel-white:
Addresses #2255

**Summary**

* removes `Content-Type` as a detection mechanism for request bodies
* adds more detailed error when the proxy detects a request body in a GET/HEAD request, which `node-fetch` does not support

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

### DIFF
--- a/test-harness/specs/proxy/proxy.get-with-body.txt
+++ b/test-harness/specs/proxy/proxy.get-with-body.txt
@@ -1,0 +1,20 @@
+====test====
+Making a GET request with request body through the proxy server rejects with 501
+====spec====
+swagger: "2.0"
+paths:
+  /status/204:
+    get:
+      produces:
+        - text/plain
+      responses:
+        204:
+          description: No Content
+====server====
+proxy -p 4010 ${document} http://httpbin.org
+====command====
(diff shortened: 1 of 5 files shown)


test-harness/specs/proxy/proxy.get-with-body.txt
```diff
@@ -0,0 +1,20 @@
...
+    get:
+      produces:
+        - text/plain
+      responses:
+        204:
+          description: No Content
+====server====
+proxy -p 4010 ${document} http://httpbin.org
+====command====
+curl -i http://localhost:4010/status/204 -X GET --header 'Content-Type: application/json' --data '{}'
```
EdVinyard, line 16: I've never tried to do this before.  Does `'{}'` get serialized to a zero-length request body?

EdVinyard (approved):
Other than a single question for my own curiosity, this looks good to me.

test-harness/specs/proxy/proxy.get-with-body.txt
daniel-white: yea - its a verbatim string here - so the request body is just `{}`

PR #2292, opened 2023-05-16
Fix/1895 prefer header validation proxy

brendarearden:
Addresses #1895 

**Summary**

Added a  non required field on the BaseConfig type to ensure request specific configuration would be available to Prism Proxy.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Screenshots**

https://github.com/stoplightio/prism/assets/19455540/fb5be96b-902d-478d-b6c3-e6829b0ce87a

Screenshot 2023-05-16 at 11 57 07 AM

**Additional context**

Screenshot 2023-05-16 at 11 56 55 AM

### DIFF
--- a/packages/cli/src/commands/mock.ts
+++ b/packages/cli/src/commands/mock.ts
@@ -22,16 +22,8 @@ const mockCommand: CommandModule = {
         },
       }),
   handler: parsedArgs => {
-    const {
-      multiprocess,
-      dynamic,
-      port,
-      host,
-      cors,
-      document,
-      errors,
-      verboseLevel
-    } = (parsedArgs as unknown) as CreateMockServerOptions;
+    const { multiprocess, dynamic, port, host, cors, document, errors, verboseLevel } =
+      parsedArgs as unknown as CreateMockServerOptions;
 
     const createPrism = multiprocess ? createMultiProcessPrism : createSingleProcessPrism;
     const options = { cors, dynamic, port, host, document, multiprocess, errors, verboseLevel };
--- a/packages/core/src/factory.ts
+++ b/packages/core/src/factory.ts
@@ -66,7 +66,7 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
       components.mock({
         resource,
         input: { data, validations },
-        config: config.mock || {},
+        config: config.requestConfig || config.mock || {},
       })(components.logger.child({ name: 'NEGOTIATOR' }));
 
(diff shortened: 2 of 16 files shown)


packages/http/src/mocker/index.ts
```diff
@@ -122,7 +122,9 @@ function runCallbacks({
         pipe(
           callbacks,
           A.map(callback =>
-            runCallback({ callback, request: parseBodyIfUrlEncoded(request, resource), response })(logger.child({ name: 'CALLBACK' }))()
+            runCallback({ callback, request: parseBodyIfUrlEncoded(request, resource), response })(
```
brendarearden, line 125: prettier changes

packages/core/src/factory.ts
```diff
@@ -66,7 +66,7 @@ export function factory<Resource, Input, Output, Config extends IPrismConfig>(
       components.mock({
         resource,
         input: { data, validations },
-        config: config.mock || {},
+        config: config.requestConfig || config.mock || {},
```
daniel-white, line 69: should we blend these objects together?

brendarearden: I went ahead and changed it so that proxy and mock both use the config.mock. Since we were using mock: false to determine if its a proxy or not, I went ahead and introduced isProxy to take its place.

packages/cli/src/commands/mock.ts
```diff
@@ -22,16 +22,8 @@ const mockCommand: CommandModule = {
...
       }),
   handler: parsedArgs => {
-    const {
-      multiprocess,
-      dynamic,
-      port,
-      host,
-      cors,
-      document,
-      errors,
```
brendarearden, line 32: prettier changes

packages/core/src/types.ts
```diff
@@ -20,14 +20,15 @@ type IPrismBaseConfig = {
...
+  isProxy: boolean;
 };
 
 export type IPrismMockConfig = IPrismBaseConfig & {
   mock: object;
 };
 
 export type IPrismProxyConfig = IPrismBaseConfig & {
-  mock: false;
+  mock: object;
```
daniel-white, line 31: any more specific type here?

brendarearden: during our pair review / refactoring session, we set this to unknown

daniel-white (approved):

Issue #2008, opened 2022-03-11
'multipart/form-data' will always return a validation error


SeMuell:
## Describe the bug

The validation always fails whenever a `multipart/form-data` content type is used in the requestBody. This prevents the normal usage of prism without any other proxy server (like nginx) on which such requests can be filtered out.

## To Reproduce

1. Given this OpenAPI document (stored as testapi.yaml) 
```yaml 
openapi: '3.1.0'
paths:
  /path:
    post:
      responses:
        200:
          content:
            text/plain:
              example: ok
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                testData:
                  type: string
              required:
                - "testData"
```
3. Run this CLI command `.\prism-cli-win.exe mock -h 127.0.0.1 .\testapi.yaml`
4. Write a non-empty post request to `http://127.0.0.1/path` with included form-data in the request body
5. See response error 
```json
 {
    "type": "https://stoplight.io/prism/errors#UNPROCESSABLE_ENTITY",
    "title": "Invalid request",
    "status": 422,
    "detail": "Your request is not valid and no HTTP validation response was found in the spec, so Prism is generating this error for you.",
    "validation": [
        {
            "location": [
                "body"
            ],
            "severity": "Error",
            "code": "type",
            "message": "must be object"
        }
    ]
}
```

## Expected behavior

It is expected that the response defined in the example is returned.

## Additional context
As one of the last PR #1990 the validation is skipped when an empty request is send. As a simple workaround this could also be implemented for the case multipart/form-data is used.

daniel-white:
thanks @SeMuell. i'm looking into this - i notice we have several feature requests closed #432 #1012 - i'll leave this open as we discuss internally.

yuki0410-dev:
Our project also needs `multipart/form-data` support because we use `multipart/form-data` to upload files.
How about responding in the following ways?

```ts
packages/http-server/src/server.ts
//~~~

import * as formidable from 'formidable';

//~~~

function parseRequestBody(request: IncomingMessage) {
  // if no body provided then return null instead of empty string
  if (
    // If the body size is null, it means the body itself is null so the promise can resolve with a null value
    request.headers['content-length'] === '0' ||
    (request.headers['content-type'] === undefined &&
      request.headers['transfer-encoding'] === undefined &&
      request.headers['content-length'] === undefined)
  ) {
    return Promise.resolve(null);
  }

[...]

PR #2321, opened 2023-07-03
Support for multipart/form-data in the request body

ilanashapiro:
Addresses #2008

**Summary**

Currently, validation always fails whenever a multipart/form-data content type is used in the request body. This prevents the normal usage of prism without any other proxy server (like nginx) on which such requests can be filtered out. Now, users can successfully use multipart/form-data content type in the request body.

**Checklist**

- The basics
- Tests
  - [X] Added or updated
  - [ ] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A

**Screenshots**

Given the following file called examples/testapi.yaml:
```
openapi: '3.1.0'
paths:
  /path:
    post:
      responses:
        200:
          content:
            text/plain:
              example: ok
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                testData:
                  type: string
              required:
                - "testData"
```
Writing a non-empty post request to http://127.0.0.1/path with included form-data in the request body now produces the following, successfully returning the response defined in the example:
image
image

In contrast, running this same command on the old version of Prism gives a 422 error:
image
image


ilanashapiro:
@rattrayalex @RobertCraigie

packages/http-server/src/__tests__/body-params-validation.spec.ts
```diff
@@ -796,5 +860,20 @@ describe('body params validation', () => {
...
+      test('returns 422', async () => {
+        const response = await makeRequest('/multipart-form-data-body-required', {
+          method: 'POST',
+          body: new URLSearchParams({
+            status: 'new',
+          }).toString(),
+          headers: { 'content-type': 'multipart/form-data' },
+        });
+
+        expect(response.status).toBe(200);
```
chohmann, line 874: the test says it returns 422, but the assertion says it returns 200. What was the intent for this test? It seems like you're trying to test when passing an invalid multipart form data since the body is using URLSearchParams which is not multipart form data.

test-harness/specs/validate-body-params/multipart-allow-empty-value-2.oas2.txt
```diff
@@ -0,0 +1,28 @@
+====test====
+If I have an operation with a parameter with the allowEmptyValue set to false to false
```
chohmann, line 2:
```suggestion
If I have an operation with a parameter with the allowEmptyValue set to false
```

test-harness/specs/validate-body-params/multipart-allow-empty-value-2.oas2.txt
```diff
@@ -0,0 +1,28 @@
...
+and I send a request with a correcly encoded parameter but with an empty key
```
chohmann, line 3:
```suggestion
and I send a request with a correctly encoded parameter but with an empty key
```

test-harness/specs/validate-body-params/multipart-allow-empty-value.oas2.txt
```diff
@@ -0,0 +1,30 @@
+====test====
+If I have an operation with a parameter with the allowEmptyValue set to false to false
```
chohmann, line 2:
```suggestion
If I have an operation with a parameter with the allowEmptyValue set to true
```

test-harness/specs/validate-body-params/multipart-allow-empty-value.oas2.txt
```diff
@@ -0,0 +1,30 @@
...
+and I send a request with a correcly encoded parameter and with the parameter empty
```
chohmann, line 3:
```suggestion
and I send a request with a correctly encoded parameter and with the parameter empty
```

test-harness/specs/validate-body-params/multipart-form-data-allow-reserved-ok.oas3.txt
```diff
@@ -0,0 +1,34 @@
...
+                  type: string
+            encoding:
+              reserved:
+                contentType: text/plain
+                allowReserved: true
+                style: form
+====server====
+mock -p 4010 ${document}
+====command====
+curl -i -X POST http://localhost:4010/path -H "Content-Type: multipart/form-data" --data "reserved=:/?#[]@!$&'()*+,;"
```
chohmann, line 29: For all of these curl requests you need to use `-F` instead of `--data` as that will actually pass form data as multipart.

chohmann (changes requested):
Thank you for submitting this PR!

The main issue we see with this PR is the test harness tests don't actually send multipart form data as expected.

Each of those tests needs to use the `-F` flag instead of `--data` in the curl command.

ilanashapiro:
@chohmann All requested changes are now implemented. Thank you for pointing out that I needed to pass data with the -F flag rather than the --data flag! I initially misunderstood that setting "Content-type: multipart/form-data" was not sufficient. And apologies about the typos in tests I added to the test harness -- I copied these tests from existing files in the application/x-www-form-urlencoded test suite and didn't proofread them first. I have now fixed all the typos in these as well as in the original files. I also added some regex parsing logic for multipart form data to handle additional cases. Please let me know if you'd like any further changes! Thanks!

packages/http/src/validator/validators/body.ts
```diff
@@ -54,6 +54,15 @@ export function splitUriParams(target: string) {
   }, {});
 }
 
+export function parseMultipartFormDataParams(target: string) {
+  const pairs = target.split(/--.*\s*.*="/).filter(element => element);
```
EdVinyard, line 58: I think this code assumes that the boundary between multipart entities is `--` followed by _any_ string.  However, the boundary between multipart entities is defined, in each individual request, by the request's `Content-Type` header's `Boundary` parameter.  This is because the client must choose a value that doesn't occur _within_ any multipart entity that it will transmit.  For example, consider the effect of the following example on your implementation.

```
curl -v -F a=b -F c='--="' <hostname>
```

Also consider any form value that includes `\r\n\r\n`.  

[RFC 2046 § 5.1.1](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1) describes the boundary as...

> a line consisting entirely of two hyphen characters ("-", decimal value 45) followed by the boundary parameter value from the Content-Type header field, optional linear whitespace, and a terminating CRLF.

I think you'll need to either:
1. implement this as the standard describes, with unit tests (not `test-harness` integration tests) to validate that it works with a variety of input, OR
2. rely on a library which parses multipart data according to the standard.

EdVinyard (changes requested):
We really appreciate your update, but I think there are some cases the `parseMultipartFormDataParams()` function doesn't yet handle regarding the boundary between multipart entities.

packages/http/src/validator/index.ts
```diff
@@ -75,17 +76,22 @@ const validateInputBody = (
     checkRequiredBodyIsProvided(requestBody, body),
     E.map(b => [...b, caseless(headers || {})] as const),
     E.chain(([requestBody, body, headers]) => {
+      // parse boundary string from content-type in case media type is multipart/form-data
+      const multipart = require('parse-multipart-data');
+      const contentTypeHeader = headers.get('content-type');
+      const multipartBoundary = contentTypeHeader ? multipart.getBoundary(contentTypeHeader) : "";
```
ilanashapiro, line 82: Here, we parse the boundary parameter value from the Content-Type header field as specified by https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1. This is later passed into the parse-multipart-data npm package to correctly parse the body.

packages/http/src/mocker/index.ts
```diff
@@ -137,19 +137,25 @@ function runCallbacks({
...
 */
 function parseBodyIfUrlEncoded(request: IHttpRequest, resource: IHttpOperation) {
-  const mediaType = caseless(request.headers || {}).get('content-type');
-  if (!mediaType) return request;
+  const contentTypeHeader = caseless(request.headers || {}).get('content-type');
+  if (!contentTypeHeader) return request;
+  
+  // parse boundary string from content-type in case media type is multipart/form-data
+  const multipart = require('parse-multipart-data');
+  const multipartBoundary = multipart.getBoundary(contentTypeHeader);
```
ilanashapiro, line 145: Also here, we parse the boundary parameter value from the Content-Type header field as specified by https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1. This is later passed into the parse-multipart-data npm package to correctly parse the body in the function parseMultipartFormDataParams.

packages/http/src/validator/index.ts
```diff
@@ -75,17 +76,22 @@ const validateInputBody = (
...
+      const mediaType = contentTypeHeader ? contentTypeHeader.replace(new RegExp(";\\s*boundary=" + multipartBoundary), "") : contentTypeHeader;
```
ilanashapiro, line 83: After end-end and unit testing, the content-type header appeared in both formats: "multipart/form-data;boundary=--...." and "multipart/form-data; boundary=--...." (with the space after the semicolon). Thus, I'm handling possible whitespace between the media type and the boundary string.

packages/http/src/mocker/index.ts
```diff
@@ -137,19 +137,25 @@ function runCallbacks({
...
+  const mediaType = contentTypeHeader.replace(new RegExp(";\\s*boundary=" + multipartBoundary), "");
```
ilanashapiro, line 146: See comment in packages/http/src/validator/index.ts

packages/http/src/validator/validators/body.ts
```diff
@@ -54,6 +55,30 @@ export function splitUriParams(target: string) {
...
+        code: 415,
+        severity: DiagnosticSeverity.Error,
+      },
+    ]);
+  }
+
+  const multipart = require('parse-multipart-data');
+  const bufferBody = Buffer.from(target, "utf-8");
+
+  // the parse-multipart-data package requires that the body is passed in as a buffer, not a string
```
ilanashapiro, line 73: See: https://github.com/nachomazzara/parse-multipart-data/issues/14#:~:text=I%20was%20having%20a%20similar%20problem%20with%20a%20different%20payload.%20I%20opened%20up%20the%20dist/%20and%20realized%20that%20the%20author%20named%20the%20body%20param%20multipartBodyBuffer.%20I%20was%20passing%20the%20body%20as%20a%20string%2C%20but%20when%20I%20changed%20to%20pass%20it%20as%20a%20Buffer%20I%20got%20results!

And also the source code: https://github.com/nachomazzara/parse-multipart-data/blob/master/src/multipart.ts#L34C80-L34C80

packages/http/src/validator/validators/body.ts
```diff
@@ -54,6 +55,30 @@ export function splitUriParams(target: string) {
...
+  const parts = multipart.parse(bufferBody, multipartBoundary);
+
+  return parts.reduce((result: Dictionary<string>, pair: string) => {
+    result[pair['name']] = pair['data'].toString();
```
ilanashapiro, line 77: then convert back from Buffer to String for further processing by Prism

packages/http/src/validator/validators/body.ts
```diff
@@ -54,6 +55,30 @@ export function splitUriParams(target: string) {
   }, {});
 }
 
+export function parseMultipartFormDataParams(target: string, multipartBoundary?: string) {
+  if(!multipartBoundary) {
+    const error = "Boundary parameter for multipart/form-data is not defined or generated in the request header. Try removing manually defined content-type from your request header if it exists.";
```
ilanashapiro, line 60: The boundary string MUST be non-empty if this is multipart/form-data (which it is if we're in this function). Thus, we return with error. According to https://bobbyhadz.com/blog/error-multipart-boundary-not-found-in-express-js, https://github.com/postmanlabs/postman-app-support/issues/191, https://stackoverflow.com/questions/40351429/formdata-how-to-get-or-set-boundary-in-multipart-form-data-angular, this can occur when the user is manually setting the content-type, which means the server is not generating it (and the boundary string). Thus, the possible fix is suggested in the error message.

packages/http/src/validator/validators/body.ts
ilanashapiro: Hi Ed, thank you so much for your feedback. I discovered a library called parse-multiform-data which I'm installing as an npm package that seemed to be the most reliable method. I made changes according to this, such as parsing the boundary parameter from the Content-Type header field in order to pass it to the library for parsing the body. I left some comments on the recent changes that hopefully help clarify some of the decisions I made when implementing this. Thanks again and please let me know if you'd like any other changes!

packages/http/src/validator/index.ts
```diff
@@ -75,17 +76,22 @@ const validateInputBody = (
     checkRequiredBodyIsProvided(requestBody, body),
     E.map(b => [...b, caseless(headers || {})] as const),
     E.chain(([requestBody, body, headers]) => {
+      // parse boundary string from content-type in case media type is multipart/form-data
+      const multipart = require('parse-multipart-data');
```
EdVinyard, line 80: We'd prefer to use `import` statements rather than dynamic `require` whenever possible.

packages/http/src/validator/index.ts
EdVinyard: 1. I'm concerned about your use of a regular expression here.  Notice that the boundary may contain any 7-bit ASCII character, which I think includes many characters with special meanings inside a regular expression (e.g., `.`, `*`, `(`).  Consider using a library to parse the Content-Type header in its entirety.  Here's a very simple example using [whatwg-mimetype](https://github.com/jsdom/whatwg-mimetype):

	```TypeScript
	import MIMEType from 'whatwg-mimetype';
	
	const t = new MIMEType('multipart/form-data; boundary=.*.?.+zzz');
	console.log(t?.essence);
	console.log(t.parameters.get('boundary'));
	```

3. It looks like this block of code is a duplicate of what you added in `mocker/index.ts`.  Could it be factored out?

EdVinyard (dismissed):

packages/http/src/validator/validators/body.ts
```diff
@@ -47,11 +49,35 @@ export function deserializeFormBody(
 }
 
 export function splitUriParams(target: string) {
-  return target.split('&').reduce((result: Dictionary<string>, pair: string) => {
+  return E.right(target.split('&').reduce((result: Dictionary<string>, pair: string) => {
```
ilanashapiro, line 52: Lifting to Either monad in order to match the return type of parseMultipartFormDataParams, which is of type Either as it may return an error.

packages/http/src/validator/validators/body.ts
ilanashapiro: (Returning an error seemed to me a better solution than just returning an empty dictionary as it keeps the user more informed. Thus, we the return type is the Either monad).

packages/http/src/validator/validators/body.ts
```diff
@@ -70,12 +96,12 @@ export function findContentByMediaTypeOrFirst(specs: IMediaTypeContent[], mediaT
...
 }
 
-function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema, target: string, prefix?: string, bundle?: unknown) {
+function deserializeAndValidate(content: IMediaTypeContent, schema: JSONSchema, target: string, prefix?: string, multipartBoundary?: string, bundle?: unknown) {
   const encodings = get(content, 'encodings', []);
-  const encodedUriParams = splitUriParams(target);
 
   return pipe(
-    validateAgainstReservedCharacters(encodedUriParams, encodings, prefix),
+    content.mediaType === "multipart/form-data" ? parseMultipartFormDataParams(target, multipartBoundary) : splitUriParams(target),
```
ilanashapiro, line 103: Since parseMultipartFormDataParams returns E.left if the boundary string is empty, the return type is Either, and I wrapped the result of splitUriParams in the Either monad as well to match the return type of parseMultipartFormDataParams. Thus, we move this statement into the expression pipeline.

packages/http/src/mocker/index.ts
```diff
@@ -137,19 +139,24 @@ function runCallbacks({
...
     O.chainNullableK(request => request.body),
     O.chainNullableK(body => body.contents),
     O.getOrElse(() => [] as IMediaTypeContent[])
   );
-
-  const encodedUriParams = splitUriParams(request.body as string);
+  
+  const requestBody = request.body as string;
+  const encodedUriParams = pipe(
+    mediaType === "multipart/form-data" ? parseMultipartFormDataParams(requestBody, multipartBoundary) : splitUriParams(requestBody),
```
ilanashapiro, line 157: parseMultipartFormDataParams returns an error wrapped in E.Left if the boundary sting is empty. Here I convert that to a default value (i.e. empty dictionary) if that occurs. I'm not sure if I'm supposed to be doing anything else here -- this same behavior (unwrapping the monad using defaults) is implemented for findContentByMediaTypeOrFirst just below on line 169, so I attempted to emulate that.

ilanashapiro:
Hi Ed, thanks for your feedback, I am now doing all Content-Type header parsing using the whatwg-mimetype library you recommended, and the dynamic require statements were changed to imports. As for the duplicate code block, I moved it to a small utility function called parseMIMEHeader I created in validators/headers.ts. Finally, I also attempted to improve the error reporting for empty multipart boundary strings by wrapping parseMultipartFormDataParams in the Either monad and returning a descriptive error (which meant I also needed to wrap splitUriParams in the Either monad to maintain equivalent types).

chohmann (approved):
Thank you for your contribution and for working through our reviews with us!

chohmann:
This is included in version 5.2.0 of Prism.

rattrayalex:
Hooray, thanks Ilana, Ed, and Chelsea!

Issue #2033, opened 2022-04-07
Error: Schema size exceeded


hainessss:
Hello,

I am getting `Error: Schema size exceeded` 

And I can't figure out why.

repro script here
```
const { getHttpOperationsFromSpec } = require('@stoplight/prism-cli/dist/operations');
const { createClientFromOperations } = require('@stoplight/prism-http/dist/client');

(async function main () {
  let operations = await getHttpOperationsFromSpec('src/mocks/swagger.json');

  operations = operations.filter(operation => operation.path === '/api/candidate-role')

  console.log('operations', operations)
  const client = createClientFromOperations(operations, {
    mock: true,
  });

  client.get('/api/candidate-role').then(console.log);
})();
```

swagger schema here:
https://prod.betterleap-api.com/api-docs-0001-json

Thanks

chohmann:
@EdVinyard can you try to reproduce?

EdVinyard:
I've reproduced this issue.  We're bumping into a limit hard-coded into `json-schema-sampler` that limits the number of "nodes" in the response to 1000: https://github.com/stoplightio/json-schema-sampler/blob/master/src/traverse.js#L20

Switching Prism into dynamic mock mode gives me a different error, so I suggest we need to:
1. (at least) return a much more descriptive error, and
2. (probably) identify exactly what it is about this schema that breaks `json-schema-sampler`'s mocking.

hernanpis-inditex:
Hello,

I am working on generating mocks for openapi specifications for which sometimes have a very large response schema and I also get the error:

```
{
  "type": "https://stoplight.io/prism/errors#UNKNOWN",
  "title": "Schema size exceeded",
  "status": 500,
  "detail": ""
}
```

Are there any open working lines in order to modify by configuration these default values that are currently hardcoded?
https://github.com/stoplightio/json-schema-sampler/blob/066d49b574246ae3ea5aa821a14e3a3f31836962/src/json-schema-sampler.js#L6

I urgently need a solution to this problem and will probably work on a fork as long as the configuration of these values is not enabled.

Thanks!

ryotrellim:
Todo:
- Improve error messaging
- Create CLI parameter for expanding the node limit

ryotrellim:
[...]

PR #2327, opened 2023-07-17
feat(http): detect complex schema error, improve error message

daniel-white:
Addresses #2033

**Summary**

* upgrade json-schema-sampler to get better error type
* rejigger to allow prism specific error to be thrown

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

**Screenshots**

**Additional context**

### DIFF
--- a/docs/guides/06-errors.md
+++ b/docs/guides/06-errors.md
@@ -264,6 +264,20 @@ paths:
 
 **Explanation:** This error occurs when the current request specifies a `Content-Type` that isn't supported by corresponding HTTP Operation. In the case there is no request body or `Content-Length` is 0, the `Content-Type` header is ignored.
 
+
+### SCHEMA_TOO_COMPLEX
+
+**Message: Unable to generate [body|header] for response. The schema is too complex to generate.**
+
+**Returned Status Code: `500`**
+
+**Explanation:** This error occurs when part of the response can't be generated using JSON Schema Sampler. JSON Schema Sampler has been configured to use a limit of 2500 "ticks". A "tick" is loosely defined as any instance of a JSON Schema schema or subschema. This includes a/an single:
+* Property
+* Object
+* Array item (each in an array between `minItems` and `maxItems`, defaulting to 1)
+* Combiner item (members of `allOf`, `oneOf`, `anyOf`)
+ 
+...all recursively, unless a `$ref` has already been visited (in which case, the subschema is skipped). As a workaround, consider using [dynamic response mocking](./11-dynamic-response-with-faker.md).
 ## Unknown error
 
 In case you get an `UNKNOWN` error, it likely means this particular edge case isn't handled. If you encounter one of these, open an [issue](https://github.com/stoplightio/prism/issues/new?labels=bug&template=bug_report.md).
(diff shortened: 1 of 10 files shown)


packages/http/src/mocker/generator/JSONSchema.ts
```diff
@@ -98,6 +102,26 @@ export function sortSchemaAlphabetically(source: any): any {
...
+}
+
+export class GeneratorError extends Error {}
+
+export class SchemaTooComplexGeneratorError extends GeneratorError {
+  constructor(operation: IHttpOperation, public readonly cause: Error) {
+    super(
+      `The operation ${operation.method.toUpperCase()} ${
+        operation.path
+      } references a JSON Schema that is too complex to generate.`
```
EdVinyard, line 124: Is there a way to state precisely in what way/dimension the spec is too complex here?  As a Prism user who sees this, what can I do to work-around the problem or simplify the spec to make it fall within the boundary?

EdVinyard (changes requested):
Everything else looks good here, but I'd love to see a really specific error message that suggests a work-around or fix for users.  I'd be happy to brainstorm that with you, or maybe even bring in our docs folks to wordsmith it once we can nail the description of what limit/boundary you've surpassed.

packages/http/src/mocker/generator/JSONSchema.ts
daniel-white: we'd have to make json-schema-sampler and json-schema-faker to provide that context. the sampler code is all vanilla js so its pretty terse to follow

EdVinyard: Is this a case where we could add to the Prism docs we publish, and include a URL here for more info?  I know we haven't done this before, but I'm just trying to find _some_ way to give users information they can act on (or at least understand).  I just don't think, as a Prism user, I'd be able to understand "too complex to generate" enough to solve my own problem.

EdVinyard (approved):

docs/guides/06-errors.md
```diff
@@ -264,6 +264,20 @@ paths:
...
 **Explanation:** This error occurs when the current request specifies a `Content-Type` that isn't supported by corresponding HTTP Operation. In the case there is no request body or `Content-Length` is 0, the `Content-Type` header is ignored.
 
+
+### SCHEMA_TOO_COMPLEX
+
+**Message: Unable to generate [body|header] for response. The schema is too complex to generate.**
+
+**Returned Status Code: `500`**
+
+**Explanation:** This error occurs when part of the response can't be generated using JSON Schema Sampler. JSON Schema Sampler has been configured to use a limit of 2500 "ticks". A "tick" is loosely defined as any instance of a JSON Schema schema or subschema. This includes a/an single:
```
RaySalty, line 274: Recommend: "This error occurs when part of the response can't be generated using **the** JSON Schema Sampler. **The** JSON Schema Sample **has** a limit of 2500 ticks."

"Has been configured" makes it feel open to changing.

docs/guides/06-errors.md
```diff
@@ -264,6 +264,20 @@ paths:
...
+* Property
+* Object
+* Array item (each in an array between `minItems` and `maxItems`, defaulting to 1)
+* Combiner item (members of `allOf`, `oneOf`, `anyOf`)
+ 
+...all recursively, unless a `$ref` has already been visited (the subschema is skipped). As a workaround, consider using [dynamic response mocking](./11-dynamic-response-with-faker.md).
```
RaySalty, line 280: If "(**in which case**, the subschema is skipped)" is a correct understanding of the use of `$ref`, then recommend adding the additional language to read even easier.

RaySalty (approved):

docs/guides/06-errors.md
daniel-white: it is open to changing

Issue #2046, opened 2022-04-22
Mock response contains random fields


zkorhone:
## Context
Generated mock response does not follow schema

## Current Behavior
Given OpenAPI doc:

```
openapi: 3.0.3
info:
  title: Subscription Query API
  version: '1.0'
  description: Unified interface for querying Subscriptions A and B
servers:
  - url: 'http://localhost:3000/subscription-queries'
paths:
  /active_subscriptions:
    get:
      summary: Get active subscriptions
      tags: []
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/ActiveSubscription'
      operationId: get-active_subscriptions
      description: Get list of active subscriptions
      parameters:
        - schema:
            type: boolean
          in: query
          name: module_support
          description: 'When true return subscriptions that could theoretically have modules. When false or undefined, do not filter result set'
components:
  schemas:
    ActiveSubscription:
      title: ActiveSubscription
      type: object
      description: Active Subscription
      x-examples: {}
      x-internal: false
      properties:
        actions:
          type: object
          description: Actions available for caller
          properties:
            availableModules:
              type: string
              description: Allowed to query for available modules
```

Prims generates response:

```
[
  {
    "actions": {
      "availableModules": "ex",
      "dolore_3d4": -18659865,
      "in_aa3": false
    }
  },
  {
    "actions": {
      "availableModules": "irure aliqua Ut",
      "irure_f_": 13851100.013715371,
      "in_d3": "exercitation tempor nisi esse et",
      "nullaca3": 24122535.056890324,
      "esseb62": -46474555
    },
    "exercitationec_": 68917320.29432371
  },
  ....
```

## Expected Behavior
Prism generates response that follows specification:

```
[
  {
    "actions": {
      "availableModules": "ex",
    }
  },
```

## Possible Workaround/Solution
None

## Steps to Reproduce
Start Stoplight Studio, open provided OpenAPI document & use Try it against Mock Server

## Environment

- Version used: Stoplight Sutdio 2.8.0-stable.7492.git-69a3957 (please make this easier to copy+paste!)
- Environment name and version (e.g. Chrome 39, node.js 5.4): -
- Operating System and version (desktop or mobile): Windows 10 (Desktop)
- Link to your environment/workspace/project: -

sreenadh-farmface:
I also got this. this happens when there is only one key in the response. just adding another key will fix the issue.
[...]

PR #2355, opened 2023-08-10
add json schema faker fill properties param

brendarearden:
Addresses #2046 
**Summary**
In an effort to make it easier to set the json schema faker `fillProperties` to false, without having to modify individual schemas, a new cli parameter has been added for mocking. 

With this change you can now run  `prism mock -d --json-schema-faker-fillProperties=false api.oas3.yaml` to ensure additional properties that are not explicitly named in your spec aren't returned when mocking. Setting this value with the command line will take priority over the value set in `x-json-schema-faker`

**Additional Background**
When using dynamic mocking, if `additionalProperties` is not set on each `properties` in a schema, json schema faker will fill in additional properties due to Prism's current [global setting](https://github.com/stoplightio/prism/blob/8f03c80f3643d2e213eeb36d51422a98faa7e762/packages/http/src/mocker/generator/JSONSchema.ts#L42) for `fillProperties` being set to true. 

Adding a `x-json-schema-faker` object at the top level of your schema allows you to change Prism's global settings per schema if you don't want to set it at the cli level. See [documentation](https://docs.stoplight.io/docs/prism/9528b5a8272c0-dynamic-response-generation-with-faker#configure-json-schema-faker) for further details

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Screenshots**

**Additional context**

### DIFF
--- a/packages/cli/src/commands/mock.ts
+++ b/packages/cli/src/commands/mock.ts
@@ -20,13 +20,29 @@ const mockCommand: CommandModule = {
           boolean: true,
           default: false,
         },
+        'json-schema-faker-fillProperties': {
+          description: 'Generate additional properties when using dynamic generation.',
+          default: undefined,
+          boolean: true,
+        },
       }),
   handler: parsedArgs => {
-    const { multiprocess, dynamic, port, host, cors, document, errors, verboseLevel } =
+    parsedArgs.jsonSchemaFakerFillProperties = parsedArgs['json-schema-faker-fillProperties'];
+    const { multiprocess, dynamic, port, host, cors, document, errors, verboseLevel, jsonSchemaFakerFillProperties } =
       parsedArgs as unknown as CreateMockServerOptions;
 
     const createPrism = multiprocess ? createMultiProcessPrism : createSingleProcessPrism;
-    const options = { cors, dynamic, port, host, document, multiprocess, errors, verboseLevel };
+    const options = {
+      cors,
+      dynamic,
+      port,
+      host,
+      document,
+      multiprocess,
+      errors,
+      verboseLevel,
+      jsonSchemaFakerFillProperties,
+    };
 
     return runPrismAndSetupWatcher(createPrism, options);
   },
--- a/packages/cli/src/extensions.ts
+++ b/packages/cli/src/extensions.ts
@@ -5,20 +5,33 @@ import * as JSONSchemaFaker from 'json-schema-faker';
 import type { JSONSchemaFakerOptions } from 'json-schema-faker';
 import { resetJSONSchemaGenerator } from '@stoplight/prism-http';
 
-export async function configureExtensionsFromSpec(specFilePathOrObject: string | object): Promise<void> {
+export async function configureExtensionsUserProvided(
+  specFilePathOrObject: string | object,
+  cliParamOptions: { [option: string]: any }
+): Promise<void> {
   const result = decycle(await new $RefParser().dereference(specFilePathOrObject));
 
   resetJSONSchemaGenerator();
 
   forOwn(get(result, 'x-json-schema-faker', {}), (value: any, option: string) => {
-    if (option === 'locale') {
-      // necessary as workaround broken types in json-schema-faker
-      // @ts-ignore
-      return JSONSchemaFaker.locate('faker').setLocale(value);
-    }
+    setFakerValue(option, value);
+  });
 
+  // cli parameter takes precidence, so it is set after spec extensions are configed
+  for (const param in cliParamOptions) {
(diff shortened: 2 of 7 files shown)


packages/cli/src/extensions.ts
```diff
@@ -5,20 +5,33 @@ import * as JSONSchemaFaker from 'json-schema-faker';
...
+
+  // cli parameter takes precidence, so it is set after spec extensions are configed
+  cliParamOptions.forEach(param => {
+    if (param.cliValue !== undefined) {
+      setFakerValue(param.cliOption, param.cliValue);
     }
+  });
+}
 
+function setFakerValue(option: string, value: any) {
```
EdVinyard, line 28: I like the effect of factoring this out.  To me it seems like `configureExtensionsUserProvided()` is now at a more consistent level of detail/abstraction.

packages/cli/src/util/createServer.ts
```diff
@@ -149,6 +150,7 @@ type CreateBaseServerOptions = {
   multiprocess: boolean;
   errors: boolean;
   verboseLevel: string;
+  jsonSchemaFakerFillProperties: boolean;
```
EdVinyard, line 153: I'm not sure if this is the right place, but I don't see a better opportunity to add a JS Doc comment that describes what effect this option is intended to have.

packages/cli/src/util/createServer.ts
```diff
@@ -71,7 +71,8 @@ const createSingleProcessPrism: CreatePrism = options => {
 
 async function createPrismServerWithLogger(options: CreateBaseServerOptions, logInstance: pino.Logger) {
   const operations = await getHttpOperationsFromSpec(options.document);
-  await configureExtensionsFromSpec(options.document);
+  const jsonSchemaFakerCliParams = [{ cliOption: 'fillProperties', cliValue: options.jsonSchemaFakerFillProperties }];
```
EdVinyard, line 74: Why an _array_ of objects (`{ cliOption: string; cliValue: any }[]`) here rather than an object (perhaps `{ [option: string]: any }` or something more specific derived from `CreateBaseServerOptions`)?

EdVinyard (approved):
I have only questions and minor suggestions; feel free to use them or ignore them.

packages/cli/src/commands/mock.ts
```diff
@@ -20,13 +20,29 @@ const mockCommand: CommandModule = {
...
+      port,
+      host,
+      document,
+      multiprocess,
+      errors,
+      verboseLevel,
+      jsonSchemaFakerFillProperties,
+    };
 
     return runPrismAndSetupWatcher(createPrism, options);
```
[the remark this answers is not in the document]
brendarearden: parsedArgs are all typed as unknown, so yes its due to them being so loosely typed

packages/cli/src/util/createServer.ts
brendarearden: I had originally done some mapping functionality that was made easier with an array, but I was able to refactor that out and didn't think to change the data type to something else. I have updated it to an object as I do prefer that suggestion.

packages/cli/src/util/createServer.ts
brendarearden: added a comment in the most recent changes

packages/cli/src/commands/mock.ts
```diff
@@ -20,13 +20,29 @@ const mockCommand: CommandModule = {
           boolean: true,
           default: false,
         },
+        'json-schema-faker-fillProperties': {
+          description: 'Generate additional properties when using dynamic generation.',
+          //default value is set in the generator and should not be set here
+          boolean: true,
```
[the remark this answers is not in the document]
brendarearden: https://github.com/stoplightio/prism/pull/2355/files#diff-d29590a9846852cedd88e73844eaf66c37b0d412a7b598200679b2c0105d4f66R16 sets the faker values found in the spec and then https://github.com/stoplightio/prism/pull/2355/files#diff-d29590a9846852cedd88e73844eaf66c37b0d412a7b598200679b2c0105d4f66R22 has an undefined check as to not override any set values from the spec unnecessarily.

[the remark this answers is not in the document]
brendarearden: after talking through this with you more, I have gone ahead and set the default value to undefined to make it very clear. https://github.com/stoplightio/prism/pull/2355/files#diff-03af80d46458a7d7520e8d4707fe9d26a4e47e329320c85548bdb95c63439471R25

daniel-white (approved):

Issue #2346, opened 2023-08-03
Incorrect `Request` prefix when converting AJV errors


Drew-Kimberly:
`Request` is hardcoded to prefix the error message here:

https://github.com/stoplightio/prism/blob/9661bf7d484cbc8be2d2457b21727a95d7a2807d/packages/http/src/validator/validators/utils.ts#L84

This leads to a misleading message when the validation error comes from the Response.

## Context

Misleading error message causes developers to look for problem in wrong schema.

## Current Behavior

Given the following error:
```ts
{ path: ['response', 'body'], code: 'required' }
```

We get the following error message:

> Request body must have required property '...'

## Expected Behavior
Given the following error:
```ts
{ path: ['response', 'body'], code: 'required' }
```

We should get the following error message:

> Response body must have required property '...'

## Possible Workaround/Solution

The prefix should be derived from the `path` when converting the AJV error.

## Steps to Reproduce

1.
2.
3.
4.

## Environment

- Version used: `5.2.0`
- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your environment/workspace/project:

chohmann:
This has been released in version 5.3.0.

PR #2358, opened 2023-08-16
Fix/2346 ajv error message prefix

chohmann:
Addresses #2346 

**Summary**

Passes along the already existing concept of validation context (input/output) to `convertAjvErrors` function to have the error message properly say it's in the Request or Response.

To test manually, I mocked a spec that returned either `bird` or `lizard` for pet type. Then called `prism proxy` with a spec that required type be either `cat` or `dog`  and configured it to forward requests to the prism mock server so that the response from the mock server would always be invalid for the spec used by the proxy.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

**Screenshots**
Violation Logs
image

Violation Header
image


packages/http/src/validator/index.ts
```diff
@@ -186,7 +200,7 @@ export const validateOutput: ValidatorFn<IHttpOperation, IHttpResponse> = ({ res
           )
         ),
         validateBody(element.body, response.contents || [], ValidationContext.Output, mediaType, bundle),
-        validateHeaders(element.headers || {}, response.headers || [], bundle)
+        validateHeaders(element.headers || {}, response.headers || [], ValidationContext.Output, bundle)
```
daniel-white, line 203: was this an optional parameter? can we remove its optionality?

packages/http/src/validator/validators/utils.ts
```diff
@@ -81,7 +87,7 @@ export const convertAjvErrors = (errors: NonEmptyArray<ErrorObject>, severity: D
       const path = prefix ? [prefix, ...errorPath] : errorPath;
       const errorPathType = errorPath.length > 0 ? (prefix == 'body' ? 'property ' : 'parameter ') : '';
       const errorSourceDescription =
-        'Request ' +
```
daniel-white, line 84: is this whitespace important?

packages/http/src/validator/index.ts
chohmann: It wasn't a parameter to the header validation function before at all. I add it in this PR and it is not optional. It was only being passed to the body validator function before for some reason, and not the other validators 🤷‍♀️

daniel-white: i see it now.

daniel-white (approved):

Issue #2376, opened 2023-09-01
Allow Binary Requests To Succeed in Prism Proxy Mode


travisgosselin:
There are varying behaviors I have noticed when working with `multipart/form-data` request bodies behind the prism proxy. First off, I understand that according to your documentation you do not support `multipart/form-data` or binary files:
https://docs.stoplight.io/docs/prism/1593d1470e4df-concepts#content-negotiation

I'd like to understand what "support" means specifically for `proxy` mode? For example, does it mean that you do not validate or log any discrepancies in these request/response bodies, but still forward the bodies through? Or does it mean that the proxy is expected to completely fail if it sees a content type of `multipart/form-data`? 

In some observation, I have noticed that:

### V4.14.1 of the Prism Proxy
- When making a request with an `xlsx` file it succeeds and proxies the file through (though it doesn't validate the schema as I would expect). When passing large `csv` or `txt` files it also works and proxies it along when executing the local CLI on a Windows OS.
- When executing the same parameters via Docker container `stoplight/prism:4.14.1`, the exact same series of requests fail:
```
/usr/src/prism/node_modules/split2/index.js:44
      push(this, this.mapper(list[i]))
                      ^
SyntaxError: Unexpected token � in JSON at position 259901
    at Transform.parse [as mapper] (<anonymous>)
    at Transform.transform [as _transform] (/usr/src/prism/node_modules/split2/index.js:44:23)
    at Transform._read (/usr/src/prism/node_modules/readable-stream/lib/_stream_transform.js:166:10)
    at Transform._write (/usr/src/prism/node_modules/readable-stream/lib/_stream_transform.js:155:83)
    at doWrite (/usr/src/prism/node_modules/readable-stream/lib/_stream_writable.js:390:139)
    at writeOrBuffer (/usr/src/prism/node_modules/readable-stream/lib/_stream_writable.js:381:5)
    at Transform.Writable.write (/usr/src/prism/node_modules/readable-stream/lib/_stream_writable.js:302:11)
    at Socket.ondata (node:internal/streams/readable:754:22)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
```

I expect that there is a platform-specific reason this functions properly on Windows but not within the Linux-based container provided?

### V5.3.1 of the Prism Proxy
When making a request with an `xlsx` file it will fail with the following error and the proxy will shut down even while executing locally on the same Windows platform:
```
Recv failure: Connection was reset
```
[...]

PR #2387, opened 2023-09-18
keep encoded value if decoding fails.

chohmann:
#2376 

**Summary**

Handles when uri decoding binary data fails.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

**Screenshots**

Using the following curl request:
```
curl -X POST -H "Content-Type: multipart/form-data" -F "profileImage=@/Users/chelsea/Downloads/1x1.png" "http://127.0.0.1:4010/anything"
```
the behavior is as follows for the different versions of Prism:

**Version 4.14.1**
image

**Version 5.3.1**
image

**Local With Fixes**
image

### DIFF
--- a/packages/http/src/validator/validators/body.ts
+++ b/packages/http/src/validator/validators/body.ts
@@ -87,7 +87,14 @@ export function parseMultipartFormDataParams(
 
 export function decodeUriEntities(target: Dictionary<string>) {
   return Object.entries(target).reduce((result, [k, v]) => {
-    result[decodeURIComponent(k)] = decodeURIComponent(v);
+    try {
+      // NOTE: this will decode the value even if it shouldn't (i.e when text/plain mime type).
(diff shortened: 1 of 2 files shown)


packages/http/src/validator/validators/body.ts
```diff
@@ -87,7 +87,14 @@ export function parseMultipartFormDataParams(
 
 export function decodeUriEntities(target: Dictionary<string>) {
```
daniel-white, line 88: no content type available when this is called?

chohmann: nope.

daniel-white (approved):

chohmann:
This has been released in version 5.3.2.

Issue #2393, opened 2023-09-26
Feature: support Optional security and AND security scheme groups


dotslashderek:
Open API supports both optional and AND chaining with security schemes.  This isn't supported correctly by Prism currently.

AND support is missing entirely.

Optional is supported, but we'd like to change that support.  Currently, the support is such that if Optional is included as a valid scheme, you can hand prism any authorization header (malformed, from a security type that isn't supported explicitly, etc) - and this will return a 200.

Given an endpoint that supports, say, Basic OR None, this should be interpreted as - either don't include an auth header (NONE case), or include one that is proper for the Basic. 

If you hand an auth header for Bearer, the response should be 401 (currently 200).
If you hand in a mangled Basic header, the response should be 401 (currently 200).

Here are some basic tests to use to begin working with the optional case:

```

    it('passes the validation with Basic', () => {
      assertRight(
        validateSecurity({
          element: { ...baseRequest, headers: { authorization: `Basic ${token}` } },
          resource: { security: securityScheme },
        })
      );
    });

    it('fails the validation with no Auth', () => {
      assertLeft(
        validateSecurity({
          element: { ...baseRequest },
          resource: { security: securityScheme },
        })
      );
    });

    it('fails with an invalid credentials error', () => {
      assertLeft(
        validateSecurity({
          element: { ...baseRequest, headers: { authorization: 'Basic abc123' } },
          resource: { security: securityScheme },
        }),
        res =>
          expect(res).toStrictEqual([
            {
              code: 401,
              message: 'Invalid security scheme used',
              severity: DiagnosticSeverity.Error,
              tags: [],
            },
          ])
      );
    });

    it('fails with an invalid security scheme error', () => {
      assertLeft(
        validateSecurity({
          element: { ...baseRequest, headers: { authorization: 'Bearer abc123' } },
          resource: { security: securityScheme },
        }),
        res =>
          expect(res).toStrictEqual([
            {
              code: 401,
              tags: ['Basic realm="*"'],
              message: 'Invalid security scheme used',
              severity: DiagnosticSeverity.Error,
            },
          ])
      );
    });
  });

  describe('when security scheme uses optional Basic authorization', () => {
    const securityScheme: HttpSecurityScheme[][] = [
      [
        {
          id: faker.random.word(),
          scheme: 'basic',
          type: 'http',
          key: 'sec',
          extensions: { 'x-test': faker.random.word() },
        },
      ],
      [],
    ];

    it('passes the validation with Basic', () => {
[...]

PR #2395, opened 2023-09-29
feat: better optional auth support

billiegoose:
Addresses #2393

**Summary**

> Optional is supported, but we'd like to change that support. Currently, the support is such that if Optional is included as a valid scheme, you can hand prism any authorization header (malformed, from a security type that isn't supported explicitly, etc) - and this will return a 200.
>
> Given an endpoint that supports, say, Basic OR None, this should be interpreted as - either don't include an auth header (NONE case), or include one that is proper for the Basic.

I found it a bit tricky to cover every case, so I just covered the Authorization header case, which covers Basic, Bearer, OpenId, OAuth2.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

### DIFF
--- a/packages/http/src/validator/validators/security/__tests__/security.spec.ts
+++ b/packages/http/src/validator/validators/security/__tests__/security.spec.ts
@@ -447,7 +447,7 @@ describe('validateSecurity', () => {
               code: 401,
               message: 'Invalid security scheme used',
               severity: DiagnosticSeverity.Error,
-              tags: ['Bearer', 'Basic realm="*"'],
+              tags: ['Bearer', 'None'],
             },
           ])
       );
--- a/packages/http/src/validator/validators/security/__tests__/security.spec.ts
+++ b/packages/http/src/validator/validators/security/__tests__/security.spec.ts
@@ -397,6 +397,63 @@ describe('validateSecurity', () => {
     });
   });
 
+  describe('when security scheme is optional', () => {
+    const securityScheme: HttpSecurityScheme[][] = [
+      [
+        {
+          id: faker.random.word(),
+          scheme: 'bearer',
+          type: 'http',
+          key: 'sec',
+          extensions: { 'x-test': faker.random.word() },
+        },
+      ],
+      [], // yeah that's how you do optional in OpenAPI
+    ];
+
+    it('passes if no security scheme is used', () => {
+      assertRight(
+        validateSecurity({
+          element: { ...baseRequest, headers: {} },
+          resource: {
+            security: securityScheme,
+          },
+        })
+      );
+    });
+
+    it('passes the validation', () => {
+      assertRight(
+        validateSecurity({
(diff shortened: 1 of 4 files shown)


packages/http/src/validator/validators/security/handlers/none.ts
```diff
@@ -0,0 +1,12 @@
...
+import { when } from './utils';
+import { Dictionary } from '@stoplight/types';
+import { IHttpRequest } from '../../../../types';
+
+// Makes sure there aren't any auth headers
+function isNoAuth(inputHeaders: Dictionary<string>) {
+  return get(inputHeaders, 'authorization') == undefined;
+}
+
+export const none = (input: Pick<IHttpRequest, 'headers' | 'url'>) => when(isNoAuth(input.headers || {}), 'None');
```
billiegoose, line 12: Do y'all think I should try to make this more elaborate? Like, somehow invert the other security schemes to check there's no API Key?

packages/http/src/validator/validators/security/__tests__/security.spec.ts
```diff
@@ -617,4 +674,168 @@ describe('validateSecurity', () => {
...
+          expect(res).toStrictEqual([
+            {
+              code: 401,
+              message: 'Invalid security scheme used',
+              severity: DiagnosticSeverity.Error,
+              tags: ['OAuth2', 'Bearer', 'OpenID'],
+            },
+          ])
+      );
+    });
```
dotslashderek, line 839: Duplicates?

P0lip (approved):
Looks good from the functional perspective, but I'd wait for someone from Tacocats for their final say.

packages/http/src/validator/validators/security/handlers/none.ts
dotslashderek: Ahhh... to check and see if they're trying to pass something that matches an api key as a query param? I'm guessing that's why you're taking the url in with the headers?

I don't think so.  Extraneous query params just tend to be tolerated and ignored, I guess, and while we could probably identify the case that it's specifically "trying" to use an API Key when it shouldn't I imagine that would only happen (in our tooling) if you had a improper security scheme AND combination that included Optional... that's the only time we try to mix / match in terms of headers / query params for different schemes.

dotslashderek (approved):
Looks good @billiegoose - not sure about the two error tests - don't think you need to do more work w/ validating noAuth state

packages/http/src/validator/validators/security/__tests__/security.spec.ts
billiegoose: huh. they weren't supposed to be duplicates lololol

chohmann (approved):
LGTM! The only thing I think might be helpful is one or two test harness tests, but I don't think they're strictly necessary. Thanks for adding this!

Issue #1934, opened 2021-10-30
Prism fails to send a response if validation pattern contains non-ascii characters


RomanKornev:
## Describe the bug

Prism fails to send a response if specification's validation pattern contains non-ascii characters. Request is stuck forever.

## To Reproduce

1. Given this OpenAPI document `spec.yaml`
```yaml
swagger: '2.0'
info:
  title: bug
  version: 1.0.2
schemes:
  - https
basePath: /
produces:
  - application/json
consumes:
  - application/json
definitions:
  GenericError:
    type: object
    required:
      - httpCode
    properties:
      httpCode:
        type: string
        example: '400'
        maxLength: 3
    additionalProperties: false
  Info:
    additionalProperties: false
    properties:
      city:
        type: string
        pattern: '^[А-ЯЁа-яё]+([- ][А-ЯЁа-яё]+)*$'
    type: object
paths:
  /:
    post:
      consumes:
        - application/json
      produces:
        - application/json
      parameters:
        - in: body
          name: body
          description: body
          required: true
          schema:
            $ref: '#/definitions/Info'
      responses:
        '422':
          schema:
            $ref: '#/definitions/GenericError'
      operationId: create
```
2. Run this CLI command `prism mock spec.yaml`
3. Send POST to localhost:4010
```
{
	"city": "fdgfg"
}
```
4. Request is stuck forever
5. See error message in logs
```
[10:29:45 AM] › [HTTP SERVER] post /applications ℹ  info      Request received
[10:29:45 AM] ›     [NEGOTIATOR] ℹ  info      Request contains an accept header: application/json
[10:29:45 AM] ›     [VALIDATOR] ⚠  warning   Request did not pass the validation rules
[10:29:45 AM] ›     [NEGOTIATOR] ✔  success   Found response 422. I'll try with it.
[10:29:45 AM] ›     [NEGOTIATOR] ⬤  debug     Unable to find a content with an example defined for the response 422
[10:29:45 AM] ›     [NEGOTIATOR] ✔  success   The response 422 has a schema. I'll keep going with this one
[10:29:45 AM] ›     [NEGOTIATOR] ✔  success   Responding with the requested status code 422
(node:22298) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["sl-violations"]
    at ServerResponse.setHeader (_http_outgoing.js:561:3)
[...]

PR #2400, opened 2023-10-09
fix: Eliminated the possible inclusion of non-ascii characters in the…

Lingihn:
Addresses #1934

**Summary**

The problem arises because HTTP headers cannot include non-ascii characters. `convertAjvErrors` produces the final error text, which is then passed into sl-violations header.

I removed the ability to view regular expression patterns. (this is exactly what they are complaining about in issue)
Also, I disabled the ability to view allowed values. Our project uses enum with Russian characters, like this:

> metering_device_model_execution:
>   type:        string
>   description: Исполнение
>   example:     электронный
>   enum:
>     - электронный
>     - индукционный
>     - комбинированный
>     - модульный

If there is a mismatch, the contents of the enum are added to the sl-violations error text.
Enum stores part of the business logic, and getting rid of them is impossible if the base language is not English.

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [X] N/A
- Event Tracking
  - [X] N/A
- Error Reporting
  - [X] N/A
  
  ### Failed to run tests
It's probably something to do with the system or something else. Because, even without making changes, the tests did not run with this error:
```
  PS C:\Users\Offlu\OneDrive\Документы\GitHub_Projects\prism> yarn test
yarn run v1.22.19
$ node --expose-gc ./node_modules/.bin/jest
C:\Users\Offlu\OneDrive\Документы\GitHub_Projects\prism\node_modules\.bin\jest:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
^^^^^^^
SyntaxError: missing ) after argument list
    at Object.compileFunction (node:vm:360:18)
    at wrapSafe (node:internal/modules/cjs/loader:1055:15)
    at Module._compile (node:internal/modules/cjs/loader:1090:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1180:10)
    at Module.load (node:internal/modules/cjs/loader:1004:32)
    at Function.Module._load (node:internal/modules/cjs/loader:839:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
error Command failed with exit code 1.
```

I would be grateful if you could help me figure this out.

### DIFF
--- a/packages/http/src/validator/validators/utils.ts
+++ b/packages/http/src/validator/validators/utils.ts
@@ -77,30 +77,32 @@ export const convertAjvErrors = (
   context: ValidationContext,
   prefix?: string
 ) =>
-  pipe(
-    errors,
-    map<ErrorObject, IPrismDiagnostic>(error => {
-      const allowedParameters = 'allowedValues' in error.params ? `: ${error.params.allowedValues.join(', ')}` : '';
-      const detectedAdditionalProperties =
-        'additionalProperty' in error.params ? `; found '${error.params.additionalProperty}'` : '';
-      const errorPath = error.instancePath.split('/').filter(segment => segment !== '');
-      const path = prefix ? [prefix, ...errorPath] : errorPath;
-      const errorPathType = errorPath.length > 0 ? (prefix == 'body' ? 'property ' : 'parameter ') : '';
-      const errorSourceDescription =
-        `${context === ValidationContext.Input ? 'Request' : 'Response'} ` +
-        (prefix ? `${prefix} ` : '') +
-        errorPathType +
-        errorPath.join('.').trim() +
-        (errorPath.length > 0 ? ' ' : '');
-
-      return {
-        path,
-        code: error.keyword || '',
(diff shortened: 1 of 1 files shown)


chohmann:
@Lingihn thank you for this PR! However, we'd prefer to see this issue solved by MIME encoding, preferably UTF-8, the sl-violations header instead of stripping this from the header altogether. This is in accordance with [Http spec](https://httpwg.org/specs/rfc8941.html#rfc.section.3.3.3).

From that link above:
> When it is necessary for a field value to convey non-ASCII content, a Byte Sequence ([Section 3.3.5](https://httpwg.org/specs/rfc8941.html#binary)) can be specified, along with a character encoding (preferably UTF-8 [[STD63]](https://httpwg.org/specs/rfc8941.html#STD63)).

We'd also like to see new or updated tests covering any new changes.

chohmann:
We're going to close this PR since we haven't heard from you in a month. Please re-open if you want to address our PR comments.

PR #2401, opened 2023-10-09
feat: better validation for optional auth

billiegoose:
I messed up the PR merge for https://github.com/stoplightio/prism/pull/2395 because I'm so used to the defaults in `platform-internal`. Sorry about that. Take 2!


chohmann:
This was released in version 5.4.0.

PR #2408, opened 2023-10-16
Implement --ignoreExamples CLI flag for Prism to treat spec as though it doesn't have examples

ilanashapiro:
Addresses comments on PR #2383

**Summary**
This PR introduces a new CLI flag --ignoreExamples. When present, and in static mode, the flag tells Prism to treat the spec as if it doesn't have any response examples defined. In this case, Prism continues to do what it already does in static mode when no response example are present, which is to return an example that has not been generated using json-schema-faker, but was created by Prism. However, when the --ignoreExamples flag is present, and in dynamic mode, the flag is essentially ignored, since in dynamic mode, examples are not consulted and json-schema-faker is used to generate a response based on the schema defined in the spec. A use may wish to use this flag when there are potentially invalid response examples defined in the schema that the user doesn't want to get returned. 

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Screenshots**
Consider the following schema:
```
openapi: 3.0.2
paths:
  /pets:
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                '$ref': '#/schemas/Pet'
              examples:
                invalid_cat:
                  summary: An invalid example of a cat
                  value:
                    name: 456
                    status: "new"
                invalid_dog:
                  summary: An invalid example of a dog
                  value:
                    id: 135
                    name: 123
                valid_dog:
                  summary: A valid example of a dog
                  value:
                    id: 135
                    name: "dog"
                    status: "available"

schemas:
  Pet:
      type: object
      required:
      - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
          example: doggie
        status:
          type: string
          enum:
          - available
          - pending
          - sold
```
By mocking this schema with the --ignoreExamples command line flag, we successfully ignore the given examples and the following example statically generated by Prism is returned:
image
image

Without the --ignoreExamples flag, we see how the invalid_cat example gets returned instead:
image
image

However, if we have the --ignoreExamples flag AND the --dynamic flag, the --dynamic flag overrides the --ignoreExamples and the example is dynamically generated:
image
image


ilanashapiro:
@rattrayalex @chohmann

chohmann (approved):
This looks great! Thanks for being patient with us as we worked to understand what your suggestion was.

rattrayalex:
Thanks so much Chelsea!! What's the procedure for merging?

chohmann:
I'll merge and release on Monday 👍

chohmann:
This was just released in version 5.5.0 of Prism 🎉

PR #2476, opened 2024-03-15
STOP-243 add fake IncomingMessage class 

EdVinyard:
STOP-243

**Summary**

Add a fake IncomingMessage class so we can reuse existing Prism code that assumes the request came in from a socket.

This is intended for inclusion in @brendarearden 's work on STOP-243 rather than direct inclusion into the Prism codebase.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Screenshots**
N/A

**Additional context**
N/A

### DIFF
--- a/packages/http/src/utils/__tests__/fakeIncomingMessage.spec.ts
+++ b/packages/http/src/utils/__tests__/fakeIncomingMessage.spec.ts
@@ -1,0 +1,84 @@
+import { FakeIncomingMessage } from '../fakeIncomingMessage';
+import * as typeIs from 'type-is';
+
+describe('FakeIncomingMessage', () => {
+  it('fools isType when there is a body', () => {
+    // Arrange
+    const body = toUtf8('{}');
+    const fake = new FakeIncomingMessage({
+      method: 'POST',
+      url: 'http://example.com/',
+      headers: {
+        'Content-Type': 'application/json',
+        'Content-Encoding': 'UTF-8',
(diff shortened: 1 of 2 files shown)


packages/http/src/utils/fakeIncomingMessage.ts
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable prettier/prettier */
```
EdVinyard, line 1: Sadly, we can't disable the formatting for a block of code, only a single line or a whole file.  Allowing all the unimplemented getters and setters at the end of the class to be compactly formatted makes it far easier to see what's going on than when they're spread out over 50+ lines.  I'd rather have the type checking @daniel-white suggested than the code format enforcement in this case (since I can only have 2 out of 3).

packages/http/src/utils/__tests__/fakeIncomingMessage.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { FakeIncomingMessage } from '../fakeIncomingMessage';
+import * as typeIs from 'type-is';
```
EdVinyard, line 2: @brendarearden , I'd be happy to add more tests here if you know of other ways we need to "fake" `IncomingMessage` behavior, but I think `typeIs` and `hasBody` were the main ones.  What others did I miss?

PR #2515, opened 2024-04-19
chore: Update Dockerfile to node:18

landrito:
Addresses #2514

**Summary**

Upgrades base image in Dockerfile to `node:18.20.1` to resolve vulnerabilities flagged on the docker image.

**Checklist**

- The basics
  - [x] I tested these changes manually in my local or dev environment
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [ ] I added event tracking and followed the event tracking guidelines
  - [x] N/A
- Error Reporting
  - [ ] I reported errors and followed the error reporting guidelines
  - [x] N/A

**Screenshots**

N/A

**Additional context**

N/A

PR #2536, opened 2024-05-28
chore: Update Dockerfile to node:18.20.3

pieterocp:
Follow up from #2515 

**Summary**

Appeases security tooling since the [updates to node](https://nodejs.org/en/blog/release/v18.20.3) causes some tools to moan.

**Checklist**

- The basics
    - Ran the build locally, used it in my project and it seems to be working fine.
- Tests
  - [ ] Added or updated
  - [X] N/A
    - As far as I am aware, there's no tests for using the projects container in an end to end fashion?
- Event Tracking
    - (I cannot find any documentation for this so I guess N/A?)
  - [X] N/A
- Error Reporting
  - [Z] N/A

**Additional context**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18.20.3 as compiler
+FROM node:18 as compiler
 
 WORKDIR /usr/src/prism
 
@@ -8,7 +8,7 @@ COPY packages/ /usr/src/prism/packages/
 RUN yarn && yarn build
 
 ###############################################################
-FROM node:18.20.3 as dependencies
+FROM node:18 as dependencies
 
 WORKDIR /usr/src/prism/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18.20.1 as compiler
+FROM node:18.20.3 as compiler
```
brendarearden, line 1:
```suggestion
FROM node:18 as compiler
```

Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY packages/ /usr/src/prism/packages/
 RUN yarn && yarn build
 
 ###############################################################
-FROM node:18.20.1 as dependencies
+FROM node:18.20.3 as dependencies
```
brendarearden, line 11:
```suggestion
FROM node:18 as dependencies
```

Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN if [ $(uname -m) != "aarch64" ]; then curl -sfL https://gobinaries.com/tj/no
 RUN if [ $(uname -m) != "aarch64" ]; then node-prune; fi
 
 ###############################################################
-FROM node:18.20.1-alpine
+FROM node:18.20.3-alpine
```
brendarearden, line 37:
```suggestion
FROM node:18-alpine
```

brendarearden (approved):

PR #2538, opened 2024-06-03
fix: circular reference issue resolved

SB-rohitdesai:
Mock server not working In case of circular references

**Summary**

Getting 500 Error or Sometimes might need to wait a long time before getting this error when spec having circular references
**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

### DIFF
--- a/packages/http/src/mocker/__tests__/HttpMocker.spec.ts
+++ b/packages/http/src/mocker/__tests__/HttpMocker.spec.ts
@@ -92,7 +92,7 @@ describe('mocker', () => {
                 {
                   id: 'example-1',
                   key: 'invalid_1',
-                  value: 'invalid input 1',
+                  value: 'invalid input 2',
                 },
                 {
                   id: 'example-2',
@@ -100,6 +100,19 @@ describe('mocker', () => {
                   value: 'invalid input 2',
                 },
               ],
+              encodings: [],
+            },
+          ],
+        },
+        {
+          id: '500',
+          code: '500',
+          headers: [],
+          contents: [
+            {
+              id: 'contents',
+              mediaType: 'application/json',
+              examples: [{ id: 'example-1', key: 'internalServerError', value: { message: 'Internal Server Error' } }],
               encodings: [],
             },
           ],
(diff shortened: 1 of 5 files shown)


bhaskarsontakke (approved):
looks good to me.

brendarearden (changes requested):
could you add a test to cover the 500 scenario?

brendarearden:
is this supposed to include a version of json-schema-ref-parser that has your [optimization changes](https://github.com/stoplightio/json-schema-ref-parser/pull/62) in it? https://github.com/stoplightio/json-schema-ref-parser/pull/62 in it?

SB-rohitdesai:
> could you add a test to cover the 500 scenario?

Test case added

SB-rohitdesai:
> is this supposed to include a version of json-schema-ref-parser that has your [optimization changes](https://github.com/stoplightio/json-schema-ref-parser/pull/62) in it? [stoplightio/json-schema-ref-parser#62](https://github.com/stoplightio/json-schema-ref-parser/pull/62) in it?

Not required. currently we just merging this PR only

bhaskarsontakke (approved):

bhaskarsontakke (approved):

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -676,5 +689,31 @@ describe('mocker', () => {
         });
       });
     });
+
+    describe('should return 500 error', () => {
```
brendarearden, line 693: This test does not seem relevant to the PR that has been opened. Could you please add a test that covers the original circular reference problem and show that it is no longer throwing a 500?

brendarearden (changes requested):

packages/http/src/mocker/__tests__/HttpMocker.spec.ts
```diff
@@ -676,5 +689,142 @@ describe('mocker', () => {
...
+        assertRight(mockResult, result => {
+          expect(result).toMatchObject({
+            statusCode: 500,
+            body: { message: 'Internal Server Error' },
+          });
+        });
+      });
+    });
+
+    describe('With __bundled__ should get response', () => {
```
brendarearden, line 719: This test is passing with and without your changes so it still doesn't seem to be testing the initial bug you were seeing / the fix hasn't changed anything. It should be removed or changed to properly capture the issue you are seeing and fixing.

We have [test harness tests](https://github.com/stoplightio/prism/blob/master/test-harness/specs/references/E2E-Recursive-Object-Handling.oas3.txt) for circular references and those have been passing in the pipeline so it seems like circular references are getting resolved in most cases. Maybe a test harness test for the specific scenario that is causing you not to see circular references get resolved would help demonstrate the issue a bit better.

PR #2556, opened 2024-07-03
chore(ci): version bump to fix release PR loop

brendarearden:
Addresses STOP-537

**Summary**

Doing a general bump of versions to see if that resolves the master branch getting stuck in a release PR loop for the root project per this suggestion: https://stackoverflow.com/questions/75806309/release-please-creates-the-same-release-pr-after-merge. We have been seeing the same PR get created with a version bump over and over. If this does not work, we will remove the root from the release please configuration. 

Changes:
- Cleaned up change log files to be the correct changes for the version
- Reverted the root version to match the sub-packages

**Checklist**

- The basics
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A


brendarearden:
I am not super convinced that bumping the versions so they are all the same will work. I think the issue will eventually come up again because here is the history: 
- https://github.com/stoplightio/prism/pull/2498 << all packages have same version (5.6.0)
- https://github.com/stoplightio/prism/pull/2503 << all packages have same version (5.7.0)
- https://github.com/stoplightio/prism/pull/2517 << core does not bump (5.7.0), but all the other ones do (5.8.0)
https://github.com/stoplightio/prism/pull/2521 << core does bump (5.8.0.) root bumps 5.9.0, others do a patch bump, 5.8.1 and the problem starts from this point on

There must be some bug with the linked-versions config.

brendarearden:
Not the route we want to go in order to resolve this issue.

PR #2564, opened 2024-07-17
fix: mock issue resolve for similar templated requests

SB-akshaythakar:
Addresses STOP-448

**Summary**

We observe that the issue is occurring in mock server when OAS is having similar templated requests as below example. Request was getting mapped with wrong endpoint and mock server was returning 400 error code.
 
Consider following API endpoints defined in project spec. 
1.	[https://stoplight-local.com:8443/mocks/seeded-enterprise/mock/174/{initgPtyIdOrgId}/payments/{paymentId}](https://stoplight-local.com:8443/mocks/seeded-enterprise/mock/174/%7BinitgPtyIdOrgId%7D/payments/%7BpaymentId%7D)
2.	[https://stoplight-local.com:8443/mocks/seeded-enterprise/mock/174/{initgPtyIdOrgId}/payments/search](https://stoplight-local.com:8443/mocks/seeded-enterprise/mock/174/%7BinitgPtyIdOrgId%7D/payments/search)
 
we can find two paths here - 
1. /{initgPtyIdOrgId}/payments/{paymentId}
2. /{initgPtyIdOrgId}/payments/search
 
If we call search Api from mock server our request will be encoded to "/SMARTBEAR/payments/search". here "SMARTBEAR" is initgPtyIdOrgId. And if call is made to payment API, request will be encoded to - SMARTBEAR/payments/paytm123. Here SMARTBEAR is initgPtyIdOrgId and paytm123 is paymentId. But if you compare the both requests looks similar.
 
There is matchPath function in prism-http which is handling request and endpoint mapping functionality and for above scenario it is returning two endpoint objects for search API. Ideally it should return single object array. And make-payment API is placed before search Api in spec, make-Api object will be the first item in array and matchPath is using first item from array to call mock Api.
 
To solve this, I created the sorting function in http-spec where all these endpoints will be sorted on the basis their match with request path. In case we get multiple matched endpoints for any request, this sorting will make sure that first endpoint in array is close match to request.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [ ] N/A
- Error Reporting
  - [ ] N/A

**Screenshots**

**Additional context**

### DIFF
--- a/packages/http/package.json
+++ b/packages/http/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/prism-http",
-  "version": "5.8.2",
+  "version": "5.8.3",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "author": "Stoplight <support@stoplight.io>",
--- a/packages/http/src/router/index.ts
+++ b/packages/http/src/router/index.ts
@@ -2,6 +2,7 @@ import { IPrismComponents } from '@stoplight/prism-core';
 import { IHttpOperation, IServer } from '@stoplight/types';
 import * as E from 'fp-ts/Either';
 import * as A from 'fp-ts/Array';
+import * as O from 'fp-ts/Ord';
 import { pipe } from 'fp-ts/function';
 import { IHttpConfig, IHttpRequest, ProblemJsonError } from '../types';
 import {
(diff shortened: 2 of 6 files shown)


packages/http/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@stoplight/prism-http",
-  "version": "5.8.2",
+  "version": "5.8.3",
```
prafullaAtSB, line 3: Is this required?  just check if it gets incremented bu build system during release or deployment ?

packages/http/src/router/index.ts
```diff
@@ -14,16 +15,42 @@ import {
...
+// matchScore is introduced to sort the resources in case we get multiple matches for given request.
+const calculateMatchScore = (endpoint: string, request: string) => {
+  const endpointParts = endpoint.split('/');
+  const requestParts = request.split('/');
+  let score = 0;
+  for (let i = 0; i < endpointParts.length; i++) {
+    if (endpointParts[i] === requestParts[i]) {
+      score++;
+    } else if (isTemplated(endpointParts[i])) {
+      score += 0.5;
```
bhaskarsontakke, line 30: what's the purpose of this score += 0.5; ?

packages/http/package.json
SB-akshaythakar: Yes. we need to change the version as per the change.

packages/http/src/router/index.ts
SB-akshaythakar: Added score to differentiate match weightage. When templated path is found, 0.5 score is added to it and if path is not templated and it is matching with request, then score is incremented by 1. So non-template path will have more weightage compare to templated path while performing sorting.

prafullaAtSB (approved):
Changes are good to go ahead

bhaskarsontakke (approved):

PR #2654, opened 2025-01-17
feat: Support mapping repeated keys to array in form-data

lilingxi01:
**Summary**

This PR adds the support of mapping repeated form keys into arrays for validation in multipart form-data and url-encoded form-data. Using repeated keys to fulfill arrays in form-data is nowadays a popular workaround whenever an array is needed inside form-data ([reference](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-timestamp_granularities), [reference](https://lingxi.li/posts/multipart-form-data-standardization)). In the old version of Prism, repeated keys will be treated as overriding previous values, thus failing the validation.

**Checklist**

- The basics
- Tests
  - [x] Added or updated
  - [ ] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A

**Additional context**

I discovered this need when working with Stainless auto-generated SDK, where Prism is used to test the SDK. Whenever I have an array in multipart form-data, Prism simply fails with 422, but the same input easily passes with real server. Thus, I dug into the workarounds online and realized that repeating keys is an untold standard in accomplishing this need. OpenAI happens to have the exact same awkward situation with Stainless + Prism, and there is no workaround online for mock server like this, thus I decided to make Prism support it.

Interesting side note: OpenAI's library tests were luckily passing because their field is optional, and the key name is mismatched in their OpenAPI schema, thus Prism simply thought that the key was not provided and treated it as valid request. OpenAI's SDK (and per API reference) passes `timestamp_granularities[]` into the Prism while having `timestamp_granularities` as key name in the Stainless-hosted OpenAPI schema, so Prism thinks it is a valid request, but it is actually being untested. In my use case, my array field needs to be required, so I end up needing to find a workaround more seriously. With this patch, OpenAI's SDKs could then actually be tested against their actual OpenAPI schema (with `timestamp_granularities[]` as key name).


dgellow:
Hi @tesanchez, @lukaszzazulak, we would love to get a review from the Prism team here :)
Anything we can do to help speed up that process?

SergioV85 (approved):

PR #2679, opened 2025-03-27
feat: prevent crashing on parsing malformed JSONs

sohitSmartbear:
Addresses #STOP-2386

**Summary**

Critical Logging Issue with Prism Proxy

Current Behavior
Attached are our config, a video walking through how we produced the issue locally and the smallest JSON body that we could reproduce the issue. (pügememory) in the JSON contains ü character we found during testing.

 

Notes

 

We were only ever able to reproduce this with a large payload that contains the ü character.

During the debugging process we noticed the error being thrown around a missing “:” from what we could see.

In some cases, different sized requests (smaller and larger) would not cause the issue. We would insert duplicate in the attributes section like in the JSON example.

Below are the steps we used for reproducing this locally:

Running Prism Proxy Locally with our yml config
**Checklist**

- The basics
  - [ ] I tested these changes manually in my local or dev environment
- Tests
  - [ ] Added or updated
  - [x] N/A
- Event Tracking
  - [x] N/A
- Error Reporting
  - [x] N/A


bhaskarsontakke:
The changes look good to me. Also, please make sure to test the changes with various specifications files and attach the results.

sohitSmartbear:
Screenshot 2025-04-09 at 4 59 18 PM
Screenshot 2025-04-09 at 6 07 54 PM
Screenshot 2025-04-10 at 3 23 37 PM
Screenshot 2025-04-10 at 3 24 10 PM
Screenshot 2025-04-10 at 3 33 43 PM
Screenshot 2025-04-10 at 3 46 02 PM
Screenshot 2025-04-10 at 3 46 09 PM
Screenshot 2025-04-10 at 3 50 46 PM

packages/cli/src/util/createServer.ts
```diff
@@ -132,10 +133,20 @@ function pipeOutputToSignale(stream: Readable) {
...
   }
-
-  stream.pipe(split(JSON.parse)).on('data', (logLine: PrismLogDescriptor) => {
-    signale[logLine.level]({ prefix: constructPrefix(logLine), message: logLine.msg });
-  });
+  stream
+    .pipe(
+      split(chunk => {
+        try {
+          const repairedJson = jsonrepair(chunk);
```
prafullaAtSB, line 140: Do you have impact analysis, if there is repair happens to the chunk then it will not impacting any change in result?

packages/cli/src/util/createServer.ts
```diff
@@ -132,10 +133,20 @@ function pipeOutputToSignale(stream: Readable) {
...
+          return JSON.parse(repairedJson);
+        } catch (error) {
```
prafullaAtSB, line 142: why exception handling is required here. what are the chances after repair json.parse will throw exception? 
Do we have expected error handling if jsonrepair will get fails?

prafullaAtSB:
As discussed during review call, create the word document to capture the testing results for different api specifications. Specially result of other api available in the same api specification. Specially use the customer specifications mentioned in STOP-184. 
Also, update the confluence page - capture the appropriate details on root cause, solution approach and impact analysis.

PR #2788, opened 2026-04-28
fix(http): skip transport-level headers in mock response generation

### LINKED ISSUE #2787: Prism adds Content-Encoding headers from spec but doesn't compress response body
## Context

When an OpenAPI spec defines `Content-Encoding` as a response header, Prism adds this header to mock responses but doesn't actually compress the response body. This causes HTTP clients to fail when attempting to decompress the uncompressed data.

## Current Behavior

When an OpenAPI spec includes `Content-Encoding` as a response header with enum values like:

```json
"headers": {
  "Content-Encoding": {
    "schema": {
      "type": "string",
      "enum": ["gzip", "deflate", "br", "compress"]
    }
  }
}
```

Prism randomly selects one of these enum values and adds it to the response headers, but **the response body remains uncompressed**. When HTTP clients (like Python's `requests` library) see the `Content-Encoding` header, they attempt to decompress the body and fail with:

```
zlib.error: Error -3 while decompressing data: incorrect header check
```

or

```
requests.exceptions.ContentDecodingError: ('Received response with content-encoding: gzip, but failed to decode it.')
```

## Expected Behavior

Prism should **not** add `Content-Encoding` headers to responses unless it's actually compressing the response body. The `Content-Encoding` header in an OpenAPI spec is describing what the real API does - the mock server shouldn't blindly add it without implementing the actual compression.

Alternatively, Prism could compress the response body to match the header, but that seems unnecessary for a mock server (or maybe not 🤷 ).

## Possible Workaround/Solution

**Workaround 1**: Remove `Content-Encoding` from the response headers in your OpenAPI spec when using with Prism.

**Workaround 2**: Use nginx (or another reverse proxy) in front of Prism to strip the `Content-Encoding` header:

```nginx
location / {
  proxy_pass http://prism:4010;
  proxy_hide_header Content-Encoding;
}
```

**Suggested Fix**: Prism should ignore `Content-Encoding` when generating response headers, or at minimum, document that this header should not be included in OpenAPI specs used with Prism.

## Steps to Reproduce

1. Create a minimal OpenAPI spec with `Content-Encoding` in response headers (see below)

2. Start Prism:
   ```bash
   npx @stoplight/prism-cli@5.15.10 mock -h 0.0.0.0 -p 4010 prism-minimal-bug-spec.json
   ```

3. Make requests with a client that supports decompression:
   ```python
   import requests
   
   session = requests.Session()
   session.headers.update({'Accept-Encoding': 'gzip, deflate'})
   
   for i in range(10):
       try:
           r = session.get('http://localhost:4010/test')
           print(f"Run {i+1}: SUCCESS")
       except requests.exceptions.ContentDecodingError as e:
           print(f"Run {i+1}: FAILED - {e}")
   ```

4. Result: **100% of requests fail** with decompression errors

**Minimal OpenAPI Spec** (`prism-minimal-bug-spec.json`):
```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "Minimal Prism Content-Encoding Bug",
    "version": "1.0.0"
  },
  "paths": {
    "/test": {
      "get": {
        "summary": "Test endpoint",
        "responses": {
          "200": {
            "description": "Success",
            "headers": {
              "Content-Encoding": {
                "description": "The compression algorithm used",
                "schema": {
                  "type": "string",
                  "enum": ["gzip", "deflate", "br", "compress"]
                }
              }
            },
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "message": {
                      "type": "string"
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Environment

- **Prism Version**: 5.14.2 and 5.15.10 (bug exists in both)
- **Environment**: Node.js v20.19.5 and v25.9.0
- **Client**: Python 3.11+ with requests library
- **Operating System**: macOS, Linux (Docker)
- **Installation method**: `npx @stoplight/prism-cli@5.15.10`

rnagulapalle:
## Summary

When an OpenAPI spec defines transport-level headers like `Content-Encoding` as response headers, Prism's `computeMockedHeaders` generates values from the schema (e.g. picking `gzip` from an enum) but doesn't actually compress the response body. This causes HTTP clients to fail with decompression errors like `zlib.error: Error -3 while decompressing data`.

The root cause is that `computeMockedHeaders` treats every spec-defined header equally — including transport-level headers whose values must be derived from the actual response, not generated from the spec schema.

## Fix

Added a `TRANSPORT_HEADERS` set containing `content-encoding`, `content-length`, `transfer-encoding`, and `content-md5`. These are filtered out before header generation in `computeMockedHeaders`.

These headers describe transport behavior the mock server doesn't implement, so generating values for them is actively harmful.

## Changes

- `packages/http/src/mocker/index.ts` — filter transport-level headers before mock generation
- `packages/http/src/mocker/__tests__/HttpMocker.spec.ts` — added tests verifying Content-Encoding, Transfer-Encoding, and Content-Length are excluded from mocked responses while application headers like X-Request-Id are preserved

Fixes #2787

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/http/src/mocker/__tests__/HttpMocker.spec.ts
+++ b/packages/http/src/mocker/__tests__/HttpMocker.spec.ts
@@ -1,5 +1,5 @@
 import { createLogger, IPrismInput } from '@stoplight/prism-core';
-import { IHttpOperation, INodeExample, DiagnosticSeverity } from '@stoplight/types';
+import { IHttpOperation, INodeExample, DiagnosticSeverity, HttpParamStyles } from '@stoplight/types';
 import { right } from 'fp-ts/ReaderEither';
 import * as E from 'fp-ts/Either';
 import { flatMap } from 'lodash';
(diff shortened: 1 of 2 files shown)


rnagulapalle:
Hi Prism maintainers — gentle review nudge on this one. The checks are green, and the change is scoped to skipping transport-level headers when generating mock responses. Is there anything else you'd like adjusted before review?

PR #2813, opened 2026-06-26
fix(cli): restore cluster namespace import to fix startup regression

fuleinist:
## Problem

PR #2763 reports that Prism 5.15+ fails to start in any mode (mock, proxy, etc.):

```
TypeError: Cannot read properties of undefined (reading 'isPrimary')
    at createMultiProcessPrism (packages/cli/dist/util/createServer.js:29:32)
```

## Root cause

PR #2752 (Node 18 → 24 bump) changed the cluster import style from
namespace to default:

```diff
-import * as cluster from 'cluster';
+import cluster from 'node:cluster';
```

Under `module: commonjs` + `esModuleInterop: false`, the default import
compiles to `node_cluster_1.default.isPrimary`, but `node_cluster_1.default`
is `undefined` at runtime because the cluster module has no default export.
The namespace import compiles to `cluster.isPrimary`, which works because
the module IS the singleton Cluster instance.

## Fix

Restore the namespace import with a type cast to satisfy the Cluster type
from `@types/node` 24+ (which only exposes the Cluster interface via the
default export). The cast is erased at compile time; runtime behavior is
identical to a plain namespace import.

```ts
import * as clusterNS from 'node:cluster';
const cluster = clusterNS as unknown as typeof import('node:cluster').default;
```

## Verification

```js
cluster.isPrimary === true            // was undefined before fix
typeof cluster.setupPrimary === 'function'
typeof cluster.fork       === 'function'
```

Pre-existing TypeScript errors in `paths.ts`, `forwarder/index.ts`, and
`validator/index.ts` (unrelated to this fix) remain unchanged.

Fixes #2763

### DIFF
--- a/packages/cli/src/util/createServer.ts
+++ b/packages/cli/src/util/createServer.ts
@@ -2,7 +2,7 @@ import { createLogger } from '@stoplight/prism-core';
 import { IHttpConfig, IHttpRequest } from '@stoplight/prism-http';
 import { createServer as createHttpServer } from '@stoplight/prism-http-server';
 import * as chalk from 'chalk';
-import cluster from 'node:cluster';
+import * as clusterNS from 'node:cluster';
 import * as E from 'fp-ts/Either';
 import { pipe } from 'fp-ts/function';
 import * as pino from 'pino';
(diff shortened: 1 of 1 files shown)


fuleinist:
Hi maintainers — gentle ping on this PR. It fixes a startup regression (#2763) where Prism 5.15+ crashes with `TypeError: Cannot read properties of undefined (reading 'isPrimary')` due to the Node 18→24 cluster import change.

The fix is a 10-line change in a single file (`packages/cli/src/util/createServer.ts`): restores the namespace import with a type cast that's erased at compile time. All CI checks pass.

Would appreciate a review when you have a moment. Thanks!

fuleinist:
Hi maintainers — gentle ping on this PR. It restores the cluster namespace import to fix a startup regression. CI checks pass and the PR is mergeable. Would appreciate a review when you have a moment. Thanks!

fuleinist:
Ping: this is a 10-line regression fix for Prism 5.15+ startup failure (#2763). CI is green. Would appreciate a review when you have a moment — the fix restores the namespace import style that the Node 18→24 bump accidentally broke.

PR #2823, opened 2026-07-01
fix: prefer exact response media type schema

### LINKED ISSUE #2509: Proxy is not validating against the correct response schema when using accept header for versioning
Prism proxy is not validating against the correct response schema when using accept header for versioning

## Context
I cant make versioned requests using accept header and expect the prism proxy to verify the correct response schema in the openapi file.

## Current Behavior
It currently takes the first schema, for the first response listed under responses and validates the server response based on that.

## Expected Behavior
It should read the content type from the response and select the proper response schema in the openapi path.

## Possible Workaround/Solution
I dont have any good suggestions.

## Steps to Reproduce
Use the following openapi spec
```json
{
    ..... the rest of the spec .....
    
    "paths": {
        "/api/stuff/{stuffid}": {
            "get": {
                "tags": [
                    "stuff"
                ],
                "operationId": "getStuff",
                "summary": "Get stuff",
                "description": "Get some stuff",
                "parameters": [
                    {
                        "name": "stuffid",
                        "in": "path",
                        "description": "Stuff id",
                        "required": true,
                        "schema": {
                            "type": "string",
                            "example": "stuff id"
                        }
                    }
                ],
                "responses": {
                    "200": {
                        "description": "200",
                        "content": {
                            "application/vnd.stuff.v2+json": {
                                "schema": {
                                    "$ref": "#/components/schemas/StuffV2"
                                }
                            },
                            "application/vnd.stuff.v1+json": {
                                "schema": {
                                    "$ref": "#/components/schemas/StuffV1"
                                }
                            },
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/StuffV2"
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

run the proxy with errors enabled

`prism proxy stuffapi.json http://stuffserver.com --errors`

call the url

curl -v http://localhost:4010/api/stuff/fluffybear -H "Accept: application/vnd.stuff.v1+json" 

if response returns v1 as it should, the proxy will test for v2 since its first on the list of responses.

I have verified this by actually reordering responses in the spec. 

## Environment

- Version used:
- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your environment/workspace/project:

ishaanlabs-gg:
## Summary

- prefer an exact response media type match before wildcard media type matches during body schema selection
- add a regression test for vendor JSON response content where a wildcard schema appears first

Closes #2509.

## Test

- `npm test -- packages/http/src/validator/validators/__tests__/body.spec.ts --runInBand --no-watchman`

### DIFF
--- a/packages/http/src/validator/validators/__tests__/body.spec.ts
+++ b/packages/http/src/validator/validators/__tests__/body.spec.ts
@@ -53,6 +53,40 @@ describe('validate()', () => {
           )
       );
     });
+
+    it('selects an exact response media type before a wildcard match', () => {
+      const specs: IMediaTypeContent[] = [
+        {
+          id: faker.word.sample(),
+          mediaType: 'application/*+json',
+          schema: {
+            type: 'object',
+            required: ['v2Only'],
+            properties: {
+              version: { type: 'string' },
+              v2Only: { type: 'boolean' },
+            },
+          },
+          examples: [],
+          encodings: [],
+        },
+        {
+          id: faker.word.sample(),
+          mediaType: 'application/vnd.stuff.v1+json',
+          schema: {
+            type: 'object',
+            required: ['version'],
+            properties: {
+              version: { type: 'string' },
+            },
+          },
+          examples: [],
+          encodings: [],
+        },
+      ];
+
+      assertRight(validate({ version: 'v1' }, specs, ValidationContext.Output, 'application/vnd.stuff.v1+json'));
+    });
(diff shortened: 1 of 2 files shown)


ishaanlabs-gg:
Closing this duplicate in favor of #2820, which now contains the same focused fix with the cleaner minimal diff.

PR #2830, opened 2026-07-16
feat: add scarf analytics

glowcloud:


robert-hebel-sb (approved):

robert-hebel-sb:
in node 24.18 there is an option to disable install scripts via `npm deny-scripts <pkg>` command
ref: https://docs.npmjs.com/cli/v11/commands/npm-deny-scripts

similar thing is implemented in pnpm and bun as well https://pnpm.io/cli/approve-builds https://bun.com/docs/guides/install/trusted

It's more like a fun fact; it's up to you if it's worth including in the docs 😅

glowcloud:
> in node 24.18 there is an option to disable install scripts via `npm deny-scripts <pkg>` command ref: https://docs.npmjs.com/cli/v11/commands/npm-deny-scripts
> 
> similar thing is implemented in pnpm and bun as well https://pnpm.io/cli/approve-builds https://bun.com/docs/guides/install/trusted
> 
> It's more like a fun fact; it's up to you if it's worth including in the docs 😅

Leaving it for now, we can add it later on here and in other repos 😅

